PR #28, opened 2026-03-06
fix: sync MCP tools on add/remove via interrupt+sendMessage

yiw190:
## Fix: REMOTE_MCPS not dynamically updated in running agents

**Problem**: When adding/removing MCP servers from the UI, running agents kept stale tool lists until manually restarted. Deletions didn't notify the container at all; additions used a heavy restart approach that sometimes failed silently.

**Solution**: Add a new `/mcp/sync` container endpoint. On MCP add/remove, the host pushes the updated `REMOTE_MCPS` list to the container, which then interrupts the current query and sends a system message so the model picks up the new tool set on the next API call.

### What changed

1. **Before**: Deletions didn't notify the container; additions used POST /env + interrupt/restart.
2. **Now**: Both add and delete go through `POST /mcp/sync` → `interrupt()` + `sendMessage()` with the change summary. The model picks up the updated tool list on the next query.
3. **`POST /env`** left untouched for other future use; MCP changes exclusively go through `/mcp/sync`.

### Files

- `agent-container/src/server.ts` — new `POST /mcp/sync` endpoint; detects removed MCPs and calls `notifyMcpRemoved()`
- `agent-container/src/claude-code.ts` — `notifyMcpRemoved()` interrupts the current query and informs the model which tools were removed
- `src/api/routes/agents.ts` — `pushRemoteMcpsToContainer()` helper; delete + provide-remote-mcp routes call it
- `src/api/routes/remote-mcps.ts` — global MCP deletion notifies all affected containers via `Promise.allSettled`

### Result

**Add MCP**: Agent picks up new tools on next query after interrupt
image

**Delete MCP**: Agent is informed tools are removed in real-time
image


iddogino:
@yiw190 conflicts

iddogino:
@yiw190 broken -- when adding new MCPs they are not added to session
Screenshot 2026-03-09 at 6 55 16 PM

yiw190:
Fixed as well

PR #39, opened 2026-03-12
fix: add proxy usage pitfalls to agent system prompt

yiw190:
## Summary

Workaround for a Claude Code Bash tool bug ([anthropics/claude-code#29298](https://github.com/anthropics/claude-code/issues/29298)) where `$VAR` expansions silently return empty strings when a command has no `#` comment line.

- Add a `# comment` curl example to the proxy usage section in `system-prompt.md`
- Add a **Proxy pitfalls** note: always include a `#` comment line before any bash command that uses `$PROXY_TOKEN` or `$PROXY_BASE_URL`

Linear: [SUP-82](https://linear.app/datawizz/issue/SUP-82/bug-first-proxy-request-always-fails)

## Test plan
- Deploy to a test agent, ask it to call a connected account API via curl
- Verify it includes a `#` comment line before proxy commands
- Verify `$PROXY_TOKEN` and `$PROXY_BASE_URL` expand correctly

### DIFF
--- a/agent-container/src/system-prompt.md
+++ b/agent-container/src/system-prompt.md
@@ -176,6 +176,17 @@ if gmail_accounts:
     print(response.json())
 ```
 
+**Using curl with the proxy:**
+```bash
+curl -sf -H "Authorization: Bearer $PROXY_TOKEN" \
+  "$(echo -n $PROXY_BASE_URL)/${ACCOUNT_ID}/api.github.com/user/repos"
+```
+
+**⚠️ Proxy pitfalls:**
+- `$PROXY_TOKEN` only works with the proxy URL — never send it directly to external APIs.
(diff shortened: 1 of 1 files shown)


yiw190:
**Related Linear issue:** [SUP-82 — Bug - first proxy request always fails](https://linear.app/datawizz/issue/SUP-82/bug-first-proxy-request-always-fails)

**Root cause (from SUP-82):** Claude Code Bash tool uses `eval` internally. When a command has no `#` comment, `$VAR` expansions silently return empty strings ([anthropics/claude-code#29298](https://github.com/anthropics/claude-code/issues/29298)). This makes `$PROXY_TOKEN` and `$PROXY_BASE_URL` intermittently empty, causing proxy calls to fail with empty responses or auth errors.

**Workaround in this PR:** Add a `# comment` line before any bash command that uses proxy env vars, and document the pitfall in the agent system prompt.

PR #44, opened 2026-03-17
Feat/multi tab preview

apai25:
- Support for multi-tab preview
- Allow user to click through tabs (especially useful in container mode, when user input causes new tab to open)
- Allow user to "follow" agent and see which tab is currently being viewed by agent
- **Bug Fix:** In container mode, `tab new [url]` was not working due to bug in agent browser. Quick fix: Route to custom JS instead of using agent browser. Long term TODO is to upgrade agent browser, but due to major refactoring on their end, this would require a more significant effort on our part as well. Leaving this for another PR. 
Screenshot 2026-03-17 at 4 33 14 PM

### DIFF
--- a/agent-container/src/browser-command-args.ts
+++ b/agent-container/src/browser-command-args.ts
@@ -44,3 +44,13 @@ export function buildRunCommandArgs(command: string): string[] {
 
   return commandArgs;
 }
+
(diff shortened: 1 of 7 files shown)


iddogino (changes requested):
There are a couple of lint warnings in the PR -- can we fix?
There's no test coverage for the new functionality - can you asses if/what can be tested?

PR #66, opened 2026-04-04
feat: platform-backed skillset provider with org-scoped sync and publish

yiw190:
## What this does

Add **platform** as a first-class skillset provider alongside GitHub, using a **SkillsetProvider** polymorphic abstraction to encapsulate all hosting-specific logic.

### Design principles

**1. No type-switching on providers** — calling code never branches on `provider === 'platform'` or `provider === 'github'`. All provider-specific logic lives inside the concrete provider class.

**2. Shared types carry no implementation semantics** — `SkillStatus`, `AgentTemplateStatus`, and API types use provider-neutral fields (`sourceLabel`) instead of platform-specific ones (`skillsetOrgId`, `skillsetOrgName`, `currentPlatformOrgId`).

**3. Open/Closed** — adding a new provider means creating a class and registering it. No existing code needs to change.

### Security: org-scoped skill lifecycle

When a user **disconnects from an org** or **switches to a different org**, all skills from the old org are **immediately and permanently removed**:

1. **Settings cleanup** — `removePlatformSkillsets()` removes all platform skillset configs belonging to the old org from `settings.json`
2. **File cleanup** — `removePlatformSkillFiles()` scans every agent workspace, reads `.skillset-metadata.json` in each installed skill directory, and **physically deletes** any skill directory whose `providerData.orgId` doesn't match the current org. Same for agent template metadata.
3. **Startup reconciliation** — `reconcilePlatformSkillsets()` runs on every app launch to catch any stale data (e.g. app crashed mid-cleanup, legacy entries without `orgId`)

This means: if you fire an employee and revoke their org access, they lose all org skills on their next app launch — no "accessible but hidden" gray area.

#### SkillsetProvider abstraction
- `BaseSkillsetProvider` / `GithubSkillsetProvider` / `PlatformSkillsetProvider` in `shared/lib/skillset-provider/`
- Each provider owns: clone URL resolution, publish flow, sync preconditions, source display info, queue status polling
- Provider data stored in a generic `providerData` bag on `SkillsetConfig` — no top-level platform-specific fields

#### What was removed
- `getAccessibleSkillsets()` / `getAccessInfo()` / `getConfigAccessInfo()` / `getInstalledAccessInfo()` — replaced by write-time cleanup
- `getSkillsetAccessScope()` from route files — routes now use `getSettings().skillsets` directly
- `isAccessible` field — no longer exists; if a skill is in settings, it's accessible
- `skillset-helpers.ts` — deleted; contents moved into providers or inlined
- `platform-submission.ts` — deleted; `prUrl` pseudo-protocol replaced with structured `SkillsetPublishResult`
- Provider type-switching in UI (`provider === 'platform'`, `prUrl.startsWith('platform:')`) — replaced with provider-neutral display model

#### Platform provider & submission flow
- Submit via proxy (`/v1/skills/submit-update`) with queue-based status polling
- Full skill-package (multi-file) support
- Rejected submissions throw instead of encoding status in URL strings
- `SkillsetPublishResult` returns `{ successMessage, prUrl?, status?, queueItem? }`

#### Remote update & status detection
- Content hash comparison (not just `index.json` version) to detect remote changes
- Auto-sync on refresh when upstream has moved forward after PR merge
- `gitPull` hardened: checkout default branch, fetch, hard-reset to origin
- Legacy single-file hash migration for older installs

#### Source display
- `getSourceInfo()` returns `{ skillsetName, sourceLabel }` — provider-neutral
- `sourceLabel` only shown when skillset config exists (orphan skills show no source)
- UI uses `sourceLabel` directly, no `|| skillsetName` fallback

---

~30 files changed across the full stack, includes unit tests for all new behavior.


src/shared/lib/utils/skillset-helpers.ts
iddogino, line 1: This file contains mostly functions that are SkillsetProvider implementation dependent. They should reside as private functions in their respective skillsets - not in shared helpers... 
For instance:

- getPlatformOrgIdFromRepoId
- getPlatformOrgName
- ensureGhAuthenticated
etc...

src/shared/lib/utils/skillset-helpers.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  const prefix = 'Default skillset for '
+  if (config.description?.startsWith(prefix)) {
+    return config.description.slice(prefix.length).trim() || undefined
+  }
+
+  return undefined
+}
+
+export function resolvePlatformSkillsetScope(params: {
```
iddogino: What does this function do?

src/shared/lib/types/skillset.ts
```diff
@@ -124,14 +162,51 @@ export interface InstalledAgentMetadata {
...
+    skillsetName?: string
+    skillsetOrgId?: string
+    skillsetOrgName?: string
+    publishable?: boolean
+  }
+  | {
+    type: 'up_to_date'
+    skillsetId: string
+    skillsetName: string
+    skillsetOrgId?: string
```
iddogino: why would a template status have org name? That seems implementation dependent for platform. This should all be generic / polymorphic.

src/shared/lib/types/skillset.ts
```diff
@@ -30,13 +30,20 @@ export interface SkillsetIndex {
...
+
 /** A configured skillset in user settings */
 export interface SkillsetConfig {
   id: string // deterministic slug from URL
   url: string // git clone URL
   name: string // from index.json skillset_name
   description: string // from index.json description
   addedAt: string // ISO date
+  provider?: SkillProvider // defaults to 'github'
+  platformRepoId?: string // CodeStorage repo ID for platform provider
```
iddogino: Need implementation specific extensions -> not dump everything into the general config. Again -> this is bad polymorphic design

src/renderer/components/agents/status-badge.tsx
```diff
@@ -5,6 +5,9 @@ interface StatusBadgeProps {
     type: 'local' | 'up_to_date' | 'update_available' | 'locally_modified'
     skillsetId?: string
     skillsetName?: string
+    skillsetOrgId?: string
+    skillsetOrgName?: string
+    publishable?: boolean
```
iddogino: why is this here? also this type definition is repeated in a bazillion places

src/renderer/components/agents/skill-publish-dialog.tsx
```diff
@@ -160,18 +173,22 @@ export function SkillPublishDialog({
...
-                <a
-                  href={prUrl}
-                  target="_blank"
-                  rel="noopener noreferrer"
-                  className="flex items-center gap-2 text-sm text-primary hover:underline"
-                >
-                  <ExternalLink className="h-4 w-4" />
-                  {prUrl}
-                </a>
+                {!prUrl.startsWith('platform:') && (
```
iddogino: NO

iddogino (changes requested):

src/api/routes/agents.ts
```diff
@@ -100,6 +101,25 @@ import pLimit from 'p-limit'
 import * as path from 'path'
 import type { ApiAgent } from '@shared/lib/types/api'
 
+function getSkillsetAccessScope() {
```
iddogino: "Scope" is a bit misleading here — it implies authorization/permission boundaries, but what this function actually does is just partition the skillset list into "all configured" vs "accessible to current user."
A more accurate name would be something like:
- `getSkillsetLists()` — plain, says what it returns
- `getSkillsetSets()` — it's two filtered sets of the same data
- `getPartitionedSkillsets()` — describes the operation

src/api/routes/agents.ts
```diff
@@ -351,8 +371,8 @@ async function processImport(c: Context, zipBuffer: Buffer, formData: FormData)
 // Uses ?refresh=true to force a cache refresh before reading
 agents.get('/discoverable-agents', async (c) => {
   try {
-    const { skillsets } = getSettings()
-    const ssArray = skillsets || []
+    const scope = getSkillsetAccessScope()
```
iddogino: same comment - word "Scope" is confusing here

src/api/routes/agents.ts
iddogino, line 1: General file comment - use of word Scope is confusing.

src/shared/lib/services/skillset-service.test.ts
```diff
@@ -75,6 +76,15 @@ vi.mock('@shared/lib/utils/retry', () => ({
   withRetry: async (fn: () => Promise<unknown>) => fn(),
 }))
 
+const mockGetPlatformAuthStatus = vi.fn(() => ({ orgId: undefined as string | undefined }))
+vi.mock('@shared/lib/services/platform-auth-service', () => ({
+  getPlatformAuthStatus: (...args: unknown[]) => mockGetPlatformAuthStatus(...args),
```
iddogino: Check Fail: A spread argument must either have a tuple type or be passed to a rest parameter.

src/shared/lib/services/agent-template-service.test.ts
```diff
@@ -44,6 +42,15 @@ vi.mock('@shared/lib/utils/retry', () => ({
   withRetry: async (fn: () => Promise<unknown>) => fn(),
 }))
 
+const mockGetPlatformAuthStatus = vi.fn(() => ({ orgId: undefined as string | undefined }))
+vi.mock('@shared/lib/services/platform-auth-service', () => ({
+  getPlatformAuthStatus: (...args: unknown[]) => mockGetPlatformAuthStatus(...args),
```
iddogino: Check fail: A spread argument must either have a tuple type or be passed to a rest parameter.

src/shared/lib/startup.ts
```diff
@@ -26,6 +27,9 @@ export async function initializeServices() {
   // Initialize server-side analytics version
   setServerAnalyticsVersion(APP_VERSION)
 
+  // Remove platform skillsets that don't belong to the currently connected org
+  await reconcilePlatformSkillsets()
```
iddogino: Again... breaking the polymorphic design. Should be generic

src/shared/lib/types/api.ts
```diff
@@ -252,6 +251,10 @@ export interface ApiSkillsetConfig {
   skillCount: number
   agentCount: number
   addedAt: string
+  provider?: 'github' | 'platform'
+  badgeLabel?: string
+  showUrl: boolean
+  publishMode: 'pull_request' | 'hosted_submit'
```
iddogino, line 268: Why `publishMode`?

iddogino:
Round of review fixes pushed in eafd08f. What changed and why:

## 1. Per-skill network calls on hot read path
**Fix:** batched queue lookups with report-only listings.
- New `BaseSkillsetProvider.getQueueItemStatuses(ids[])` — default loops, platform overrides with one `POST /v1/skills/queue/batch` (falls back to per-id GET if the proxy returns 404, so older servers still work).
- `getAgentSkillsWithStatus` / `getAgentTemplateStatus` do a pre-pass to collect pending IDs, one batch per provider, no git ops, no metadata writes, no file copies.
- All mutation (queue transitions, file adoption, clearing `openPrUrl`, legacy-hash migration) moved into `refreshAgentSkills` / `refreshAgentTemplates`. The read path is now pure.

**Why:** prior code made N serial HTTP requests plus git fetches on every listing read; with 20 pending submissions that's 20+ round-trips on a poll that React Query fires routinely. Also eliminates write-races where the sidebar and details panel both refreshed the same metadata file concurrently.

## 2. `useAgentTemplateStatus` side-effect
**Fix:** moved the background refresh out of `queryFn` into a ref-guarded `useEffect`. Replaced `JSON.stringify(a) !== JSON.stringify(b)` with a structural `statusesEqual` (field-order-safe).

**Why:** React Query can invoke `queryFn` multiple times; firing a network side-effect inside it is an anti-pattern and could double-refresh on remount/refetch.

## 3. DELETE route's non-null assertion
Left as-is for now — the route guarantees the id exists before reaching the line. Flagged only as brittle, not broken.

## 4. `resolveCloneUrl` SSRF
**Fix:** new `src/shared/lib/utils/url-safety.ts` with `validateSafeCloneUrl` — rejects non-http(s) schemes, private/loopback/link-local/CGNAT IPs (IPv4 + IPv6 + `.local` names), and off-allowlist origins. Platform `resolveCloneUrl` now calls it with the proxy's own origin as the allowed-host prefix. 31-case unit test.

**Why:** the server returns a URL that we feed into `git clone` — defending even against a compromised/misconfigured proxy is cheap insurance against SSRF into the user's LAN / metadata endpoints.

## 5. `getSkillsetRepoDir` path sanitization
**Fix:** extended to a strict allowlist — `[^a-zA-Z0-9._-]+` → `--`, plus trim leading dots/dashes, with a `'skillset'` fallback for empty results. Protects against `..`, backslashes, null bytes, and names that resolve to `.` / `..`.

**Why:** `providerData.repoId` flows from the platform server into a filesystem path; defense-in-depth.

## 6. Merge (not replace) `providerData` in `updateSkillsetConfig`
**Fix:** `existing.providerData = { ...existing.providerData, ...nextFields }`.

**Why:** if a newer platform version adds fields the current client doesn't know about, replacing wipes them on sync.

## 7. `refreshSkillset` always `set-url`
**Fix:** removed the conditional; `git remote set-url origin $freshUrl` runs unconditionally on every refresh.

**Why:** the platform clone URL embeds a short-lived token that must be re-freshened every pull; github `set-url` with an unchanged URL is a cheap no-op. Also avoids the comparison-against-stale-ref bug flagged in the review.

## 8. Startup walk + provider encapsulation
**Fix:** this was the interesting one. Pushed `isConfigValid` / `isInstalledValid` into `BaseSkillsetProvider` so each provider owns its own "belongs to current scope?" check. Github returns `true`; platform compares `providerData.orgId` against the connected auth's orgId. Then:
- New `skillset-reconcile.ts` module does provider-polymorphic sweeps.
- `savePlatformAuth` / `revokePlatformToken` call `reconcileAfterAuthChange` on org switch (eager, via dynamic import to break the module cycle).
- `getInstalledSkillMetadata` / `getInstalledAgentMetadata` lazy-prune on read — invalid records get deleted and return `null` so the agent reverts to looking "local".
- Startup now only does the cheap in-memory settings sweep; the O(agents) filesystem walk is gone.

**Why:** the old code hard-coded `provider === 'platform'` and reached into `providerData.orgId` from outside the provider abstraction, which was exactly what this PR was trying to eliminate. New design: provider decides what's valid, reconcile is generic, startup pays only for what's cheap.

## 9. Legacy compat bridge TODO
Added `TODO(skillset-platform, remove by 2026-06-30)` to `normalizeProviderData`. Removal gate: earliest-supported client must have rolled the fields forward at least once.

## 10. `useSkillsetPublishMode` fallback
Effectively fixed by #8: when a skillset config is gone (org change), the backend returns `type: 'local'` instead of `locally_modified`, so the wrong-publish-mode submit button never renders. No UI change needed.

## 11. `gitPull` fallback visibility
**Fix:** both the `reset --hard` failure and the `git pull` fallback catches now `captureException` to Sentry with area/op tags, and log warnings to stderr.

## 12. Silent read-path mutation races
**Fix:** removed by #1 — the listing read no longer writes metadata at all, so there's nothing to race on.

## Extras not in the original review

### Zod schemas on persisted JSON
Per `CLAUDE.md`. New `src/shared/lib/types/skillset-schema.ts` with schemas for `SkillsetConfig`, `InstalledSkillMetadata`, `InstalledAgentMetadata`, `PlatformAuthSettings`. `savePlatformAuth` validates before persisting; metadata readers `safeParse` and capture schema failures to Sentry.

### Silent catches
`gitPull` fallback, queue polling (batch + per-id), platform revoke, metadata parse, template refresh — all capture to Sentry with `{ area, op }` tags and a minimal `extra` payload for correlation.

## Tests

Added:
- `platform-auth-service.test.ts`: org-switch cleanup (platform configs + installs removed, github preserved), full-disconnect cleanup, lazy-prune on stale skill read.
- `platform-provider.test.ts` (new): 14 cases covering `resolveCloneUrl` SSRF + allowed-host + schemes + unauth; batch queue + per-id fallback + empty + unauth; `isConfigValid` / `isInstalledValid`.
- `url-safety.test.ts` (new): 31 cases for private/public IPs, scheme validation, host allowlisting.
- `skillset-service.test.ts`: platform-merged queue transition (adopts remote content, clears pending id) and platform-rejected queue transition (clears pending id, preserves local modifications).

All 390 existing + new tests pass. `npx tsc --noEmit` clean. No new lint warnings in touched files.

iddogino (approved):

PR #67, opened 2026-04-04
Fix/numeric agent name crash

yiw190:
a fix for put agent name as pure numbers


iddogino (changes requested):
Need to cast to string when writing. Fix to file-storage is wrong

iddogino (approved):

PR #73, opened 2026-04-13
refactor: simplify container prompt and tool surface

yiw190:
## Summary
- make `gamut-system-prompt.md` the container's runtime prompt source so the container is driven by a single canonical prompt instead of the old `claude_code` preset-plus-append shape; keep the legacy `system-prompt.md` file in the repo unchanged, but stop treating it as the active runtime source
- clean up duplicated tool concepts in the runtime surface so the prompt and the exposed tools point to the same mental model. In particular, remove `Task` in favor of `Agent`, since they represent nearly the same subagent/delegation scope and keeping both names encourages branching in prompt wording, examples, and model behavior
- remove scheduling tools that overlap with the platform's own scheduling path. `CronCreate` / `CronDelete` / `CronList` create a second scheduling model next to the platform-managed scheduled-task flow, which makes it harder for both users and the model to understand which system is authoritative
- disallow tools that add extra workflow branches or noisy runtime behavior without being core to the coding-agent path: `Monitor` introduces streaming event behavior that changes conversation rhythm and overlaps awkwardly with background Bash jobs and scheduled triggers; `TaskOutput` is deprecated and only adds another branch in the tool surface; `EnterWorktree` / `ExitWorktree` introduce an additional isolation workflow that is heavier than the default coding path and not intended as a first-class product flow here
- tighten the remaining tool semantics in the prompt so the surviving tools are described in a way that matches actual runtime behavior. For example, `TaskStop` is kept but reframed as a kill switch for asynchronously started background jobs, rather than another generic "task" concept that could be confused with agents, todos, or scheduled tasks
- update the container build path so the new prompt source is what actually ships in the image, ensuring the runtime behavior, tool exposure, and prompt guidance move together instead of drifting apart

## Why this matters
- reduces naming splits and duplicated capability paths
- keeps the model's routing guidance aligned with the tools it can actually call
- lowers prompt noise from legacy or overlapping workflows
- makes the container closer to a one-fit-all prompt/runtime setup instead of a layered prompt plus mixed tool taxonomy

## Test plan
- [x] `docker build -t ghcr.io/skillfulagents/superagent-agent-container-base:0.3.9 ./agent-container`

### DIFF
--- a/.claude/skills/add-composio-provider/SKILL.md
+++ b/.claude/skills/add-composio-provider/SKILL.md
@@ -76,7 +76,7 @@ Add an entry to `SCOPE_DESCRIPTIONS` with end-user-friendly text for **every** s
 Add the slug to the `SERVICES` array so the agent's `search_connected_account_services` tool can discover it.
 
 ### 8. `agent-container/src/system-prompt.md`
-Add the slug to the appropriate category in the "Supported services include" line (around line 122). This is what the agent reads to know what's available without calling the search tool first.
+Add the slug to the appropriate category in the "Supported services include" line in the Requesting Connected Accounts section. This is what the agent reads to know what's available without calling the search tool first.
 
 ### 9. `scripts/download-service-icons.ts`
 Add the slug to `ALL_SLUGS`. If the Composio logos API serves the icon under a different name (test with `curl -sI https://logos.composio.dev/api/<name>`), also add a `SLUG_TO_API_NAME` mapping. Many Google services need hyphenation (`googlesheets` → `google-sheets`), but **not** all of them — `googleslides` works as-is.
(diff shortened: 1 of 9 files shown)


iddogino (changes requested):
remove Gamut references from code

iddogino:
@yiw190 do you have a diff of original CC sys prompt versus edited version?

yiw190:
Further cleanup on the system prompt.

1. Found several built-in tools that were either duplicates of what we already route through the platform, or unused on the container runtime: `Monitor`, `Cron*`, `TaskOutput`, `EnterWorktree`, `ExitWorktree`. Dropped them from the prompt and added them to `disallowedTools` in `agent-container/claude-code.ts` so the model can't reach for them. Also pruned leftover Claude Code-CLI bits that don't apply here (settings hooks, `/help`, claude-code feedback link, the Claude Code commit/PR footer).

2. Found `Task` and `Agent` overlap — same surface, `Agent` is the newer replacement and has been tested. Replaced all `Task(...)` call-site examples with `Agent(...)` and banned `Task`.

3. Rescoped `# Doing tasks` from coding-only to the full SuperAgent surface (browser, desktop, connected accounts, scheduled / webhook workflows). Coding-specific discipline moved into its own `# When writing or modifying code` section.

Also renamed `gamut-system-prompt.md` → `system-prompt-v2.md` and updated the two callers.

PR #88, opened 2026-04-17
fix: reconnect re-adds deleted platform skillsets

yiw190:
## Summary
- Extracted `syncRemoteSkillsets` from the inline `sync-remote` route handler into `skillset-service.ts`
- The existing reconnect flow (`use-platform-auth.ts`) already calls `POST /api/skillsets/sync-remote` on successful platform auth — this ensures platform skillsets that were manually deleted get re-added to settings when the user reconnects

## What was broken
After deleting platform skillsets in the UI, reconnecting agents would not bring them back because `sync-remote` was only called on platform auth success. The logic itself was correct (it re-adds missing skillsets from the remote), but it was buried in the route handler and not reusable.

## Test plan
- [ ] Connect to platform, verify skillsets appear
- [ ] Delete a platform skillset from Settings > Skillsets
- [ ] Disconnect platform, then reconnect
- [ ] Verify the deleted platform skillset is automatically re-added after reconnect

### DIFF
--- a/src/api/routes/skillsets.ts
+++ b/src/api/routes/skillsets.ts
@@ -10,7 +10,7 @@ import {
   refreshSkillset,
   getSkillsetIndex,
   removeSkillsetCache,
-  ensureSkillsetCached,
+  syncRemoteSkillsets,
 } from '@shared/lib/services/skillset-service'
 import { getSkillsetProvider } from '@shared/lib/skillset-provider'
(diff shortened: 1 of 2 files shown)


yiw190:
Closing: confirmed via debug logging that the sync-remote logic on main already works correctly. The original issue was likely a transient platform-side failure (silent auth/proxy error), not a SuperAgent bug.

PR #98, opened 2026-04-23
feat: consolidated agent connections page

westaskew:
## Summary

- New unified Agent Connections page (APIs + remote MCPs in one list) reachable from the agent home, with a "New connection" directory dialog covering both Composio APIs and MCP servers.
- Per-row actions (rename, edit permissions, discover MCP tools, test/reconnect MCP, delete) live behind a single `IntegrationRowActions` component shared across the manage and home views.
- Replaces the legacy `connected-accounts-tab.tsx` and `remote-mcps-tab.tsx` (now deleted) with `connections-tab.tsx`. Net diff: +1556 / −663 across 18 files.
- Final cleanup commit removes the dead tabs and dedupes a `safeDate` helper into `connections/utils.ts`.

## Test plan

- [ ] Open an agent → Connections section in the home renders existing OAuth + MCP connections and the empty-state illustration when none.
- [ ] Click "Manage Connections" / "Add Connection" → Agent Connections page loads; back button returns to home.
- [ ] "New connection" dialog: APIs tab connects via OAuth popup; MCPs tab adds a common server and a Custom MCP (none / bearer / OAuth auth).
- [ ] Toggle a connection on/off for the agent → row animates between "Access granted" and "Access not granted" sections; verify server state matches.
- [ ] Row actions menu: rename, edit permissions (scope/tool editor opens correctly per type), MCP test connection + reconnect, MCP discover tools, delete.
- [ ] Typecheck + lint pass.


iddogino:
Took a pass to address the review findings. Most are under-the-hood (Zod at boundaries, `Array.isArray` guards, `useCallback`-ing OAuth listeners, file renames, dropping dead props, etc.) — no need to re-approve those. Flagging the pieces that have **visible UX impact** so we're aligned before merge:

### 1. MCP status chips (`auth_required` / `error`)
New: a small pill next to the relative-date on MCP rows when `mcp.status !== 'active'`, with a tooltip carrying the server error message.

**Why:** `RemoteMcpServer.status` is already `'active' | 'error' | 'auth_required'`, and the deleted `remote-mcps-tab.tsx` rendered colored status badges. The new row dropped them, so a broken MCP looked identical to a healthy one — users had no cue to click "Reconnect". The chip restores that signal without hijacking the row layout.

### 2. Delete vs Remove — split into two actions with differentiated copy
On the agent home row, the menu now has two entries:
- **"Remove from agent"** — unassigns the connection from this agent only. Non-destructive.
- **"Delete for all agents"** — permanently deletes; destructive; red.

Each has its own confirmation modal. The Delete modal on the home view explicitly says _"If you only want this agent to lose access, cancel and pick 'Remove from agent' instead."_

**Why:** the original menu had a single "Delete" on the home that was actually a global delete (revokes access for every agent + can't be undone). It sat one click from the per-agent toggle and looked identical to the Connections page's "Remove" action, which only unassigns. High blast-radius miss: OAuth re-auth, bearer token re-entry, scope policy re-setup all lost. Clear language + two explicit actions fix that.

On the Connections page (no `agentSlug` context) the menu keeps a single "Delete" — the per-row toggle already handles per-agent access there.

### 3. Restored the X on the MCP tools dialog
Put back the standard close button in the top-right.

**Why:** a11y + consistency. The PR removed it (via `hideClose`) in favor of a refresh button, so the only way to close the dialog was Esc / outside-click. No visible close affordance breaks keyboard/SR users who expect the standard pattern, and it diverged from every other dialog in the app. I repositioned the refresh button so it doesn't overlap (`pr-14` on the content). Also deleted the `hideClose` prop on `DialogContent` entirely — it was single-use and not a safe pattern to leave lying around.

### 4. Spinner replaces the Switch during pending mutations
When a toggle mutation is in-flight the row shows a spinner instead of a disabled Switch-plus-spinner pair.

**Why:** the original rendered the Switch + a separate `<Loader2>` next to it — visually noisy and the switch could still be clicked in a rapid on→off→on sequence, which fired racing `assign`/`remove` mutations whose final server state could diverge from the optimistic UI. Replacing the control entirely while pending prevents the race and reads cleaner.

### 5. "Back to Agent Home" → "Back"
Shorter label, accurate semantics.

**Why:** `selectConnections(false)` doesn't reset session / scheduledTask / chatIntegration, so the button sometimes returned to the same sub-view you came from, not agent home. "Back" is honest about that. If we ever want "Back to Agent Home" to literally return to home, it needs to reset those selections — happy to do that instead if preferred.

### 6. Home Connections layout — now shares `IntegrationRow` + `IntegrationRowActions`
Visual end-state matches main (divided rows, no outer card) — the shared component was the right home for the logic, but I initially wrapped it in `IntegrationList` which adds the bordered card (correct for the Connections page, wrong for the home where we're already inside `HomeCollapsible`). Caught that and reverted to a bare `divide-y` wrapper — thanks @iddog for spotting on the screenshot.

**Why consolidate at all:** the PR description said `IntegrationRowActions` was shared between manage and home, but the home still had its own hand-rolled Popover + AlertDialog (`ConnectionRowItem`). The two had already diverged on behavior ("Remove" vs "Delete", "Configure" vs "Edit permissions"/"Rename"/"Discover tools"). Single source of truth avoids that drift.

---

Full list of smaller changes (non-visual): MCP OAuth completion now actually waits for `mcp-oauth-callback` before closing the dialog (extracted into `useMcpOAuthListener`), Zod-validated MCP add form with `https:` URL check, `location: 'connections_tab'` for analytics on new connections, `SettingsPageContainer`/`PageTitle` layout primitives extracted for future settings pages, `startViewTransition` helper, and `Array.isArray` guards on all API-sourced lists per the project rule.

PR #104, opened 2026-04-24
style: promote API Logs to a standalone agent page

westaskew:
## Summary
- Move the API Logs view out of the agent-settings dialog into a standalone page, routed via a new `selectedApiLogs` selection state and opened from the agent-home "API Logs" extras button.
- Lay it out in the shell pattern used by the recent crons ([#100](https://github.com/SkillfulAgents/SuperAgent/pull/100)) and connections ([#98](https://github.com/SkillfulAgents/SuperAgent/pull/98)) pages: centered `max-w-5xl`, "Back" link, `text-xl font-medium` title, rounded empty-state card.
- Render entries with the new shadcn `Table` primitive (columns: Timestamp, Duration, Source, Method, Status, Policy, Toolkit, Path/error). Errors are appended in red on the same line as the path. A More/Less button overlays the right edge of the Path cell on row hover (Less stays pinned while expanded) and toggles the inline details row.

## Test plan
- [ ] Open an agent → click **API Logs** in the extras section → the standalone page opens, breadcrumb reads `<agent> / API Logs`, "Back" returns to agent home.
- [ ] Confirm the **API Log** tab no longer appears in the agent Settings dialog.
- [ ] With logged requests, verify table renders, sort/timestamp format `MM/dd/yy, HH:mm:ss`, 15 rows per page, pagination works past page 1.
- [ ] Hover a row → "More ⌄" button appears on the right of the path cell. Click it → row expands, button becomes "Less ⌃" and stays visible without hover.
- [ ] Row with an `errorMessage`: confirm the error renders in red parentheses after the path, truncated with the rest.
- [ ] Refresh button refetches current page; loading and empty states render correctly.


iddogino:
@iddogino 2 major issues:
1) For table -- we should default to the table hook rather than custom building
2) Selection context boolean is growing too large. Make single enum.

PR #106, opened 2026-04-27
feat(auth): platform OIDC login + multi-user auth-mode

yiw190:
## Summary

Part A of the SuperAgent platform integration split.

This PR covers platform OIDC login and the auth-mode bootstrap needed for managed multi-user deployments.

## What is in this PR

- Better Auth generic OIDC provider config from `AUTH_PROVIDERS_JSON`
- `/api/auth-config` exposure of public auth providers
- Renderer login flow updates (`auth-page`, `auth-client`, test setup)
- Auth-mode session handling and the `requireAdminApproval && !isAuthMode()` fix
- Env-managed platform bootstrap:
  - `platform-auth-service` env fallback for `PLATFORM_TOKEN`
  - env-managed platform auth status / `source`
  - wizard bootstrap behavior for managed installs
  - hiding the legacy `Platform` tab in auth-mode

## What is intentionally out of scope

These stay in Part B (`feat/user-attribution` / #107):

- outbound member attribution
- `X-Platform-Member-Id` propagation
- request-scoped attribution / AsyncLocalStorage plumbing
- multi-tenant Composio, webhook, and proxy client attribution behavior

## Why split this way

This PR should stand on its own as:

- OAuth / login support
- auth-mode managed deployment bootstrap

without pulling in proxy attribution behavior.

## Test plan

- [ ] `npm run typecheck`
- [ ] `npm run test`
- [ ] OIDC login works end-to-end in `AUTH_MODE=true`
- [ ] managed installs with env-provisioned `PLATFORM_TOKEN` bootstrap correctly
- [ ] local password auth still works in non-auth-mode
- [ ] OIDC users do not hit the local pending-admin-approval flow


src/shared/lib/auth/index.ts
```diff
@@ -98,13 +107,14 @@ export function getAuth() {
...
-                // needing to recreate the Better Auth singleton.
+                // Auto-ban-pending-approval is only meaningful for standalone
+                // installs where SuperAgent owns the user list. In auth-mode
+                // the platform's `subscribed_member` table is the gatekeeper
+                // -- if the user reached this hook they were already vouched
+                // by platform OIDC, and there's no local admin to "approve".
                 const currentSettings = getSettings()
                 const currentAuth = { ...DEFAULT_AUTH_SETTINGS, ...currentSettings.auth }
-                if (result.changes === 0 && currentAuth.requireAdminApproval) {
+                if (result.changes === 0 && currentAuth.requireAdminApproval && !isAuthMode()) {
```
iddogino: @yiw190 this change is out of-scope for this feature. Users can achieve that today by enabling platform and disabling `requireAdminApproval`. There might still be an (edge) case for wanting to require admin approval even on platform.
Please remove this from this PR

src/shared/lib/auth/index.ts
```diff
@@ -33,6 +35,12 @@ export function getAuth() {
     const trustedOrigins = getTrustedOrigins()
     const settings = getSettings()
     const authSettings = { ...DEFAULT_AUTH_SETTINGS, ...settings.auth }
+    const oauthProviders = getGenericOAuthProviderConfigs()
```
iddogino, line 37: Currently, the code loads this env data lazily on first call. The issue is that if it is malformed, we don't learn about it soon enough. We should do it on app start -- `src/shared/lib/auth/startup-validation.ts:15 (validateAuthModeStartup)`. 

Drop in something like:                                                                                                                                                                                                                              
```                                                                                                                                                                                                                                                       
  // startup-validation.ts                                                                                                                                                                                                                             
  export async function validateAuthModeStartup(): Promise<void> {                                                                                                                                                                                     
    // ...existing user-table / agents-dir checks...                                                                                                                                                                                                   
                                                                                                                                                                                                                                                       
    // Validate provider config eagerly so misconfiguration surfaces at boot,
    // not on the first /api/auth/oauth2 request.                                                                                                                                                                                                      
    validateAuthProviders()
  }

  function validateAuthProviders(): void {
    const providers = getPublicAuthProviders() // already runs the env parse + readiness
    for (const p of providers) {
      if (!p.readiness.ok) {
        console.warn(`[auth] provider ${p.id} is unavailable: ${p.readiness.reasons.join(', ')}`)
      }                                                                                                                                                                                                                                                
    }
    console.log(`[auth] loaded ${providers.length} OIDC provider(s): ${providers.map(p => p.id).join(', ') || '(none)'}`)                                                                                                                              
  }                                                                                                                                                                                                                                                    
```  
  You also get the warm-up for free if you want it: add getAuth() at the end of validateAuthModeStartup and the Better Auth singleton (with the OAuth plugin) is constructed before the first request, so the first sign-in doesn't pay the init cost.

src/shared/lib/auth/provider-config.ts
```diff
@@ -0,0 +1,170 @@
...
+
+function readEnv(name: string): string | undefined {
+  const value = process.env[name]?.trim()
+  return value ? value : undefined
+}
+
+// TODO: Keep OIDC providers deployment-managed for now. If we ever expose
+// admin-managed provider config, prefer deployment-time tooling over storing
+// provider secrets in app settings.
+function resolveEnvAuthProviders(): AuthProviderSettings[] {
```
iddogino, line 146: Add ZOD schema + validation here for the data read from env var

src/shared/lib/auth/provider-config.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  toGenericOAuthConfig(): GenericOAuthProviderConfig | null {
+    const readiness = this.getReadiness()
+    if (!readiness.ok) return null
+    return {
+      providerId: this.config.id,
+      discoveryUrl: this.config.discoveryUrl?.trim() || undefined,
+      issuer: this.config.issuer?.trim() || undefined,
+      clientId: this.config.clientId!.trim(),
```
iddogino: Safe because getReadiness() was checked, but coupling the safety to a side check is fragile. Just narrow with a guard inside toGenericOAuthConfig and return null.

src/shared/lib/auth/provider-config.ts
```diff
@@ -0,0 +1,170 @@
...
+
+function getEnabledProviderDefinitions(
+  providers: AuthProviderSettings[] = resolveEnvAuthProviders(),
+): AuthProviderDefinition[] {
+  return providers
+    .filter((provider) => provider.enabled !== false)
+    .map(createProviderDefinition)
+}
+
+export function getGenericOAuthProviderConfigs(): GenericOAuthProviderConfig[] {
```
iddogino, line 182: Each getGenericOAuthProviderConfigs() (and each getPublicAuthProviders()) re-resolves and re-validates env. In the auth-config request path that's per-request. Should be memoized

src/shared/lib/services/platform-auth-service.ts
```diff
@@ -69,32 +73,70 @@ async function reconcileAfterAuthChange(): Promise<void> {
   }
 }
 
+function decodeOrgIdFromToken(token: string): string | null {
```
iddogino: Add console warning + Sentry event if JWT validation fails

src/shared/lib/services/platform-auth-service.ts
```diff
@@ -69,32 +73,70 @@ async function reconcileAfterAuthChange(): Promise<void> {
...
+    const parsed = JSON.parse(json) as { orgId?: unknown }
+    return typeof parsed.orgId === 'string' && parsed.orgId.length > 0 ? parsed.orgId : null
+  } catch {
+    return null
+  }
+}
+
+function getEnvManagedStatus(): PlatformAuthStatus | null {
+  if (!isAuthMode()) return null
+  const envToken = process.env.PLATFORM_TOKEN?.trim()
```
iddogino, line 197: should also jwt validate here

src/renderer/components/auth/auth-page.tsx
```diff
@@ -335,18 +350,77 @@ function SignUpForm({ onSwitchToSignIn, config, onPendingApproval }: { onSwitchT
 export function AuthPage({ onPendingApproval }: { onPendingApproval?: (pending?: boolean) => void } = {}) {
   const config = useAuthConfig()
   const [tab, setTab] = useState<string>('signin')
+  const [providerError, setProviderError] = useState<string | null>(null)
```
iddogino, line 363: 1 - should show loading when fetching auth configs
2 - should surface errors in loading configs properly

src/renderer/components/auth/auth-page.tsx
```diff
@@ -18,16 +18,31 @@ interface AuthConfig {
...
   allowLocalAuth: boolean
   allowSocialAuth: boolean
+  providers: AuthProviderConfig[]
   passwordMinLength: number
   passwordRequireComplexity: boolean
   requireAdminApproval: boolean
   hasUsers: boolean
 }
 
+interface AuthProviderConfig {
```
iddogino: AuthProviderConfig is a hand-typed copy of PublicAuthProviderConfig. Import the shared type from provider-config.ts (or export to a renderer-safe module) to keep them in sync.

src/renderer/components/auth/auth-page.tsx
```diff
@@ -335,18 +350,77 @@ function SignUpForm({ onSwitchToSignIn, config, onPendingApproval }: { onSwitchT
...
+        <CardContent className="space-y-4">
+          {hasProviders && (
+            <div className="space-y-2" data-testid="auth-providers">
+              {config.providers.map((provider) => (
+                <Button
+                  key={provider.id}
+                  type="button"
+                  variant="outline"
+                  className="w-full"
+                  disabled={!provider.available || providerLoadingId !== null}
```
iddogino: Show the spinner only on the clicked one and disabling just that button (less jarring).

src/renderer/components/settings/global-settings-dialog.tsx
```diff
@@ -53,9 +53,11 @@ export function GlobalSettingsDialog({
       <SettingsDialogTab id="notifications" label="Notifications" icon={<Bell className="h-4 w-4" />}>
         <NotificationsTab />
       </SettingsDialogTab>
-      <SettingsDialogTab id="platform" label="Platform" icon={<Waypoints className="h-4 w-4" />}>
-        <PlatformTab />
-      </SettingsDialogTab>
+      {!isAuthMode && (
```
iddogino: Don't hide in platform auth mode... Make read only

src/renderer/hooks/use-platform-auth.ts
```diff
@@ -7,6 +7,9 @@ import { prepareOAuthPopup } from '@renderer/lib/oauth-popup'
 
 export const PLATFORM_AUTH_CHOICE_STORAGE_KEY = 'superagent-auth-choice'
 
+// `env` means the active token came from an org-provisioned `PLATFORM_TOKEN`.
+export type PlatformAuthSource = 'settings' | 'env' | null
```
iddogino: PlatformAuthSource declared in two files.                                                                                                                                                                                  
Same type lives in platform-auth-service.ts:9. Re-export from one place.

iddogino (changes requested):
Also testing gaps:

  provider-config.test.ts:98-100 — misleading test name.                                                                                                                                                                                               
  "does not expose bundled providers before their activation env is set" — there's no concept of bundled providers in the code. The test asserts an empty array when env is unset. Rename.
                                                                                                                                                                                                                                                       
  Coverage gaps:  
  - No test for enabled: false filtering.                                                                                                                                                                                                              
  - No test for a malformed entry in an otherwise-valid array (does the bad one get dropped while good ones survive?).
  - No test for the public shape not leaking clientSecret (the whole point of the public/generic split).              
                                                                                                                                                                                                                                                       
  src/shared/lib/services/platform-auth-service.test.ts                                                                                                                                                                                                
                                                                                                                                                                                                                                                       
  platform-auth-service.test.ts:32-49 — env cleanup inline, not in afterEach.                                                                                                                                                                          
  If the assertions throw, delete process.env.PLATFORM_TOKEN and AUTH_MODE never run, leaking into the next test in the file. Move env teardown to afterEach.
                                                                                                                                                                                                                                                       
  Coverage gaps:                                                                                                                                                                                                                                       
  - No test for getEnvManagedStatus returning source: 'env' and decoded orgId.                                                                                                                                                                         
  - No test for decodeOrgIdFromToken (JWT vs opaque vs malformed payload).                                                                                                                                                                             
  - No test for "settings record wins over env" precedence — currently the surprising behavior is undocumented and untested.
                                                                                                                                                                                                                                                       
  src/renderer/components/auth/auth-page.test.tsx                                                                                                                                                                                                      
                                                                                                                                                                                                                                                       
  Coverage gaps:                                                                                                                                                                                                                                       
  - No test for an unavailable provider being rendered as disabled (available: false).                                                                                                                                                                 
  - No test for the providers fetch failing — does the form crash, render local auth, or stay blank?                                                                                                                                                   
  - No test for the OIDC-only + first-user case.

src/shared/lib/auth/index.ts
yiw190: removed

src/renderer/components/auth/auth-page.tsx
yiw190: imported from @shared/lib/auth/provider-config now

src/renderer/hooks/use-platform-auth.ts
yiw190: removed duplication

src/shared/lib/auth/provider-config.ts
yiw190: Added Zod

iddogino: @yiw190 why not just make the zod schema the whole array? Would save you from the custom loop here...

src/shared/lib/auth/index.ts
iddogino: @yiw190 addressed?

src/shared/lib/services/platform-auth-service.ts
iddogino: @yiw190 where's the valuation happening?

src/shared/lib/auth/index.ts
yiw190: yes, this was addressed in startup-validation.ts

https://github.com/SkillfulAgents/SuperAgent/pull/106/changes#diff-a253ad1a9aa53c250dccec4230d55f163a47a62b4a02f836c85085d977b1009b

src/shared/lib/services/platform-auth-service.ts
yiw190: Now it is in verifyOrgAccessTokenSigned from platform-auth-service.ts

https://github.com/SkillfulAgents/SuperAgent/pull/106/changes#diff-9f609efb6138e510f413fd1a44a3f8f3f25ce7d4964ca2536cf59f70d001a1df

yiw190:
Hi Iddo, addressed comments. Added zod schema for the whole array, and also replying on the comments to point at the other two changes.

`src/shared/lib/auth/startup-validation.ts` is the change for app start.

`src/shared/lib/services/platform-auth-service.ts` has the verification. Just noticed `verifyOidcJwt` can be totally separate, so added one file `src/shared/lib/auth/oidc-jwt.ts` for the general JWT resolver but kept platform consts + the orgId check in `platform-auth-service.ts`.

iddogino (changes requested):

PR #110, opened 2026-04-29
add triggers warning message

yiw190:


src/renderer/components/webhook-triggers/webhook-trigger-view.tsx
```diff
@@ -171,6 +177,26 @@ export function WebhookTriggerView({ triggerId, agentSlug }: WebhookTriggerViewP
         </div>
       </div>
 
+      {/* Warning when using personal Composio key */}
+      {hasLocalComposioKey && isActive && (
+        <div className="mx-6 mt-4 flex items-start gap-2 rounded-md border border-amber-200 bg-amber-50 px-3 py-2 dark:border-amber-900 dark:bg-amber-950/30">
```
iddogino: @yiw190 let's use standard alert component -- https://ui.shadcn.com/docs/components/radix/alert

iddogino (changes requested):

src/renderer/components/webhook-triggers/webhook-trigger-view.tsx
yiw190: sure!

PR #115, opened 2026-04-30
fix: auto-name Untitled agents from initial prompt (server-side)

yiw190:
## Summary
- Fix `/api/agents/generate-name` so it is handled before the agent slug middleware.
- Restore the previous client-side Untitled agent rename flow.
- Add a regression test for the route ordering bug.

## Test plan
- `npm run test:run -- src/api/routes/agents.test.ts`


iddogino:
@yiw190 I am not sure I understand from this PR's description what it actually solves for. Can you please provide a description of the bug / steps to reproduce?

PR #141, opened 2026-05-13
feat(.claude/skills): add claude-prompt-drift skill

yiw190:
## Summary

Adds a `.claude/skills/claude-prompt-drift/` skill that snapshots and diffs the `/v1/messages` request body that `claude-agent-sdk` emits, at two independent layers:

| Axis          | What it captures                                                                              |
| ------------- | --------------------------------------------------------------------------------------------- |
| `pure-claude` | The bare `{ type: 'preset', preset: 'claude_code' }` baseline — no MCP, no append, no custom tools |
| `superagent`  | What `agent-container` actually puts on the wire (skills, MCP, claudeMd, etc.)                |

Cross-version diff on each axis catches two separate kinds of drift:

- **`pure-claude` drift** → Anthropic silently changed the Claude Code preset (new skill_listing entry, new `<system-reminder>`, tool description rewritten, etc.).
- **`superagent` drift** → our overlay changed (added/removed an MCP server, edited `claudeMd`, bumped SDK without realizing it touched something else).

The diff *between* the two axes inside a single snapshot is our **known modifications** — intentional, not noise. This skill does not try to subtract them; it just tracks each axis's evolution over time.

## How it works

`capture.sh` is a one-shot orchestrator that:

1. Reads the pinned SDK version from `agent-container/package-lock.json`.
2. Builds (or reuses) the `superagent-container` image.
3. Spins up a private Docker network with three containers:
   - `proxy-*`: `node:20` running `proxy.mjs`, a pass-through HTTP proxy that dumps `body.system` / `body.messages` / `body.tools` as Markdown on first sight of each model, then forwards to `api.anthropic.com`.
   - `superagent-*`: the agent-container image, pointed at the proxy via `ANTHROPIC_BASE_URL`.
   - `baseline-*`: the **same** agent-container image, but with CMD overridden to run `pure-baseline/run.mjs` (a bare `query({ preset: 'claude_code' })` driver). Reusing this image is critical — it has the right `claude` native binary on PATH (glibc, not musl) so the SDK's spawned subprocess actually issues the wire request.
4. Fires one model call per axis.
5. Tears everything down (trap-based cleanup, verified).

Output lands in `snapshots/<sdk-version>/{pure-claude,superagent}/<model>/{system,messages,tools}.md` plus a `meta.json` recording SDK version, model, SuperAgent commit, capture timestamp.

`diff.sh <old> <new>` runs `diff -ruN` per axis (excluding `raw.json`) and exits non-zero if drift is detected — CI-friendly.

## Why this design

Two non-obvious traps that bit during testing and informed the implementation:

1. **Alpine/musl vs glibc**: my first attempt ran the baseline driver in `node:20-alpine` with `npm install`. The SDK's spawned `claude` subprocess immediately failed because `agent-container`'s own Dockerfile explicitly strips the musl variant of the native binary; the muscle/glibc mismatch makes the binary unusable. Fix: use the agent-container image itself as the runtime for both axes.
2. **Local API-key validation**: current `claude-agent-sdk@0.2.118` does light local key validation before issuing the request, so dummy keys + a synthetic install path never reach the proxy. Inside the agent-container image this path doesn't bite — the SDK proceeds and the proxy captures the body.

## What's included

- `SKILL.md` — frontmatter + full usage docs (matches the `update-claude-deps` neighbor style).
- `capture.sh`, `diff.sh` — orchestrator + diff (both `bash -n` clean, `--help` flags).
- `proxy.mjs` — pass-through capture proxy.
- `pure-baseline/run.mjs` — bare-preset SDK driver.
- `snapshots/0.2.118/` — a captured baseline for the currently-pinned SDK so the very next SDK bump has something to diff against. `raw.json` and `.seen-models.json` are `.gitignore`d (large / proxy-internal).
- `snapshots/README.md` — explains the snapshot layout and commit policy.

## Test plan

- [x] Fresh end-to-end run: `rm -rf snapshots/0.2.118 && ./capture.sh --axis both` — exit 0 in ~66s, both axes produce non-empty captures (pure-claude system 245 lines, superagent system 725 lines; tools 1871 vs 2952; messages 51 each).
- [x] `./diff.sh 0.2.118 0.2.118` reports no drift (sanity).
- [x] Cleanup trap: no leftover proxy / baseline / superagent containers; port 9876 free; private network removed.
- [x] `.gitignore` correctly excludes `raw.json` and `.seen-models.json` from the staged tree.
- [ ] Reviewer: please skim `snapshots/0.2.118/{pure-claude,superagent}/claude-opus-4-7/system.md` to confirm the captured content looks correct.

## Notes

- Skill is co-located with other SuperAgent-engineering skills (`update-claude-deps`, `update-models`, etc.) under `.claude/skills/`, not under the broader DatawizzAI/skills registry, because it is tightly tied to agent-container's Dockerfile and SDK pinning.
- No production code is touched. Adding files only.

### DIFF
--- a/.claude/skills/claude-prompt-drift/.gitignore
+++ b/.claude/skills/claude-prompt-drift/.gitignore
@@ -1,0 +1,5 @@
+pure-baseline/node_modules/
+pure-baseline/package-lock.json
+pure-baseline/package.json
+snapshots/**/.seen-models.json
+snapshots/**/raw.json
--- a/.claude/skills/claude-prompt-drift/SKILL.md
+++ b/.claude/skills/claude-prompt-drift/SKILL.md
@@ -1,0 +1,97 @@
+---
+description: Detect drift in the Claude system prompt that agent-container sends to Anthropic. Captures two axes — the bare `claude_code` preset (Anthropic baseline) and agent-container's overlaid wire request — at the SDK version pinned in `agent-container/package-lock.json`, then stores versioned snapshots under `snapshots/<sdk-version>/`. Use when bumping `@anthropic-ai/claude-agent-sdk`, debugging unexpected model behavior after an SDK upgrade, or auditing what Anthropic silently changed in the Claude Code preset.
+---
+
+# Claude Prompt Drift Check
+
+Container-orchestrated capture + diff of the `/v1/messages` request body that the claude-agent-sdk emits, at two layers:
+
+| Axis          | What it captures                                                                              |
+| ------------- | --------------------------------------------------------------------------------------------- |
+| `pure-claude` | The bare `{ type: 'preset', preset: 'claude_code' }` baseline — no MCP, no append, no custom tools |
+| `superagent`  | What `agent-container` actually puts on the wire (skills, MCP, claudeMd, etc.)                |
+
+The diff *between* the two axes in a single snapshot is **our known modifications** — intentional, not interesting. The interesting signal is each axis's drift *across SDK versions*:
+
+- `pure-claude` drift → Anthropic changed the preset (new skill_listing entry, new system-reminder, tool description rewritten).
+- `superagent` drift → our overlay changed (we added/removed an MCP server, edited claudeMd, bumped SDK).
+
+## When to use
+
+- After bumping `@anthropic-ai/claude-agent-sdk` in `agent-container/package.json`.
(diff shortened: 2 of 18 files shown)


yiw190:
Reopening on a clean branch off main — this one accidentally carried the browser stable-ref commit from exp/agent_broswer_change.

PR #146, opened 2026-05-15
feat(skillset): platform fallback to HTTPS zipball when git unavailable

yiw190:
## Summary
When `provider: 'platform'` and a local git CLI is not available, fall back to HTTPS tar.gz archive download instead of failing with "Git is not installed."

Pairs with platform PR https://github.com/SkillfulAgents/platform/pull/7 which adds the server-side `/v1/skills/archive` endpoint (returns `application/gzip` / tar.gz, not zip).

## What changed
`PlatformSkillsetProvider`:
- Probes git CLI availability once at module load (`execFileSync('git', ['--version'])`); result cached in `cachedGitAvailable`
- `usesGitCache` is now conditional: `true` when git is available (preserves existing fast incremental `git pull` refreshes), `false` only as fallback
- New `populateCache`: downloads + extracts the proxy tar.gz with bearer auth, `withRetry` 3x, maps `404` / `401-403` / `5xx` to friendly errors, guards against `__MACOSX` entries and path-traversal entries
- New `refreshCache`: re-downloads to tmp dir and atomically swaps
- New `isCacheReady`: checks `.skillset-cache-meta.json` (matches the marker the public provider already uses for non-git caches from SUP-168)
- `setPlatformGitAvailabilityForTesting` exported for deterministic tests

## Test plan
- [x] 9 new unit tests covering happy path, auth missing, 404/401/403/5xx mapping, SSRF (proxy on private IP), `__MACOSX` + path-traversal stripping, refresh atomic swap, and both git-available / git-missing branches
- [x] 314 existing skillset tests still green
- [x] `npm run lint` clean, `tsc --noEmit` clean
- [x] End-to-end against staging proxy:
  - Launched SuperAgent with a fake git shim (`PATH=/tmp/fakegit:\$PATH`) so the probe returns false
  - Installed a platform skillset via the UI
  - New cache directory \`~/Library/Application Support/Superagent-Dev/skillset-cache/skillsets--<orgId>--<skillset>/\` contained \`.skillset-cache-meta.json\` with \`{\"provider\":\"platform\",...}\` and **no \`.git/\`**

## Notes
- The git path is unchanged — existing \`resolveCloneUrl\` + \`/v1/skills/git-url\` flow still runs when git is available.
- \`publishUpdate\` already goes through \`POST /v1/skills/submit-update\` and never touches the local \`.git/\`, so editing + publishing works in both modes without any changes.
- Refresh in archive mode is a full re-download (vs. incremental \`git pull\` in git mode). Acceptable for current skillset sizes (KB).
- The archive fallback always pulls the skillset's default branch. The server's \`?ref=\` query param is **not** exposed in this mode — git mode is the only way to install a non-default ref. This is acceptable today because the UI has no ref-selection surface; if/when that changes we can plumb \`ref\` through \`SkillsetProviderRef\`.


yiw190:
Follow-up commit: paired with platform#7 — switched from `/v1/skills/zipball` (adm-zip) to `/v1/skills/archive` (tar.gz). Client decoder now uses the standard `tar` package. Path-traversal protection migrated from manual `destPath.startsWith` to `tar.x` strict mode + filter. All 26 unit tests pass with tar.c-built tar.gz fixtures. End-to-end retested via curl against redeployed staging proxy.

PR #159, opened 2026-05-21
fix(triggers): scope webhook poll to this host's active trigger ids

yiw190:
## Bug

Desktop + cloud SuperAgent instances (or any two SuperAgent processes sharing the same platform `member_id`) racing on the proxy's `/v1/webhook-events/poll`. The winner claimed and acked every pending event for the member, including ones whose `composio_trigger_id` was only registered locally on the other host. Loser host's triggers silently never fired.

## Repro (pre-fix)

1. Two SuperAgent hosts authenticating against the same platform member (e.g. two opaque keys with the same `access_key.member_id`, or one opaque key + one cloud org-JWT `::sub_X`).
2. Register `ti_A` only on host A, `ti_B` only on host B.
3. Both hosts run `TriggerManager.pollAndProcess` near-simultaneously.
4. Winner claims both events. The trigger lookup for the foreign `ti_*` fails locally → event is acked and dropped.

### `webhook-trigger-service.ts`

New `getActiveComposioTriggerIds(): string[]` — a simple `SELECT composio_trigger_id FROM webhook_triggers WHERE status='active' AND composio_trigger_id IS NOT NULL`. **No join on `account` / `connected_accounts`, no per-member filter** — the access key (opaque) or org JWT (cloud) already binds the host to its member at the proxy, and host-local triggers are exactly the ones this host can route. Crucially, this works in opaque-mode hosts that have **no platform `account` row** in local SQLite (AUTH_MODE=false), which an earlier per-member filter would have broken.

### `webhook-events-client.ts`

`pollAndClaimEvents(memberId)` now always sends `{ trigger_ids: getActiveComposioTriggerIds() }`. No token-shape branching — the race is on `member_id`, not on bearer type, so the fix must apply uniformly.

`TriggerManager` is **untouched** — same signature, same iteration pattern.

## Backward compatibility

`memberId` parameter is unchanged. Old proxy builds that ignore the body still resolve `target_member_id` from `authCtx` and claim all pending rows; new proxy uses the `trigger_ids` filter. Mixed-mode rollout works without coordination.

## Test plan

- [x] **`webhook-events-client.test.ts`** (new): 3 cases — org JWT mode sends `trigger_ids` + `::memberId` bearer, opaque mode sends `trigger_ids` without `::` suffix, empty trigger list still sent
- [x] **`webhook-trigger-service.test.ts`**: new case verifies `getActiveComposioTriggerIds()` returns only active rows with set composio id, skips paused and unset
- [x] **Local manual**: started Electron dev with `[webhook-events-client] POST /poll … trigger_ids=[\"ti__mM9RQdq_jJB\"]` confirmed; on receiving 2 pending events from staging, both processed (`[TriggerManager] Processing 2 event(s) for member local`)
- [x] **Staging live**: with this client + new proxy deployed, two hosts on the same member claimed disjoint rows per the proxy log `event=webhook_poll claimed=1 trigger_ids_count=…` pattern (see paired platform PR)

## Self-check

- [x] No type-switching on provider / token shape
- [x] No type declared in two places
- [x] No silent catches added
- [x] Tests: env teardown in \`afterEach\`, edge cases covered (paused / unset composio id)
- [x] \`typecheck\` clean, vitest green
- [x] Reproduce-before-fix done — race observed in staging logs prior to fix
- [x] No multi-line JSDoc on internal helpers

## Out of scope

- Mixed-mode race window (one host upgraded, one not) is accepted per platform PR rollout.
- Per-member \`trigger_ids\` narrowing (cloud multi-member host) is a future optimization; current "send all local active" is correct, just slightly wasteful.

### DIFF
--- a/src/shared/lib/services/webhook-events-client.test.ts
+++ b/src/shared/lib/services/webhook-events-client.test.ts
@@ -1,0 +1,77 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const mockGetPlatformAccessToken = vi.fn()
+const mockDecodeOrgIdFromToken = vi.fn()
+const mockGetActiveComposioTriggerIds = vi.fn()
+const mockFetch = vi.fn()
+let originalFetch: typeof globalThis.fetch
+
+vi.mock('@shared/lib/services/platform-auth-service', () => ({
+  getPlatformAccessToken: () => mockGetPlatformAccessToken(),
+}))
+
+vi.mock('@shared/lib/platform-attribution', () => ({
+  decodeOrgIdFromToken: (token: string) => mockDecodeOrgIdFromToken(token),
+}))
+
+vi.mock('@shared/lib/services/webhook-trigger-service', () => ({
+  getActiveComposioTriggerIds: () => mockGetActiveComposioTriggerIds(),
+}))
+
+vi.mock('@shared/lib/platform-auth/config', () => ({
+  getPlatformProxyBaseUrl: () => 'https://proxy.test',
+}))
+
+import { pollAndClaimEvents } from './webhook-events-client'
+
+describe('webhook-events-client', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    originalFetch = globalThis.fetch
+    mockGetPlatformAccessToken.mockReturnValue('token-value')
+    mockFetch.mockResolvedValue(
+      new Response(JSON.stringify({ events: [], realtime: null }), { status: 200 })
+    )
+    globalThis.fetch = mockFetch
+  })
+
+  afterEach(() => {
+    globalThis.fetch = originalFetch
+  })
+
+  it('sends local active trigger_ids in org JWT mode with ::memberId bearer', async () => {
+    mockDecodeOrgIdFromToken.mockReturnValue('org_1')
+    mockGetActiveComposioTriggerIds.mockReturnValue(['ti_host_a', 'ti_host_b'])
+
+    await pollAndClaimEvents('sub_member_1')
+
+    const [, init] = mockFetch.mock.calls[0]
+    const body = JSON.parse((init as RequestInit).body as string)
+    expect(body.trigger_ids).toEqual(['ti_host_a', 'ti_host_b'])
+    const headers = (init as RequestInit).headers as Headers
+    expect(headers.get('Authorization')).toBe('Bearer token-value::sub_member_1')
+  })
+
+  it('sends local active trigger_ids in opaque key mode (no ::memberId suffix)', async () => {
+    mockDecodeOrgIdFromToken.mockReturnValue(null)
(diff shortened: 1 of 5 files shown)


src/shared/lib/services/webhook-trigger-service.ts
```diff
@@ -56,6 +53,22 @@ export function getDistinctPlatformMemberIdsForActiveTriggers(): string[] {
...
+    .from(webhookTriggers)
+    .where(
+      and(
+        eq(webhookTriggers.status, 'active'),
+        isNotNull(webhookTriggers.composioTriggerId),
+      ),
+    )
+    .all()
+    .map((r) => r.composioTriggerId!)
+    .filter(Boolean)
```
iddogino: the isNotNull SQL clause already guarantees no nulls, and the ! assertion on line 68 acknowledges that. The trailing .filter(Boolean) is dead code, no?

src/shared/lib/services/webhook-events-client.test.ts
```diff
@@ -0,0 +1,71 @@
...
+import { pollAndClaimEvents } from './webhook-events-client'
+
+describe('webhook-events-client', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    mockGetPlatformAccessToken.mockReturnValue('token-value')
+    mockFetch.mockResolvedValue(
+      new Response(JSON.stringify({ events: [], realtime: null }), { status: 200 })
+    )
+    globalThis.fetch = mockFetch
```
iddogino, line 35: globalThis.fetch = mockFetch in beforeEach but there's no afterEach restoring the original fetch. If other test suites run in the same Vitest worker after this file, they'll inherit the mock.

 Suggested fix:
```
  let originalFetch: typeof fetch
  beforeEach(() => { 
    originalFetch = globalThis.fetch
    // ...
    globalThis.fetch = mockFetch
  })
  afterEach(() => {
    globalThis.fetch = originalFetch
  })
```

src/shared/lib/scheduler/trigger-manager.ts
```diff
@@ -132,6 +132,10 @@ class TriggerManager {
         memberIds = ['local']
       }
 
+      if (memberIds.length > 0) {
+        console.log(`[TriggerManager] poll memberIds=${JSON.stringify(memberIds)}`)
```
iddogino: Seems very noisy, no ?

iddogino (changes requested):

src/shared/lib/scheduler/trigger-manager.ts
yiw190: agreed, this one should be deleted

src/shared/lib/services/webhook-trigger-service.ts
yiw190: cleaned

PR #170, opened 2026-05-26
clean: skill install dialog

yiw190:


src/shared/lib/container/base-container-client.ts
iddogino, line 1: @yiw190 what is this??

iddogino (changes requested):

PR #171, opened 2026-05-26
Redesign global Connections settings + add detail page

westaskew:
## Summary
- Restyles the global **Connections** settings tab: groups the two default policies into one bordered card with compact toggles + clearer fallback copy, sorts the list A→Z, drops the Protected / Agents dropdown pills, and surfaces the same info as inline subtitle metadata ("Used by N agents" / "Not in use", compact "Nh ago" timestamps). Rows show a right-chevron on hover.
- Adds a new **ConnectionDetailPage**: clicking a row opens a dedicated view (hides the default settings page header, breaks out to ~5xl width at lg). Two columns:
  - **Agents column** — a new sectioned mode of `ConnectionAgentsList` splitting into *Agents With Access* / *Agents Without Access*.
  - **Permissions column** — the scope (OAuth) or tool (MCP) policy editor inline.
  - Header has Rename (icon-only) + Delete (icon + text) as outline buttons in place of the popover.
- Refactors the three editor dialogs to expose inline `*Body` / `*List` panels so the new detail page can reuse them. The existing Dialog wrappers keep working — message-item and post-OAuth flows are unchanged.
- Adds small primitives used along the way: `useHideSettingsHeader` hook on `SettingsPage`, `subtitleTrailing` + `onActivate` slots on `ConnectionRow`, `layout="buttons"` mode on `IntegrationRowActions`, and a `formatCompactDistance` helper for short relative timestamps.

## Test plan
- [ ] Open Settings → Connections; confirm the two default-policy rows render in one bordered card and toggles still persist.
- [ ] Confirm rows are sorted A→Z and that the subtitle reads `API · <name> · Nm ago · Used by N agents` (or `Not in use`).
- [ ] Hover a row; confirm the right chevron fades in.
- [ ] Click a row; the new detail page appears, the "Connections" page header + "+ New connection" button are hidden, and the layout is two columns at desktop widths.
- [ ] Toggle an agent's access in the Agents column; row jumps between *With Access* / *Without Access*.
- [ ] Change a scope/tool policy and click **Save Policies**; verify persistence.
- [ ] Use **Rename** to rename the connection, **Delete** to delete it; verify the delete confirmation copy.
- [ ] Click the back chevron / **Connections** breadcrumb; list re-appears with the global header restored.
- [ ] Confirm the per-agent connections page (Agent → Connections tab) still renders normally (still uses the dialog editors).
- [ ] Confirm the post-OAuth `ScopePolicyEditor` dialog still pops after connecting a new account from the **+ New connection** flow.


westaskew:
This PR was split into **3 stacked, independently-reviewable PRs**, rebased onto current `main` and verified locally (`npm run typecheck && npm run lint` both clean). Review/merge in order:

| PR | head → base | What |
|----|----|----|
| #243 | `connections/primitives` → `main` | reusable row primitives (additive, low-risk) |
| #244 | `connections/editor-panels` → `connections/primitives` | extract inline editor panels (pure refactor) |
| #245 | `connections/detail-page` → `connections/editor-panels` | the actual list redesign + detail page |

Same 10 files / same net change as here (+784/−413), just separated so the mechanical refactors land apart from the user-facing redesign. `scope-policy-editor.tsx` was re-split against current `main`'s editor.

🤖 Split out with Claude Code.

PR #178, opened 2026-06-02
feat: cloud workspace layout for remote runtime

yiw190:
## What this adds
Host-side workspace and session handling for cloud deployments, where agent workspaces live on a shared PVC root instead of under `~/.superagent`. Complements the Kubernetes remote runtime (#177) but is independent: gated on `GAMUT_CLOUD_WORKSPACE_ROOT` / `REMOTE_RUNTIME`, so local behavior is unchanged when unset.

## Changes
- **`data-dir.ts` / `file-storage.ts`**: resolve the agent workspace under `GAMUT_CLOUD_WORKSPACE_ROOT/<slug>` when set; `CLAUDE.md` stays on the local metadata path (`getLocalAgentWorkspaceDir`).
- **`session-service.ts`**: under the cloud workspace root there is no metadata sidecar, so parse the session JSONL to recover name/timestamps; `getSession` falls back to metadata-only when the JSONL doesn't exist yet. Also handles array (multimodal) message content when deriving the session name.
- **`agent-service.ts`**: expose `runtime` (`'cloud' | 'local'`) on `ApiAgent` and ensure the `CLAUDE.md` parent dir exists (cloud splits the workspace and metadata directories).
- **`types/api.ts`**: add optional `runtime?: 'local' | 'cloud'` to `ApiAgent`.

## Relationship to other PRs
Touches a disjoint set of files from #177 (K8s runtime) and #176 (per-member account sync). Each merges to main independently without breaking the others.

## Test plan
- [x] `tsc --noEmit` clean (only the pre-existing unrelated `pdf-renderer.tsx`/`react-pdf` error)
- [x] `eslint` clean
- [x] `vitest run` — session-service (80), file-storage (67), agent-service (32) green
- [ ] Manual: set `GAMUT_CLOUD_WORKSPACE_ROOT`, verify session list names/timestamps and `runtime: 'cloud'` on the agent API

### DIFF
--- a/src/shared/lib/config/data-dir.ts
+++ b/src/shared/lib/config/data-dir.ts
@@ -39,6 +39,11 @@ export function getAgentsDataDir(): string {
  * Get the workspace directory for a specific agent.
  */
 export function getAgentWorkspaceDir(agentId: string): string {
+  const cloudWorkspaceRoot = process.env.GAMUT_CLOUD_WORKSPACE_ROOT
+  if (cloudWorkspaceRoot) {
+    return path.join(path.resolve(cloudWorkspaceRoot), agentId)
+  }
+
   return path.join(getAgentsDataDir(), agentId, 'workspace')
 }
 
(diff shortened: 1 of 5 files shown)


yiw190:
Folded into #177 — the cloud workspace layout now ships in the same PR as the Kubernetes runtime.

PR #179, opened 2026-06-02
feat: add platform k8s runtime

yiw190:
## Summary
- Add a platform-managed Kubernetes runtime client that creates per-agent Pods and Services using the in-cluster Kubernetes API.
- Preserve the existing SuperAgent workspace layout under `SUPERAGENT_DATA_DIR`; agent pods mount the matching `agents/{slug}/workspace` subPath at `/workspace`.
- Route container HTTP, WebSocket, and host callback URLs through runtime-provided endpoints so local runtimes keep localhost behavior and in-cluster agents use service DNS/public host URLs.

## Test plan
- [x] `npx tsc --noEmit` (clean aside from the pre-existing `pdf-renderer` / `react-pdf` issue)
- [x] `npm run test:run -- src/shared/lib/container/platform-k8s-runtime.test.ts src/shared/lib/container/client-factory.test.ts src/shared/lib/container/container-manager.test.ts src/shared/lib/container/container-manager-concurrency.test.ts src/shared/lib/config/settings.test.ts`
- [ ] Manual in-cluster deploy with infra env update: verify pod create, health, browser stream, host callbacks, and teardown

## Notes
- Runner ID remains `kubernetes`; file/class naming uses platform k8s runtime to make clear this is the platform-managed runtime, not a generic arbitrary Kubernetes client.
- Kubernetes labels under `gamut.cloud/*` are intentionally unchanged because infra NetworkPolicy/controller selectors depend on them.
- Hard-kill GC via ownerReferences remains a follow-up; graceful shutdown uses async `client.stop()` and deletes Pod/Service.
- Companion infra PR: https://github.com/SkillfulAgents/infra/pull/2


src/shared/lib/container/client-factory.ts
```diff
@@ -72,6 +74,7 @@ const RUNNER_DISPLAY_NAMES: Record<ContainerRunner, string> = {
   podman: 'Podman',
   lima: 'Built-in Runtime',
   wsl2: 'Built-in Runtime',
+  kubernetes: 'SuperAgent Cloud',
```
yiw190: not sure here

src/shared/lib/container/platform-k8s-runtime.ts
```diff
@@ -0,0 +1,692 @@
...
+    // host-app dies (covers the crash path stopSync() can't clean up).
+    const ownerRef = await resolveOwnerReference(kube.namespace)
+    await deleteResource(`/api/v1/namespaces/${kube.namespace}/pods/${this.podName()}`)
+    await deleteResource(`/api/v1/namespaces/${kube.namespace}/services/${this.serviceName()}`)
+    await createResource(`/api/v1/namespaces/${kube.namespace}/services`, buildAgentServiceManifest(kube, this.serviceName(), this.podName(), ownerRef))
+    await createResource(`/api/v1/namespaces/${kube.namespace}/pods`, buildAgentPodManifest(kube, this.podName(), this.config, options?.envVars ?? {}, ownerRef))
+
+    await waitForPodReady(kube.namespace, this.podName(), 300_000)
+
+    if (!(await this.waitForHealthy(60_000, CONTAINER_INTERNAL_PORT))) {
```
iddogino, line 125: What happens if it doesn't become healthy in time? It can be left dangling behind...

The next `start()` exits early at src/shared/lib/container/platform-k8s-runtime.ts:106 because Kubernetes reports the pod as Running/Ready, so the manager can cache it as running even though the agent app is unusable. This is the same failure mode the base client’s **SUP-212** cleanup guards against. Delete the pod/service before throwing, or re-check health before the early return.

src/shared/lib/container/platform-k8s-runtime.ts
```diff
@@ -0,0 +1,692 @@
...
+      const logs = await this.getLogs(30)
+      const logsSnippet = logs ? `\n\nPod logs:\n${logs}` : ''
+      throw new Error(`Platform k8s agent pod ${this.podName()} failed to become healthy${logsSnippet}`)
+    }
+  }
+
+  async stop(_options?: StopOptions): Promise<StopResult> {
+    this.terminateWebSocketConnections()
+    const { namespace } = getKubeConfig()
+    await deleteResource(`/api/v1/namespaces/${namespace}/pods/${this.podName()}`)
```
iddogino, line 142: Pod deletion is asyncrhonous -> we should wait for pod to actually be gone, not just for deletion request to be accepted.

`containerManager.stopContainer()` can mark the agent stopped while the process is still terminating, and `deleteAgent()` then removes the workspace at src/shared/lib/services/agent-service.ts:334. Poll until the pod is actually gone.

Bonus: consider treating metadata.deletionTimestamp as stopped/not-routable in `getInfoFromRuntime()`.

src/shared/lib/container/platform-k8s-runtime.ts
```diff
@@ -0,0 +1,692 @@
...
+  protected getRunnerCommand(): string {
+    return 'kubernetes'
+  }
+
+  static isEligible(): boolean {
+    return isRunningInKubernetes()
+  }
+
+  static async isAvailable(): Promise<boolean> {
+    return Boolean(resolveKubeConfigOrNull())
```
iddogino: Should we should validate `HOST_PUBLIC_URL` here as well? container-manager.ts:519 always calls `getHostApiBaseUrl()` and would fail if not set...

iddogino (changes requested):

PR #205, opened 2026-06-05
feat: steer agent to subagents over the Workflow tool

yiw190:
## What this fixes

SDK bump `0.3.144 → 0.3.154` (commit `4c5c40a4`, May 28) added the **`Workflow`** tool to the `claude_code` preset. Because `agent-container` doesn't disallow it, the main agent now has `Workflow` available automatically — but nothing in our `system-prompt.md` tells the agent *when* to use it. Found sometimes the agents creates workflow for some easy task, which is not nessesary.

The idea of the PR is to default agent to use easy solutions first then escalate.


yiw190:
Should add better UI instead of prompt change

PR #216, opened 2026-06-06
fix(security): scheme-allowlist openExternal in Electron main (SUP-214)

iddogino:
### Bug
The Electron main process forwarded any renderer-/popup-supplied string straight to `shell.openExternal` with **no scheme validation**:
- `src/main/index.ts` `ipcMain.handle('open-external', ...)` → `await shell.openExternal(url)`
- `src/main/index.ts` `mainWindow.webContents.setWindowOpenHandler(({ url }) => { ... shell.openExternal(url) })`

`shell.openExternal` asks the OS to launch a registered handler, so an attacker- or agent-controlled URL with `file:`, `javascript:`, `data:`, or a custom app protocol (`myapp://`) could launch local apps / privileged flows outside the browser sandbox.

### Fix
Added `src/main/safe-open-external.ts`: a Zod-validated boundary (per CLAUDE.md) that parses the input string, constructs `new URL(url)`, and forwards only an allowlist of safe web schemes — `https:`, `http:` (OAuth / localhost redirects), and `mailto:`. Everything else, plus non-string / empty / malformed input, is logged and dropped **without throwing**. Both call sites now route through `safeOpenExternal`; the existing `/api/agents/.../files/` download special case in `setWindowOpenHandler` is preserved ahead of it.

### Regression test (failing → green)
`src/main/safe-open-external.sup214.test.ts` (21 cases) mocks `electron`'s `shell.openExternal` and asserts the Zod allowlist drives the decision: `https:`/`http:`/`mailto:` are opened; `file:///Applications/Calculator.app`, `javascript:alert(1)`, `data:...`, `myapp://`, `vbscript:`, `ftp:` are rejected (shell never called); and non-string / empty / malformed input is rejected without throwing. The test fails on the pre-fix tree (no validation helper exists) and passes after the fix.

### Changed files
- `src/main/safe-open-external.ts` (new helper)
- `src/main/safe-open-external.sup214.test.ts` (new test)
- `src/main/index.ts` (route both call sites through `safeOpenExternal`)

### Validation
- New test: 21/21 pass; related `src/main` suite (auto-updater) still green.
- `npx tsc --noEmit`: clean. `npx eslint` on changed files: 0 errors.

Status: waiting for review.

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -37,6 +37,7 @@ import { getSettings } from '@shared/lib/config/settings'
 import { detectAllProviders } from './host-browser'
 import { registerUpdateHandlers, initAutoUpdater, updateAutoUpdaterWindow } from './auto-updater'
 import { enableKeepAwake, disableKeepAwake, cleanupKeepAwake, restoreKeepAwakeOnStartup } from './keep-awake'
+import { safeOpenExternal } from './safe-open-external'
 
 // In dev mode, use a separate data directory to avoid mixing with production data.
 // Setting app.name before getPath('userData') changes the resolved directory.
@@ -359,8 +360,10 @@ function createWindow() {
       mainWindow?.webContents.downloadURL(url)
       return { action: 'deny' }
     }
-    // For other URLs (OAuth, external links), open in system browser
-    shell.openExternal(url)
+    // For other URLs (OAuth, external links), open in system browser — but only
+    // after scheme validation so a popup can't ask the OS shell to launch
+    // file:/javascript:/custom-protocol handlers (SUP-214).
+    void safeOpenExternal(url)
     return { action: 'deny' }
   })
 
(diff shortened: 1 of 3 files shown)


iddogino:
Heads-up + fix pushed re: a regression I found while you were reviewing. The initial allowlist (http/https/mailto) would have silently broken two **first-party** flows that legitimately pass non-web schemes to `openExternal`:

- **computer-use permission helper** → `x-apple.systempreferences:…` (opens macOS System Settings)
- **iMessage setup link** → `sms:…&body=/setup`

**Finder is unaffected** — folder reveal uses `shell.openPath` (`show-in-folder` IPC) and `shell.showItemInFolder`, neither of which this PR touches.

Fix: two-tier guard — the popup `setWindowOpenHandler` (untrusted web/agent content) stays strict (web-only), while the `open-external` IPC (first-party app UI) also allows `sms:`/`tel:`/`x-apple.systempreferences:`. These hand a structured request to a fixed OS handler (user still confirms), unlike `file:`/custom protocols. 34 test cases cover both tiers. Let me know if you'd rather keep a single strict allowlist and instead route those two flows through `shell.openPath`/a dedicated IPC.

iddogino:
Refined per the review point: **moved `sms:`/`tel:` into the shared (popup-reachable) allowlist** alongside `mailto:`. They're the same user-confirmed "communication composer" class — agent content surfacing a `tel:`/`sms:` link now works, consistent with `mailto:` already being allowed there. The only scheme that stays **app-IPC-only** is `x-apple.systempreferences:` (no legitimate use from agent/web content).

Net allowlists now:
- popup / untrusted (`setWindowOpenHandler`): `http`, `https`, `mailto`, `sms`, `tel`
- app IPC (`open-external`): the above **+** `x-apple.systempreferences:`

Residual (documented in code): `sms:`/`tel:` can prefill a premium-rate number/shortcode — a minor, **user-gated** social-engineering angle (the user sees the number and confirms before sending), same exposure profile as the long-allowed `mailto:`. Shout if you'd rather keep `tel:` stricter given the toll-fraud angle.

PR #221, opened 2026-06-06
fix(chat-integrations): track queue settledness so cleanup evicts (SUP-234)

iddogino:
### Bug
`ChatIntegrationManager.cleanupResolvedQueues` tried to detect a settled `Promise` by attaching `.then(() => isSettled = true)` and reading the local flag on the very next synchronous line. Per ECMAScript, `.then` reactions are enqueued as microtasks and never run during the synchronous `.then()` call, so `isSettled` was always `false`. The companion `promise === Promise.resolve()` check compared each stored promise against a brand-new `Promise.resolve()`, which can never match a `.then()`-chained queue value produced by `enqueueMessage`/`enqueueSSEEvent`. The delete branch was therefore unreachable, and `messageQueues` grew unbounded — one lingering entry per active chat and per SSE stream — across long-running chat-integration usage. The 10-minute sweep ran but was a no-op.

### Fix
Stop inspecting native Promise state synchronously (no such API exists in standard JS) and track settledness explicitly:
- `messageQueues` values are now `QueueEntry { promise: Promise<void>; settled: boolean }`.
- `enqueueMessage`/`enqueueSSEEvent` chain off the previous entry's `promise` and attach a `.finally` that flips `settled` and self-evicts the entry **only if the map still holds that exact reference** — a newer enqueue replaces the value with a fresh, unsettled entry (chained off the old one) that must not be evicted.
- `cleanupResolvedQueues` is now a correct periodic backstop that deletes entries whose `settled` flag is `true`. Since each map value is by definition the current entry for its key, in-flight (unsettled) entries are never evicted.
- `removeIntegration`/`stop` operate on keys only and are unaffected by the value-shape change.

### Tests (failing-now-green regression)
New file `src/shared/lib/chat-integrations/chat-integration-manager.sup234.test.ts` drives the private queue mechanics (handlers stubbed so no container/agent code runs):
1. removes an already-settled **chat** message queue entry during cleanup (mirrors the Linear repro assertion) — failed before the fix, passes now.
2. removes an already-settled **SSE** queue entry during cleanup — failed before the fix, passes now.
3. retains an **in-flight** (never-resolving) entry — guard against over-eviction.
4. retains the key when a newer **pending** message replaces a settled one under the same key — the identity race guard.

All 362 tests in `src/shared/lib/chat-integrations/` pass (including the e2e suite that exercises the real `enqueueMessage`/`enqueueSSEEvent` paths). `npx tsc --noEmit` clean; `eslint` 0 errors on changed files (the one remaining warning is a pre-existing unrelated `startsWith` containment check in `sendDeliveredFile`, untouched here).

### Changed files
- `src/shared/lib/chat-integrations/chat-integration-manager.ts`
- `src/shared/lib/chat-integrations/chat-integration-manager.sup234.test.ts` (new)

Status: waiting for review.

### DIFF
--- a/src/shared/lib/chat-integrations/chat-integration-manager.sup234.test.ts
+++ b/src/shared/lib/chat-integrations/chat-integration-manager.sup234.test.ts
@@ -1,0 +1,115 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
+import { chatIntegrationManager } from './chat-integration-manager'
+
+// ---------------------------------------------------------------------------
+// SUP-234 — cleanupResolvedQueues never removes settled promises.
+//
+// `cleanupResolvedQueues` tried to detect a settled Promise by attaching
+// `.then(() => isSettled = true)` and reading `isSettled` on the very next
+// synchronous line. Per ECMAScript, `.then` reactions are enqueued as
+// microtasks and never run during the synchronous `.then()` call, so the flag
+// is always `false` and the delete branch is unreachable. The companion
+// `promise === Promise.resolve()` check can never match a `.then()`-chained
+// queue value either. As a result `messageQueues` grows unbounded.
+//
+// These tests drive the manager's private message-queue mechanics directly
+// (handlers stubbed so no container/agent code runs) and assert that:
+//   1. a settled chat message queue entry is removed on cleanup,
+//   2. a settled SSE queue entry is removed on cleanup,
+//   3. an in-flight (never-resolving) entry is retained,
+//   4. when a newer pending message replaces a settled one under the same key,
+//      the (now unsettled) current entry is retained — the race guard.
+// ---------------------------------------------------------------------------
+
+interface ManagerInternals {
+  messageQueues: Map<string, unknown>
+  cleanupResolvedQueues: () => void
+  enqueueMessage: (integrationId: string, message: { chatId: string; text?: string }) => void
+  enqueueSSEEvent: (integrationId: string, chatId: string, event: unknown) => void
+  handleIncomingMessage: (integrationId: string, message: unknown) => Promise<void>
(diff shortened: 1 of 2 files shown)


iddogino:
Pushed the three refinements you asked for, plus hardening from a two-pass adversarial review:

**1. Dropped the redundant sweep.** Self-eviction makes a periodic scan unreachable (a settled entry is deleted by its own `.finally` before `settled` is ever observable), so `cleanupResolvedQueues`, the `queueCleanupInterval` timer, and `QUEUE_CLEANUP_INTERVAL_MS` are gone. The `settled` flag only existed for the sweep, so `QueueEntry` collapses back to `Map<string, Promise<void>>` and the helper is now `scheduleQueueEviction` (`.finally` + map-identity guard). One mechanism, not two.

**2. Fixed `removeIntegration`.** The old `delete(id)` never matched the composite keys; now it sweeps by prefix. Documented the one behavior change inline: on the reconnect path it force-drops an in-flight chain rather than serializing the first post-reconnect message behind a handler tied to the dead connection — a conscious trade-off (correctness reviewer flagged it as sign-off-worthy, not a bug).

**3. Hardened tests** (now 7): the identity-guard test now enqueues both messages *before* flushing so it actually exercises the race (previously it was equivalent to the in-flight test); added a **volume** test (many keys → drains to 0), a **`removeIntegration`** test using prefix-colliding ids `intg`/`intg2` to lock in the `:` delimiter, and a **rejecting-handler** test (the `.catch` keeps the tail resolving, so a failed handler still self-evicts).

Verification: adversarial correctness review came back **solid** (no leak / no lost work / ordering preserved / prefix cleanup safe / no dangling refs). I mutation-tested the delimiter guard (drop the `:` → the new test fails as expected, confirming it's not vacuous). `tsc` + `eslint` clean (0 errors; the lone warning is the pre-existing `deliver_file` line, handled in #229). Full chat-integrations suite green.

PR #222, opened 2026-06-06
fix(auth): give app admins the EntityAgentRole ACL bypass (SUP-204)

iddogino:
### Bug
In `AUTH_MODE`, the generic `EntityAgentRole()` factory in `src/api/middleware/auth.ts` (used by scheduled-task, webhook-trigger, and chat-integration routes) checks the caller's per-agent ACL role but never short-circuits for app admins. App admins (`role==='admin'`) with no explicit `agentAcl` row therefore got a `403` on these entity routes, even though `AgentRead`/`AgentUser`/`AgentAdmin` (and `HasNotificationAccess`/`OwnsAccountByParam`/`OwnsMcpByParam`) let them through. For an admin with no ACL row, `getUserAgentRole` returns `null` and `hasMinRole(null, minRole)` is `false`, yielding the 403. The permissions model was inconsistent.

### Fix
In `EntityAgentRole`'s returned middleware, after `const user = getUser(c)` and before the `getUserAgentRole`/`hasMinRole` check, add `if (isAdmin(user)) return next()` to mirror `AgentRead`/`AgentUser`/`AgentAdmin`. The entity is already loaded and stashed on context (`c.set`) before the auth gate, so admins still get the entity in context. The not-found `404` and non-auth-mode passthrough remain ahead of the admin check, unchanged.

### Tests
New file `src/api/middleware/auth.entity-admin-bypass.sup204.test.ts` (failing before the fix, green now):
- Admin with no ACL row + strictest role `'owner'` -> `200`, entity stashed on context, and the ACL `select` is never consulted (`expect(mockSelect).not.toHaveBeenCalled()`), proving the bypass precedes any ACL lookup.
- Admin + missing entity still `404` (not-found precedes the bypass).
- Regression: a non-admin user with no ACL row still gets `403` (and the ACL lookup IS consulted).

The existing `src/api/middleware/auth.test.ts` suite (120 tests, including the `EntityAgentRole()` block) still passes unchanged.

### Validation
- New + existing auth middleware tests pass (123 total).
- `npx tsc --noEmit` clean.
- `npx eslint` clean on changed files (0 errors).

### Changed files
- `src/api/middleware/auth.ts` (one-line admin bypass in `EntityAgentRole`)
- `src/api/middleware/auth.entity-admin-bypass.sup204.test.ts` (new repro + regression test)

Status: waiting for review.

### DIFF
--- a/src/api/middleware/auth.entity-admin-bypass.sup204.test.ts
+++ b/src/api/middleware/auth.entity-admin-bypass.sup204.test.ts
@@ -1,0 +1,147 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { Hono } from 'hono'
+
+// ---------------------------------------------------------------------------
+// SUP-204 — EntityAgentRole admin ACL-bypass repro.
+//
+// In AUTH_MODE the generic EntityAgentRole() middleware (scheduled-task,
+// webhook-trigger, chat-integration routes) checked the caller's per-agent ACL
+// role but never short-circuited for app admins, unlike AgentRead/AgentUser/
+// AgentAdmin. An app admin (role==='admin') with no explicit agentAcl row got a
+// 403 on those entity routes.
+//
+// This suite mirrors the EntityAgentRole() block in auth.test.ts but exercises
+// the admin path. The mock preamble must be set up before importing the module
+// under test (vi.mock is hoisted).
+// ---------------------------------------------------------------------------
+
+const mockIsAuthMode = vi.fn<() => boolean>()
+vi.mock('@shared/lib/auth/mode', () => ({
+  isAuthMode: () => mockIsAuthMode(),
+}))
+
+const mockGetSession = vi.fn()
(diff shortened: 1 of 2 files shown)


iddogino:
Closing — the premise is inverted. SUP-204 was filed as 'app admins are wrongly 403'd on agent-scoped entity routes,' but per a product decision **ACLs are the authoritative root of agent access; the deployment/admin role does not grant agent access**. Under that model the 403 is correct behavior, and this PR (which adds the admin bypass to EntityAgentRole) would cement the wrong behavior.

The real issue is the inverse and broader: the pre-existing `if (isAdmin(user)) return next()` bypass in AgentRead/AgentUser/AgentAdmin (and the Or(IsAdmin) ownership middlewares) lets admins reach any user's agents. That'll be tracked in a new, larger ticket to remove the admin-over-agents bypass (with a narrow exception for stopping agents during infra/runtime changes). SUP-204 will be closed as not-the-issue.

PR #225, opened 2026-06-06
fix(agents): roll back agent workspace when owner ACL insert fails (SUP-207)

iddogino:
### Bug
In `AUTH_MODE`, the three agent-creation routes in `src/api/routes/agents.ts` write the agent's on-disk workspace (directory + `CLAUDE.md`) **before** inserting the owner ACL row:

- `POST /api/agents` (`createAgent` → `createOwnerAcl`)
- `POST /api/agents/import-template` (`importAgentFromTemplate` → `createOwnerAcl`, via `processImport`)
- `POST /api/agents/install-from-skillset` (`installAgentFromSkillset` → `createOwnerAcl`)

If the ACL insert throws (e.g. a transient DB failure), the surrounding `catch` only returned `500` and never deleted the just-created workspace — leaving an orphaned agent directory on disk with no owner ACL. Such an agent is hidden from the creating user in auth mode but still shows up in filesystem-backed listings and can interfere with later cleanup/import/reset.

### Fix
Introduced `createOwnerAclOrRollback(c, slug)` that wraps the owner ACL insert: on failure it best-effort calls `deleteAgent(slug)` (already imported) to roll back the on-disk workspace, then rethrows so the route still returns `500`. The cleanup is itself guarded in try/catch so a failed rollback never masks the original error. It is a no-op outside auth mode, since `createOwnerAcl` already short-circuits there. All three creation paths now use this helper.

### Tests
New regression suite `src/api/routes/agents-owner-acl-rollback.sup207.test.ts`:
- Forces the owner ACL insert to throw and asserts the route returns `500` **and** `deleteAgent` is called with the created slug — for all three routes (`/`, `/import-template`, `/install-from-skillset`).
- Asserts a failed rollback does not mask the original error (still `500`).
- Regression: happy-path `201` performs the ACL insert and does **not** roll back; non-auth mode neither inserts an ACL nor rolls back.

These tests fail on `main` (`deleteAgent` never called) and pass with this change. The closely-related `security-repro.test.ts` suite still passes. `npx tsc --noEmit` and `eslint` on the changed files are clean.

### Changed files
- `src/api/routes/agents.ts`
- `src/api/routes/agents-owner-acl-rollback.sup207.test.ts` (new)

Status: waiting for review.

### DIFF
--- a/src/api/routes/agents-owner-acl-rollback.sup207.test.ts
+++ b/src/api/routes/agents-owner-acl-rollback.sup207.test.ts
@@ -1,0 +1,349 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { Hono } from 'hono'
+
+// ---------------------------------------------------------------------------
+// SUP-207 — Auth mode: failed owner ACL insert leaves orphaned agent workspace.
+//
+// In AUTH_MODE, the three agent-creation routes create the agent's on-disk
+// workspace BEFORE inserting the owner ACL row:
+//   - POST /api/agents              (createAgent + createOwnerAcl)
+//   - POST /api/agents/import-template      (importAgentFromTemplate + createOwnerAcl)
+//   - POST /api/agents/install-from-skillset (installAgentFromSkillset + createOwnerAcl)
+// If the ACL insert throws, the route returns 500 but never deletes the
+// just-created workspace, orphaning an agent dir with no owner ACL.
+//
+// These tests drive the owner ACL insert to fail and assert the route both
+// returns 500 AND rolls back the workspace via deleteAgent(slug). Regression
+// cases assert the happy path still 201s without any rollback, and that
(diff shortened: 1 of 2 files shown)


iddogino:
Added Sentry reporting on the failure paths (per request). The rollback helper previously only `console.error`'d, and the route's outer catch returns 500 without reporting — so these were invisible in Sentry.

`createOwnerAclOrRollback` now calls `captureException` (`@shared/lib/error-reporting`):
- **ACL insert failure** → reported with `{ agentSlug, userId, rolledBack }`. Clean rollback = recovered → `warning`; failed rollback (orphan left) → `error`.
- **Rollback failure** → reported as a distinct `error` (`operation: owner-acl-rollback`) with the original error in `extra`, since it leaves an orphaned workspace needing attention.

Tests now assert `captureException` fires with the right operation/level on both failure paths and not on success. tsc + eslint clean; 6/6 tests pass.

PR #226, opened 2026-06-06
fix(host-browser): bind Chrome CDP to loopback, not all interfaces (SUP-217)

iddogino:
### Bug
`ChromeProvider.launch()` spawned Chrome with `--remote-debugging-address=0.0.0.0` and bound the Windows fallback proxy on `0.0.0.0`, exposing the **unauthenticated** Chrome DevTools Protocol on **every** interface. Anything on the LAN could fully remote-control the dedicated Chrome profile (read pages, cookies/session, drive navigation).

### Fix
Chrome CDP now binds **`127.0.0.1`** on every platform. To keep the agent container able to reach it, a lightweight TCP proxy forwards the loopback CDP port — bound to the **single host-internal bridge interface** the active runner uses, **never `0.0.0.0`**.

Each container runner reports how its containers reach the host via a new `ContainerClient.getHostBridgeIp()`:
- **null** → the container reaches the host's loopback directly (Docker Desktop forwards `host.docker.internal` to host loopback). No proxy; container uses Chrome's loopback port.
- **an IP** → the container routes through a real bridge gateway. The proxy binds exactly that IP.

| Runner | `getHostBridgeIp()` | Behavior |
|---|---|---|
| **Lima** (macOS) | VM default-gateway (e.g. `192.168.64.1`) | proxy bound to the vmnet host IP |
| **WSL2** (Windows) | WSL distro default-gateway | proxy bound to the WSL gateway (Chrome ignores the bind on Windows, so the proxy is required) |
| **Docker — Desktop** (mac/win) | null | no proxy; loopback forwarded by Docker Desktop |
| **Docker — native** (Linux) | docker0 gateway IP | proxy bound to docker0 |
| **Podman** | null (rootless slirp/pasta forwards loopback) | no proxy; rootful-bridge is a known gap |

> **Why this grew from the original PR:** the first cut only ran the proxy on Windows, which broke every gateway-routed runner — including **Lima, the majority runner** — because loopback isn't reachable through a bridge gateway. The Lima/Linux/Podman gateway detection is the same `ip route show default` the clients already run for `--add-host`, now shared with the proxy.

### Tests (`chrome-provider.sup217.test.ts`)
Driven by the bridge IP (the real decision variable):
- **no bridge IP** (Docker Desktop) → no proxy, only loopback binds; Chrome arg loopback.
- **bridge IP** (Lima / native Docker / Podman) → proxy bound to that IP, **never `0.0.0.0`**; Chrome arg loopback.
- **win32** (WSL2) → proxy bound to the gateway IP, never `0.0.0.0`; Chrome arg loopback.

All fail against `main` (which passes `0.0.0.0`). `tsc --noEmit` and `eslint` clean; full container-client suite (498 tests) green.

### Changed files
- `src/main/host-browser/chrome-provider.ts`
- `src/shared/lib/container/{types,base-container-client,lima-container-client,wsl2-container-client,docker-container-client,mock-container-client}.ts`
- `src/main/host-browser/chrome-provider.sup217.test.ts`

Status: waiting for review.

### DIFF
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/iddogino/Dropbox/Superagent/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/iddogino/Dropbox/Superagent/node_modules
--- a/src/main/host-browser/chrome-provider.sup217.test.ts
+++ b/src/main/host-browser/chrome-provider.sup217.test.ts
@@ -19,17 +19,28 @@ import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest'
 // hoisted above the imports) and by the tests.
 const h = vi.hoisted(() => {
   const listenCalls: Array<{ port: unknown; host: unknown }> = []
+  const killedPids: number[] = []
+  let failProxyListen = false
+  let proxyCloseCount = 0
 
-  function makeServer() {
+  // isProxy distinguishes the CDP forwarding proxy (created with a connection
+  // listener) from findFreePort's throwaway server (created with no args).
+  function makeServer(isProxy: boolean) {
+    const handlers: Record<string, (...a: unknown[]) => void> = {}
     const server: Record<string, unknown> = {
       listen: (port: unknown, host: unknown, cb?: unknown) => {
         listenCalls.push({ port, host })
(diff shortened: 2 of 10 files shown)


iddogino:
Questions:
1/ Does this work with DooD, Docker Desktop, Podman

PR #227, opened 2026-06-06
fix(agent-service): abort agent deletion when container stop fails (SUP-209)

iddogino:
### Bug
`deleteAgent()` wrapped `containerManager.stopContainer(slug)` in a `try/catch` with an empty handler, then **unconditionally** called `removeDirectory(agentDir)` and resolved `true`. A genuine runtime stop failure (wedged VM, unexpected stop error) was swallowed and treated identically to "container was already gone", so the host workspace was deleted while the container could still be running or in an unknown stop state.

`stopContainer` genuinely rejects on real runtime errors (`container-manager.ts` awaits `client.stop()` in a `try/finally` with no catch — see `container-manager.test.ts`: `await expect(containerManager.stopContainer('error-agent')).rejects.toThrow('unexpected error')`), so the empty catch could not distinguish "already gone" from "stop failed".

### Fix
Remove the blanket `try/catch` and let real errors propagate. The underlying container client is idempotent for already-stopped/missing containers — `base-container-client.ts` runs the `docker stop`/`rm` commands through `execWithPathSilent`, which silently swallows benign "no such container" errors — so any rejection out of `stopContainer` is genuinely abnormal. On such a rejection, deletion now aborts: `removeDirectory` runs strictly after a successful (or benignly no-op) stop, so the workspace is preserved and the failure surfaces to the API/UI. The benign "container already gone" path still deletes normally.

### Regression test (failing → green)
Added `src/shared/lib/services/agent-service.delete-stop-failure.sup209.test.ts` (dedicated file to avoid cross-branch merge conflicts), reusing the existing `vi.mock` harness header:
- **Abort path**: `mockStopContainer.mockRejectedValueOnce(new Error('runtime wedged: cannot stop container'))` → `deleteAgent('test-agent')` rejects `/runtime/` AND `agentExists('test-agent')` stays `true` (workspace survives). This test failed before the fix (`deleteAgent` resolved `true`) and passes now.
- **Happy path**: a benign no-op stop still returns `true`, calls `stopContainer`, and removes the workspace.

All 34 tests in the new file + existing `agent-service.test.ts` pass; `tsc --noEmit` and eslint on the changed files are clean.

### Changed files
- `src/shared/lib/services/agent-service.ts` — `deleteAgent()` no longer swallows stop failures
- `src/shared/lib/services/agent-service.delete-stop-failure.sup209.test.ts` — new regression tests

Status: waiting for review.

### DIFF
--- a/src/api/routes/agents.delete-ordering.sup208.test.ts
+++ b/src/api/routes/agents.delete-ordering.sup208.test.ts
@@ -28,6 +28,20 @@ import { Hono } from 'hono'
 // irreversible workspace removal we are gating.
 const mockGetAgent = vi.fn()
 const mockDeleteAgent = vi.fn()
+// SUP-209: a genuine container stop-failure surfaces from deleteAgent as this
+// typed error, which the route maps to 409. Hoisted so the mock factory and the
+// test share one class — the route's `instanceof` resolves to this same
+// stand-in via the mocked module.
+const { AgentContainerStopError } = vi.hoisted(() => ({
+  AgentContainerStopError: class AgentContainerStopError extends Error {
+    readonly slug: string
+    constructor(slug: string, cause: unknown) {
+      super(`Failed to stop the container for agent "${slug}": ${cause instanceof Error ? cause.message : String(cause)}`)
+      this.name = 'AgentContainerStopError'
+      this.slug = slug
+    }
+  },
+}))
 vi.mock('@shared/lib/services/agent-service', () => ({
   listAgentsWithStatus: vi.fn(),
   createAgent: vi.fn(),
@@ -36,6 +50,7 @@ vi.mock('@shared/lib/services/agent-service', () => ({
   updateAgent: vi.fn(),
   deleteAgent: (...args: unknown[]) => mockDeleteAgent(...args),
   agentExists: vi.fn().mockResolvedValue(true),
+  AgentContainerStopError,
 }))
 
 // --- peripheral cleanup services --------------------------------------------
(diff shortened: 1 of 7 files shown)


iddogino:
### Update: added the user-facing error path + brought the branch up to date

This PR originally only fixed the service layer (abort + preserve workspace). On review it was pointed out that the *user-visible* behavior was still poor: the route returned a generic **500**, the hook threw a hardcoded string, and both delete call sites only `console.error`'d — so the confirm dialog silently stayed open with **no toast and no explanation**. Extended the PR to close that gap:

- **`agent-service.ts`** — `deleteAgent` wraps a genuine `stopContainer` rejection in a typed **`AgentContainerStopError`** (preserving the cause message for logs).
- **`agents.ts`** — `DELETE /:id` maps that error to **`409`** with an actionable message ("Couldn't stop the agent's container … try again in a moment"); everything else stays `500`. The agent is preserved and the delete is retryable.
- **`use-agents.ts`** — `useDeleteAgent` now reads the server's error body and throws *that*, not a fixed string.
- **`agent-context-menu.tsx` / `general-tab.tsx`** — `toast.error(message)` on failure instead of a bare `console.error`.
- **Tests** — sup209 service test now asserts the rejection is an `AgentContainerStopError` (still matching the `/runtime/` cause); sup208 route test gains a case asserting `DELETE → 409` + a container-mentioning message, with peripheral cleanup still having run.

**Also merged latest `main` into the branch** — it was 47 commits behind and still on the pre-SUP-208 DELETE ordering. The merge was conflict-free (the branch only touched `agent-service.ts`, which `main` hadn't), so it now sits on the current SUP-208 handler (peripheral cleanup → workspace removal), and the 409 mapping lives in that handler.

Validation: `tsc --noEmit` + `eslint` clean; affected suites green (sup209 + sup208 = 7 tests, `agent-service` = 32).

Note on status code: chose **409 Conflict** (the target resource is in a state — wedged container — that blocks deletion) over 503 (which implies whole-service unavailability). Easy to switch if you'd prefer 503 + `Retry-After`.

PR #229, opened 2026-06-06
fix(chat-integrations): path traversal, Slack redirect token leak, archived-session restore (SUP-231, SUP-232, SUP-233)

iddogino:
Three independent security fixes in `src/shared/lib/chat-integrations/chat-integration-manager.ts`, each kept in its own function, with a faithful failing-test-now-green regression spec per issue.

## SUP-231 — Chat attachment filenames can escape the workspace uploads directory
External chat attachment filenames (Slack/Telegram/iMessage) were concatenated unsanitized into a filesystem path in `writeToWorkspace`. The `Date.now()-` prefix only absorbed the first `..` segment, so a name like `../../../oauth-token.txt` wrote OUTSIDE `<workspace>/uploads` (and `../../../../etc/cron.d/x` escaped the workspace entirely).

**Fix:** new exported `sanitizeUploadFilename` helper — takes the basename, strips path separators, `..`, leading dots and NUL bytes, and falls back to `file` when empty. After resolving, `assertPathWithinDir` (from `@shared/lib/utils/path-safety`) defensively asserts the destination stays under `uploads` before mkdir/write.

## SUP-232 — Slack file download leaks bot token across cross-origin redirects
`downloadWithAuth` manually followed 3xx redirects and re-sent the Slack `Authorization: Bearer <botToken>` header to every target with no host check, leaking the xoxb token to an attacker-controlled redirect.

**Fix:** resolve each redirect `location` against the current URL with `new URL(location, current)` and attach the token only for HTTPS requests to trusted Slack hosts (`slack.com` / `*.slack.com`). Cross-origin hops are fetched without the header (legit Slack downloads redirect to signed S3 URLs whose auth is in the query string, so this does not break them). Redirects stay `redirect: 'manual'` and the cap is retained.

## SUP-233 — Chat integration reconnect restores archived session SSE forwarding
On connect/reconnect the restore loop subscribed SSE forwarding for every row of `listChatIntegrationSessions` (which returns archived rows too), so archived/cleared/timed-out sessions were re-subscribed and could forward stale agent output back to the external chat.

**Fix:** new `listActiveChatIntegrationSessions` (adds `isNull(archivedAt)`) used by the restore loop. The shared `listChatIntegrationSessions` is intentionally left unchanged — the UI route and x-agent-chat surfaces return archived rows and filter at the call site.

## Regression tests (failing before, green after)
- `src/shared/lib/chat-integrations/chat-integration-manager.sup231.test.ts` — writeToWorkspace against a real temp workspace (no escape) + `sanitizeUploadFilename` unit cases.
- `src/shared/lib/chat-integrations/chat-integration-manager-redirect-auth.sup232.test.ts` — stubbed fetch: cross-origin redirect drops auth; same-host Slack redirect keeps it.
- `src/shared/lib/chat-integrations/chat-integration-manager-archived-restore.sup233.test.ts` — in-memory DB + mock connector drives the real restore path; `addSSEClient` wired for the active session only.

Full chat-integrations suite: 369 passing (incl. 11 new). `tsc --noEmit` clean; eslint 0 errors on changed files.

## Changed files
- `src/shared/lib/chat-integrations/chat-integration-manager.ts`
- `src/shared/lib/services/chat-integration-session-service.ts`
- 3 new test files (above)

Status: waiting for review.


iddogino:
Added a 4th, related hardening surfaced by the new `no-path-containment-startswith` lint rule (from infra #213): `sendDeliveredFile` (the `deliver_file` MCP path) used `fullPath.startsWith(path.resolve(workspaceDir))` without a trailing separator — a **reachable** sibling-prefix traversal, since `filePath` is agent-controlled (`/workspace/../<slug>-victim/secret` resolves into another agent's workspace). Migrated to `isPathWithinDir`. This clears the last path-containment warning in the repo. Containment logic is covered by the merged `path-safety` unit tests (sibling-prefix vector included); `sendDeliveredFile` isn't exported so it has no standalone integration test. Flagging for explicit review since it's outside the original SUP-231/232/233 scope — happy to split it into its own PR/issue if you'd prefer.

PR #231, opened 2026-06-06
fix(scheduler): roll back start() state when initial overdue scan fails (SUP-224)

iddogino:
### Bug
`TaskScheduler.start()` (src/shared/lib/scheduler/task-scheduler.ts) set `this.isRunning = true` *before* awaiting the initial `executeOverdueTasks()` scan, with no try/catch around it. The interval install came *after* the await. If the first scan threw — e.g. a transient SQLite error in `getDueTasks()` (an un-guarded `db.select()`), which `executeOverdueTasks`'s try/finally does not catch — `start()` rejected with `isRunning` still `true` and `intervalId` still `null`. The polling loop was never installed, and every later `start()` short-circuited on the `Already running` guard, silently disabling all scheduled tasks for the rest of the process lifetime. The only caller (`startup.ts:116`) just `.catch`-logs the rejection, so nothing ever retried.

### Fix
Wrap the initial scan and interval install in `try/catch`. On failure: clear/null any installed interval, reset `isRunning = false`, and rethrow so `startup.ts`'s `.catch` still logs. `isActive()` now reports `false` after a failed startup, and a later `start()` can install the polling loop and recover.

### Tests (failing-test-first, now green)
New file `src/shared/lib/scheduler/task-scheduler.sup224.test.ts` mocks only the `@shared/lib/services/scheduled-task-service` boundary:
- **does not stay marked running when the initial overdue scan fails** — `getDueTasks` rejects once; `start()` rejects and `isActive()` must be `false`. Failed pre-fix with `expected true to be false`.
- **allows a later start() to install the polling loop after a failed startup scan** — after the failed start, a second `start()` goes active and advancing fake timers by the poll interval (60s) re-invokes `getDueTasks`, proving the interval was installed.
- **installs the polling loop and stays active on a clean startup** — positive regression guard so the fix can't silently break normal startup.

All 3 pass post-fix; the full scheduler suite (5 files, 47 tests) stays green. `tsc --noEmit` clean; eslint clean on changed files.

### Changed files
- `src/shared/lib/scheduler/task-scheduler.ts`
- `src/shared/lib/scheduler/task-scheduler.sup224.test.ts` (new)

Status: waiting for review.

### DIFF
--- a/src/shared/lib/scheduler/task-scheduler.sup224.test.ts
+++ b/src/shared/lib/scheduler/task-scheduler.sup224.test.ts
@@ -1,0 +1,92 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+
+// ============================================================================
+// Mocks — only the scheduled-task-service boundary needs mocking.
+// When the initial overdue scan throws (or returns []), executeTask is never
+// reached, so the heavy collaborators (containerManager, messagePersister,
+// etc.) are never exercised.
+// ============================================================================
+
+const mockGetDueTasks = vi.fn()
+const mockMarkTaskExecuted = vi.fn().mockResolvedValue(undefined)
+const mockMarkTaskFailed = vi.fn().mockResolvedValue(undefined)
+const mockUpdateNextExecution = vi.fn().mockResolvedValue(undefined)
+
+vi.mock('@shared/lib/services/scheduled-task-service', () => ({
+  getDueTasks: (...args: unknown[]) => mockGetDueTasks(...args),
+  markTaskExecuted: (...args: unknown[]) => mockMarkTaskExecuted(...args),
+  markTaskFailed: (...args: unknown[]) => mockMarkTaskFailed(...args),
+  updateNextExecution: (...args: unknown[]) => mockUpdateNextExecution(...args),
+}))
+
+// Import after mocks
+import { taskScheduler } from './task-scheduler'
+
+// ============================================================================
+// Tests
+// ============================================================================
+
+describe('TaskScheduler startup resilience (SUP-224)', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  afterEach(() => {
+    // Reset the singleton so a wedged isRunning never leaks across tests.
+    taskScheduler.stop()
+    vi.useRealTimers()
+  })
+
+  it('does not stay marked running when the initial overdue scan fails', async () => {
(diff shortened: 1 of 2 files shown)


iddogino:
Sentry logging + retries needed

iddogino:
### Update: self-heal via the poll loop + Sentry reporting (answering "should we add retries / Sentry?")

Strengthened the fix beyond the original roll-back approach, and brought the branch up to date with `main` (it was 59 behind; conflict-free — `main` never touched the scheduler).

**Why the original fix wasn't enough:** `taskScheduler.start()` is called **exactly once** (`startup.ts`, fire-and-forget `.catch`-logged). The roll-back made recovery *possible* (`isActive()` → false, a later `start()` would work) but **nothing ever calls `start()` again** — so after a transient initial-scan failure the scheduler stayed down for the whole process lifetime, just reporting `false` instead of a wedged `true`.

**1. Retries — via the existing poll loop, not new machinery.** The 60s poll loop *is* the scheduler's natural retry (same code path as steady state). The only reason a startup hiccup disabled everything was that the `setInterval` install sat *after* the awaited initial scan. So now the poll loop is **installed unconditionally, first**, and the immediate catch-up scan is **best-effort** (report + continue). A transient failure self-heals on the next tick (≤60s); `start()` no longer rejects and `isActive()` stays `true`. This also subsumes the wedge fix — the loop is always installed.

**2. Sentry — yes.** The scheduler `console.error`'d everything, so prod failures were invisible. Now reports via `captureException` (`@shared/lib/error-reporting`, the main-process sink already used across `main/` + API routes) at three tagged spots:
- `phase: 'poll'` — the recurring cycle (highest value; was console-only *forever*),
- `phase: 'initial-scan'` — the startup catch-up,
- `phase: 'execute-task'` — a scheduled task failing to run (with `taskId`/`agentSlug`).

**Tests** rewritten to the new contract: `start()` resolves + stays active on a failed initial scan, the failure is reported, advancing the interval re-scans (self-heal proof); a failed poll cycle is reported without stopping the loop; clean startup reports nothing. Full scheduler suite green (6 files, **52 tests**); `tsc` + `eslint` clean.

**Pre-existing, out of scope (worth its own ticket):** `markTaskExecuted`/`updateNextExecution` run *after* `createSession`, so a crash in that window re-creates a session on the next scan (at-least-once). Re-scanning here has the same semantics as normal polling — no new risk — but the dedupe deserves a separate fix.

PR #238, opened 2026-06-06
fix(notifications): scope auth-mode notifications by user_id (SUP-227)

iddogino:
### Bug
In `AUTH_MODE`, the notification service scoped list/count/mark-read queries by the caller's accessible-agent ACL only and **never** by the per-user owner column `notifications.user_id`, while `createNotification` left `user_id` null. Consequences for users who share an agent:
- They could read each other's notification **bodies and history** (`listNotifications`, `countNotifications`, unread queries all filtered by `agentSlug` only).
- They cleared each other's unread/read state through a single shared `isRead` bit (`markSessionNotificationsRead` / `markAllAsRead` — opening a shared session marked *everyone's* rows read).
- `POST /:id/read` and `DELETE /:id` were gated by `HasNotificationAccess`, which only checked agent role — a teammate could flip or delete a notification they did not own.

### Fix (per-recipient fan-out — the smaller change matching the existing `user_id` column intent)
- **`notification-service.ts`**: `createNotification` gains a `userId` owner param; new `getAgentAclUserIds` (inverse of `getAccessibleAgentSlugs`). Every auth-mode query now *also* filters by `eq(notifications.userId, userId)` — `listNotifications`, `countNotifications`, `listUnreadNotifications`, `getUnreadCount`, `getSessionIdsWithUnreadNotifications`, `getUnreadNotificationsByAgents`, `markSessionNotificationsRead`, `markAllAsRead`. Non-auth behavior (`userId` undefined) is unchanged.
- **`notification-manager.ts`**: session lifecycle events have no single owning user, so in auth mode the event is fanned out to one row per agent ACL member; non-auth mode keeps a single null-owner row. The global SSE broadcast omits a per-user `notificationId` in auth mode (each recipient owns a distinct row).
- **`auth.ts` (`HasNotificationAccess`)**: non-admins must now **own** the row via `isOwnedByCaller` from `@shared/lib/auth/ownership`; non-owned/missing rows return `404` so existence does not leak. Admins retain bypass.
- **`agents.ts`**: thread the scoped `userId` into `enrichAgentsWithSummary` → `getUnreadNotificationsByAgents` and into `getSessionIdsWithUnreadNotifications`, so sidebar/list unread indicators are per-user.
- **`global-notification-handler.tsx`**: when the broadcast carries no `notificationId` (auth-mode fan-out), the mid-view auto-mark falls back to the user-scoped `read-by-session` endpoint so the badge still clears.

### Tests (failing-then-green regression)
New `src/shared/lib/services/notification-service-user-scoping.sup227.test.ts` (in-memory sqlite + drizzle harness). Seeds `agentAcl` so userA and userB both access `shared-agent`, inserts `notif-a` (owner userA) and `notif-b` (owner userB) on the same agent+session, and asserts behavior that **failed on the pre-fix code and now passes**:
- `listNotifications(50, userB)` → `['notif-b']` (was `[notif-a, notif-b]`)
- `getUnreadCount(userB) === 1` (was `2`)
- `markSessionNotificationsRead(session, userB)` marks only `notif-b`; `getUnreadCount(userA)` stays `1` (no shared isRead bit)
- per-user `getUnreadNotificationsByAgents` / `getSessionIdsWithUnreadNotifications` scoping
- route/middleware: `POST /api/notifications/:notifA/read` as userB → `403/404` and `notif-a.isRead` stays false; userB marking their own row → `200`
- non-auth-mode regression guards (userId undefined) preserved

Also updated the stale agent-role `HasNotificationAccess` unit tests in `auth.test.ts` to the new ownership semantics.

Note: notifications are stored as typed columns (no JSON (de)serialization at the DB boundary), so no new Zod schema was required; the SSE `actions` payload remains validated by the existing `NotificationActionsArraySchema`.

### Validation
- New + closely-related existing suites green (`auth.test.ts`, `security-repro.test.ts`, `agents.test.ts`, `notification-manager.test.ts`, `notification-service.test.ts`): 317 passing.
- `npx tsc --noEmit` clean; `eslint` on changed files clean.

### Changed files
- `src/shared/lib/services/notification-service.ts`
- `src/shared/lib/notifications/notification-manager.ts`
- `src/api/middleware/auth.ts`
- `src/api/routes/agents.ts`
- `src/renderer/components/notifications/global-notification-handler.tsx`
- `src/api/middleware/auth.test.ts`
- `src/shared/lib/services/notification-service-user-scoping.sup227.test.ts` (new)

Status: waiting for review.

### DIFF
--- a/src/api/middleware/auth.test.ts
+++ b/src/api/middleware/auth.test.ts
@@ -777,11 +777,9 @@ describe('Auth Middleware', () => {
       expect(mockSelect).not.toHaveBeenCalled()
     })
 
-    it('allows when user has access to notification agent', async () => {
-      // First query returns notification with agentSlug
-      mockLimit.mockResolvedValueOnce([{ agentSlug: 'test-agent' }])
-      // Second query returns user's role on that agent
-      mockLimit.mockResolvedValueOnce([{ role: 'user' }])
+    it('allows when user owns the notification (SUP-227)', async () => {
+      // Notifications are per-user rows; the caller must own this one.
+      mockLimit.mockResolvedValueOnce([{ userId: 'user-1' }])
 
       const app = new Hono()
       setUser(app, { id: 'user-1', role: 'user' })
@@ -791,18 +789,16 @@ describe('Auth Middleware', () => {
       expect(res.status).toBe(200)
     })
 
-    it('returns 403 when user has no access to notification agent', async () => {
-      // First query returns notification with agentSlug
-      mockLimit.mockResolvedValueOnce([{ agentSlug: 'test-agent' }])
-      // Second query returns no role (no access)
-      mockLimit.mockResolvedValueOnce([])
+    it('returns 404 when the notification is owned by another user (SUP-227)', async () => {
+      // Row exists but belongs to a teammate — agent ACL access is not enough.
+      mockLimit.mockResolvedValueOnce([{ userId: 'user-2' }])
 
       const app = new Hono()
       setUser(app, { id: 'user-1', role: 'user' })
       app.get('/:id', HasNotificationAccess(), (c) => c.json({ ok: true }))
 
       const res = await request(app, '/notif-1')
-      expect(res.status).toBe(403)
+      expect(res.status).toBe(404)
     })
 
     it('returns 404 when notification does not exist', async () => {
--- a/src/api/middleware/auth.ts
+++ b/src/api/middleware/auth.ts
@@ -5,6 +5,7 @@ import { runWithRequestUser } from '@shared/lib/platform-attribution'
 import { db } from '@shared/lib/db'
 import { agentAcl, connectedAccounts, remoteMcpServers, notifications } from '@shared/lib/db/schema'
 import { validateProxyToken } from '@shared/lib/proxy/token-store'
+import { isOwnedByCaller } from '@shared/lib/auth/ownership'
 
 // Lazy import to avoid pulling in better-auth ESM at import time
 let _getAuth: (() => ReturnType<typeof import('@shared/lib/auth/index').getAuth>) | null = null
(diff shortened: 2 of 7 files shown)


iddogino:
I think the current state is actually desired - even in a shared state -- if I see that the agent finished (and mark as read) - like doesn't require somebody else's attention. Seems to me current behavior is correct, and the user_id columns should just be removed as unused. What am I missing?

iddogino:
Closing in favor of #246.

After review, the premise behind this PR — that notifications should be per-user — doesn't hold. Every notification type is a session-lifecycle fact on a shared agent (`session_complete`, `session_waiting`, ...), and its body is a strict subset of session content that any ACL member can already see by opening the session. So:

- **No confidentiality benefit:** per-user scoping adds nothing — message access is gated by the same per-agent ACL, with no per-session sub-ACL.
- **Shared read-state is the desired behavior, not a bug:** a teammate acknowledging a "waiting"/"complete" item *should* clear it for the team (shared worklist). Per-user fan-out would *contradict* that intent.
- **`user_id` was dead, not half-built:** added speculatively in the auth-support pass (comment copied from `connected_accounts`), never set by the sole producer, never read by any query — every deployed row is NULL.
- **Cross-agent isolation never depended on `user_id`:** it's enforced by `getAccessibleAgentSlugs()` + `HasNotificationAccess`, both unchanged.

The cost of fan-out (N× write amplification, a NULL-`user_id` branch on every query, a wider leak surface where any forgotten filter = cross-user read) isn't justified absent a *targeted* notification type (@-mention / per-approver), which isn't on the roadmap.

**#246** instead drops the dead column, documents the shared-scope decision at the schema, and adds a regression test locking in (1) cross-agent isolation and (2) shared read-state — so the model can't silently regress. The agent-role `HasNotificationAccess` gate on the by-id routes is kept verbatim.

PR #250, opened 2026-06-10
feat: queue messages while the agent is working

iddogino:
## What

Users can now send messages while the agent is mid-turn instead of the composer being disabled. Sent messages appear as **"queued" ghosts** (reduced opacity + label) at the bottom of the thread, and **materialize** into regular transcript messages when the agent loop picks them up. Works for multiple queued messages (FIFO) and for peers in shared sessions.

## How

**Composer** (`message-input`, `composer-action-button`)
- Stop + Send render together while active; Enter/click queues instead of being blocked
- Model/effort selectors are locked mid-turn — changing them would interrupt/restart the running query — and queued sends strip runtime options from the payload

**Optimistic ghosts** (`pending-message.ts`, `main-content`, `message-list`)
- Pending messages are per-session **arrays**; each send carries a client-generated uuid through POST → container → SDKUserMessage → session JSONL
- Materialization matches by uuid first, then falls back to trimmed text + time window: the CLI **re-ids mid-turn input on enqueue** (`queued_command.source_uuid` is regenerated), so queued messages never come back under the client uuid
- Each persisted message id is claimed at most once, so two identical queued texts cannot be cleared by one persisted copy
- The 10s safety-net cleanup now only runs while idle (queued ghosts may legitimately wait minutes); a failed POST drops the ghost immediately and restores the input text

**Transcript** (`session-service`, `message-transform`)
- The CLI records mid-turn messages as `queued_command` **attachment** entries, not `user` entries — previously they were invisible in history (and lost on reload) even though the agent acted on them. They are now converted to synthetic user entries (id = `source_uuid`, timestamp = pickup time)
- Converted messages are tagged `queued` and excluded from all turn-boundary logic: elapsed-time dividers, delivered-files grouping, running-tool detection, and pending-request dismissal — a steering message does not end the turn it appears in

**Multi-user** (`agents.ts`, `use-message-stream`)
- The `user_message` SSE broadcast now carries `uuid` + `queued` (computed server-side before `markSessionActive`); peer ghosts are an array with the same uuid/text materialization lifecycle
- Known limitation: sender attribution (`messageAuthor`) joins on the client uuid, which the CLI discards for mid-turn input, so a peer's *queued* message materializes without their name label

**Mock container** (`mock-container-client`)
- Tracks per-session busy state (scenario start → `result`); messages sent while busy take a steering path that mirrors the real CLI exactly — no user entry, then a `queued_command` attachment with a **regenerated** `source_uuid` at pickup, followed by assistant output. New `SlowWorkScenario` (`"work slowly"`) holds a ~5s working window for tests

## Testing

- **E2E** (`e2e/specs/message-queueing.spec.ts`): mid-turn send → queued ghost → materialization → ordering; multiple queued messages drain in send order. Plus chat-flow / cross-session / model-selection / tool-rendering regressions (20 specs) against the new mock busy behavior — all green
- **Unit**: ~25 new tests — queued-send payload shape, ghost rendering/ordering, uuid + fallback + claim matching, turn-boundary exclusions, `queued_command` conversion, host uuid passthrough + queued broadcast, and 7 tests for the `use-message-stream` peer-message lifecycle
- Verified against a real session: queued steering message now persists, materializes, and renders without a spurious turn divider


iddogino:
## Code-review round (7-angle local review, verified findings)

A multi-agent review (line-by-line, removed-behavior, cross-file, reuse, simplification, efficiency, altitude — each finding independently verified) surfaced 9 confirmed issues, all fixed in 92ddcbe9:

**Design change vs. the original description:** message UUIDs are now **always server-generated** — `POST /messages` returns `{ uuid, queued }` and create-session returns `initialMessageUuid`; the composer keys ghosts by a local correlation id and re-keys from the response. Clients can no longer choose the `messageAuthor` PK (closing a cross-user misattribution vector), and `onConflictDoNothing` is gone so collisions hard-fail again.

Other fixes:
- Server strips `model`/`effort` on mid-turn sends (`wasQueued`) — a stale client can no longer interrupt an in-flight turn with runtime options.
- Queued messages lost to an interrupt show a dismissible "Not delivered" notice instead of vanishing silently.
- Text-fallback ghost matching is gated to queued sends / pre-response sends; turn-starting ghosts can't be cleared by an older identical-text message.
- Peer-ghost text matching is bounded by arrival time; the sender's own SSE echo is pruned immediately (it was suppressing peers' typing indicators for the whole turn).
- Deleting a queued message works (`removeMessage` resolves `attachment.source_uuid`); session naming/counts include queued messages; activity-indicator elapsed time skips them after refresh.
- Scroll re-pin keyed on seen ghost ids (fixes a send-races-removal miss).
- Cleanup: one ghost render path, one `isTurnStartingUserMessage` predicate across all five turn-boundary sites, memoized peer filtering.

Refuted during verification (no change needed): stuck-forever ghosts (covered by reconnect re-sync + ping safety net), resume-time duplicate ids (replay is stream-only; using `source_uuid` as the id is what *dedupes* it), and version-skew on uuid-less broadcasts (renderer always ships with the host).

Validation: typecheck/lint clean, all affected unit suites pass (505 tests), E2E queueing + chat-flow + cross-session 10/10.

iddogino:
## Added: cancel queued messages (acca5be8)

Queued ghosts now have a **Cancel** action. Verified by a live spike against CLI 2.1.170 first: the SDK's undocumented `cancel_async_message` control request (`query.cancelAsyncMessage(uuid)` in sdk.mjs) drops a pending message from the CLI command queue **by the uuid we send with it**, returns `false` when it lost the race to pickup, and a cancelled message leaves zero transcript trace.

Flow: ghost Cancel button (enabled once the POST response delivers the server uuid) → `DELETE /api/agents/:id/sessions/:sessionId/queued-messages/:uuid` → container endpoint → splice from our MessageQueue buffer or `cancelAsyncMessage`. `cancelled: false` = too late; the ghost is left to materialize normally. Mock container registers steering injections per-uuid (pickup delay widened to 1.2s) so the E2E cancel window is deterministic.

Tests: 3 route tests, 3 message-list tests, 1 E2E (queue mid-turn → cancel → message never materializes, no steering ack, turn completes). Full queueing + chat-flow E2E green.

iddogino:
﻿## Idle lifecycle rework (pushed in `ac23bdd8`)

Live testing surfaced a premature-idle bug: queue a message while the agent streams its **final** response and the session flips idle (working indicator off, completion notification fires) before the queued turn runs.

**Root cause:** the message arrives too late to be steered into the running query, so the CLI completes the query at the first `result` **without emitting any `session_state_changed` event** — a CLI run starts in `running` and only publishes its first transition (the final `idle`) at the very end. The container then starts a continuation run for the queued message ~70ms later. Any host-side design that *discovers* state-event support from the first observed event therefore can''t engage in time.

**Design now:**
- The runtime''s `session_state_changed:''idle''` is the authoritative end-of-session signal; a `result` alone doesn''t end the session.
- The container announces `capabilities { session_state_events: true }` on WebSocket connect, before relaying any SDK message (WS is FIFO), so the host is in authority mode from the session''s first turn.
- All the previous host-side heuristics (queued-uuid shadow set, idle-suppression flag, activity-clear, fallback timer) are deleted — net code removal in the persister.
- Containers without the hello degrade gracefully to the original result-driven idle.

**Regression coverage:** `queued-message-idle-replay.test.ts` replays a real capture of the bug (transcript payloads + container log ordering from the repro session) through `MessagePersister` and asserts the session stays active across the inter-query gap with exactly one idle. Verified the test fails on the pre-fix code. Also validated end-to-end against the live app via the session SSE stream (queued POST mid-stream → no `session_idle` until the queued turn settles).

⚠️ **Heads-up for anyone testing this branch:** rebuild the agent container image (`npm run build:container`) and recreate the container — a stale container silently falls back to legacy idle behavior.

PR #275, opened 2026-06-16
feat: stale-session prompt (branch-with-summary on idle + large context)

JeremyBischoff:
## Summary

Adds a "stale-session" prompt: when a user sends into a session that is both idle for a while **and** holding a large context, a modal offers to start fresh instead of silently re-reading 100k+ tokens on every message. Goal is to cut per-message cost and teach the user when a session has gotten expensive.

The prompt fires only when **both** gates pass:

- idle > 6h (last activity), **AND**
- current context > 100k tokens

It is suppressed when the session is dismissed (per-session "won't ask again"), awaiting input (parked on a permission), or actively running. Brand-new sessions never prompt.

## What the modal offers

1. **Continue from a summary** (recommended) — branches to a new session carrying a short AI summary of the prior thread, so the new session starts cheap but keeps continuity. The carried context renders as a collapsed "Continued from previous session" card with a **View original session** link back to the source.
2. **Start a new topic** — fresh session, your message sent verbatim, nothing carried over.
3. **Send here anyway** — sends into the stale session and won't ask again for it.

Backdrop click or Esc cancels and restores the typed draft.

## How it works

- **Trigger** (`stale-session-trigger.ts`): pure `evaluateStalePrompt` (AND gate + suppressors), gated at the composer submit.
- **Cost/occupancy** (`message-cost.ts`): `currentContextTokens` + `estimateNextMessageCostUsd`, reusing the existing pricing table.
- **Summary** (`session-summary-service.ts`): budgeted-recency slice of the transcript (reuses a prior `compact_boundary` summary when present), Haiku via the configured provider, composes the injected first message for the branched session.
- **Branch** (`POST /sessions/branch`): builds the summary, starts the container, creates the session, returns the initial-message uuid for optimistic render. Real summary failure returns `502 summary_failed`.
- Sessions are file-based: the dismissal flag lives in `session-metadata.json`. **No DB migration.**

Thresholds (6h / 100k / summary budget) are code-level constants in `stale-session-config.ts`, deliberately tunable from real usage (see the spec's open calibration items).

## Pre-merge hardening (from an adversarial review pass)

- Removed a 15s client timeout that wrapped the whole branch request (summary + container cold-start + session init) and mislabeled slow-but-successful branches as failures.
- Summarizer JSON parsing tolerates markdown-fenced output (Haiku wraps JSON in a ```json fence).
- Branch-card split anchors on the genuine (last) transcript-path line and charset-validates the back-link id, so a path string or `---` inside the LLM summary can't mis-split the card or forge the back-link target.
- Modal locks the other two actions while one is in flight (was creating duplicate branched sessions on an impatient second click).
- Optimistic render: the typed message shows immediately on branch/new-topic instead of a blank bubble.
- Extracted the pricing primitive into a Node-free leaf so the renderer graph no longer pulls `fs`/`path`/`p-limit`.

## Testing

- Unit: full suite green (5193 passed, rebased onto current main). New coverage for the trigger gate, summary parsing (incl. fenced + forged-path cases), `loadTranscript`, cost helpers, and the back-link.
- E2E (`e2e/stale-session-prompt.spec.ts`): trigger/suppression/actions/dismissal, backdrop-cancel, and new-topic delivery. 40 passed.
- Typecheck + lint clean.

## Known follow-ups (low severity, intentionally deferred)

- The "Continued from previous session" card is detected from message text (`startsWith(SENTINEL)`); a user could craft a message in their own session that renders a lookalike card. Self-targeted, within-agent only. Proper fix is a trusted metadata flag.
- Branch routes are agent-scoped (matching every other session route), not session-owner-scoped.
- `loadTranscript` reads its own transcript without a Zod parse, consistent with the existing `transformMessages` pattern.


JeremyBischoff:
Superseded by #276 — switched from a fork PR to a same-repo PR so CI runs with the signing secrets (build-mac/build-win/comment can't run on a fork). Identical branch and commits.

PR #276, opened 2026-06-16
feat: stale-session prompt (branch-with-summary on idle + large context)

JeremyBischoff:
## What this does

When a conversation goes stale (idle 6h+ AND context over ~100k), a footer toast offers starting a new conversation under the same agent, with or without a carried summary. Keeps long-running agents fast and cheap without losing continuity.

## Flow

- Toast appears at rest on a stale conversation (suppressed while active, awaiting input, or view-only).
- **Start with Summary:** prune the source transcript (actions-in, dumps-out) -> summarize to plain markdown (via the active provider's configured summarizer model) -> land on the agent's new-chat composer with the composer carried over and a collapsed summary card. No session exists yet. On send, a new session is created seeded with the summary, and a back-link context card renders in the thread.
- **Start fresh:** carry the composer into a clean new chat, no summary.
- Session naming and the optimistic echo always use the bare user message, never the seed.

## Key decisions

- **Lazy create-on-send.** No orphan sessions; mirrors Start fresh.
- **Model and provider agnostic.** The summarizer resolves its model via `resolveActiveProviderModel(summarizerModel, 'summarizer')` against the active provider's catalog, falling back to that provider's summarizer default. Nothing is hard-coded to Haiku or Anthropic, so the flow rides whatever provider and model the user has configured. Matches the other summarizer call sites (naming, agent-template, skillset, scheduled-tasks).
- **Build vs buy.** No drop-in Anthropic vendor fits: compaction / context-editing is observe-only and cannot be triggered on demand from the SDK, so it cannot drive a user-initiated, pre-navigation summarize. We prune locally and summarize. Output is plain markdown rather than Structured Outputs because it feeds both a human-readable card and the next session's seed; a JSON envelope added parsing for no gain and reintroduced a fence-parse failure the earlier version hit.
- **Splitter untouched.** The post-send context card is drawn by the existing message-transform splitter, which anchors on the last in-container `.jsonl` path line. buildSeed preserves that line; the splitter is not modified.
- **Auth + boundaries.** New endpoints use the same AgentUser scope as the sibling create; `fromSessionId` is regex-scoped to the agent dir and checked for existence; request/response are Zod-validated.

## Since West's UI review

West's polish pass (reframed copy, spacing, always-open row blurbs, forward chevrons, rounded corners) is rebased in (725917ad). User-facing vocabulary is "conversation" throughout.

## Reconciled with main (model catalog)

Re-merged main, which now carries the per-provider model catalog (#286) and the wizard e2e rebrand fix (#290). Conflict resolution kept the first-session Opus default working under the new catalog model (a bare family alias the popover resolves to the family's latest), and the summarizer migration above rides the catalog that #286 introduced. The re-merge conflict resolution and the summarizer migration each passed an independent opus 4.8 + codex adversarial review with zero Critical/Important findings.

## Review hardening

Ran a 2x2 adversarial audit (opus 4.8 + codex, across coverage and principles). Fixed: a real bug where Start fresh could inherit a stale staged summary; summarize now 404s on a missing source transcript; dropped a decorative schema duplicate and dead code; tightened the empty-summary guard. Added coverage for the buildSeed shape, bare-message naming, the alwaysShowBlurb branch, and the Start-fresh regression.

## Testing

typecheck clean, lint 0 errors. Feature unit suites and the stale-session e2e (9 specs, E2E_MOCK) green after the re-merge. Manually smoke-tested the toast, both popovers, the Start-with-Summary happy path, and the Start-fresh-after-abandoned-summary path.


src/renderer/components/ui/dialog.tsx
```diff
@@ -26,8 +26,8 @@ DialogOverlay.displayName = DialogPrimitive.Overlay.displayName
 
 const DialogContent = React.forwardRef<
   React.ElementRef<typeof DialogPrimitive.Content>,
-  React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content>
->(({ className, children, onKeyDown, ...props }, ref) => (
+  React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content> & { showCloseButton?: boolean }
```
iddogino, line 29: Not sure we want to hide the close button in this feature -> could be a good escape hatch (confused user just wanting to close out of the modal)

src/renderer/components/agents/agent-home/agent-home.tsx
```diff
@@ -84,7 +86,21 @@ export function AgentHome({ agent, onSessionCreated }: AgentHomeProps) {
   const [sessionSearch, setSessionSearch] = useState('')
   const [sessionSort, setSessionSort] = useState<SortOrder>('newest')
   const { data: sessionsData } = useSessions(agent.slug)
-  const composerOptions = useComposerOptions()
+  // First-session Opus default: when the session list has finished loading
+  // and is empty, the brand-new agent's first session should default to Opus
+  // regardless of the user's "Default Model" setting. Passed as a preferred
+  // family — the user can still pick a different model in the dropdown.
+  const isFirstSession = Array.isArray(sessionsData) && sessionsData.length === 0
```
iddogino, line 93: Deprecated

src/renderer/components/layout/session-chat-column.tsx
```diff
@@ -48,79 +74,214 @@ export function SessionChatColumn({
 
   const renderCtx: RenderContext = { sessionId, agentSlug, readOnly: isViewOnly }
 
+  // --- Stale-session detection (continuous, not send-gated) ---
```
iddogino, line 77: That is a lot of logic here.
1/ Can it be simplified? (best case). Any business logic we can streamline? This is a nightmare...
2/ If not - fallback - needs to be extracted to a standalone hook / function

src/renderer/components/layout/session-view.tsx
```diff
@@ -143,6 +143,9 @@ export function SessionView({ agentSlug, sessionId }: SessionViewProps) {
             onMessageSent={onMessageSent}
             onMessageUuidAssigned={onMessageUuidAssigned}
             onMessageFailed={onPendingMessageAppeared}
+            lastActivityAt={session?.lastActivityAt ? new Date(session.lastActivityAt) : null}
+            contextUsage={contextUsage}
+            onStartFresh={() => void navigate({ to: '/agents/$slug', params: { slug: agentSlug } })}
```
iddogino, line 148: Already passing agent slug can't we use that internally to compute the link ? Versus needing an extra prop

src/renderer/components/messages/composer-options.tsx
```diff
@@ -53,10 +53,19 @@ export interface UseComposerOptionsArgs {
...
   initialModel?: string
+  /**
+   * Bare family alias (e.g. 'opus') for the initial model when neither
+   * `initialModel` nor a prior user selection applies. Wins over the user's
+   * "Default Model" setting. Used by AgentHome to start brand-new agents on
+   * Opus. Only consulted until the selector is seeded — once the user picks,
+   * their choice takes over. The popover resolves the alias to the family's
+   * latest via findCatalogModel.
+   */
+  preferredFamily?: string
```
iddogino, line 64: We deprecated families since this PR opened. Need to be brought inline with newer standards...

JeremyBischoff:
Superseded. #451 shipped the stale-session toast and #1044 carries the summary fork.

PR #277, opened 2026-06-16
feat(telegram): Bot API 10.1 rich messages

JeremyBischoff:
## Summary

Adopt Telegram Bot API 10.1 Rich Messages in the Telegram connector. Agent markdown now renders as native rich content (real tables, code blocks, lists) instead of the old HTML/`<pre>` approximation. DM responses stream via animated rich drafts; groups stream via throttled rich edits. The legacy HTML path is kept as a per-send fallback and a global rollback flag.

## What changed

- Bump grammy to `^1.44.0` (`@grammyjs/types` 3.28.0) for the typed rich API.
- A near-identity markdown -> `InputRichMessage` passthrough converter (Telegram parses GFM server-side; we do not build `RichBlock` objects).
- Typed `sendRichMessage` / `sendRichMessageDraft` / `editMessageText` with `rich_message` (no `as never` casts).
- DM streaming: animated `sendRichMessageDraft`, committed to a real message on finalize.
- Group/channel streaming: throttled rich `editMessageText`, with HTML fallback on any rich-send error.
- "Thinking…" indicator: Telegram's native `RichBlockThinking` block (`<tg-thinking>`), sent via `sendRichMessageDraft` (DM-only; groups use the typing action). A ~1s keep-alive re-send (`THINKING_REFRESH_MS`) covers the ~30s draft expiry, and the draft shares the streaming `draft_id` so the response replaces it. The block's text is a plain ✨: the recommended AIActions custom emoji are omitted because Telegram strips bot-sent custom emoji unless the bot's owner has Telegram Premium, which would leave a worse fallback for most users.
- `markdownToTelegramHtml` retained as the rich-send error fallback and the `richMessages` rollback flag.

## Testing

- Unit: 410 chat-integration tests (converter golden corpus, DM draft + group streaming, finalize/split, native `<tg-thinking>` draft + keep-alive re-send + group typing action, HTML fallback + 4096 re-split).
- Typecheck and lint clean.
- Full unit suite passes modulo the pre-existing webhook/scheduler DB-isolation flakiness, which is unrelated to this branch (no files in those areas are touched).
- No Playwright e2e: the connector is backend-only with no UI change.
- Manually exercised against a live Telegram bot: DM streaming, markdown rendering, the native thinking indicator, and tool-call pills.

## Adversarial review

Ran a multi-agent adversarial review over the diff (concurrency, correctness, Bot-API misuse, resource leaks, security, test gaps). Confirmed findings, all addressed here:

- Two thinking-timer leaks: `clearChatSessionById` and `removeIntegration` deleted a managed session without `stopThinking`, leaking the 1s keep-alive interval and orphaning the indicator (the bot stays connected). Now mirror `stop()` / `teardownManagedSession()`.
- The group first streaming send bypassed the HTML fallback; now routed through `sendRichOrHtml`.
- The HTML fallback chunked at the 32768 rich ceiling but sends to the 4096 `sendMessage` sink; long messages on the rollback/fallback path were rejected. Now re-split at 4096.
- Added the missing `editRichOrHtml` HTML-edit fallback test.

## Known limitations

- The native `RichBlockThinking` block is used for the "Thinking…" *indicator* only. Using it to render the model's actual reasoning, and collapsible `RichBlockDetails`, is out of scope per the design spec. Math still renders, because Telegram parses the notation server-side from the passed-through markdown.
- Old-client rendering for pre-10.1 Telegram apps is undocumented and untrappable (there is no plain-text field on `InputRichMessage`). Mitigated by the `richMessages` rollback flag.
- A DM draft response can be lost if the session is torn down mid-stream before finalize, or on a >30s streaming stall; the ephemeral draft is by design and the self-heal path already tells the user to retry.
- An over-4096 HTML *edit* on the rollback path is best-effort (an edit cannot be split into multiple messages); the final-message *send* path is fully chunked.


src/shared/lib/chat-integrations/base-connector.ts
```diff
@@ -68,6 +68,14 @@ export abstract class ChatClientConnector {
   /** Show typing / processing indicator. */
   abstract showTypingIndicator(chatId: string): Promise<void>
 
+  /**
+   * Tear down any "thinking" indicator owned by showTypingIndicator (e.g. a
+   * posted placeholder message) when the response is about to take over. Default
+   * no-op for connectors whose typing indicator is ephemeral (typing actions,
+   * reactions).
+   */
+  async clearThinking(_chatId: string): Promise<void> {}
```
iddogino: Is there an opportunity to unify more between the thinking and typing indicator mechanisms? 
Maybe we create a more abstract "startShowWorking(chatId)" and "stopShowWorking(chatId)" ? 

I don't love that we have these as two parallel concepts now.. Feels implementation specific

src/shared/lib/chat-integrations/chat-integration-manager.ts
```diff
@@ -125,6 +125,8 @@ export interface ManagedConnector {
   }
   currentToolInput: string
   pendingToolMessages: Array<{ messageId: string; text: string }>
+  /** Animation timer editing the posted "Thinking…" message until the first token streams. */
+  thinkingTimer: ReturnType<typeof setInterval> | null
```
iddogino: Does this belong here or in implementation classes? Feels implementation specific

src/shared/lib/chat-integrations/chat-integration-manager.ts
```diff
@@ -193,6 +195,7 @@ class ChatIntegrationManager {
     // Clean up all chat session SSE subscriptions
     for (const [, session] of this.chatSessions) {
       session.sseUnsubscribe?.()
```
iddogino: The combo:

```
session.sseUnsubscribe?.()
stopThinking(session)
```

should appear together often. Maybe create a single 'stopSession' function that wraps them both ?

src/shared/lib/chat-integrations/chat-integration-manager.ts
```diff
@@ -1441,6 +1446,39 @@ export async function processSSEEvent(
...
+ */
+export function startThinking(managed: ManagedConnector): void {
+  stopThinking(managed)
+  managed.connector.showTypingIndicator(managed.chatId).catch(() => {})
+  managed.thinkingTimer = setInterval(() => {
+    managed.connector.showTypingIndicator(managed.chatId).catch(() => {})
+  }, THINKING_REFRESH_MS)
+}
+
+export function stopThinking(managed: ManagedConnector): void {
```
iddogino: `stopThinking` is called on every `stream_delta`. The if guards only `clearInterval`, `clearThinking()` runs every token. For Telegram this is a harmless no-op (it doesn't override the base method), so no bug today. But the base class documents `clearThinking` as a hook that may "actively remove an indicator" (e.g. delete a placeholder message via an API call). The moment Slack/iMessage implements it non-trivially, this becomes one API call per streamed token.

Move the call inside the guard so it only fires on the thinking→streaming transition

src/shared/lib/chat-integrations/telegram-connector.ts
iddogino: Should this be rich format aware as well? Hardcoded to HTML right now...

src/shared/lib/chat-integrations/telegram-connector.ts
```diff
@@ -436,46 +502,39 @@ export class TelegramConnector extends ChatClientConnector {
...
         }
 
         return lastMessageId
       }
 
       case 'secret_request': {
-        const text = `<b>Secret requested:</b> <code>${this.escapeHtml(event.secretName)}</code>\n${event.reason ? `\nReason: ${this.escapeHtml(event.reason)}` : ''}\n\nPlease reply with the secret value.`
-        const sent = await this.bot.api.sendMessage(chatId, text, { parse_mode: 'HTML' })
-        return String(sent.message_id)
+        const text = `**Secret requested:** \`${event.secretName}\`\n${event.reason ? `\nReason: ${event.reason}` : ''}\n\nPlease reply with the secret value.`
```
iddogino: The migration replaced escapeHtml(event.secretName) etc. with raw interpolation into markdown:
```typescript
const text = `**Secret requested:** \`${event.secretName}\` ...`
```
 With `skip_entity_detection` defaulting OFF, interpolated values (secretName, filePath, toolName, summary, q.question) are parsed by Telegram. A value containing a backtick/asterisk breaks the code span or formatting. 

It's not full-on XSS — this is the user's own (semi-trusted) agent content — so severity is low, but a safety layer was silently removed. A small escapeMarkdown() helper for the structural literals, or passing `skip_entity_detection: true` on these would solve.

src/shared/lib/chat-integrations/telegram-connector.ts
```diff
@@ -2,28 +2,37 @@
...
+  skipEntityDetection?: boolean
 }
 
 // ── Telegram limits ─────────────────────────────────────────────────────
 
-const MAX_MESSAGE_LENGTH = 4096
 const FIRST_POLL_BATCH_DELAY_MS = 500
+const RICH_DRAFT_SENTINEL_PREFIX = 'draft:'
+// Q4 A/B switch: true = stream rich intermediates (group path); false = stream legacy HTML, finalize rich.
+const STREAM_RICH_INTERMEDIATES = true
```
iddogino: Is this a dead flag?

src/shared/lib/chat-integrations/telegram-rich-message-schema.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Zod schema for Telegram Bot API 10.1 InputRichMessage — the typed contract for
+ * a rich-message payload, where exactly one of `html` | `markdown` is present.
+ * The converter builds this shape directly; the schema is the source of truth for
+ * the `InputRichMessage` type and is exercised by the converter tests.
+ */
+import { z } from 'zod'
+
+export const inputRichMessageSchema = z
```
iddogino: I think GrammY exports a `InputRichMessage` type - should we adopt that rather than duplicate ?

src/shared/lib/chat-integrations/base-connector.ts
iddogino: ### Proposed shape:

```typescript
  // base-connector.ts
  /**
   * Signal the agent is working (processing, no output yet). The connector owns
   * the representation AND any keep-alive needed to survive provider-side expiry.
   * Idempotent — safe to call repeatedly for the same chat.
   */
  async startWorking(_chatId: string): Promise<void> {}
  /** Stop the working indicator as the response takes over. Idempotent. */
  async stopWorking(_chatId: string): Promise<void> {}
```
The key move: the keep-alive timer moves into the connector, right next to the activeDrafts map it already drives. Telegram keeps a Map<chatId, Timer> and runs its own 1s interval; Slack and iMessage just  add/remove with no timer.

### What that buys you:
  - Manager gets simpler — thinkingTimer, THINKING_REFRESH_MS, startThinking/stopThinking all disappear; call sites become connector.startWorking(chatId) / stopWorking(chatId).
  - Timer is co-located with the state it drives, both in the connector, both keyed by chat. No more split-brain.
  - Slack stops churning (no more 1s no-op re-polls).
  - Fixes the inter-segment known-limitation for free: today stream_start does a one-shot showTypingIndicator with no heartbeat, so a >30s gap between segments drops the indicator. Under startWorking (self-keep-alive), it survives.

iddogino (changes requested):

src/shared/lib/chat-integrations/base-connector.ts
JeremyBischoff: Done in `ff1e1066`. Went with your shape: `startWorking`/`stopWorking` on the base connector, and the keep-alive timer now lives in the Telegram connector next to its draft state. Manager loses `thinkingTimer`, `THINKING_REFRESH_MS`, `startThinking` and `stopThinking`. Slack just adds/removes its reaction; iMessage sends one typing action with no timer, so a long idle gap there won't refresh. Flag if that matters.

src/shared/lib/chat-integrations/chat-integration-manager.ts
JeremyBischoff: Gone in `ff1e1066`. Removed from `ManagedConnector`; the timer lives in the Telegram connector now, keyed by chat, next to its draft map.

src/shared/lib/chat-integrations/chat-integration-manager.ts
JeremyBischoff: Done in `ff1e1066`. Added a `stopSession()` helper wrapping the unsubscribe plus the indicator stop; all four teardown sites and `teardownManagedSession` call it now.

src/shared/lib/chat-integrations/chat-integration-manager.ts
JeremyBischoff: Fixed in `ff1e1066`. The stop now fires once on the thinking-to-streaming transition, guarded on the empty accumulator, instead of on every token. It's the connector's `stopWorking`, which is idempotent.

src/shared/lib/chat-integrations/telegram-connector.ts
JeremyBischoff: Done in `e3895a91`. Routed the confirmation edit through `editRichOrHtml` so it matches the connector's mode instead of hardcoding HTML. Pulled the handler into a named `handleCallbackQuery` (like `handleTextMessage`) and added a test; dropped the now-unused private `escapeHtml`.

src/shared/lib/chat-integrations/telegram-connector.ts
JeremyBischoff: Done in `10565dd4`. Added `escapeMarkdown()` for the interpolated values, plus a `codeSpan()` fence for `secretName`/`filePath` so a backtick can't break the code span either. Didn't use `skip_entity_detection`: it only governs auto-entity detection, not markdown structure, so it wouldn't fix this.

src/shared/lib/chat-integrations/telegram-connector.ts
JeremyBischoff: Yes, dead. Removed in `297cc3de`; it was always true, so both conditions collapse to `this.useRich`. Removed it rather than wiring to config. Shout if you wanted it kept as a real toggle.

src/shared/lib/chat-integrations/telegram-rich-message-schema.ts
JeremyBischoff: Done in `1118c619`. Adopted grammy's `InputRichMessage` from `grammy/types` and deleted our duplicate schema. It's the exact type the raw API expects, so it's also more correct.

src/shared/lib/chat-integrations/base-connector.ts
```diff
@@ -65,8 +65,19 @@ export abstract class ChatClientConnector {
...
+   * expiry. Idempotent — safe to call repeatedly for the same chat.
+   */
+  abstract startWorking(chatId: string): Promise<void>
+
+  /**
+   * Stop the working indicator as the response takes over. Idempotent — safe to
+   * call repeatedly. Default no-op for connectors whose indicator is ephemeral
+   * and self-expires.
+   */
+  async stopWorking(_chatId: string): Promise<void> {}
```
iddogino, line 80: nit: why is it `_chatId` here and `chatId` for startWorking above?

iddogino (changes requested):

src/shared/lib/chat-integrations/base-connector.ts
JeremyBischoff: They differ because one uses the value and one doesn't. `stopWorking`'s base version is an empty default no-op, so it ignores `chatId`, and the `_` is what tells the linter that's intentional. `startWorking` is abstract with no body, so there's nothing for the linter to check, and the real implementations actually use `chatId`, so it keeps the plain name.

iddogino: Fair enough

iddogino (approved):

PR #293, opened 2026-06-20
Telegram chat-integration allowlist (access control for personal bots)

JeremyBischoff:
# Telegram chat-integration allowlist (access control for personal bots)

## What

Adds per-chat access control to chat integrations (Telegram first). A bot is private by default: it only responds to chats the owner has approved. Unapproved chats are told once that approval is needed, then stay silent until approved.

## Why

Today any chat that finds a connected bot can talk to the agent and spend tokens. For personal bots that is wrong by default. This makes a bot private unless the owner opts a chat in, with a low-friction path for the common case (the owner's own first DM auto-allows).

## How it works

- New `chat_integration_access` table: one row per (integration, externalChatId) with status `pending | allowed | denied`, plus `requireApproval` on the integration (default **true**).
- Inbound gate (`decideInboundAccess` / `isChatAllowed`) runs before any token spend, before `/clear` and `/start`:
  - First **private** contact when zero allowed rows exist bootstraps to `allowed` (auto first contact), so the owner is not locked out of their own bot.
  - Otherwise the chat is recorded `pending` and gets exactly one "needs approval" notice; further messages are silent.
- Revoke re-checks before **every** spend site (new-chat, existing-session send, and the send-time self-heal), so a revoke landing mid-flight cannot spend.
- Fail-closed SSE forwarding: agent output is dropped for any chat that is not currently allowed, even if a session is live.
- Owner-only management: approve / deny / revoke and the make-public toggle require the `owner` role (in auth mode; local single-user mode is unaffected).
- UI: pending/allowed/denied lists with per-row loading and error states, Approve-primary / Deny-neutral actions, an owner-only "N pending" badge on the agent's Remote Chat list, and a require-approval toggle with a "public bot" warning.

## Hardening (review passes)

After the initial build, three adversarial whole-branch passes (opus + codex, against our code principles and prior PR feedback) drove these fixes:
- Make-public is owner-only at every layer: a dedicated `PATCH /:integrationId/require-approval` (owner role); `requireApproval` removed from the `user`-role create route **and** the create service path (new integrations are always private; the schema default is the backstop); enabling approval reconciles and tears down previously-public live sessions.
- `GET /:integrationId/access` owner-gated (it exposes requester identity + message previews shown only to owners).
- Fail-closed `isChatAllowed` recheck in `handleSSEEvent`; revoke re-check before the send-time self-heal spend (the one path upstream's self-heal refactor left ungated); Zod-validate `requireApproval`.
- Approval UI scoped to Telegram (the gate is Telegram-only) so the toggle + pending badge aren't shown for providers they don't protect.
- Tests: role-aware middleware mock with `user`→403 assertions on every owner route; create-ignores-requireApproval; mid-flight-revoke self-heal e2e.
- DRY: `useChatAccessAction(verb)` factory; `AccessActionButton`/`chatLabel` extraction. Cleanups: removed unreachable code + a dead `.catch`; reused the `preview()` helper.
- Migration cleanup: dropped the duplicated seed const (test reads the shipped `0024 .sql`), added the missing `0024` snapshot. The `0024` journal `when` is pinned by a test (raising it re-runs the non-idempotent seed and crashes existing DBs).

## Rebase notes

Rebased onto current `main` (through #286 model catalog + #290 wizard-test fix), so this branch is up to date and the PR diff is only this feature. Conflicts were all in the chat-integration surface and resolved to keep upstream's refactors AND the feature's gating: new-chat extraction (`startNewChatSession`), callback extraction (`handleCallbackQuery` re-threaded `chatId`), `/send` gate + working-indicator rename, reconnect restore (`listActiveChatIntegrationSessions` query filter + the `isChatAllowed` loop filter as complementary layers), the URL-router migration in the home/view components, and the model-select refactor in `integration-settings-menu.tsx` (kept #286's `SettingsModelSelect` and dropped the now-unused `useSettings`, kept our owner-gated require-approval toggle).

## Deferred / accepted (not bugs)

- OS/dropdown push notification for new pending requests: deferred; shipped a lightweight in-app pending **badge** instead. A real push needs a `notifications.session_id`-nullable table rebuild + a new type. Separate ticket.
- Bootstrap re-opens when zero allowed rows exist: kept by design.
- Denied rows count toward the 100 pending+denied cap: kept.
- Access enforcement is Telegram-only by design (Slack/iMessage forward unconditionally); the approval toggle + badge are shown only for Telegram. Extending the gate to other providers is a follow-up.
- Owner-gating in multi-user auth mode is code- and review-verified but not manually exercised (needs a multi-user web setup).

## Testing

- `npm run typecheck`: clean.
- `npm run lint`: 0 errors.
- Scoped feature suite (`chat-integration chat-access x-agent-chat telegram-connector base-connector imessage-connector settings`): 830 passed.
- e2e (`E2E_MOCK=true npx playwright test`): 203 passed, 0 failed.
- Manual: private allow/deny/revoke, group behavior, bootstrap, pending notice, make-public toggle + reconcile, owner-only badge. (Group @mention has a Telegram-side privacy-mode limitation unrelated to this change; replies and commands work.)


src/shared/lib/db/schema.ts
```diff
@@ -415,6 +415,7 @@ export const chatIntegrations = sqliteTable('chat_integrations', {
 
   // Behavior settings
   showToolCalls: integer('show_tool_calls', { mode: 'boolean' }).notNull().default(false),
+  requireApproval: integer('require_approval', { mode: 'boolean' }).notNull().default(true),
```
iddogino, line 418: Will that trigger approval workflows with Slack / iMessage?
**iMessage** -> mapped to private phone number so not relevant (only 1:1)
**Slack** -> adding _optional_ approvals would actually be a nice feature (not everyone in the org can message the HR agent, for instance)

iddogino (changes requested):

src/shared/lib/db/schema.ts
JeremyBischoff: Verified it does not trigger approvals for Slack or iMessage. The one real cross-provider thing the allowlist introduced was the new /start greeting firing on all providers instead of reaching the agent., so I scoped it to Telegram and added regression tests. Will revisit the rest when we refactor and work on the Slack and iMessage integrations but should be all good now for this PR.

src/shared/lib/db/migrations/0024_seed_chat_access.sql
```diff
@@ -0,0 +1,12 @@
+INSERT INTO chat_integration_access
```
iddogino: This migration works, but as noted by subsequent tests - it isn't idempotent. 
Could make it idempotent with a simple `INSERT OR IGNORE` -- which will make this more robust IMO

src/shared/lib/db/migrations/chat-access-migration.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  it('seeds allowed access only for active telegram sessions, deduped', () => {
+    seedFixtures()
+    sqlite.exec(SEED_SQL)
+    const rows = sqlite.prepare(`SELECT integration_id, external_chat_id, status, approval_source FROM chat_integration_access`).all() as Array<Record<string, unknown>>
+    expect(rows).toHaveLength(1)
+    expect(rows[0]).toMatchObject({ integration_id: 'int-tg', external_chat_id: 'chat-1', status: 'allowed', approval_source: 'migration' })
+  })
+
+  it('is not idempotent — re-running the seed violates the unique index', () => {
```
iddogino: The exact test we can get rid of by actually making the migration idempotent...

src/shared/lib/db/schema.ts
iddogino: One other question / concern -- when people upgrade, we automatically migrate all their existing chats as approved (correct). So if the have 3 slack chats - all will be approved (which is pointless / inert today, but it'll be there).
If someone new messages the bot after the upgrade - a 4th chat ID - that will _not_ be migrated (b/c the one time migration already ran) - so there won't be an approval row (correct?).
Again - this is inert now, but let's say in two weeks we add the Slack approvals feature - would the 4th chat (created post migration) fail?

src/shared/lib/chat-integrations/telegram-connector.ts
```diff
@@ -161,6 +161,9 @@ export class TelegramConnector extends ChatClientConnector {
       const photos = ctx.message.photo
       if (!photos || photos.length === 0) return
 
+      const chatType = (ctx.chat as { type: string }).type
```
iddogino: Why cast as string? Isn't ctx.chat already typed by grammy SDK? Note you are re-casting down from string to `'private' | 'group' | 'supergroup'` down in lune 183...

iddogino: Confirmed:  @grammyjs/types: ctx.chat.type is already typed "private" | "group" | "supergroup" | "channel".... So we are actually missing channel here

src/shared/lib/chat-integrations/telegram-connector.ts
```diff
@@ -585,26 +593,23 @@ export class TelegramConnector extends ChatClientConnector {
     const text = ctx.message?.text
     if (!text || !ctx.chat) return
 
-    // Handle /start command — greet the user instead of forwarding to the agent
-    if (text === '/start') {
-      ctx.reply('Hello! Send me a message and I\'ll forward it to your agent.').catch(() => {})
-      return
-    }
+    const chatType = (ctx.chat as { type: string }).type
```
iddogino: Again.... Why string casting vs built in types ? 

**Principle**: when a library already gives you a discriminated union, narrow with a runtime check and let the compiler do the rest — an as cast is you overriding the compiler, which is exactly when bugs hide.

src/shared/lib/services/chat-integration-access-service.ts
```diff
@@ -0,0 +1,116 @@
...
+      SELECT ?, ?, ?, 'private', 'allowed', 'auto_first_contact', ?, ?, ?, ?, ?, ?, ?
+      WHERE NOT EXISTS (SELECT 1 FROM chat_integration_access WHERE integration_id = ? AND status = 'allowed')
+    `).run(crypto.randomUUID(), args.integrationId, args.externalChatId,
+      args.chatName ?? null, args.userId ?? null, args.userName ?? null, preview(args.preview),
+      now, now, now, args.integrationId)
+    if (info.changes === 1) return { action: 'forward', sendNotice: false, status: 'bootstrapped' }
+  }
+
+  // not bootstrapped → record pending unless capped
+  const nonAllowed = db.select().from(chatIntegrationAccess)
```
iddogino: Nit: you are fetching all the rows just to count. Much more efficient to just count in DB

iddogino (changes requested):

src/shared/lib/services/chat-integration-access-service.ts
JeremyBischoff: Done in `8e072558` - switched to a DB `COUNT()` so it no longer loads every pending/denied row just to count them.

src/shared/lib/db/migrations/0024_seed_chat_access.sql
JeremyBischoff: Done in `36cb8a09` - added `INSERT OR IGNORE`, so re-running the seed just does nothing instead of erroring on the unique index. Dropped the non-idempotency test since it no longer applies; kept the journal-timestamp test.

src/shared/lib/db/migrations/chat-access-migration.test.ts
JeremyBischoff: Removed this test in `36cb8a09` - with the seed now idempotent (`INSERT OR IGNORE`), the case it documented no longer applies.

src/shared/lib/chat-integrations/telegram-connector.ts
JeremyBischoff: Done in `c1c3bb1f` - dropped the `as` casts and used grammy's built-in chat types, and narrowed `title` with `'title' in ctx.chat`. This also turned up something: the `channel` guards in the photo/document handlers can never actually run, since channel posts come in as `channel_post`, not `message` - so grammy already types those handlers without `channel`, and the compiler now confirms it. Removed the two channel-drop tests that were testing that impossible case. `handleTextMessage` keeps its guard, since that one is reachable.

src/shared/lib/chat-integrations/telegram-connector.ts
JeremyBischoff: Applied your principle here too in `c1c3bb1f` - dropped the cast for grammy's built-in type and let the compiler narrow it.

src/shared/lib/db/schema.ts
JeremyBischoff: Right now there's a one-time migration that runs on Telegram chats only - it auto-adds anyone who's already interacted with the bot onto the allowlist, so nobody gets locked out when they upgrade. Slack and iMessage are untouched by it.

New chats don't use that one-time step at all. They go through the live gate: the owner's first DM bootstraps to allowed (so they're never locked out of their own bot), and every chat after that is pending until the owner approves it. Slack and iMessage are untouched by the gate too - they still forward everything.

So when we add Slack approvals later, we'll need a one-time backfill for Slack, exactly like we're doing for Telegram here - otherwise existing Slack chats would have no row and get locked out the moment the gate turns on. Nothing to change for this PR.

iddogino (approved):

PR #305, opened 2026-06-22
fix(usage): price GPT models incl. 272K cliff in usage dashboard

yiw190:
## What was broken

The local **usage dashboard** (`usage-service.ts`) shows $0 for every GPT run.

`loadDailyUsageData` uses `entry.costUSD` when present, else computes cost from `model-pricing.json`. But `costUSD` is **almost never in the transcripts** — only **2 of 130** fixture files carry it; normal assistant entries have just `{ timestamp, type, requestId, message }`. So `calculateCost(model, …)` keyed off `model-pricing.json` is the *actual* cost path — and that table was **Claude-only**. GPT models (`gpt-5.4`, `gpt-5.5`, `openai/gpt-5.*`) miss the lookup → cost `0`.

## What changed

- Add `gpt-5.4` / `gpt-5.5` (Platform bare ids) and `openai/gpt-5.4` / `openai/gpt-5.5` (OpenRouter ids) to `model-pricing.json`.
- Apply the **same 272K long-context cliff the proxy bills at** (platform PR #104): once a request's prompt input (`inputTokens + cacheReadTokens`, since input excludes cache reads in this shape) crosses 272K, the whole request reprices **2× input / 1.5× output**.
- Export `calculateCost` and unit-test it directly (the cliff is pure logic).
- Claude and unknown models are unchanged.

### Rates (mirror of the proxy table)

| Model | input ≤272K | input >272K | output ≤272K | output >272K |
|---|---|---|---|---|
| gpt-5.4 | $2.50 | $5.00 | $15.00 | $22.50 |
| gpt-5.5 | $5.00 | $10.00 | $30.00 | $45.00 |

## Tests

`usage-service.test.ts` adds: GPT priced (was $0), >272K reprice, cache-reads-count-toward-threshold, exactly-272K stays short, Claude stays linear, unknown → 0. All 27 tests pass; typecheck clean for touched files.

## Known follow-up (out of scope)

`scripts/fetch-model-pricing.ts` regenerates `model-pricing.json` from LiteLLM **claude-only**, so a regen would wipe these manual GPT entries (same as the proxy's generator). Tracked separately.

## Test plan

- [ ] Run a GPT-5.5 agent, open the usage dashboard → non-zero cost.
- [ ] A GPT run with >272K context shows the doubled rate.

### DIFF
--- a/src/shared/lib/services/model-pricing.json
+++ b/src/shared/lib/services/model-pricing.json
@@ -130,5 +130,57 @@
     "output": 15,
     "cacheCreation": 3.75,
     "cacheRead": 0.3
+  },
+  "gpt-5.4": {
+    "input": 2.5,
+    "output": 15,
+    "cacheCreation": 2.5,
+    "cacheRead": 0.25,
+    "longContext": {
+      "thresholdTokens": 272000,
+      "input": 5,
+      "output": 22.5,
+      "cacheCreation": 5,
+      "cacheRead": 0.5
+    }
+  },
+  "gpt-5.5": {
+    "input": 5,
+    "output": 30,
+    "cacheCreation": 5,
+    "cacheRead": 0.5,
+    "longContext": {
+      "thresholdTokens": 272000,
+      "input": 10,
+      "output": 45,
+      "cacheCreation": 10,
+      "cacheRead": 1
+    }
+  },
+  "openai/gpt-5.4": {
+    "input": 2.5,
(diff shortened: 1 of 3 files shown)


yiw190:
Folded into #304 — same concern (GPT 272K long-context pricing handling in SuperAgent). The usage-service commit was cherry-picked onto #304's branch.

PR #307, opened 2026-06-22
feat(telegram): Mini App dashboards - open an agent dashboard from Telegram

JeremyBischoff:
## What

Let a Telegram user open and fully interact with one of an agent's dashboards from inside Telegram. Two surfaces:

- A `share_dashboard` agent tool that pushes a dashboard into a chat as a tappable "Open dashboard" web_app button, or a plain-text message when there's no public HTTPS URL to point at.
- A Telegram Mini App that verifies Telegram `initData`, mints a short-lived scoped `tg_dash` cookie, and serves the dashboard in-app on the artifact/llm/stt routes, with an "open in external browser" handoff.

Inside the Mini App the dashboard is fully interactive - including dashboards with their own server-side backends - matching the in-app experience (see Auth model).

## Why

Agents already build dashboards; this makes them reachable where the user already is - their Telegram chat - without a separate login.

## Reviewing this PR

This diff reads as large (`+5.6k`) but most of it is not hand-written logic:

- **~1,000 lines** of production code - the part to review closely.
- **~2,300 lines** of tests (unit + e2e).
- **~2,300 lines** of a generated Drizzle migration snapshot (`migrations/meta/`) - machine-written, now marked `linguist-generated` so GitHub collapses it.

The auth/crypto foundation (initData verification + the scoped `tg_dash` cookie + session middleware) is **included here, not a separate PR**: its security is emergent across the cookie, the middleware, and the `api/index.ts` mount, so it only reviews correctly as one unit. initData verification uses `@tma.js/init-data-node` and the cookie is signed with `jose` - both behind swappable seams - rather than hand-rolled crypto.

What remains here is one vertical feature in three layers - suggested reading order:

1. **Serve** - `routes/telegram-miniapp.ts` plus the cookie wiring on existing routes (the `middleware/auth.ts` short-circuit and the `api/index.ts` mount with its bare-path exclusion). This is where the authorization-scope risk lives - see the Auth model below.
2. **Deliver** - the `share_dashboard` agent tool, the connector card/button, the `x-agent-chat` route.
3. **Deploy-safety** - HTTPS availability gating, `SHARE_DASHBOARD_ENABLED`, owner attribution, and the backfill migration (the only existing-data change).

## Scope and limitations (please read)

The interactive "Open dashboard" button only works when the server is reachable at a **public HTTPS URL** (`getPlatformBaseUrl()`). That maps to deployment type:

- **Hosted platform:** has a public HTTPS URL, so the button + Mini App work.
- **BYOK / self-hosted / Electron / localhost:** no public URL by default, so sharing falls back to a **plain-text message** naming the dashboard. The Mini App button does not work there unless the operator deliberately exposes the server at a public HTTPS endpoint and sets `PLATFORM_BASE_URL`.

So in practice this is a **platform (publicly-reachable) feature**; BYOK/desktop users get graceful plain-text degradation, not the button - normal messaging is unaffected. A second precondition for the button is that the integration has an owner (`createdByUserId`); a one-time migration in this PR backfills that field for integrations created before it existed, and the connector still refuses to send a button that would dead-end on tap.

## Auth model (the part worth scrutinizing)

When a user taps the dashboard button:

1. The Mini App POSTs Telegram's signed `initData` to `/session`; the server verifies the HMAC with the bot token (constant-time) plus freshness. This proves the Telegram user id is real - auth before authz.
2. The user must be a non-archived, approved (not-revoked) member of the integration; the dashboard must belong to that agent; the integration must have an owner.
3. The server mints a signed, HttpOnly, SameSite=Lax, 15-min `tg_dash` cookie carrying the **integration owner's** id plus a scope of exactly one agent + one dashboard. There are no per-Telegram-user accounts - the cookie acts as the owner, and the tight scope is what makes that safe.
4. `TelegramDashboardSession` re-validates the cookie and re-checks the scope on every request, acting as the owner. It is scoped to that one dashboard on that one agent and is **never mounted on, and cannot reach,** the agent's conversations, secrets, tools, other dashboards, or any other agent (those 401).
5. On the dashboard's content sub-paths the cookie grants the **same access an in-app `AgentRead` viewer has - including writes** (POST/PUT/DELETE) to the dashboard's own backend, which runs in the agent's sandboxed container - so Telegram and in-app match. The one hard exclusion is the bare `:artifactSlug` **management path** (AgentAdmin `DELETE`/`PATCH`): the cookie never applies there, for any method, via an encoding-safe decoded-path check.
6. In the Mini App the cookie silently re-mints (~every 10 min, re-running the full check, so a revocation takes effect within ~10 min). The "open in browser" handoff uses a 120s single-use link token and a non-refreshing 15-min session.

## How it was verified

- Full e2e suite green; agent-container unit tests green; root typecheck + lint clean.
- The initData HMAC, the signed cookie codec, the per-route `isChatAllowed` gate, and the artifact-mount scoping are unit-tested; the e2e exercises the real `/session` -> cookie -> artifact-read flow end to end. Added unit coverage for the `/session` rejection branches (no-owner, dashboard-not-found, agent-mismatch, not-bound, malformed), the `/share-dashboard` telegram-only and not-connected paths, deterministic owner attribution, and rejection of future-dated `initData`.
- The cookie's access scope is unit-tested: writes to a dashboard sub-path are authorized (matching in-app), while the bare AgentAdmin management path is excluded for every method and encoding (`%2F`, `%252F`, dot-segments). A cross-model security audit with an independent Hono routing reproduction confirmed no host-side mutation is reachable via the cookie and no encoding/routing evasion reaches the management routes.
- Smoke-tested live: in web mode behind a public HTTPS tunnel the button appears and the Mini App loads and stays interactive over HTTPS; in the Electron desktop app the `share_dashboard` tool is correctly absent and the agent declines honestly, with no dead button or fabricated link. The plain-text fallback is covered for both the no-URL and no-owner cases.

## Notes for the reviewer

- The cookie acts as the integration owner; access is limited to the bound Telegram user(s), scoped to the one shared dashboard on one agent - it can't reach the agent's other dashboards, the destructive management routes, or any other agent's data. Writes are confined to that dashboard's container-sandboxed backend, matching what an in-app viewer can do.
- The owner who administers the Telegram allowlist is the same trust root the cookie acts as, so the feature introduces no principal outside the agent's existing owner boundary.


JeremyBischoff:
Closing. Too far behind main to rebase. SUP-252 tracks the idea, pending whether the mobile app already covers it.

PR #310, opened 2026-06-23
feat(container): AWS Lambda MicroVM agent runtime

yiw190:
## What

Adds `LambdaMicroVmRuntimeClient` — a new `ContainerClient` runner that runs cloud agents as **AWS Lambda MicroVMs** instead of k8s pods. Registered in `client-factory` as `'lambda-microvm'`; **callers are unchanged** (polymorphism, no external type-switching).

This is the SuperAgent half of the MicroVM agent-runtime track. It does **not** remove the k8s agent path (kept as deprecated fallback); host-app stays on k8s.

## Why

agent k8s pods need warm-pool/balloon node pre-warming to hit ~2s cold start (costly, operationally heavy). MicroVMs resume from snapshot in seconds, bill per-second, isolate at VM level, and self-expose an HTTPS endpoint (no ALB/Ingress). suspend/resume is also a natural per-agent scale-to-zero.

All blockers were verified out-of-band on AWS (us-east-2 POC) before writing this: S3 Files workspace mount, suspend/resume mount survival, WS bidirectional stream + egress SG isolation, and the real `agent-container` image building + booting in a MicroVM (`/health` OK, real `createSession` streaming through the auth-proxy).

## What changed

- **`lambda-microvm-runtime.ts`** (config + auth-proxy + client):
  - `start` = RunMicrovm, `getInfoFromRuntime` = GetMicrovm (`SUSPENDED`/`SUSPENDING` treated as running — idlePolicy auto-resume wakes on next request). Terminal/`ResourceNotFound` states drop local state; **transient errors keep last-known state** (don't orphan a live VM).
  - `stop`: background auto-sleep (`escalateToForceStop:false`) is a no-op (AWS idlePolicy owns idle); plain stop → **SuspendMicrovm** (preserve memory+disk + proxy so the next request auto-resumes ~2s); explicit shutdown → TerminateMicrovm. Suspend is idempotent and never throws into the sweep loop.
  - `LocalAuthForwardProxy`: host loopback proxy injecting `X-aws-proxy-auth`/`X-aws-proxy-port` for HTTP **and** WebSocket upgrade, single-flight token refresh. Lets `BaseContainerClient` (global `fetch`/`new WebSocket`) talk to a MicroVM unchanged.
    - HTTP requests replay across the suspend→resume 502 window.
    - **WS upgrade wakes a suspended VM via the HTTP resume-retry path before opening the unreplayable stream.**
    - **Upstream HTTP request + WS connect phase are guarded by a timeout** so a silently hung socket can't wedge the resume loop; the WS idle timeout is cleared on handshake so a quiet-but-live stream isn't killed.
  - zod-validated, memoized runtime config + eligibility gate; SDK client memoized per region.
- **agent env delivery** — `RunMicrovm` has no env field and the full env exceeds the 4096B `runHookPayload` cap, so env is stashed host-side (`agent-bootstrap-env-store`) and the VM fetches it at boot via `GET /api/agent-bootstrap/:slug/env` (Bearer `PROXY_TOKEN`, slug-matched). The store is **re-fetchable until agent teardown**, so a lost/retried boot fetch still succeeds; cleared on stop/terminate.
- **`base-container-client.ts`** — `buildAgentEnv` centralizes the env merge; `platform-k8s-runtime` refactored to use it (removes a duplicated merge rather than adding a 3rd copy).
- **`client-factory.ts`** — registers the runner.

## SDK constraint handled

- `ListMicrovms` has **no tag filter** → `agentId → microvmId` is kept in-process, with unique-per-start `clientToken` + `idlePolicy` (idle→suspend→suspended-timeout→terminate) reclaiming orphans after a host-app restart. Tradeoff documented in code.

## Test plan

- [x] Unit (vitest): env boundary/missing-field/defaults/memoization, lifecycle (run / state-mapping / terminate / auto-sleep suspend + state preserved for resume / no-op-when-running / transient-error keeps state), eligibility, factory registration, `runHookPayload` (bootstrap + mount) construction, unique `clientToken` per start, `LocalAuthForwardProxy` HTTP + WS wire behavior incl. **WS resume-retry** and **upstream timeout**, bootstrap-env **re-fetch + teardown clearing**.
- [x] lint clean; no new `src/` typecheck errors.
- [x] Gated real-AWS e2e (`RUN_MICROVM_E2E=1`): `/health` through proxy; cold-start series; auto-sleep → auto-resume (state preserved). Verified against `poc2-agent-mvm` on us-east-2.

## Notes

- `package-lock.json` churn is from adding `@aws-sdk/client-lambda-microvms` (+ `@smithy`/`@aws-sdk` transitive deps).

### DIFF
--- a/src/shared/lib/container/lambda-microvm-runtime.ts
+++ b/src/shared/lib/container/lambda-microvm-runtime.ts
@@ -1,0 +1,640 @@
+import http from 'http'
+import https from 'https'
+import tls from 'tls'
+import net, { AddressInfo } from 'net'
+import { randomUUID } from 'crypto'
+import { z } from 'zod'
+import {
+  LambdaMicrovmsClient,
+  RunMicrovmCommand,
+  GetMicrovmCommand,
+  SuspendMicrovmCommand,
+  TerminateMicrovmCommand,
+  CreateMicrovmAuthTokenCommand,
+} from '@aws-sdk/client-lambda-microvms'
+import { BaseContainerClient, CONTAINER_INTERNAL_PORT } from './base-container-client'
+import type { ContainerConfig, ContainerInfo, ContainerStats, StartOptions, StopOptions, StopResult } from './types'
+import { getSettings } from '@shared/lib/config/settings'
+import { captureException } from '@shared/lib/error-reporting'
+import { setBootstrapEnv, clearBootstrapEnv } from './agent-bootstrap-env-store'
+
+// RunMicrovm caps runHookPayload at 4096 bytes. We only put a small bootstrap
+// credential + mount params here; the full agent env is fetched at boot (see
+// start()). This guard backstops an unexpectedly large payload.
+const RUN_HOOK_PAYLOAD_MAX_BYTES = 4_096
+const AUTH_TOKEN_EXPIRATION_MINUTES = 60
+// Max wait to kick a (possibly suspended) VM back to a serveable agent before a real
+// request. Covers auto-resume (~2-10s) with headroom; returns as soon as /health is ok.
+const RESUME_KICK_TIMEOUT_MS = 60_000
+// Gap between proxy retries while a suspended VM wakes (it 502s for ~2-3s).
+const RESUME_RETRY_DELAY_MS = 400
+// Idle timeout on a single upstream exchange (HTTP request, or the WS connect
+// handshake). Guards against a silently hung socket that never errors or 502s,
+// which would otherwise wedge the resume-retry loop forever. Disabled once a WS
+// handshake completes (a live stream is idle by design).
+const UPSTREAM_IDLE_TIMEOUT_MS = 30_000
+
+// ---------------------------------------------------------------------------
+// Runtime config (env-driven, zod-validated, memoized)
+// ---------------------------------------------------------------------------
+
+function allIngressConnectorArn(region: string): string {
+  return `arn:aws:lambda:${region}:aws:network-connector:aws-network-connector:ALL_INGRESS`
+}
+
+const microvmRuntimeSchema = z.object({
+  region: z.string().min(1),
+  imageArn: z.string().min(1),
+  imageVersion: z.string().min(1).optional(),
+  executionRoleArn: z.string().min(1),
+  // Per-org egress connector (gates "agent A only talks to app A" via its SG).
+  egressConnectorArn: z.string().min(1),
+  ingressConnectorArn: z.string().min(1).optional(),
+  agentPort: z.coerce.number().int().positive().default(CONTAINER_INTERNAL_PORT),
+  // Total VM lifetime cap (AWS hard max 28_800 = 8h). Default to the max so a
+  // suspended VM survives "until killed" — the next request auto-resumes it and
+  // only the 8h ceiling force-terminates an untouched VM.
+  maxDurationSeconds: z.coerce.number().int().positive().max(28_800).default(28_800),
+  // How long a suspended VM is kept before AWS terminates it. Maxed so suspend
+  // lasts "until killed" within the 8h lifetime cap.
+  suspendedSeconds: z.coerce.number().int().positive().max(28_800).default(28_800),
+  logGroup: z.string().min(1).optional(),
+  // Per-org S3 Files workspace mount, passed to the image's run hook so the
+  // supervisor mounts /workspace. All three required together or none (no mount).
+  fsId: z.string().min(1).optional(),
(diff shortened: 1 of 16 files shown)


src/shared/lib/container/lambda-microvm-runtime.ts
```diff
@@ -0,0 +1,539 @@
...
+    // hook; omit it entirely when there's nothing to pass so hook-less images still run.
+    const env = this.buildAgentEnv(options?.envVars)
+    const mount = config.fsId && config.accessPoint && config.mountTargetIp
+      ? { fsId: config.fsId, accessPoint: config.accessPoint, mountTargetIp: config.mountTargetIp }
+      : undefined
+    const runHookPayload = Object.keys(env).length > 0 || mount
+      ? JSON.stringify(mount ? { env, mount } : { env })
+      : undefined
+    const payloadBytes = runHookPayload ? Buffer.byteLength(runHookPayload, 'utf8') : 0
+    if (payloadBytes > RUN_HOOK_PAYLOAD_MAX_BYTES) {
```
iddogino, line 440: The env has all MCPs, connected accounts, etc... Won't we hit this limit? Might need to fetch this at boot if this is a hard limit.

Idea -- we put everything in .env file anyway - can we just load it into env on boot (so that its just one command versus adding them all and busting this cap)

src/shared/lib/container/lambda-microvm-runtime.ts
```diff
@@ -0,0 +1,539 @@
...
+    try {
+      const mvm = await getMicrovmClient(config.region).send(
+        new GetMicrovmCommand({ microvmIdentifier: state.microvmId }),
+      )
+      // SUSPENDED is "running on demand": idlePolicy.autoResumeEnabled wakes it
+      // on the next request through the proxy.
+      if (mvm.state === 'RUNNING' || mvm.state === 'SUSPENDED' || mvm.state === 'SUSPENDING') {
+        return { status: 'running', port: state.proxyPort }
+      }
+      return { status: 'stopped', port: null }
```
iddogino, line 530: If AWS terminates the instance - we properly set to stopped but don't do any cleanup... 

Normal lifecycle: agent suspends (auto-sleep) → state + proxy kept → after `suspendedDurationSeconds`/8h cap, AWS terminates the VM. The next `getInfoFromRuntime` (`:427-430`) sees `TERMINATED`, returns `{status:'stopped'}` **but does not call `cleanupLocal()`** — unlike the `ResourceNotFound` branch (:433) and `suspendForAutoSleep` (:497-499). The stale `agentStates` entry (with a live `http.Server`) lingers. Then `ensureRunning` (`container-manager.ts:480,493`) sees `stopped` → calls `start()` → `agentStates.set(...)` at `:383` **overwrites the stale state without stopping the old proxy** → leaked loopback port + fd + listener, accumulating each terminate→restart cycle.
   - Fix: in `getInfoFromRuntime`, `cleanupLocal()` on terminal states (mirror the NotFound branch), and/or have `start()` call `this.cleanupLocal()` before creating the new proxy. The latter is the robust belt-and-suspenders fix.

src/shared/lib/container/lambda-microvm-runtime.ts
```diff
@@ -0,0 +1,539 @@
...
+    } catch (error) {
```
iddogino, line 531: swallows any non-NotFound error (throttling, transient network) to `{status:'stopped'}`. Combined with #2, a transient error during the `start()` pre-check (`:328`) also overwrites live state and orphans the running VM. Same root fix as #2 helps; consider treating transient errors as "keep last known state" rather than `stopped`.

iddogino (changes requested):

src/shared/lib/container/lambda-microvm-runtime.ts
yiw190: Fixed. Now full env no longer goes through runHookPayload; it is stored host-side and fetched by the MicroVM during boot via the bootstrap endpoint.

src/shared/lib/container/lambda-microvm-runtime.ts
yiw190: Fixed. Terminal states now call cleanupLocal(), and start() clears stale local state before installing a new proxy.

src/shared/lib/container/lambda-microvm-runtime.ts
yiw190: Fixed. Non-NotFound GetMicrovm errors now keep the last known running state instead of marking the VM stopped.

PR #318, opened 2026-06-25
Chat-integration UI overhaul: cron-page-style connector page

JeremyBischoff:
## What this does

Reorganizes the chat-integration connector page (Telegram / Slack / iMessage) to mirror the scheduled-task (cron) page, reusing its frame and components. Pure renderer change - no API, schema, or backend behavior changes.

- **Page frame:** cron-style `SettingsPageContainer` + `PageTitle` (a max-w-5xl document that scrolls as a whole), with the provider icon, inline rename, and Pause/Resume + Clear in the header.
- **Conversation History (Read-Only):** the left rail is replaced by a dropdown that switches conversations (labelled by date/time, with the chat name shown only when more than one chat exists). The thread is a read-only mirror: the sender is lifted to a label, there are no edit controls, the panel is muted/washed (legible in both light and dark mode), and it scrolls internally only when the conversation is long.
- **Settings / Access / Danger:** non-collapsible `DetailCard`s (the cron pattern) instead of collapsible sections. Session timeout is a dropdown ("New conversation after inactivity"), used in the setup dialog too. Access is a binary allowed/Blocked model (pending and denied both read as Blocked, since the bot ignores both), with the blocked group ordered by first-contact time so chats still waiting surface above old denials. Connection state is folded into Settings; the active/paused control is a header Pause/Resume button.

## Why

Follow-up to design review: get the connector page as close to the cron page as possible, maximize component reuse, and trim to essentials - one read-only signal, no duplicate messaging, minimal chrome.

## Changes since the last review (the three-pane version)

- Three-pane full-bleed layout -> a single cron-style document (max-w-5xl).
- Conversations left rail -> a dropdown in the section header.
- Collapsible side panel -> non-collapsible DetailCards.
- Read-only thread: dropped the banner / footer / dim overlay; sender lifted to a label and the inline `[name]:` prefix removed; washed panel tuned per theme (light-mode wash, dark-mode contrast preserved).
- Provider icon added to the title; status is now a Pause/Resume button; Delete uses the standard Button.
- Session-timeout free-text input -> dropdown (on the page and in the setup dialog).
- Extracted `ui/SectionHeader`, now shared with the cron Run History.

## Review-council pass

A follow-up review pass tightened the implementation with no behavior change (verified frontend-only):

- **Leaner decomposition:** removed a dead `conversation-ordering` module and inlined its one live sort; folded the single-purpose model-card wrapper into the side panel; dropped an unreachable access-admin-only path (`canSeeSidePanel` is just `canManage`, since an admin is always a user).
- **Completed the blocked-group ordering:** a chat blocked while it was active kept its recent-activity sort key and could still jump above a chat still waiting; blocked rows now stay keyed to first-contact time (regression-tested).
- **Snappier settings:** the model/effort select and the settings toggles update instantly instead of lagging a refetch (optimistic cache write on update).
- **Tests:** unified ~6 copy-pasted fixtures into a shared factory, collapsed repetitive cases with `it.each`, extracted an e2e setup helper, and added coverage for the Unblock/Block access wiring and the conversation window switcher.

Net: about -190 lines and -4 files versus the pre-pass branch, all frontend.

## Since the last review (the cron-page version)

Addresses the review on the cron-page version, plus two features on top. Still frontend-only.

Fixes:

- Inbox title took a no-access chat's *oldest* window name (Slack / iMessage), because `buildChatRows` overwrote the title on every window - now seeded from the newest window, falling through to an older *named* window when the newest is unnamed.
- `parseSenderPrefix` mis-split a DM literally starting `[TODO]: ...` into sender "TODO"; it now requires the backslash the connector always writes, so unescaped brackets stay user text. (Legacy pre-escape group transcripts render the prefix inline - cosmetic, no data loss.)
- Deduped the settling predicate (`isSettling`) and the session-timestamp formatter, each previously copied twice; a stuck "Connecting…" now fast-polls only while foregrounded.

New:

- **Inline-rename** an integration from its title (the `InlineEditableTitle` the agent page uses), replacing the rename that lived in the now-removed dropdown menu. Owner-only; read-only for viewers.
- **Blocked / pending request view:** a pending or denied chat has no conversation window, so its first message (the chat request) was captured but shown nowhere. Those rows now open the conversation modal to a request view - the chat title, the first message as a bubble, and the decision inline (Approve / Deny for a pending request, Unblock for a denied one) - and the preview shows in the row subtitle for list-level triage. Extracted a shared `AccessActions` so the row and the modal share one implementation.

pr-ready-loop audit (frontend-only):

- Lock both Approve and Deny while either mutation is in flight (a fast Approve-then-Deny raced two conflicting writes to the same access row).
- Keyboard: an Enter / Space on an access button no longer bubbles to the row and opens the dialog instead of acting.
- Added characterization tests for the above, plus blocked-row openability, the request-preview subtitle, the `windows > 0` thread guard (a denied chat with history opens its thread, not the request view), decide-then-close, denied-no-windows, and the rename-error toast.

## Reuse

`SettingsPageContainer`, `PageTitle`, `InlineEditableTitle`, `DetailCard`, `Select`, `ServiceIcon`, `SessionThread`, and the new shared `SectionHeader`. The shared `MessageList` / `MessageItem` gained an optional `readOnly` flag (default off) so the live session view is unchanged.

## Gating

Permission gating is unchanged: viewers are read-only; `canManage` gets the Settings card + Clear; admins get Access.

## Tests

- Unit: new conversation-history-section + clear-conversation-button suites; updated view / settings-card / side-panel / access / danger tests; shared test factories and `it.each` consolidation; added Unblock/Block wiring and conversation-window-switcher coverage. All message + chat-integration suites pass.
- E2E: chat-integration-overhaul spec updated to the new structure, with a shared create-agent-with-integration helper.

## Verification

Pixel-checked the live app (real data dir) in both light and dark mode, for Telegram and Slack integrations, and re-verified after each review pass.


src/shared/lib/chat-integrations/utils.ts
```diff
@@ -3,11 +3,50 @@
...
+/**
+ * The single status the user sees for a chat integration, derived from the
+ * persisted lifecycle (`status`) plus the live transport (`connected`):
+ *
+ *   Paused      — toggled off; won't connect.
+ *   Connecting  — toggled on, wire not up yet.
+ *   Listening   — toggled on and the wire is up.
+ *   Error       — toggled on but a connect attempt failed; retrying.
+ */
+export type ChatIntegrationState = 'paused' | 'connecting' | 'working' | 'error'
```
iddogino, line 21: **Code Smell** - this feels like we should just have a `<ChatIntegrationPill>` component. We are exporting shared UI priors then re-implementing it in two places...

src/renderer/globals.css
```diff
@@ -133,6 +133,25 @@ html[data-keyboard-open] [data-composer-footer] {
   background-color: rgba(255, 255, 255, .3);
 }
 
+/* Chat-integration read-only thread: replace the pale native overlay scrollbar with
```
iddogino: **Code Smell** - having component specific styling in global.css is an anti pattern. Should either inline it or create a general class (reader-scrollbar) in global.css and use that in the component

src/renderer/components/messages/message-item.tsx
```diff
@@ -214,15 +214,24 @@ interface MessageItemProps {
...
   useRenderTracker('MessageItem')
   const isUser = message.type === 'user'
   const isAssistant = message.type === 'assistant'
 
   const rawText = message.content.text
-  const { cleanText: textAfterFiles, attachedFiles } = isUser && rawText ? parseAttachedFiles(rawText) : { cleanText: rawText, attachedFiles: [] }
+  // Read-only chat mirror: the connector prefixes incoming messages with an
+  // escaped "\[sender]: " so the agent can attribute them. Lift that into the
+  // sender label instead of showing it inline. Live sessions never carry it.
+  const prefixMatch = readOnly && isUser && rawText ? rawText.match(/^\\?\[([^\]]+)\]:\s*/) : null
```
iddogino: Extract logic to a function so it can be unit tested...

e2e/specs/chat-integration-connector-page.spec.ts
iddogino, line 1: rename test file - overhaul means nothing in 3 weeks... should just describe what it does

iddogino (changes requested):

src/shared/lib/chat-integrations/utils.ts
JeremyBischoff: Done in `c0ad5bc3` - extracted a `ChatIntegrationPill` that owns the label + dot treatment, keyed on a derived `ChatIntegrationState`. Both the Status card and the agent-home tag render it now instead of their own span. Visuals unchanged - the card keeps its dot/pulse, the home tag stays the compact plain badge (size/showDot props).

src/renderer/globals.css
JeremyBischoff: Done in `e53ebdde` - moved it out of the component-specific `.chat-thread` selector into a reusable `.reader-scrollbar` class, same shape as the `.code-scrollbar` one right above it, applied on the read-only panel. Went with the general class over inlining since a `::-webkit-scrollbar` pseudo-element can't be done in Tailwind. Dropped the now-dead `chat-thread`.

src/renderer/components/messages/message-item.tsx
JeremyBischoff: Done in `4f37c5db` - pulled the `[sender]:` prefix parsing into `parseSenderPrefix`, next to the `parseAttachedFiles`/`parseMountedFolders` siblings it's used with, and added unit tests. No behavior change.

e2e/specs/chat-integration-connector-page.spec.ts
JeremyBischoff: Done in `28a4ba46` - renamed to `chat-integration-connector-page.spec.ts` and updated the describe block to match.

PR #321, opened 2026-06-25
fix(container): make MicroVM stop suspend-only

yiw190:
## Summary
- Makes MicroVM stop semantics suspend-only so stopped agents can resume warmly.
- Uses app auto-sleep settings as the sole source for AWS idlePolicy timing.
- Retries proxy HTTP requests across the brief suspend-to-resume 502 window.

## Test plan
- `npx vitest run src/shared/lib/container/lambda-microvm-runtime.test.ts`


yiw190:
Closing: this change was moved into PR #310.

PR #323, opened 2026-06-26
fix(chat): unify the working/"Thinking" indicator on the session_awaiting_input signal

JeremyBischoff:
**TL;DR** — the chat "Thinking…" indicator now settles off one signal (`session_awaiting_input`) instead of the chat manager guessing from a per-request-type allowlist. Kills a recurring "perpetual Thinking…" bug class on Telegram/Slack. ~90 lines of real logic across 2 files; the rest is tests.

### Where to look

| File | Δ | What | Review weight |
|---|---|---|---|
| `chat-integration-manager.ts` | +74/-14 | the consumer: one settle case, drop the allowlist, re-arm on resolve, proxy-review settle | **the meat** |
| `message-persister.ts` | +16 | emit the signal on the per-session stream + cover subagent browser-input | 2 small hunks |
| `*.test.ts`, `mock-connector.ts` | +212 | coverage | skim |

Start with the two source files; the tests just lock in the behaviors.

### The problem

The chat manager subscribes **per-session**, but `session_awaiting_input` was broadcast **global-only** — so it never saw the signal and fell back to enumerating eight `*_request` types. Anything the allowlist missed (subagent browser-input, proxy-review approvals) left the indicator spinning until the 5-min watchdog. Shipped and patched twice (last for Telegram in #308).

### The fix — one source of truth

`markSessionAwaitingInput` now also broadcasts on the per-session stream, and the chat manager settles off that one signal.

**`message-persister.ts`**
- Emit `session_awaiting_input` on the per-session SSE stream too (global stays for sidebar/promotion).
- Raise awaiting in `handleBrowserInputRequestTool` so subagent/sidechain browser-input is covered.

**`chat-integration-manager.ts`**
- Extract `settleIndicator` (stop indicator + clear watchdog); `settleTurn` routes through it.
- One `session_awaiting_input` case → `settleIndicator`; drop the per-type `clearWorkingWatchdog` from the eight `*_request` cases.
- Re-arm on a confirmed interactive-response resolve (Telegram → snappy; Slack → existing stream_start fallback). Re-checks `isChatAllowed` after the await.
- Settle proxy-review approval cards, which route through the global path and miss the per-session signal.

The Gamut app is untouched — it already read the request card directly and behaved correctly. This brings Telegram/Slack/iMessage up to match it. Connectors unchanged.

### Two deliberate trade-offs (worth a look)

1. **Dropped the per-type watchdog-clear on purpose.** Keeps the 5-min idle watchdog armed as a backstop: a future card path that forgets to raise awaiting degrades to a ≤5-min settle, never perpetual. Near-moot — the only long non-awaiting flow (auto-approved host scripts) is capped ~30-35s.
2. **Slack's resolve-time re-arm is a follow-up, not this PR.** Slack doesn't report a chatId on interactive responses, and its chatId feeds the `isChatAllowed` gate — so closing that gap is auth-scope-adjacent and lands separately. Slack keeps its existing stream_start re-arm meanwhile (no behavior change).

### Verified
- Full chat-integrations + message-persister suites green; typecheck + lint clean.
- Whole-branch cross-model review (opus + codex): both SHIP, all probe areas confirmed.
- Manual smoke: _pending_


JeremyBischoff:
Superseded by a ground-up redesign. The working/"Thinking" indicator is being reworked as a single derived predicate (projection model: showThinking = isActive && !isAwaitingInput && !isStreaming, owned/emitted by the persister, reconciled trivially per surface) on refactor/working-indicator-projection — replacing the event-reaction approach this PR refined. Live testing surfaced several independent failure modes (cold-start subscribe race, message-to-blocked-agent hang) that confirmed the per-handler approach can't be made airtight; the projection model fixes the class. Closing in favor of that.

PR #324, opened 2026-06-26
feat(telegram): Mini App auth foundation - verify initData + scoped dashboard cookie

JeremyBischoff:
## What

The auth foundation for the Telegram Mini App dashboards feature, split out of #307 for focused review. Three security primitives, no routes:

- `verifyInitData` - verifies Telegram WebApp `initData` (HMAC signature + freshness) via the maintained `@tma.js/init-data-node` library, wrapped in a stable `Result` seam so the vendor stays swappable, then reads `user.id` from the verified blob.
- `signDashboardCookie` / `verifyDashboardCookie` - a short-lived signed `tg_dash` cookie scoped to a single dashboard (HttpOnly, SameSite, `Secure` honoring `x-forwarded-proto`).
- `TelegramDashboardSession` middleware - authenticates a Mini App request from that cookie and sets the request user.

## Why

#307 (the full Mini App dashboards feature) is large. Per review feedback, this carves out the security boundary so it can be reviewed on its own - it is the part where a missed bug is costliest. It lands as library + middleware (no caller yet); the routes that mint and consume the cookie follow in #307, which will rebase to drop these files once this merges.

## How it was verified

Typechecks against `main` with nothing else from the feature branch present. 28 unit tests across the three units (initData verification, cookie sign/verify, session middleware) pass in isolation.

## Notes for the reviewer

This is the auth/crypto layer - the highest-value part to read closely. initData verification is delegated to `@tma.js/init-data-node` rather than hand-rolled; we use its `validate()` (signature + freshness) but deliberately not `parse()`, which requires a Bot API 8.0 `signature` field and a full Telegram `User` we don't need. Two library quirks are documented in `init-data.ts`: its error classes fail `instanceof` (so we discriminate on `e.name`), and it does not upper-bound a future `auth_date` (accepted, since a valid signature already proves authenticity).

The cookie is scoped to one dashboard and grants the same read-and-write access an in-app viewer has inside that dashboard's sandbox; the bare-path admin routes stay excluded by the mount that lands in #307, not here.


JeremyBischoff:
Closing — folded back into #307 per discussion with Iddo (the split's only real cost was reviewing the auth in isolation from its mount). The two refinements from this PR — initData via @tma.js/init-data-node and the jose-signed cookie — are now commits on #307, which is reopened as a draft with the whole feature reviewable together.

PR #330, opened 2026-06-26
refactor(chat): project session activity onto the working indicator

JeremyBischoff:
## What & why

The chat "working" indicator showed a single hardcoded **"Thinking…"** no matter what the agent was actually doing, even while running a tool or compacting context. This replaces it with a labeled projection of the real activity, mirroring the desktop app's activity indicator so the chat surfaces stay honest.

There is now **one source of truth for chat**: the persister computes a single `SessionActivity` from a session's streaming state (`computeActivity`) and emits one `session_activity` signal. The manager reconciles one indicator off that signal; each connector only *renders* it. Telegram labels its native draft, Slack reacts coarsely (ignores the label by design), the desktop app keeps its own richer indicator.

## What the indicator shows (state -> activity -> label)

| Session state (checked in order) | Activity | Telegram draft | Desktop app |
|---|---|---|---|
| `!isActive` | `idle` | (none) | (hidden) |
| `isAwaitingInput` | `awaiting` | (none; question card is the surface) | "Waiting for input…" |
| `isCompacting` | `compacting` | 🗜 Compacting… | "Compacting…" |
| `isRetrying` | `retrying` | 🔄 Retrying… | "Retrying…" |
| `currentThinking` | `thinking` | ✨ Thinking… | "Thinking…" |
| `isStreaming && !currentToolUse` | `streaming` | (none; streamed reply is the surface) | (indicator stays; text streams below) |
| otherwise (active; tool run / gap) | `working` | ✨ Working… | "Working…" / todo activeForm |

Busy precedence (`awaiting > compacting > retrying > thinking > working`) matches the app exactly. `idle` / `awaiting` / `streaming` are surface-yield states that show no placeholder. The only app/chat divergence is *which widget* each surface uses during `streaming` and `awaiting`, not what they claim the agent is doing.

## Notable fixes

- **Cold-start:** a first message that makes the agent ask via `AskUserQuestion` no longer leaves "Thinking…" stuck under the question card (`awaiting` yields the placeholder).
- **Compaction lifecycle:** the chat indicator now leaves "🗜 Compacting…" the moment compaction finishes (it previously stayed stale until the next token), and a turn that ends mid-compaction no longer wedges the next turn's label to "Compacting…". Both are brought in line with the desktop app's reset behavior.

## Test plan

- **Unit, full suite green (5775 passed).** Persister projection precedence (including the streaming-vs-`compacting`/`retrying` ordering that would otherwise contradict the app); manager reconcile off `session_activity`; Telegram per-activity label + mid-turn relabel; the existing PR #308 watchdog / error-settle invariants migrated to `session_activity`.
- **Manual Telegram smoke (DM).** Working/streaming/idle; tool-gap shows "Working…" (not "Thinking…"); cold-start `AskUserQuestion` shows the card with no lingering placeholder; "Thinking…" only during actual extended thinking. All pass.

## Out of scope

Sending a *new* message while a question is awaiting still deadlocks the session ("Working…" forever, message lost). That is a separate session-cancel fix on `fix/message-cancels-awaiting-input`, not this PR. The indicator here is honest about that state (the session genuinely is stuck), so the fix belongs in message dispatch, not the indicator.

## Follow-ups: connector label fidelity

The `session_activity` signal is connector-agnostic and reaches every integration; the `startWorking(chatId, activity)` seam is already in place. Today only Telegram renders the activity as a visible label. Slack and iMessage act on the same signal (their reaction / typing indicator turns on and off from it, so they get the cold-start and streaming-yield fixes too) but do not vary by status.

- **Slack: deferred to a follow-up.** Could map activity to a reaction emoji, but a per-activity swap churns reactions (remove + re-add), so it should be scoped to the slow, notable states (Compacting, Retrying) with working/thinking folded into one generic busy reaction. Not in this PR.
- **iMessage: intentionally not planned.** The indicator is the native typing bubble, which is binary by protocol (no label to vary). Reflecting status would require sending status text (5-edit limit, intrusive), a worse UX than the clean typing indicator. Left binary by design.

### DIFF
--- a/src/shared/lib/container/message-persister.ts
+++ b/src/shared/lib/container/message-persister.ts
@@ -1749,12 +1749,14 @@ class MessagePersister {
     switch (event.type) {
       case 'message_start':
         state.currentText = ''
-        state.isStreaming = true
+        state.isStreaming = false // text hasn't started; set true at the first text token below
         state.currentToolUse = null
         state.currentThinking = false
-        state.isRetrying = false // a streaming message means any retry resolved
-        // Emit after the resets so the activity reads 'streaming' (text owns the
-        // reply surface), not a stale 'thinking'/'working' from the prior block.
+        state.isRetrying = false // the response is flowing now, so any retry resolved
+        // Emit after the resets. Activity reads 'working' (active, nothing surfaced
+        // yet) — 'streaming' is deferred to the first text token (content_block_delta)
+        // so a message that opens with a tool call doesn't flip streaming→working and
+        // churn connectors (Slack reaction remove+re-add, iMessage typing-bubble blink).
         this.emitActivityState(sessionId)
         this.broadcastToSSE(sessionId, { type: 'stream_start' })
         break
@@ -1780,8 +1782,8 @@ class MessagePersister {
           state.currentThinking = true
           this.broadcastToSSE(sessionId, { type: 'thinking_start' })
         }
-        // A tool block projects 'working'; a thinking block 'thinking' — both
-        // differ from the 'streaming' set at message_start, so re-project.
+        // A tool block projects 'working', a thinking block 'thinking'. A text block
+        // keeps 'working' (deduped) until its first token sets 'streaming' below.
         this.emitActivityState(sessionId)
         break
 
@@ -1792,6 +1794,11 @@ class MessagePersister {
             type: 'stream_delta',
             text: event.delta.text,
           })
+          // The streamed reply now owns the surface → project 'streaming' (deferred
+          // from message_start so a tool-first message never churns connectors).
+          // Idempotent + deduped, so the per-token cost is one cheap compare.
+          state.isStreaming = true
+          this.emitActivityState(sessionId)
         } else if (event.delta?.type === 'thinking_delta' && event.delta.thinking) {
           // Stream summarized reasoning text so the UI can accumulate it for "View thinking"
           this.broadcastToSSE(sessionId, {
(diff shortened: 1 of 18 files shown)


src/shared/lib/container/message-persister.ts
```diff
@@ -1518,6 +1592,10 @@ class MessagePersister {
         state.isStreaming = true
         state.currentToolUse = null
         state.currentThinking = false
+        state.isRetrying = false // a streaming message means any retry resolved
+        // Emit after the resets so the activity reads 'streaming' (text owns the
```
iddogino: `message_start` emits streaming before we know the block type, so a message that opens with a tool call (no preamble text) flips streaming→working here → at 1625. Telegram absorbs it (the draft persists), but Slack removes+re-adds its reaction and iMessage blinks the typing bubble once per tool-first assistant message in an agentic turn. Consider emitting streaming only on the first text content_block_start/text_delta instead of optimistically at message_start.

iddogino (changes requested):

JeremyBischoff:
@iddogino addressed in ec1c6544. Moved the `streaming` projection off `message_start` onto the first `text_delta`: `message_start` now reads `working` (deduped against the `working` from `markSessionActive`), so a tool-first message stays `working` start to finish - no streaming→working flip, nothing for Slack to remove+re-add or iMessage to blink. Text-first messages flip to `streaming` on the first token. Two tests cover it: a tool-first message never emits `streaming`, and `streaming` only appears once real text starts.

PR #334, opened 2026-06-29
feat(models): model-specific disallowed tools + capability warnings

yiw190:
## What & why

Some models don't support certain capabilities (e.g. web search, image input). Today the picker only *warns* about missing web search and doesn't actually stop the agent from calling the tool, and there's no signal at all for image-only gaps. This wires model capability flags from the catalog through to the agent container so unsupported tools are disallowed, and surfaces capability gaps in the picker.

## What changed

- **Mechanism**: the host resolves per-model banned tools from the active provider's catalog and sends them on session create; the container merges them into the SDK `disallowedTools`. This mirrors the existing `modelPromptHints` plumbing (host computes → `/sessions` body → `session-manager` → `ClaudeCodeProcess`).
- **Web search enforcement**: `supportsWebSearch === false` → ban `WebSearch` / `WebFetch` (the main agent's Anthropic-native web tools).
- **UI warnings**: the model picker warns when the selected model lacks web search (existing) or image input (new).

## Deliberately scoped out (follow-up)

Image-emitting tools — `browser_screenshot`, `browser_get_state`, `computer_screenshot`, `start_dashboard` — are **not** banned here. They belong to the browser/dashboard/computer **subagents**, and their image results target the *subagent's own model* (e.g. dashboard-builder runs on Opus), not the main model. Banning them by main-model capability would needlessly break dashboards/browsing. Image input is therefore **warning-only** in this PR; per-subagent image-tool gating is a separate follow-up.

## Test plan

- [x] `agent-container` suite green (incl. new `bannedTools` → `disallowedTools` merge test)
- [x] app tests green: `resolve-model` (web-search ban), `model-family-list` (image warning)
- [x] typecheck clean (app + agent-container), lint clean

Draft: opening for early review of the mechanism + scope before the subagent follow-up.

### DIFF
--- a/agent-container/src/claude-code.ts
+++ b/agent-container/src/claude-code.ts
@@ -299,6 +299,8 @@ export interface ClaudeCodeProcessOptions {
   claudeSessionId?: string;
   userSystemPrompt?: string;
   modelPromptHints?: string[];
+  /** Host-resolved tools this model can't use (e.g. WebSearch/WebFetch when web search is unsupported). */
+  unsupportedTools?: string[];
   availableEnvVars?: string[];
   model?: string;
   browserModel?: string;
@@ -322,6 +324,7 @@ export class ClaudeCodeProcess extends EventEmitter {
   private model: string | undefined;
   private browserModel: string | undefined;
   private dashboardBuilderModel: string | undefined;
+  private unsupportedTools: string[];
   private maxOutputTokens: number | undefined;
   private maxThinkingTokens: number | undefined;
   private maxTurns: number | undefined;
@@ -346,6 +349,7 @@ export class ClaudeCodeProcess extends EventEmitter {
     this.model = options.model;
     this.browserModel = options.browserModel;
     this.dashboardBuilderModel = options.dashboardBuilderModel;
+    this.unsupportedTools = options.unsupportedTools ?? [];
     this.maxOutputTokens = options.maxOutputTokens;
     this.maxThinkingTokens = options.maxThinkingTokens;
     this.maxTurns = options.maxTurns;
@@ -451,6 +455,9 @@ export class ClaudeCodeProcess extends EventEmitter {
           'CronCreate', 'CronDelete', 'CronList',
           'ScheduleWakeup', 'RemoteTrigger', 'PushNotification',
           'EnterWorktree', 'ExitWorktree',
+          // Host-resolved tools the selected model can't use (e.g.
+          // WebSearch/WebFetch without web-search support). Applies query-wide.
+          ...this.unsupportedTools,
         ],
         // Request summarized thinking so reasoning text streams to the UI. Without an
         // explicit `display`, Opus 4.8/4.7 default to `omitted` — thinking_delta events
--- a/agent-container/src/claude-code.ts
+++ b/agent-container/src/claude-code.ts
@@ -1,4 +1,4 @@
-import { query, Query, SDKUserMessage } from '@anthropic-ai/claude-agent-sdk';
+import { query, Query, SDKUserMessage, type HookInput } from '@anthropic-ai/claude-agent-sdk';
 import type { UUID } from 'crypto';
 import { EventEmitter } from 'events';
 import * as fs from 'fs';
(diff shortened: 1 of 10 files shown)


agent-container/src/claude-code.ts
```diff
@@ -451,6 +455,9 @@ export class ClaudeCodeProcess extends EventEmitter {
           'CronCreate', 'CronDelete', 'CronList',
           'ScheduleWakeup', 'RemoteTrigger', 'PushNotification',
           'EnterWorktree', 'ExitWorktree',
+          // Host-resolved tools the selected model can't use (e.g.
+          // WebSearch/WebFetch without web-search support). Applies query-wide.
+          ...this.unsupportedTools,
```
iddogino: **Image/web tool bans are applied query-wide, so they also strip the subagents' tools, contradicting the design**. The PR description says image-emitting tools are "not banned here… warning-only," but the code does ban them via disallowedTools when `supportsImageInput === false`. The code comment claims subagents are "unaffected" because they have their own tool lists — but `disallowedTools` is a global deny ("Applies query-wide," per the sibling comment; SDK doc: tools "cannot be used, even if they would otherwise be allowed"). The banned names exactly match the web-browser subagent's tools (mcp__browser__browser_screenshot, browser_get_state, WebSearch — claude-code.ts:505-506) and the dashboard-builder subagent's core tool (mcp__dashboards__start_dashboard — claude-code.ts:521). Failure: user picks a text-only main model but keeps a vision-capable browser/dashboard model → browsing screenshots and dashboard verification break entirely, the exact opposite of the stated intent. Separately, since OpenRouter marks every non-Claude model supportsWebSearch:false, WebSearch gets banned in the browser subagent for essentially all non-Claude OpenRouter selections.

src/shared/lib/container/resolve-model.ts
```diff
@@ -43,3 +66,24 @@ export function getContainerModelPromptHints(resolvedModel: string | undefined):
...
+/**
+ * Tools the given (already-resolved) wire model can't use, based on the active
+ * provider's catalog capabilities: web tools when web search is unsupported,
+ * image-emitting tools when image input is unsupported. Empty for a model that
+ * supports both or whose capabilities are unknown (assume supported).
+ */
+export function getContainerUnsupportedTools(resolvedModel: string | undefined): string[] {
+  if (!resolvedModel) return []
+  try {
+    const def = getModelDefinition(resolvedModel, getActiveLlmProvider().id)
```
iddogino: Gating silently no-ops for models not found in the catalog or not explicitly selected. getModelDefinition is an exact-id lookup; a pass-through version pin (resolver step 3) or options.model === undefined (provider default) both yield undefined → [], so no ban is applied even if that effective model is incapable. Documented as "unknown → assume supported," but it means a pinned text-only model, or a user-override default pointed at an incapable model, runs un-gated while the container still serves it.

src/renderer/components/messages/model-family-list.tsx
```diff
@@ -173,6 +173,15 @@ export function ModelFamilyList({
           <span>Web search and fetch aren’t available on this model.</span>
         </div>
       )}
+      {resolved?.supportsImageInput === false && (
+        <div
+          data-testid="model-no-image-warning"
+          className="mx-1 mb-1 flex items-start gap-1.5 rounded-sm bg-amber-500/10 px-2 py-1 text-[11px] text-amber-600 dark:text-amber-500"
```
iddogino: Three near-identical amber warning <div>s. Web-search, image, and long-context warnings share identical wrapper markup + TriangleAlert; this PR pasted the third. A tiny PickerWarning component (or a {predicate, testId, text}[] map) collapses them and prevents a fourth paste.

src/shared/lib/container/resolve-model.ts
```diff
@@ -43,3 +66,24 @@ export function getContainerModelPromptHints(resolvedModel: string | undefined):
...
   }
 }
+
+/**
+ * Tools the given (already-resolved) wire model can't use, based on the active
+ * provider's catalog capabilities: web tools when web search is unsupported,
+ * image-emitting tools when image input is unsupported. Empty for a model that
+ * supports both or whose capabilities are unknown (assume supported).
+ */
+export function getContainerUnsupportedTools(resolvedModel: string | undefined): string[] {
```
iddogino: Duplicated scaffold + double catalog lookup. getContainerUnsupportedTools copy-pastes the if(!model) return [] + try/catch-getModelDefinition + empty-fallback shape from getContainerModelPromptHints (the comment even says "Same test-context fallback as…"). And base-container-client.ts:1005-1006 now resolves the model definition twice per create (getModelPromptHints and getContainerUnsupportedTools each call getModelDefinition → getEffectiveCatalog).

iddogino (changes requested):

PR #351, opened 2026-06-30
debug(microvm): bootstrap env lifecycle logging

yiw190:
## What

Temporary, value-free `[bootstrap-debug]` instrumentation to root-cause why a MicroVM agent's boot-time `GET /api/agent-bootstrap/:slug/env` returns **404** on an ECS Fargate host-app, while a later `curl` with the same `PROXY_TOKEN` returns **200** with a full env.

## Why

On the ECS+Fargate host-app POC (gamut-infra Milestone 0), every fresh MicroVM agent logs:
```
bootstrap env fetch error: HTTP Error 404
spawning agent with 0 env overrides
```
So the agent starts with no env (no `CLAUDE_CONFIG_DIR`, `PROXY_BASE_URL`, etc.) → transcript not persisted to S3, UI stuck on "Working...". Verified the route/token/slug/store are all correct *after the fact* (200 on manual curl), so the failure is timing/lifecycle, not static config. These logs align `set` / `read-hit` / `read-miss` / `clear(+caller)` against the boot fetch.

## Changes (all temporary; remove before merge)

- `agent-bootstrap-env-store.ts`: log set / read-hit / read-miss / clear (with caller stack), incl. `pid` + current store keys. No env values.
- `routes/agent-bootstrap.ts`: log enter / 401 / 403 / 404-no-env / 200.
- `lambda-microvm-runtime.ts` `start()`: log enter / before+after RunMicrovm / stashed-bootstrap-env.

## Not for merge

Pure diagnostics for `debug/microvm-bootstrap-env-logging`. Will be reverted once the root cause is confirmed.

### DIFF
--- a/src/api/routes/agent-bootstrap.ts
+++ b/src/api/routes/agent-bootstrap.ts
@@ -7,24 +7,42 @@ import { readBootstrapEnv } from '@shared/lib/container/agent-bootstrap-env-stor
 
 const agentBootstrap = new Hono()
 
+// TEMP debug (debug/microvm-bootstrap-env-logging): trace the boot fetch outcome
+// without leaking token/env. Remove before merge.
+function bootstrapRouteDebug(agentSlug: string, outcome: string, extra?: Record<string, unknown>): void {
+  try {
+    console.log(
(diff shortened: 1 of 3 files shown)


yiw190:
bug found. Not related, PR closed.

PR #358, opened 2026-06-30
feat(mobile): "Install Gamut" PWA banner on the home page

iddogino:
## What

Adds an **"Install Gamut"** banner to the home page for the mobile web/PWA build, to market install and make it one tap where the platform allows. No service worker (option A) — so this ships now without the SW's network-only review.

- `usePwaInstall()` (`hooks/use-pwa-install.ts`) — reports `{ isStandalone, canPrompt, promptInstall, method }`. Captures Chromium's `beforeinstallprompt` (preventing the default mini-infobar), listens for `appinstalled` + `display-mode: standalone` changes, and classifies the install method per platform. No-ops under Electron.
- `PwaInstallBanner` (`components/home/pwa-install-banner.tsx`) — mounted at the top of the home content.

## Install path per platform

| Platform | Banner |
|---|---|
| Chromium with a captured prompt | real one-tap **Install** button → `prompt()` → `userChoice` |
| iOS Safari | coach: tap **Share** → **Add to Home Screen** (iOS has no install API) |
| iOS non-Safari (Chrome/FF/in-app) | "Open in Safari, then Add to Home Screen" |
| Other (e.g. Android w/o a prompt) | "Open your browser menu → Add to Home Screen" |

> The one-tap button is **forward-compatible**: Chromium only fires `beforeinstallprompt` once the site is installable, which currently needs a service worker. When that SW lands, the button lights up automatically — no change here. Until then it gracefully shows instructions.

## Behavior / safety

- Hidden once installed (`display-mode: standalone` / iOS `navigator.standalone`), reacting live so it disappears on install without a reload.
- Dismissible; the choice persists in `localStorage`.
- **Desktop-safe:** renders nothing under Electron (`isElectron()`) or on a non-mobile viewport (`useIsMobile()`), so the desktop app is untouched.

## Test plan

- `tsc --noEmit` clean, `eslint` clean
- New `pwa-install-banner.test.tsx` (6 tests): hidden on desktop / when standalone / when dismissed; iOS coaching copy with no button; one-tap button fires `promptInstall`; dismissal persists. Existing `home-page.test.tsx` still green (13).


iddogino:
---

## Update: HDR-prominent styling

The banner now has a prominent animated treatment:
- **Every device:** an animated multi-color gradient border + soft colored glow.
- **Apple EDR displays (Safari + HDR screen):** the box surface and a colored rounded outline glow **brighter than SDR white** — true HDR.

**How the HDR works (so the two `<video>` tags in a banner make sense):** the ultrawhite is raw HDR-encoded video rendered *directly* (two tiny PQ/HEVC clips, ~6 KB + ~18 KB in `public/`). The catch that drove the implementation: a `backdrop-filter`, or *any* translucent/`z-index`/`isolation` layer over the video, **clamps its HDR back to SDR**. So the content card sits over the box video with **no stacking context**, and there's deliberately no `isolation` on the wrapper. All HDR is gated behind `@media (dynamic-range: high)`; non-HDR devices get the plain SDR gradient. Mobile-web only — hidden on desktop/Electron.

**Tuning knobs** (in `pwa-install-banner.css` → `@media (dynamic-range: high)`): box/halo/outline `opacity` and the ring thickness.

PR #359, opened 2026-06-30
feat(web-search): vendor-swappable web search providers

JeremyBischoff:
# feat(web-search): vendor-swappable web search providers

## What

Adds a host-side provider seam so agents can run web search through a configured vendor (Exa, Parallel, You.com, or Firecrawl) instead of Claude's Anthropic-native server-side search. Native stays the default; a configured vendor wins on every model, including Claude.

This is the canonical second instance of the platform's plug-in pattern (`BaseLlmProvider` / `BaseSttProvider` / `BaseContainerClient`), applied to web search.

## Why

- Native web search is Anthropic-server-side, so non-Anthropic models (OpenRouter GPT/GLM, etc.) get no web search at all today. A vendor-backed tool works across every model.
- Vendors can be better than native on control (result count, date windows, domain filters) and on retrieval quality, and move cost from Anthropic server-tool billing to direct vendor billing.

## How it works

- **Seam:** `BaseWebSearchProvider` owns the shared credential plumbing (settings-then-env), the transport (timeout + retry + HTTP-status classification), and the result clamp. Each vendor subclass only builds its request and maps its response (Zod-parsed at the boundary, no `as`-casts). Registry is an eager `Record<vendorId, provider>` (compile-time-fixed set, like LLM/STT) so a missing vendor is a compile error.
- **Activation (container, per session):** the active provider id (non-secret) is resolved once at `createSession` and threaded like `browserModel`. When set, the container disables native `WebSearch` (`disallowedTools`), registers an in-container `mcp__web__web_search` MCP tool, and repoints the `web-browser` subagent. No vendor never sees a key.
- **Execution (host):** `mcp__web__web_search` RPCs to `/api/x-agent/web/search`, which has its own `validateProxyToken` gate, reads the active vendor + key live, dispatches, then enforces the operator allow/deny site policy host-side and re-caps hits/snippet length before results reach the agent.
- **Settings:** a "Web Search" tab picks the provider and stores per-vendor API keys (validated via `/api/settings/validate-web-search-key`).

## Scope

Search only. Web fetch (`BaseWebFetchProvider`) is a deliberate later slice.

## Testing

- Unit: provider mapping/error/retry/validate per vendor, response-schema parsing, host route (auth + caps + policy), settings round-trip, allowed-sites filter. Full suite green.
- E2e: settings + navigation pass.
- Live-validated per vendor (Exa, Parallel, You.com, Firecrawl) through the real UI: each fired through the host route, mapped real hits, and the per-vendor date behavior matched the adapters (Firecrawl correctly dateless).

## Known limitations (accepted for v1)

- **Global-only, frozen at session creation.** The provider is a global setting threaded into a session at creation and restored on resume (identical to `browserModel`). Changing it mid-session doesn't re-route an already-running session until a new session. Corollary: a running *native* session never routes through the host, so a newly-set allowed-sites blocklist doesn't apply to it until a new session.
- Allowed-sites matching is exact-host (no subdomain/case normalization).
- The You.com settings key is stored as `youComApiKey` (renaming would orphan saved keys).
- `webSearchProvider` is stored without a Zod enum (native fallback mitigates an unknown value).

## Deferred follow-ups (not this PR)

- `BaseWebFetchProvider` seam + SSRF DNS-resolution guard.
- Per-search cost attribution (`callerSlug`).
- Allowed-sites UI editor + subdomain/case handling.
- OSS / self-host (SearXNG).


JeremyBischoff:
Closing in favor of splitting this into two smaller, more reviewable PRs: (1) the web-search provider seam + Exa reference (the pattern), and (2) Parallel/You.com/Firecrawl as pure extensions. Same code, split so the architecture can be reviewed on a minimal surface first.

PR #361, opened 2026-07-01
feat(web-search): web search provider seam + Exa

JeremyBischoff:
# feat(web-search): web search provider seam + Exa

## What

Adds a host-side provider seam so agents can run web search through a configured vendor instead of Claude's Anthropic-native server-side search. Native stays the default; a configured vendor wins on every model, including Claude. This PR ships the seam plus one reference vendor (Exa); Parallel/You.com/Firecrawl follow as pure extensions in a stacked PR.

This is the platform's plug-in pattern (`BaseLlmProvider` / `BaseSttProvider` / `BaseContainerClient`) applied to web search.

## Why

- Native web search is Anthropic-server-side, so non-Anthropic models (OpenRouter GPT/GLM, etc.) get no web search today. A vendor-backed tool works across every model.
- Vendors can beat native on control (result count, date windows, domain filters) and move cost from Anthropic server-tool billing to direct vendor billing.

## How it works

- **Seam:** `BaseWebSearchProvider` owns credentials (settings-then-env) + transport (timeout / retry / status-classification / result clamp). `ExaWebSearchProvider` only builds its request and Zod-maps its response (no casts). An eager `Record` registry means adding a vendor to the union without wiring it is a compile error.
- **Replace, not intercept:** native `WebSearch` is an Anthropic server-side tool we can't re-point. When a vendor is active, the container (once, at session creation) disables native `WebSearch`, registers an in-container `mcp__web__web_search` MCP tool, and repoints the web-browser subagent. No vendor set leaves native untouched.
- **Execution:** the MCP tool RPCs to `/api/x-agent/web/search` (its own `validateProxyToken` gate); the host reads the active vendor + key live, dispatches, enforces the operator allow/deny site policy host-side, and caps hits/snippet length. Keys never enter the container.
- **Settings:** a Web Search tab picks the provider and stores the Exa key (validated via `/api/settings/validate-web-search-key`).

### How a search flows, end to end

1. The model calls the in-container `mcp__web__web_search` tool.
2. The tool HTTP-POSTs the query to the host route `/api/x-agent/web/search` (Bearer `PROXY_TOKEN`). It carries no key and no vendor logic - a thin pipe.
3. The host authenticates, Zod-validates the body, resolves the active provider, and calls `provider.search()` - the only place the vendor is hit and the only place the API key is read.
4. The vendor's raw JSON is Zod-parsed, mapped to the normalized `WebSearchHit[]`, filtered by the operator allow/deny policy, and capped (50 hits / 2000-char snippets), all host-side.
5. The host returns `{ hits, warnings }`; the tool formats it into the text the model reads (and the `Links:` line the UI renderer parses).

Keys and policy enforcement live entirely on the host side of that HTTP hop.

## A note on the base for review

`BaseWebSearchProvider`'s shared transport (`fetchSearchJson` / `runValidation` / `clampNumResults`) is consumed by one vendor here, but it's the seam the follow-up vendors use unchanged - it was validated against all four before this PR was split out. It's the shared engine, not premature abstraction.

## Scope

Search only (web fetch is a later slice). Exa only - Parallel/You.com/Firecrawl land as a stacked follow-up that touches only the provider layer, nothing in the host route, activation, or base.

## Testing

- Unit: provider mapping/error/retry/validate, response-schema parsing, host route (auth + caps + policy), settings round-trip, allowed-sites filter, registry. Full suite green.
- E2e: settings + navigation pass.
- Live-validated end to end against the real Exa API through the UI (an agent ran real searches through the host route).

## Known limitations (accepted for v1)

- Global-only, frozen at session creation (same shape as `browserModel`); a mid-session change doesn't re-route a running session. A running native session never routes through the host, so a newly-set allowed-sites blocklist doesn't apply to it until a new session.
- Allowed-sites matching is exact-host (no subdomain/case normalization).
- `webSearchProvider` is stored without a Zod enum (native fallback mitigates an unknown value).

## Deferred (later slices)

`BaseWebFetchProvider` + SSRF DNS guard · per-search cost attribution · allowed-sites UI editor · OSS/SearXNG.


src/api/routes/web-search.ts
iddogino, line 1: x-agent was meant to be a collection of routes for the x-agent (agent talking to agent) feature. We already added chat there (which is a small crime), but we shouldn't propagate the mess by calling this x-agent as well...

src/api/index.ts
```diff
@@ -168,6 +169,7 @@ app.get('/api/llm/anthropic-sdk.js', (c) => {
 app.route('/api/agents', agents)
 app.route('/api/x-agent', xAgent)
 app.route('/api/x-agent/chat', xAgentChat)
+app.route('/api/x-agent/web', xAgentWeb)
```
iddogino: x-agent is wrong prefix
`/api/web-search`?

agent-container/src/system-prompt.md
```diff
@@ -43,7 +43,7 @@ Your tools come in sets. Depending on configuration, all tool definitions may be
 
 This catalog is an index: sets that have a dedicated section further down include a pointer to it, otherwise a one-line description is given here. Tools from remote MCP servers the user has connected appear in an additional runtime-injected "Remote MCP Servers (Available)" section; treat each connected server as another set.
 
-- **File system, shell, web** — `Read` / `Write` / `Edit` / `Bash` / `WebFetch` / `WebSearch`. The standard agent core. Search files and file contents with `find` / `grep` / `rg` via Bash.
+- **File system, shell, web** — `Read` / `Write` / `Edit` / `Bash` plus web search and web fetch tools. The standard agent core. Search files and file contents with `find` / `grep` / `rg` via Bash.
```
iddogino: These are common enough in workflows that I wonder if we want to add a conditional branch in the sys-prompt based on wether or not they appear (leave as before if using built, add actual MCP if we use that)

iddogino (changes requested):

iddogino (approved):

PR #363, opened 2026-07-01
Per-task overlap guard: at most one live run per recurring schedule (SUP-329)

iddogino:
## Summary

Part of **SUP-326** (cron hardening). A recurring task whose run outlasts its interval currently accumulates overlapping sessions — every poll spawns a fresh session without checking whether the previous run of the *same task* is still in flight. This is the root cause of the "pile of parallel tasks → exhaust resources" failure in the parent.

This adds a per-task overlap guard: before firing a recurring task, if its previous run is still **actively progressing**, hold the fire instead of spawning a second concurrent session.

## How it works

**Occupied predicate** (in `executeTaskInner`, crons only):

```
isSessionActive(lastSessionId) && !isSessionAwaitingInput(lastSessionId)
```

- A run parked on **user input** is *not* occupied — it re-fires. An offline scheduled run has nobody to answer a question, so a parked run must not wedge the task forever.
- A run with a `run_in_background` Bash/workflow still executing stays `isActive=true / isAwaitingInput=false` (it emits `session_waiting_background`), so it **is** occupied and blocks a new run.

**Hold semantics:** when occupied we do **not** advance `nextExecutionAt` — the task stays due, the next poll re-checks, and it fires the freshest run the moment the slot frees (re-anchoring forward via `getNextCronTime(now)`). Because `nextExecutionAt` is a single scalar, at most one pending fire can exist per task — coalesce-to-latest is automatic, no queue.

**Skip tracking:** each held cycle bumps `consecutiveSkips` / `lastSkippedAt`; a successful fire resets both. Consumed by the observability sub-task (**SUP-333**, which this blocks).

## Scope

Crons only. Webhooks/triggers are out of scope — they need buffer-not-skip semantics (dropping a webhook = data loss), tracked separately.

## Changes

- **`db/schema.ts`** + **migration `0026`** — additive `consecutive_skips` (int, notNull, default 0) and `last_skipped_at` (nullable) columns on `scheduled_tasks`. `ADD COLUMN` with defaults, safe on existing rows.
- **`scheduled-task-service.ts`** — new `recordTaskSkip(taskId)`; `updateNextExecution` resets the skip counters on a fire.
- **`task-scheduler.ts`** — the overlap guard, gated on `isRecurring && lastSessionId`.

## Tests

- **`task-scheduler.overlap-guard.test.ts`** — control-flow unit tests (overlap-skip, awaiting-input-fires, background-occupied-skips, hold-then-fire re-anchor, crons-only, no-prior-session).
- **`task-scheduler.overlap-guard-replay.test.ts`** — **composition test**: drives the *real* `messagePersister` with the real `background-bash-premature-idle` capture, then invokes the *real* scheduler guard against that live state (HOLD while bash pending → FIRE once settled). Mutation-verified: disabling the guard makes it fail.
- **`scheduled-task-service.test.ts`** — skip increment/reset through a real in-memory SQLite DB + the real migration.

Full scheduler + service + persister suites: **276/276 green**. Typecheck + lint clean.

## Accepted residuals (per ticket)

- After an app restart, in-memory streaming state is empty → the task fires fresh. Correct: the prior container session is gone too, no zombie.
- A genuine runaway loop (busy, never awaiting input) blocks its task indefinitely — surfaced by the skip-threshold notification in SUP-333.

### DIFF
--- a/src/api/routes/scheduled-tasks.test.ts
+++ b/src/api/routes/scheduled-tasks.test.ts
@@ -56,6 +56,7 @@ vi.mock('@shared/lib/container/container-manager', () => ({
 
 const mockMessagePersister = vi.hoisted(() => ({
   isSessionActive: vi.fn(),
+  isSessionAwaitingInput: vi.fn(),
   subscribeToSession: vi.fn(),
   markSessionActive: vi.fn(),
 }))
@@ -149,6 +150,7 @@ function createTask(overrides: Record<string, unknown> = {}) {
     scheduleExpression: '0 9 * * 1-5',
     status: 'pending',
     isRecurring: true,
+    lastSessionId: null,
     model: null,
     effort: null,
     ...overrides,
@@ -170,6 +172,7 @@ describe('scheduled-tasks route', () => {
       { id: 'session-idle', name: 'Idle session' },
     ])
     mockMessagePersister.isSessionActive.mockImplementation((sessionId: string) => sessionId === 'session-active')
+    mockMessagePersister.isSessionAwaitingInput.mockReturnValue(false)
     mockCreateSession.mockResolvedValue({ id: 'container-session-1' })
     mockEnsureRunning.mockResolvedValue({ createSession: mockCreateSession })
     mockGetSecretEnvVars.mockResolvedValue(['GITHUB_TOKEN'])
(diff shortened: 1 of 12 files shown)


iddogino:
## Review finding addressed (P1) — `493d8245`

**Finding:** a `recordTaskSkip()` failure on a held cycle escaped `executeTaskInner` as an execution failure. The recurring-task catch responds with `updateNextExecution(task.id, nextTime, '')`, which advances the schedule, resets the skip streak, and — worst — blanks `lastSessionId`, gating the guard off entirely on the next poll (`task.lastSessionId` falsy → predicate never evaluated). A single failed observability write could therefore disarm the guard and allow the exact overlap it exists to prevent.

**Fix:** the skip write is now contained inside the guard — reported via `captureException` (`phase: 'record-skip'`) but never escalated; the hold proceeds regardless. The skip counter is best-effort observability; the hold itself requires no DB write to be correct.

**Regression tests** (TDD — both failed against the pre-fix code on exactly the predicted assertion):
- skip-write rejection → `updateNextExecution` NOT called, `markTaskFailed` NOT called, no second session, failure still reported
- guard stays armed through the failed write → fires exactly once when the slot frees on a later poll

Scheduler + service + persister suites: 278/278 green.

iddogino:
## Re-review + hardening pass — `63dc34ab`

A multi-agent re-review of this PR (8 finder angles, adversarial verification per candidate) surfaced 3 correctness gaps and 3 cleanups, all fixed:

### Correctness

1. **Failure-path disarm (generalizes the earlier P1 fix).** Any transient throw while a task was held — e.g. an FS error in the pre-guard reconcile lookup, which runs every held poll — reached the recurring-failure catch, which advanced the schedule, wiped the hold streak, and blanked `lastSessionId`, disarming the guard. Worst variant: a fire failing after `createSession` (prompt already executing) but before `registerSession` looped an untracked orphan session per fire. **Fix:** new `rescheduleAfterFailure()` advances `nextExecutionAt` only (guard pointer, streak, and execution stats preserved); post-`createSession` failures record the real session id before propagating, arming the guard against the orphan.
2. **Run-now bypassed the guard.** A held task shows a stale past "next run" timestamp, making Run Now the natural user response — and it spawned unconditionally, creating exactly the overlap being held back (and rebinding `lastSessionId` away from the still-running session). **Fix:** 409 under the same occupied predicate; awaiting-input still passes through, matching the scheduler.
3. **Phantom hold streaks.** `resume`/`reset` re-anchor the schedule (abandoning the held fire) and manual runs rebind `lastSessionId`, but none cleared `consecutiveSkips`/`lastSkippedAt` — the future wedge-detection consumer would flag healthy tasks. **Fix:** all three clear the streak.

### Cleanups

- `recordTaskSkip` → single atomic SQL increment (`sql\`consecutive_skips + 1\``, same pattern as webhook-trigger `fireCount`), dropping the read-modify-write.
- Tracker IDs removed from test comments; duplicated failure-cascade narration trimmed.

### Verified non-issues (checked and cleared by the review)

- **Restart fails-open**: every orderly shutdown path (quit, crash handlers, auto-update, SIGTERM, dev-server close) stops all containers via `containerManager.stopAll()`, so the ticket's accepted residual is sound; only a hard-kill on Docker/Lima leaves a brief window, bounded by the auto-sleep reaper.
- **Pending proxy reviews counting as occupied**: correct — reviews self-resolve within the 5-min timeout (unlike unbounded input parks), and they're agent-scoped so they can't be attributed to the guarded session anyway.
- **Awaiting-input fire race**: outcome is identical under an atomic guard; it reduces to the ticket's mandated fire-while-parked semantics.

471 tests green across scheduler/service/routes/persister groups; typecheck + lint clean.

PR #374, opened 2026-07-02
feat(chat): consolidate idle conversations into durable agent memory + recap

JeremyBischoff:
## Summary
When a chat conversation (Telegram/Slack/iMessage) goes idle past its `sessionTimeout`, its context is currently lost. This adds a background sweep that consolidates the finished conversation into (1) durable, typed memories written into the agent's persistent memory store, and (2) a short recap that seeds the next conversation in that chat. It fires even for abandoned conversations, which the lazy rotation path never catches.

## How it works
- **Rotate (sweep):** off the existing 5-min health-check interval, `selectRotationTargets` finds active conversations idle past the timeout and archives+rotates them (sets `rotated_at`), mirroring the lazy `resolveActiveSession` path including connector teardown.
- **Consolidate (sweep):** `selectConsolidationTargets` picks archived+rotated, still-allowed conversations (oldest-first, capped 10/tick) and calls `consolidateConversation`, which reads the transcript, makes one structured-output LLM call, writes the memory files + a `MEMORY.md` pointer index, and stashes the recap. `consolidated_at` is committed atomically (`WHERE consolidated_at IS NULL`).
- **Recap injection:** the next new conversation's system prompt is seeded with the prior conversation's recap (most-recent timeout-rotation only, no stale fallback), framed as reference-only.

**Session lifecycle (the eligibility crux):**

| State | `archived_at` | `rotated_at` | `consolidated_at` | Sweep action |
|---|---|---|---|---|
| Active | null | null | null | rotate **if** idle past timeout |
| Timeout-rotated | set | set | null | **consolidate** |
| Consolidated | set | set | set | none (excluded) |
| `/clear` / self-heal / revoke | set | null | null | none (never consolidated or recapped) |

Consolidation only ever runs on the timeout-rotated state, so a consolidated row can't be revived and non-timeout archives are never mined.

## Migration `0026` - three nullable columns on `chat_integration_sessions`
- **`rotated_at`** (`timestamp_ms`, null) - set **only** when a session is archived by a `sessionTimeout` rotation. This is the marker that distinguishes a timeout rotation from a `/clear` / self-heal / revoke archive (all of which leave it null). It gates **both** consolidation eligibility and recap eligibility. It has to be a stored column because `archived_at` alone can't say *why* a session was archived, and `isSessionTimedOut` only looks at `updated_at` - so without an explicit marker an old `/clear`'d row would be mistaken for a timed-out conversation and wrongly consolidated.
- **`recap`** (`text`, null) - the continuity summary the consolidation produces, stashed on the row so the *next* conversation in that chat can be seeded with it. Null until consolidated; empty string on a terminal-empty commit. It lives on the session row because the recap belongs to the conversation that ended, and `getLatestTimeoutRecap` reads it off the most-recent archived row.
- **`consolidated_at`** (`timestamp_ms`, null) - the atomic commit marker. Null = not yet consolidated (a sweep candidate); set = done. Tracked **separately from `archived_at`** because a conversation is rotated at one time and consolidated later (the sweep runs afterward), and the `WHERE consolidated_at IS NULL` guard is what makes the at-least-once sweep idempotent - no double-consolidation, no double-commit.

Existing rows default all three to null and behave correctly (excluded from consolidation until they time out and rotate).

## Key decisions
- **Model:** consolidation reuses the user-settable **summarizer** model (default `haiku`) - the same seam the other background-summary call sites use - rather than a hidden dedicated default. Honors the user's model setting; structured output verified on that tier. (The earlier dedicated `consolidator` `ModelPurpose` was removed.)
- **Trigger:** the existing manager sweep, not a new scheduler or per-conversation timer.
- **Default timeout:** new integrations default to a 6-hour session timeout, so rotation + consolidation are on by default (blank = never disables both). Existing integrations are unchanged. The feature is fully gated on this: with the timeout set to never, nothing rotates and consolidation never runs.
- **Shared memory is intentional:** memories are per-agent and shared across that agent's conversations (single-user / personal-bot is the primary scope).

## Security & robustness
- The transcript is **untrusted input**. It's fenced with a per-call `crypto.randomUUID()` nonce so content can't break out into instruction position; the existing memory index is fenced the same way (so prior model-written descriptions can't act as a second-order injection while still being visible for correct reuse); the prompt frames both as data, not instructions, and instructs memory bodies to be neutral observations.
- **Injection posture (by design):** the summarizer faithfully records a preference the transcript shows the user stating and the assistant agreeing to. This is inherent to any model (verified: haiku and sonnet behave identically), so a stronger model doesn't help. The primary defense against a malicious behavioral instruction is the **conversation agent refusing it in-band** (verified live - the real bot declines, so the transcript records a refusal), backed by memories being human-inspectable files. On a multi-user bot this shares one user's mined facts agent-wide, consistent with the intentional shared-memory design.
- **Durability:** memory files and the `MEMORY.md` index are written atomically (temp + fsync + rename), so a mid-write failure can never truncate an existing memory or the index.
- **Failure discipline:** deterministic failures (empty transcript, refusal, truncation, unparseable output, 400/413/422) commit an empty fallback so a bad conversation isn't retried forever; transient errors (auth/rate-limit/5xx/network, transient FS errnos) rethrow and retry next tick. Memory parsing is element-wise, so one bad entry doesn't discard the rest.

## Blast radius
- **No new routes, middleware, or auth-scope changes.** The one access-control touch is a re-check of the existing `isChatAllowed` predicate before a durable write.
- **No `agent-container` change**, so no image rebuild.
- Exactly one hot-path change: `resolveActiveSession` now calls rotate (tagging `rotated_at`) instead of archive-only. Everything else is the background sweep or the LLM call.

## Diff footprint
The `+4,003` headline is mostly auto-generated. The real reviewable surface is ~720 lines of production code (461 of it in the one new service file) against ~900 lines of tests.

| Category | Added | Removed |
|---|---|---|
| Production (`.ts`, non-test) | +722 | -16 |
| Tests (`*.test.ts`) | +909 | -106 |
| Drizzle (auto-generated) | +2,372 | 0 |

Of the drizzle total, `meta/0026_snapshot.json` alone is 2,362 lines - regenerated wholesale by `drizzle-kit`, never hand-edited. The migration itself is 3 lines of SQL.

## Testing
- Unit + integration + e2e across the consolidation service, sweep selectors, session lifecycle, migration, and providers (green in the affected suites; typecheck + lint clean).
- Live-smoked on the real Telegram bot: idle conversation consolidated → typed memory files + `MEMORY.md` index written → a fresh conversation recalled the banked facts and was seeded with the recap. Create/update/dedup decisions verified against the live model.

## Known & accepted (by-design / low)
- A user returning within one sweep interval of the timeout boundary gets no recap for that one conversation (best-effort continuity).
- An `output_config`-incompatible summarizer model degrades to an empty consolidation (shared behavior with the other summarizer call sites).
- Under sustained disk failure, orphaned `.tmp` files are bounded and never read.


JeremyBischoff:
Closing. Too far behind main to rebase. SUP-348 tracks the idea for a fresh design.

PR #388, opened 2026-07-03
feat(composer): swap stop/send button based on composer content while agent works

westaskew:
## Summary

While the agent is working (or background tasks linger), the composer action slot now shows a single context-sensitive button instead of stop + send side by side:

- **Empty composer** → Stop button
- **User types anything** → swaps to Send (labelled "Queue message" mid-turn, "Send message" when only background tasks remain); sending mid-turn queues as before
- **Text cleared** → swaps back to Stop

## Implementation

- `useMessageComposer` exposes a new `hasContent` flag (non-whitespace text, attachments, or active voice recording — the same content predicate `canSubmit` already used, so the two can't drift).
- `ComposerActionButton` renders Stop only when `(isActive || isWaitingBackground) && !hasContent`; otherwise the Send button.
- The swap deliberately keys off `hasContent`, not `canSubmit`: with text typed but sending blocked (offline, runtime starting), the user sees a disabled Send rather than the button confusingly flipping back to Stop.

## Testing

- `composer-action-button.test.tsx` rewritten for the swap behavior, including the type→send / clear→stop round-trip and both background-waiting states; `message-input.test.tsx` gains an integration test typing/clearing real input (60 unit tests pass).
- E2E: `message-queueing.spec.ts` and `background-bash.spec.ts` updated to assert stop-only before typing and the swap after; all 11 tests across the affected specs (message-queueing, background-bash, stop-session) pass in mock mode.
- Typecheck + lint clean for the changed files.

### DIFF
--- a/src/renderer/components/messages/composer-action-button.tsx
+++ b/src/renderer/components/messages/composer-action-button.tsx
@@ -4,6 +4,7 @@ import { Button } from '@renderer/components/ui/button'
 interface ComposerActionButtonProps {
   isActive: boolean
   isWaitingBackground: boolean
+  hasContent: boolean
   canSubmit: boolean
   isSending: boolean
   isInterrupting: boolean
@@ -13,62 +14,48 @@ interface ComposerActionButtonProps {
 export function ComposerActionButton({
   isActive,
   isWaitingBackground,
+  hasContent,
   canSubmit,
   isSending,
   isInterrupting,
   onInterrupt,
 }: ComposerActionButtonProps) {
-  // While the agent works (or background tasks linger) show Stop alongside
-  // Send — messages sent mid-turn are queued and picked up by the agent loop.
-  if (isActive || isWaitingBackground) {
+  // While the agent works (or background tasks linger) the slot holds a single
+  // button: Stop while the composer is empty, swapping to Send as soon as the
+  // user types — messages sent mid-turn are queued and picked up by the agent
+  // loop. Clearing the composer brings Stop back.
+  if ((isActive || isWaitingBackground) && !hasContent) {
     const stopLabel = isWaitingBackground ? 'Stop background processes' : 'Stop the agent'
-    const sendLabel = isWaitingBackground ? 'Send message' : 'Queue message'
     return (
-      <div className="flex items-center gap-2">
-        <Button
-          type="button"
-          size="icon"
-          variant="outline"
-          className="h-[34px] w-[34px]"
-          onClick={onInterrupt}
-          disabled={isInterrupting}
-          aria-label={stopLabel}
-          title={stopLabel}
-          data-testid="stop-button"
-        >
-          {isInterrupting ? (
-            <Loader2 className="h-4 w-4 animate-spin" />
-          ) : (
-            <Square className="h-3.5 w-3.5 fill-current" />
-          )}
-        </Button>
-        <Button
-          type="submit"
-          size="icon"
-          className="h-[34px] w-[34px]"
-          disabled={!canSubmit || isSending}
-          aria-label={sendLabel}
-          title={sendLabel}
-          data-testid="send-button"
-        >
-          {isSending ? (
-            <Loader2 className="h-4 w-4 animate-spin" />
-          ) : (
-            <ArrowUp className="h-4 w-4" />
-          )}
-        </Button>
-      </div>
+      <Button
+        type="button"
+        size="icon"
+        variant="outline"
+        className="h-[34px] w-[34px]"
+        onClick={onInterrupt}
+        disabled={isInterrupting}
+        aria-label={stopLabel}
+        title={stopLabel}
+        data-testid="stop-button"
+      >
+        {isInterrupting ? (
+          <Loader2 className="h-4 w-4 animate-spin" />
+        ) : (
+          <Square className="h-3.5 w-3.5 fill-current" />
+        )}
+      </Button>
     )
   }
 
+  const sendLabel = isActive && !isWaitingBackground ? 'Queue message' : 'Send message'
   return (
     <Button
       type="submit"
       size="icon"
       className="h-[34px] w-[34px]"
       disabled={!canSubmit || isSending}
-      aria-label="Send message"
-      title="Send message"
+      aria-label={sendLabel}
+      title={sendLabel}
       data-testid="send-button"
     >
       {isSending ? (
--- a/src/renderer/components/messages/composer-action-button.tsx
+++ b/src/renderer/components/messages/composer-action-button.tsx
@@ -4,7 +4,10 @@ import { Button } from '@renderer/components/ui/button'
 interface ComposerActionButtonProps {
   isActive: boolean
   isWaitingBackground: boolean
+  /** User-entered composer content (restored drafts and a wordless live recording don't count). */
   hasContent: boolean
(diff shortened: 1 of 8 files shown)


src/renderer/components/messages/composer-action-button.tsx
```diff
@@ -13,62 +14,48 @@ interface ComposerActionButtonProps {
...
   onInterrupt,
 }: ComposerActionButtonProps) {
-  // While the agent works (or background tasks linger) show Stop alongside
-  // Send — messages sent mid-turn are queued and picked up by the agent loop.
-  if (isActive || isWaitingBackground) {
+  // While the agent works (or background tasks linger) the slot holds a single
+  // button: Stop while the composer is empty, swapping to Send as soon as the
+  // user types — messages sent mid-turn are queued and picked up by the agent
+  // loop. Clearing the composer brings Stop back.
+  if ((isActive || isWaitingBackground) && !hasContent) {
```
iddogino: **Blocker: double-clicking Send mid-turn interrupts the agent.**

`handleSubmit` clears the message synchronously before awaiting the POST (`use-message-composer.ts:177` — no intervening `await` in the no-attachment/no-voice path), so `hasContent` flips false between the two clicks of a double-click. Because both branches render the same `Button` component type at the same tree position, React mutates the *same DOM node* in place into an **enabled** Stop (`disabled` only by `isInterrupting`). The second click — or an impatient re-click on a slow network, since the `isSending` spinner is now unreachable mid-turn and there's no feedback the first click registered — fires `onInterrupt` and kills the very turn the user just queued into. Pre-PR the second click hit a `disabled` Send harmlessly.

Cheap mitigation: suppress the swap (or disable Stop) while `sendMessage.isPending`.

src/renderer/components/messages/composer-action-button.tsx
```diff
@@ -13,62 +14,48 @@ interface ComposerActionButtonProps {
...
     const stopLabel = isWaitingBackground ? 'Stop background processes' : 'Stop the agent'
```
iddogino: **Blocker: Stop can be unreachable without the user ever typing — and there's no other stop affordance.**

`useMessageComposer` seeds `message` from the draft store at mount (`use-message-composer.ts:29-30`, key `session:${sessionId}`), so re-entering a busy/runaway session with a saved draft shows Send-only from first render; the user must select-all-delete their draft (which also wipes the stored copy) just to reach Stop. The only other interrupt UIs are `StopSessionButton` on interactive-request cards and the crashed-card Dismiss — no keyboard shortcut exists.

The background-waiting case is the sharpest: a lingering/runaway background task keeps the container awake, and a draft leaves zero UI path to "Stop background processes" — exactly the scenario the previous two-button layout served. (Drafts are in-memory, so this is within one app run, not across restarts.)

src/renderer/components/messages/composer-action-button.tsx
```diff
@@ -13,62 +14,48 @@ interface ComposerActionButtonProps {
...
-          data-testid="send-button"
-        >
-          {isSending ? (
-            <Loader2 className="h-4 w-4 animate-spin" />
-          ) : (
-            <ArrowUp className="h-4 w-4" />
-          )}
-        </Button>
-      </div>
+      <Button
```
iddogino: **A11y: keyboard focus drops to `<body>` on every swap.**

Stop and Send are separate elements in separate return branches with no shared identity, so when the focused button unmounts (activating Send clears the message; typing replaces a focused Stop; a failed send's `setMessage(content)` restore flips it back), `document.activeElement` falls to body and a keyboard/screen-reader user loses their place.

Rendering one `<Button>` whose `type`/`onClick`/label/testid vary with the mode fixes this *and* makes the double-click hazard above impossible, since the node keeps its `disabled` semantics through the transition.

src/renderer/components/messages/composer-action-button.tsx
```diff
@@ -13,62 +14,48 @@ interface ComposerActionButtonProps {
...
+        {isInterrupting ? (
+          <Loader2 className="h-4 w-4 animate-spin" />
+        ) : (
+          <Square className="h-3.5 w-3.5 fill-current" />
+        )}
+      </Button>
     )
   }
 
+  const sendLabel = isActive && !isWaitingBackground ? 'Queue message' : 'Send message'
```
iddogino: **Cleanup: queueing predicate duplicated across two files.**

`isActive && !isWaitingBackground` here is the exact expression `message-input.tsx:80` computes as `queued` to decide the real send semantics (strip model/effort, mark the ghost queued). If the queueing rule ever changes, the label silently diverges from behavior. Consider passing a single derived `willQueue` prop down from the container instead.

src/renderer/hooks/use-message-composer.ts
```diff
@@ -197,7 +197,8 @@ export function useMessageComposer(options: UseMessageComposerOptions) {
     }
   }
 
-  const canSubmit = (!!message.trim() || attachments.length > 0 || voiceInput.isRecording) && !isUploading && !submitDisabled
+  const hasContent = !!message.trim() || attachments.length > 0 || voiceInput.isRecording
```
iddogino: **Voice dictation hides Stop before any words exist.**

`voiceInput.isRecording` counts as `hasContent`, so the moment the mic starts (empty transcript) the slot flips to an enabled Send. Pressing it stops the recording and, with an empty transcript, silently no-ops (`handleSubmit` bails at line 129) — recording killed, nothing sent, agent still running. It's one confusing press deep rather than a permanent trap (Stop reappears after the no-op), but consider excluding `isRecording` from the button-mode predicate while keeping it in `canSubmit`.

Minor naming note: this exported `hasContent` shadows the same-named local inside `handleSubmit` (line 128) which has deliberately different semantics (post-voice `effectiveMessage`, no `isRecording`). Renaming the local (e.g. `hasSubmittableContent`) would prevent a future editor from updating one notion of "has content" and missing the other.

iddogino (changes requested):
Thanks — the implementation is tidy and the tests genuinely cover the new swap behavior, but the single-slot swap has some real behavioral consequences that I'd like addressed before merge. Findings 1–3 below (inline) are the blockers; they all stem from the swap happening in-place with no allowance for in-flight sends/interrupts or restored drafts.

One structural change addresses most of it: render a **single button element** whose `type`/handler/label/testid vary (instead of two separate `<Button>`s in separate return branches), keep it in Stop mode while `isInterrupting`, and disable rather than swap while `sendMessage.isPending`.

Two findings can't be anchored inline because the lines aren't in the diff:

**A. `src/renderer/components/messages/message-input.tsx:101` — typing during a pending interrupt hides all interrupt feedback and enables Send.** After clicking Stop (spinner shown, disabled via `isInterrupting`), one keystroke unmounts that state and renders a fully enabled "Queue message" Send — neither `submitDisabled` (`sendMessage.isPending || isOffline || !isRuntimeReady`) nor the Send branch accounts for `interruptSession.isPending`. The user can queue a message against a session mid-teardown with zero indication an interrupt is in flight (the server does recompute `queued` authoritatively and the client reconciles, but whether an interrupted turn ever consumes that queued message is unverified).

**B. `e2e/pages/session.page.ts:239` — `isAgentWorking()` is now a latent false-negative.** It equates stop-button visibility with "agent working", which is no longer true whenever the composer has text. Zero callers today, but please update or delete it before a future spec trips on it.

Whether restored drafts should suppress Stop (inline finding on the stop branch) is a product call worth an explicit decision rather than an accident of `hasContent`.

PR #389, opened 2026-07-03
fix(telegram): dedupe redelivered updates; await disconnect on reconnect

JeremyBischoff:
## What

Rarely, the Telegram bot delivers a reply twice. Root cause: Telegram's `getUpdates` long polling is at-least-once. An update stays in Telegram's queue until we confirm it (by advancing the offset), so any interruption between handling an update and confirming it - a reconnect on resume, a brief poller overlap - makes Telegram redeliver it. We had no idempotency, so a redelivery ran a second agent turn and the user saw the whole response twice.

## Changes

- **Idempotency gate.** New `UpdateDeduplicator` (bounded, exact-match, count-based), owned by the manager per integration so it survives the connector teardown/recreate of a reconnect - which is exactly when the redelivery arrives. The connector checks `update_id` at the top of each message-bearing handler, before first-poll batching, and drops repeats (breadcrumb, not error). Callback queries are deduped too, since a redelivered multi-select toggle would otherwise flip a checkbox back off and corrupt the answer.

- **No message loss during teardown.** A disconnecting connector now ignores incoming updates, so a teardown-window update isn't recorded in the shared deduplicator and then dropped as a "duplicate" by the new connector.

- **No poller overlap on reconnect.** Reconnect now awaits the old connector's `disconnect()` before starting the new poller, so two pollers never share the token (removes 409 thrash and a rare lost-update window). The await is bounded, so a hung stop can't wedge the reconnect, and the connections-map delete is guarded by identity so a concurrent reconnect can't orphan a live poller. `stop()` stays fire-and-forget so app quit is snappy.

## Not covered (on purpose)

- The gate is in-memory, so it catches same-process redeliveries; a cross-restart redelivery is avoided instead by grammY resuming at the last confirmed offset.
- Two processes on one bot token still double-deliver (separate notepads). That's "don't run two pollers," not something dedup can fix.
- General per-integration reconnect serialization (a broader lifecycle race) is a separate follow-up.

## Verification

- Unit tests for the deduplicator, connector dedup on text/photo/document/callback, the disconnecting guard, the bounded awaited disconnect, and the concurrent-reconnect identity guard (verified to fail without the fix).
- Full chat-integrations suite green (589 tests). Typecheck and lint clean.

### DIFF
--- a/src/shared/lib/chat-integrations/chat-integration-manager-duplicate-delivery.test.ts
+++ b/src/shared/lib/chat-integrations/chat-integration-manager-duplicate-delivery.test.ts
@@ -1,13 +1,14 @@
 /**
  * Duplicate-delivery hardening: a per-integration UpdateDeduplicator owned by the manager
- * (survives a reconnect), and a bounded awaited disconnect so two pollers never overlap.
+ * (survives a reconnect, resets on a bot-token change, dropped on permanent delete), and a
+ * bounded awaited disconnect so two pollers never overlap.
  */
 
 import { describe, it, expect, vi, afterEach } from 'vitest'
 import { chatIntegrationManager } from './chat-integration-manager'
 import { UpdateDeduplicator } from './update-deduplicator'
 
-// Reach the manager singleton's private seams (matches the sup231 spec's pattern).
+// Reach the manager singleton's private seams (as the sibling manager tests do).
 const mgr = chatIntegrationManager as any
 
 function fakeConn(id: string, disconnect: () => Promise<void>) {
@@ -25,25 +26,54 @@ afterEach(() => {
   vi.useRealTimers()
   mgr.connections?.clear?.()
   mgr.deduplicators?.clear?.()
+  mgr.disconnectedSince?.clear?.()
+  mgr.consecutiveFailures?.clear?.()
 })
 
 describe('ChatIntegrationManager — per-integration deduplicator', () => {
   it('uses a distinct deduplicator per integration', () => {
-    const a = mgr.getOrCreateDeduplicator('int-dedup-2')
-    const b = mgr.getOrCreateDeduplicator('int-dedup-3')
+    const a = mgr.getOrCreateDeduplicator('int-dedup-2', 'tok-a')
+    const b = mgr.getOrCreateDeduplicator('int-dedup-3', 'tok-b')
     expect(a).toBeInstanceOf(UpdateDeduplicator)
     expect(b).not.toBe(a)
   })
 
+  it('reuses the same deduplicator for the same integration + token', () => {
+    const a = mgr.getOrCreateDeduplicator('int-reuse', 'tok-a')
+    const b = mgr.getOrCreateDeduplicator('int-reuse', 'tok-a')
+    expect(b).toBe(a)
+  })
+
   it('removeIntegration preserves the deduplicator so a redelivery after reconnect is still dropped', async () => {
-    const dedup = mgr.getOrCreateDeduplicator('int-survive')
-    dedup.isDuplicate('9001') // an update handled before the reconnect
+    const dedup = mgr.getOrCreateDeduplicator('int-survive', 'tok-a')
+    dedup.markDelivered('9001') // an update delivered before the reconnect
     mgr.connections.set('int-survive', fakeConn('int-survive', () => Promise.resolve()))
 
     await mgr.removeIntegration('int-survive') // the reconnect teardown
 
-    expect(mgr.getOrCreateDeduplicator('int-survive')).toBe(dedup) // same instance survives
+    expect(mgr.getOrCreateDeduplicator('int-survive', 'tok-a')).toBe(dedup) // same instance survives
     expect(dedup.isDuplicate('9001')).toBe(true) // and still remembers the pre-reconnect update
+  })
+
+  it('resets the deduplicator when the bot token changes (a new bot must not inherit old update ids)', () => {
+    const old = mgr.getOrCreateDeduplicator('int-token', 'tok-old')
+    old.markDelivered('7001')
+
+    const fresh = mgr.getOrCreateDeduplicator('int-token', 'tok-new') // token changed → new notepad
+    expect(fresh).not.toBe(old)
+    expect(fresh.isDuplicate('7001')).toBe(false) // does not inherit the old bot's ids
+  })
+
+  it('forgetDeduplicator drops the notepad on permanent delete (no leak, no stale carry-over)', () => {
+    const dedup = mgr.getOrCreateDeduplicator('int-del', 'tok-a')
+    dedup.markDelivered('5001')
+
+    mgr.forgetDeduplicator('int-del')
+
+    expect(mgr.deduplicators.has('int-del')).toBe(false) // evicted, not leaked
+    const recreated = mgr.getOrCreateDeduplicator('int-del', 'tok-a')
+    expect(recreated).not.toBe(dedup) // a fresh id reused later starts clean
+    expect(recreated.isDuplicate('5001')).toBe(false)
   })
 })
 
@@ -90,5 +120,30 @@ describe('ChatIntegrationManager — removeIntegration connection identity', () 
     await p
 
     expect(mgr.connections.get('int-race')).toBe(connB) // must not orphan connB
+  })
+
+  it('leaves the swapped-in connection sibling state intact (does not wipe the new connection)', async () => {
+    let releaseDisconnect: () => void = () => {}
+    const connA = fakeConn('int-race2', () => new Promise<void>((r) => { releaseDisconnect = r }))
+    const connB = fakeConn('int-race2', () => Promise.resolve())
+    mgr.connections.set('int-race2', connA)
+
+    const p = mgr.removeIntegration('int-race2') // captures connA, awaits its disconnect
+    mgr.connections.set('int-race2', connB)      // concurrent reconnect swaps in connB...
+    mgr.disconnectedSince.set('int-race2', 123)  // ...and records fresh sibling state for it
+    releaseDisconnect()
+    await p
+
+    // The old teardown must not wipe the new connection's state, same as it must not delete connB.
+    expect(mgr.disconnectedSince.get('int-race2')).toBe(123)
+  })
+
+  it('still clears sibling state on a genuine teardown (no swap)', async () => {
+    mgr.connections.set('int-plain', fakeConn('int-plain', () => Promise.resolve()))
+    mgr.disconnectedSince.set('int-plain', 456)
+
+    await mgr.removeIntegration('int-plain')
+
+    expect(mgr.disconnectedSince.has('int-plain')).toBe(false) // cleaned when this teardown owns the id
   })
 })
--- a/src/shared/lib/chat-integrations/chat-integration-manager-duplicate-delivery.test.ts
+++ b/src/shared/lib/chat-integrations/chat-integration-manager-duplicate-delivery.test.ts
@@ -1,0 +1,94 @@
+/**
+ * Duplicate-delivery hardening: a per-integration UpdateDeduplicator owned by the manager
+ * (survives a reconnect), and a bounded awaited disconnect so two pollers never overlap.
+ */
+
+import { describe, it, expect, vi, afterEach } from 'vitest'
+import { chatIntegrationManager } from './chat-integration-manager'
+import { UpdateDeduplicator } from './update-deduplicator'
+
+// Reach the manager singleton's private seams (matches the sup231 spec's pattern).
+const mgr = chatIntegrationManager as any
+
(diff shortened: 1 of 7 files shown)


iddogino (commented):
In-depth review (multi-angle + adversarially verified). Core design is right — manager-owned dedup surviving reconnects, exact-match not watermark, count-bound not TTL, and I verified against grammY's `stop()` source that the bounded 3s disconnect **cannot** produce a sustained dual-poller (the loop is aborted before the confirmation `getUpdates`; a "hung stop" is a dying one-shot request), so that tradeoff holds as stated.

Two confirmed findings should hold merge though: the deduplicator records an update id at **accept** time, but two async paths can then fail to **deliver** — and since the dedup survives the reconnect, it suppresses the redelivery that used to recover the message. The PR converts a recoverable transient loss into a permanent one in exactly the teardown window it hardens. Details inline (findings 1–2), plus six smaller items.

Also checked and clean: all four `bot.on` registrations are guarded (verified against `allowed_updates`), every production `TelegramConnector` construction passes the dedup, `.catch`-before-race prevents late unhandled rejections, and test coverage of the intended behaviors is genuinely good.

src/shared/lib/chat-integrations/telegram-connector.ts
```diff
@@ -797,7 +806,26 @@ export class TelegramConnector extends ChatClientConnector {
...
+      // message be traced (the one failure mode a dedup gate can introduce).
+      addErrorBreadcrumb({ category: 'chat-integration', message: 'Dropped duplicate Telegram update', data: { updateId } })
+      return true
+    }
+    return false
+  }
+
   private handleTextMessage(ctx: GrammyContext): void {
+    if (this.disconnecting) return
+    if (this.isDuplicateUpdate(ctx)) return
```
iddogino: **[1 — confirmed, message loss]** First-poll buffered messages are permanently lost on a teardown-window reconnect.

This line records the `update_id` in the reconnect-surviving deduplicator **before** the first-poll branch below buffers the text (no emit yet). On teardown the manager unsubscribes first (`disconnectConnection` runs `conn.messageUnsubscribe?.()` before `connector.disconnect()`), then `disconnect()` calls `flushAllPendingBatches()` — which, unlike the timer path (`if (!this.disconnecting) flushBatch(...)`), has **no** `disconnecting` check — so the batch is emitted into an empty handler list and vanishes. The id stays recorded, so when Telegram redelivers to the new connector, `isDuplicateUpdate` drops it: the message is gone for good.

Pre-PR the redelivery recovered this message; now the loss is permanent. This is the D1 failure mode the `disconnecting` guard targets — but the guard only covers updates that *arrive* during teardown, not ones already buffered.

Suggested invariant: an id enters the deduplicator only once its message has actually been handed off (record at successful emit, or un-record on failure-to-deliver).

src/shared/lib/chat-integrations/telegram-connector.ts
```diff
@@ -185,6 +186,8 @@ export class TelegramConnector extends ChatClientConnector {
 
     // Handle photo messages (images sent directly, not as documents)
     this.bot.on('message:photo', async (ctx) => {
+      if (this.disconnecting) return
+      if (this.isDuplicateUpdate(ctx)) return
```
iddogino: **[2 — confirmed, message loss]** Same record-then-lose gap via the post-await window (here and in the document handler).

The `disconnecting`/`isDuplicateUpdate` checks run at handler entry, then the handler `await`s `bot.api.getFile(...)`. If teardown lands during that round-trip, the eventual `emitMessage` fires after the manager unsubscribed (the `catch` around `getFile` swallows errors and still emits with `url: ''`), so the file message is lost — while the already-recorded id suppresses the redelivery on the new connector. Narrower window than the batching case, same root cause: id recorded at accept, delivery not guaranteed.

A post-await `if (this.disconnecting) return` before `emitMessage` narrows it, but the record-at-handoff invariant from finding 1 fixes both sites at once.

src/shared/lib/chat-integrations/chat-integration-manager.ts
```diff
@@ -168,6 +180,8 @@ class ChatIntegrationManager {
   // chain settles (see scheduleQueueEviction), so the map stays bounded.
   private messageQueues: Map<string, Promise<void>> = new Map()
   private lastSessionTouch: Map<string, number> = new Map()
+  // Per-integration dedup notepad; owned here so it survives a connector reconnect (see UpdateDeduplicator).
+  private deduplicators: Map<string, UpdateDeduplicator> = new Map()
```
iddogino: **[3 — confirmed, low]** This map leaks on permanent integration deletion. `removeIntegration` intentionally preserves the entry (correct for reconnect), but the DELETE route calls `removeIntegration` → `deleteChatIntegration` and nothing evicts it — only `stop()` clears the map. One orphaned Set of up to 1000 ids per deleted Telegram integration, for process lifetime.

Fix: `this.deduplicators.delete(id)` in the permanent-delete path (not in `removeIntegration`, to keep the reconnect-survival property).

src/shared/lib/chat-integrations/chat-integration-manager.ts
```diff
@@ -438,6 +458,16 @@ class ChatIntegrationManager {
     }
   }
 
+  /** The per-integration dedup notepad, created on first use and reused across reconnects. */
+  private getOrCreateDeduplicator(integrationId: string): UpdateDeduplicator {
```
iddogino: **[4 — plausible, low]** A botToken change keeps the old bot's dedup ids. The PATCH reconfigure path (`removeIntegration` → `addIntegration`) keeps the same integration id, so a *different* bot inherits up to 1000 of the old bot's update ids. Update-id sequences are per-bot with arbitrary large starts, so a false-duplicate drop of a real message is theoretically possible (though astronomically unlikely). Cheap fix: reset the deduplicator when the token changes — retention is only meaningful for a same-bot reconnect.

src/shared/lib/chat-integrations/chat-integration-manager.ts
```diff
@@ -268,11 +284,15 @@ class ChatIntegrationManager {
...
       }
     }
-    // Remove the connection
+    // Remove the connection. Await the teardown so a reconnect (removeIntegration →
+    // connectIntegration) fully stops the old poller before starting the new one.
     const conn = this.connections.get(id)
     if (conn) {
-      this.disconnectConnection(conn)
-      this.connections.delete(id)
+      await this.disconnectConnection(conn)
```
iddogino, line 306: **[5 — confirmed, low]** This await serializes teardowns on the paths that loop over integrations: `reconnectAll` (system resume — the one case where `bot.stop()` genuinely can stall on a half-open socket) now has a worst case of 3s × N before the last bot resumes polling. The teardowns are independent; `Promise.allSettled` over them would collapse resume to ~3s total. Low priority at typical 1–3 integrations.

src/shared/lib/chat-integrations/chat-integration-manager.ts
```diff
@@ -268,11 +284,15 @@ class ChatIntegrationManager {
...
+      // Delete by identity, not key: the await above is a yield point, so a concurrent
+      // reconnect may have swapped in a new connector for this id — deleting by key would
+      // orphan that live, still-subscribed poller (same guard as the messageQueues eviction).
+      if (this.connections.get(id) === conn) this.connections.delete(id)
```
iddogino, line 310: **[6 — plausible, low]** The identity guard protects `connections` but not its siblings: `disconnectedSince.delete(id)`, `consecutiveFailures.delete(id)`, and the `messageQueues` sweep below still run unconditionally after the yield point, so when the identity check detects a swapped-in connection they wipe the *new* connection's state. Not reachable as harm today (those maps are only written by the 5-minute health loop, which can't repopulate within the 3s window), but moving them inside the same `if (this.connections.get(id) === conn)` block is cheap defensive symmetry.

src/shared/lib/chat-integrations/telegram-connector.ts
```diff
@@ -614,6 +619,10 @@ export class TelegramConnector extends ChatClientConnector {
    *  and only resolves on Done. The callback is answered inside each branch so the empty-Done
    *  toast can fire (a callback can be answered exactly once). */
   private async handleCallbackQuery(ctx: GrammyContext): Promise<void> {
+    if (this.disconnecting) return
+    // Dedupe callback redeliveries too: a redelivered multiSelect toggle would flip a
+    // checkbox back off (its callback mapping isn't consumed on tap), corrupting the answer.
+    if (this.isDuplicateUpdate(ctx)) return
```
iddogino: **[7 — cleanup]** The `disconnecting` + `isDuplicateUpdate` guard pair is now copy-pasted into all four handlers. A single `bot.use((ctx, next) => { if (this.disconnecting) return; if (this.isDuplicateUpdate(ctx)) return; return next() })` registered before the `.on()` handlers would centralize the gate and automatically cover any future handler (`message:video`, voice, …). Coverage is complete today (all four registrations vs `allowed_updates: ['message','callback_query']`), but it's opt-in per handler with no test forcing the invariant.

src/shared/lib/chat-integrations/chat-integration-manager-duplicate-delivery.test.ts
```diff
@@ -0,0 +1,94 @@
+/**
+ * Duplicate-delivery hardening: a per-integration UpdateDeduplicator owned by the manager
+ * (survives a reconnect), and a bounded awaited disconnect so two pollers never overlap.
+ */
+
+import { describe, it, expect, vi, afterEach } from 'vitest'
+import { chatIntegrationManager } from './chat-integration-manager'
+import { UpdateDeduplicator } from './update-deduplicator'
+
+// Reach the manager singleton's private seams (matches the sup231 spec's pattern).
```
iddogino: **[8 — convention]** Issue ID in a code comment (`sup231`) — repo convention is no Linear/issue IDs in code comments or test filenames; keep the explanatory text, drop the ID (PR/commit messages are fine).

src/shared/lib/chat-integrations/telegram-connector.ts
JeremyBischoff: Done in `7f38411a` - dedup records at handoff now, not accept. The first-poll batch marks its ids only on a real flush, so a teardown flush leaves them unrecorded and the redelivery recovers the message.

src/shared/lib/chat-integrations/telegram-connector.ts
JeremyBischoff: Done in `7f38411a` - same record-at-handoff fix. Photo/document re-check `disconnecting` after `getFile` and bail before emit, recording only on a successful emit.

src/shared/lib/chat-integrations/chat-integration-manager.ts
JeremyBischoff: Done in `7f38411a` - added `forgetDeduplicator(id)` on the delete path only; `removeIntegration` still keeps the notepad for reconnect survival.

src/shared/lib/chat-integrations/chat-integration-manager.ts
JeremyBischoff: Done in `7f38411a` - the map is keyed by `{dedup, botToken}` and resets when the token changes, so a new bot can't inherit the old ids.

src/shared/lib/chat-integrations/chat-integration-manager.ts
JeremyBischoff: Done in `7f38411a` - `reconnectAll` tears down concurrently via `Promise.allSettled`, each still awaited internally, so resume isn't 3s x N.

src/shared/lib/chat-integrations/chat-integration-manager.ts
JeremyBischoff: Done in `7f38411a` - moved the sibling cleanups under the same identity guard. Guarded on the swap specifically rather than nesting inside `if (=== conn)`, so the no-connection case still clears stale state.

src/shared/lib/chat-integrations/telegram-connector.ts
JeremyBischoff: Done in `7f38411a` - centralized into one `bot.use` gate (covers future handlers); it does the check only, recording moved to handoff per 1-2. Added a forcing test across all handler types.

src/shared/lib/chat-integrations/chat-integration-manager-duplicate-delivery.test.ts
JeremyBischoff: Done in `7f38411a` - dropped the `sup231` id (reworded, since it cross-referenced the sibling test's pattern). A few other existing `SUP-xxx` refs around - left them; happy to do a separate cleanup if you want.

JeremyBischoff:
Closing. Parked under SUP-486 with the other Telegram work. The duplicate-reply bug is still live on main.

PR #398, opened 2026-07-06
fix(env): fail-closed .env reads + line-preserving merges — stop the secret-wiping race

iddogino:
## The bug (observed in prod, authmode)

Providing a secret via `request_secret` wiped `/workspace/.env` down to just the new key — every other secret (SUPABASE_URL, service keys, …) gone.

**Root cause chain:**
1. The provide-secret route double-writes: host `setSecret()` (lock → merge → atomic rename), then container `POST /env` milliseconds later.
2. Since the atomic-write work (#320), every host write is a temp-file **rename** — a new inode each time. On the shared S3 File Gateway NFS, the container pod's cached lookup then hits **ESTALE** instead of seeing the new file.
3. The container's `updateEnvFile` had a bare `catch {}` around its read: *any* error meant "file doesn't exist yet, start fresh." It merged the one new key into an empty view and durably, atomically wrote the wipe.

The wiped file's fingerprint confirms this path: no host header, container-style quoted value, and mode `0600` — which `writeFileAtomic` only applies when its `stat` of the target *also* failed.

## Fixes

- **Fail-closed reads** (`agent-container/src/env-file-store.ts`): `readEnvFileOrNull` retries transient errors across the NFS attribute-cache window (~2.3s) and **throws** on anything persistent that isn't ENOENT. `POST /env` now 500s loudly instead of wiping; only a confirmed-absent file means "create".
- **ENOENT-scoped `existingMode` stat** in all four atomic writers (host `file-storage.ts` + container `atomic-file.ts`, async + sync): an unstattable target fails the write instead of silently resetting perms.
- **Line-preserving container merge**: `upsertEnvContent` edits only the target key's line — the host's header, display-name comments, and every other line survive byte-for-byte. File creation is 0o666 (umask-applied) matching the host, so cross-uid writes keep working.
- **Escape/unescape symmetry** (found by the new cross-implementation stress test): host `serializeEnvFile` escaped `"` but `parseEnvFile` never unescaped, so quoted/JSON secrets gained an escape level on **every** read-modify-write cycle. Both serializers and the parser now do symmetric `\\` / `\"` / `\n` handling, backward-compatible with existing files.

## Tests

- Wipe regression: host-format `.env` + container upsert → all prior bytes survive; persistent ESTALE → update rejects, file untouched, lock released.
- `secrets-service.crossprocess.test.ts`: host `setSecret` and container `updateEnvFileEntry` (the two real prod writers) interleaved 20-deep on one file — no key lost, no stray locks/temps, display names and header survive both writers.
- `*.statscope.test.ts` (host + container): non-ENOENT stat errors fail the write and leave the target untouched; ENOENT still creates with the requested mode.
- Round-trip suite: values with quotes/JSON/backslashes/newlines survive repeated serialize→parse cycles unchanged.
- Full runs green: host 6294 unit / container 420 unit / tsc both / lint 0 errors / secrets E2E specs 17 passed.

## Not changed (deliberately)

`POST /env` keeps writing the file: the CONNECTED_ACCOUNTS and REMOTE_MCPS callers have no host-side write, and non-shared-FS runtimes need it. The residual double-write in provide-secret is now harmless (fail-closed + line-preserving).

### DIFF
--- a/agent-container/src/atomic-file.statscope.test.ts
+++ b/agent-container/src/atomic-file.statscope.test.ts
@@ -1,0 +1,89 @@
+/**
+ * existingMode stat scoping in the atomic writers: a non-ENOENT stat error
+ * (NFS ESTALE after a cross-client rename, EIO) means "state unknown", not
+ * "file absent" — the write must FAIL instead of proceeding as a create, which
+ * would silently reset the target's permissions to the caller's default mode
+ * (the 0o666 → 0o600 flip observed in the /workspace/.env wipe).
+ *
+ * Lives in its own file with a module-level fs mock: Node's builtin module
+ * properties are non-configurable, so vi.spyOn cannot patch fs.statSync.
+ */
+import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
+import * as path from 'path';
+import * as os from 'os';
+
+const statFailure = vi.hoisted(() => ({ active: false }));
+
+vi.mock('fs', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('fs')>();
+  const estale = () =>
+    Object.assign(new Error('ESTALE: stale file handle'), { code: 'ESTALE' });
+  return {
+    ...actual,
+    statSync: ((...args: Parameters<typeof actual.statSync>) => {
+      if (statFailure.active) throw estale();
+      return actual.statSync(...args);
+    }) as typeof actual.statSync,
+    promises: {
+      ...actual.promises,
+      stat: (async (...args: Parameters<typeof actual.promises.stat>) => {
+        if (statFailure.active) throw estale();
+        return actual.promises.stat(...args);
+      }) as typeof actual.promises.stat,
+    },
+  };
+});
+
+import * as fs from 'fs';
+import { writeFileAtomic, writeFileAtomicSync } from './atomic-file';
+
+let tmpDir: string;
+
+beforeEach(() => {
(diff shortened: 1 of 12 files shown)


iddogino:
**Second commit — revised root cause, found via deterministic local repro (Linux DooD).**

After the first commit, providing a secret from the request card failed with `EACCES: permission denied, open '/workspace/.env'` while the UI path worked. The workspace showed `-rw------- 1 root root .env` — which cracks the original mystery too:

The rename-based atomic write **transfers file ownership to whichever process wrote last**, while "preserve existing mode" carries a stray `0o600` forward. The .env started 0600-owned-by-container (old container create mode); the host's first rename write made it 0600-owned-by-host → the container can no longer even read it. Before the fail-closed fix, that EACCES became "start fresh" → the wipe. After it, POST /env correctly errors — which is exactly what surfaced. No NFS staleness required (ESTALE remains a covered secondary on the File Gateway, but the owner-flip lockout reproduces on plain Linux).

Fix: the agent .env is deliberately a two-uid file, so both writers now pass a new `forceMode` option to `writeFileAtomic` (host + container, async + sync) — every write re-asserts **exact 0o666** via fchmod on the temp fd (not umask-reduced, not preserved). This also self-heals any already-stuck 0600 file on the next successful write from either side (e.g. one secret edit from the UI).

Tests: stuck-0600 heal tests for host `setSecret` and container `updateEnvFileEntry`, forceMode override/create tests for all four writers, and the one existing assertion that encoded the old umask-reduced expectation updated to the forced-exact-0666 contract. Full suites re-run green: host 6320, container 422, tsc + lint clean.

PR #400, opened 2026-07-06
fix(env): force 0o666 on the agent .env — atomic renames flip ownership and can lock the other writer out

iddogino:
Follow-up to #398, which turned the silent secret wipe into a loud failure — this closes the failure itself, and revises the root cause based on a deterministic local repro.

## The repro (Linux DooD, observed after #398)

Providing a secret from the request card fails:

```
[provide-secret] Failed to set env var: {"error":"EACCES: permission denied, open '/workspace/.env'"}
```

while adding the same secret from the Secrets UI works. The workspace shows `-rw------- 1 root root .env`.

## Revised root cause (simpler than NFS staleness)

The rename-based atomic write (#320) **transfers file ownership to whichever process wrote last**, while its "preserve existing mode" carries a stray `0o600` forward:

1. `.env` starts `0o600` owned by the **container** user (the old container create-mode).
2. The host's first rename-write makes it `0o600` owned by the **host** uid.
3. The container can now no longer even *read* the file. Before #398's fail-closed read, that EACCES became "file doesn't exist, start fresh" → **the wipe**. After #398, `POST /env` correctly errors — which is exactly what surfaced.

No NFS involvement required — this reproduces on plain Linux DooD. (The ESTALE retry from #398 remains useful as a transient-error belt on the File Gateway.)

## Fix

The agent `.env` is deliberately a **two-uid file**, so its mode is a contract, not a preference. New `forceMode` option on `writeFileAtomic` (host `file-storage.ts` + container `atomic-file.ts`, async + sync): re-assert the exact requested mode via fchmod on the temp fd — not umask-reduced, not preserved-from-existing. The three `.env` writers (`setSecret`, `deleteSecret`, `updateEnvFileEntry`) pass `{ mode: 0o666, forceMode: true }`.

This also **self-heals** an already-stuck 0600 file on the next successful write from either side (one secret edit from the UI is enough). Every other `writeFileAtomic` caller keeps the existing preserve-mode behavior.

## Tests

- Stuck-0600 heal regressions: host `setSecret` and container `updateEnvFileEntry` restore 0o666 over a poisoned file.
- `forceMode` override + exact-mode-on-create tests for all four atomic writers.
- Updated the one assertion that encoded the old umask-reduced expectation to the forced-exact-0666 contract.
- Full suites green on this base: host 6320 / container 422, tsc + lint clean.

### DIFF
--- a/agent-container/src/atomic-file.test.ts
+++ b/agent-container/src/atomic-file.test.ts
@@ -93,4 +93,21 @@ describe('writeFileAtomic (async)', () => {
     expect(fs.statSync(target).isDirectory()).toBe(true);
     expect(leftoverTmpFiles(tmpDir)).toEqual([]);
   });
+
+  // forceMode: for two-uid files (the agent .env) the mode must be applied on
+  // EVERY write — the rename transfers ownership, so preserving a stray
+  // restrictive mode locks the other writer out permanently.
+  it.runIf(process.platform !== 'win32')('forceMode overrides an existing restrictive mode', async () => {
+    const p = path.join(tmpDir, 'shared.env');
+    fs.writeFileSync(p, 'A=1\n');
+    fs.chmodSync(p, 0o600);
+    await writeFileAtomic(p, 'A=2\n', 0o666, { forceMode: true });
+    expect(fs.statSync(p).mode & 0o777).toBe(0o666);
+  });
+
+  it.runIf(process.platform !== 'win32')('forceMode applies the exact mode on create (not umask-reduced)', async () => {
+    const p = path.join(tmpDir, 'fresh.env');
+    await writeFileAtomic(p, 'A=1\n', 0o666, { forceMode: true });
+    expect(fs.statSync(p).mode & 0o777).toBe(0o666);
+  });
 });
--- a/agent-container/src/atomic-file.test.ts
+++ b/agent-container/src/atomic-file.test.ts
@@ -110,4 +110,33 @@ describe('writeFileAtomic (async)', () => {
     await writeFileAtomic(p, 'A=1\n', 0o666, { forceMode: true });
     expect(fs.statSync(p).mode & 0o777).toBe(0o666);
   });
+
+  // Ownership preservation: the rename replaces the inode, which would hand
+  // the file to this process's uid:gid. Group is the one dimension testable
+  // without root — a process may chgrp a file it owns to any of its
+  // supplementary groups.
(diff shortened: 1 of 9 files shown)


iddogino:
Added best-effort **owner + group preservation** to all four atomic writers (suggested follow-through on the ownership-flip finding): the existing target's uid:gid is restored via fchown on the temp fd before the rename lands, matching npm `write-file-atomic` semantics. This protects every other rewritten file in privileged-writer setups (e.g. a root host process rewriting user-owned workspace files in DooD).

Why this *complements* rather than replaces `forceMode` for the .env: chown to another uid requires CAP_CHOWN, so preservation is inherently best-effort — the container's non-root `claude` writer can never give the file back to the host's uid. For a file two uids must both read and write, the only uid-independent contract is the forced world-RW mode; for everything else, preservation is now the default behavior.

Tests: group-restoration across atomic overwrites (async + sync, host + container) — group is the one ownership dimension assertable without root, via chgrp to a supplementary group. Full suites green: host 6300 / container 424, tsc + lint clean.

iddogino:
**Third commit — startup auto-heal for already-poisoned agents.**

Question that prompted it: what happens to agents whose .env is *already* 0600/owner-flipped from the buggy version? Answer before this commit: nothing auto-fixed them —

- **Host-owned 0600**: container locked out until the next host secret write (the request-card flow does heal post-forceMode, but only if someone provides a secret).
- **Container-owned 0600 + non-root host**: a hard wedge. `getSecretEnvVars` at session start throws EACCES → session creation 500s → the container never boots → the only process that *can* chmod the file (its owner) never runs. Secrets UI 500s and fail-closed `setSecret` refuses too, so there was no recovery path at all.

Now both sides self-heal on agent start — each side can always chmod a file it owns, and the two poisoned states are each owned by exactly one side:

1. **Container boot**: `healEnvFilePermissions('/workspace/.env')` — best-effort chmod back to 0666.
2. **Host session start** (`getSecretEnvVars`): chmod-heal when owner; if the file is *still* unreadable (EACCES/EPERM), start the session with no secret names instead of 500ing — names-only degradation, logged loudly — so the container boots and heals the file it owns; the next session recovers fully. All mutating paths (`setSecret`/`deleteSecret`/`updateEnvFileEntry`) remain strictly fail-closed.

Tests: heal fixes 0600→0666 without touching content / no-op on correct or absent files (container), heal-on-session-start + wedge-avoidance degradation (host). Full suites green on both packages.

PR #405, opened 2026-07-07
fix: eliminate permanently stuck chat working indicators

JeremyBischoff:
## What

Kills every cause of a permanently stuck "Working…" chat indicator except the two known-open ones (subagent input requests and proxy review, which have their own PRs). Since the indicator became a faithful pull-based projection of session state, any state-tracking bug shows up as a light that never turns off - across Telegram, Slack, iMessage, and the desktop sidebar. This branch closes that class at the state layer instead of patching surfaces.

## How

Seven commits, layered:

1. **Telegram surface teardown** - a no-stream terminal turn now blanks the un-committed draft in place; the first streamed token hands the draft to the stream instead (`yieldingToStream`).
2. **Turn-start route ordering** - `isActive` can no longer outlive a failed send: the existing-session send routes clear host state when the send never reached the container.
3. **Slack reaction teardown** - reaction tracking is dropped only on confirmed removal (Slack reactions never expire on their own), added optimistically on ambiguous failures, and per-chat writes are serialized.
4. **Grace-after-result backstop** - when a turn's `result` was seen but the settling `idle` never arrives (runtime hang in the wind-down), the host settles itself after 30s, completion notification included. The gate is derived and re-evaluated on every message: it arms only when the runtime's last word was "work done" and disarms on any turn activity, so it cannot fire during a genuinely running turn - verified against a real queued-continuation capture where the next turn starts 73ms after the previous result.
5. **Lossless container-to-host stream** - every relayed message carries a per-epoch sequence number; the host keeps an (epoch, seq) cursor and passes it on every re-subscribe, so a reconnect replays exactly what the gap swallowed (a lost `idle` or background-task terminal used to pin the session forever). An epoch mismatch means the old container process died: the host settles immediately, drops the dead background tasks, and fires the completion notification only if the reply's result was actually seen. Socket close/error handlers got identity guards (a replaced socket's late close used to be able to orphan the new socket and double-deliver).
6. **Audited residuals** - a whole-branch adversarial audit (two independent model sweeps plus coverage and conformance passes) surfaced and this branch fixes: two pre-existing paths where a dead subscription entry made `isSubscribed()` lie forever (the next send then ran with no stream attached - unfixably stuck); new sessions missing a fast first turn's terminal events because `createSession` starts the turn before the host subscribes (all six new-session paths now attach with a from-start replay); half-open sockets dying silently (the host never writes on the stream socket, so it now pings and terminates on a missed pong, routing into the normal reconnect-plus-replay recovery); and a set of boundary hardenings (cursor clamps, URL-parse edge, envelope narrowing).
7. **Coverage pins** for the regression directions none of the existing tests could see.

## Compatibility / rollout

- The stream protocol changes are version-skew safe in both directions: an old container image sends no epoch and the host behaves exactly as before; an old host ignores the new fields and attaches live-only. The E2E mock deliberately speaks the legacy protocol, pinning that path.
- `agent-container/` changed, so the fix only reaches live agents after an image rebuild + container recreate.

## Verification

- 6355 unit/integration tests green, typecheck and lint clean, rebased onto current main.
- New suites: grace state machine (10, fake-timer, incl. a real-capture replay), cursor/reconcile (11), connection recovery (5), client socket lifecycle + liveness (8), container replay/parse (15), plus route, Slack, and Telegram pins.
- Live smoke on a rebuilt image (normal round trip + sleep-the-laptop-mid-background-task) tracked separately before marking ready.

### DIFF
--- a/agent-container/src/server.ts
+++ b/agent-container/src/server.ts
@@ -1,6 +1,7 @@
 import { Hono } from 'hono';
 import { serve } from '@hono/node-server';
 import { SessionManager } from './session-manager';
+import { parseStreamCursor } from './stream-replay';
 import { CreateSessionRequest, SendMessageRequest } from './types';
 import type { UUID } from 'crypto';
 import * as http from 'http';
@@ -1648,11 +1649,7 @@ server.on('upgrade', (request: http.IncomingMessage, socket: any, head: Buffer) 
     const sessionId = sessionMatch[1];
     // Optional resume cursor: the host's last-processed position. Replay picks
     // up exactly after it (same epoch) so no message is lost across reconnects.
-    const cursorEpoch = url.searchParams.get('epoch');
-    const sinceSeqRaw = url.searchParams.get('since_seq');
-    const sinceSeq = sinceSeqRaw !== null && Number.isInteger(Number(sinceSeqRaw))
-      ? Number(sinceSeqRaw)
-      : null;
+    const { epoch: cursorEpoch, sinceSeq } = parseStreamCursor(url.searchParams);
 
     wss.handleUpgrade(request, socket, head, (ws: WebSocket) => {
       handleWebSocketConnection(ws, sessionId, cursorEpoch, sinceSeq);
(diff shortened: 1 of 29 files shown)


JeremyBischoff:
Closing in favor of the split PRs, as planned in #564.

PR #412, opened 2026-07-07
feat(chat): /stop + stall nudge to recover hung turns (Telegram docked Stop keyboard)

JeremyBischoff:
## What

Chat integrations get a way to recover a hung agent turn. If a turn gets stuck, you can stop it from the chat instead of being stranded on "Working…" forever.

Two visible pieces:
- **`/stop`** (and a docked **⏹ Stop** button on Telegram) interrupts the current turn. The conversation stays - your next message runs as a fresh turn.
- A **stall nudge**: after 7 minutes of silence on a turn the agent still owes a reply on, the chat sends one message noting it might be stuck and pointing at `/stop`.

## Why

Follow-up to #339. That made the working indicator honest (it stops claiming to be busy when it isn't), but it left a genuinely stuck turn showing "Working…" forever with no way out from chat - matching the desktop app, whose only recovery is its Stop button. This gives chat that same Stop, and the nudge is what makes an otherwise-invisible command discoverable.

## What changed, in three layers

**1. The recovery.** The app Stop button's interrupt logic was pulled out into one shared helper (`interrupt-session.ts`), so chat `/stop` runs the *exact same* code as the app button, not a parallel copy. That helper always settles the turn locally even against a dead or wedged container, so Stop works no matter what the container does, bounded at 15s. `/stop` uses it; `/clear` now also interrupts a busy turn before it archives; and the stall nudge is a plain silence timer that never touches the indicator.

**2. The Telegram draft change.** This is the part worth explaining. A streaming reply is sent as a Bot API *draft*, and a live draft blocks the user from sending to that chat until it finalizes - so a typed `/stop` could not get through mid-turn. The fix: during working/thinking, the indicator is no longer a draft, it is a **real editable message** (with the ⏹ Stop button docked on it), so `/stop` reaches the bot in exactly the phase where turns hang. To be clear, draft streaming is **not** removed - the reply still streams as a smooth draft. Only the *indicator* moved off drafts, because that is the phase that has to accept a Stop.

**3. Concurrency.** `/stop` skips the serial per-chat queue (a priority lane) so it can preempt an in-flight turn instead of waiting behind a cold container start. Because it now runs concurrently with normal-message handling, two races are closed explicitly: a per-chat `stopInFlight` latch keeps a double-tap from double-interrupting or trailing a contradictory ack, and a per-chat interrupt epoch keeps a follow-up message that was already mid-flight when you hit Stop from re-activating the turn you just stopped (the follow-up is abandoned before the container send, again before `markSessionActive`, and before the session-gone self-heal could recreate it as a fresh session).

## How it was verified

- 621 unit/e2e tests, typecheck and lint clean. The `/stop` e2e block covers interrupt/ack/settle, the group `/stop@bot` form, nothing-running, sibling-chat isolation, stale-error suppression, the access gate, and a dead container; dedicated tests pin the concurrent-`/stop` guard and the three resurrection races (a follow-up racing Stop mid-send, before the send, and via session-gone self-heal).
- Manual smoke on live Telegram across ~12 cases: mid-turn stop, multi-tool flicker, a question card mid-turn, a long silent tool, queue-then-stop, rapid multi-tap, a mid-stream follow-up, `/clear`, an error turn, and a long overflowing reply.

## Notes for the reviewer

- Known limitation (accepted): the working indicator is now a real message instead of an auto-expiring draft, so a rare teardown or restart race can leave a stale "Working…" message in the chat. The Stop button self-heals on the next turn.
- `/stop`, like the app Stop button, interrupts the agent's foreground work; a background task it spawned can outlive the stop (same behavior as the app).
- Known limitation (accepted): the interrupt epoch above tracks chat `/stop`, not the desktop app's Stop button (which interrupts the same session through a different path). So the rare case where the app Stop is pressed at the exact moment a chat follow-up is mid-send can still re-activate - a pre-existing, narrow cross-surface edge, out of scope here.


JeremyBischoff:
Closing. #993 reworked interrupts under this. Chat /stop is a fresh PR if we still want it.

PR #413, opened 2026-07-07
feat: branded .agent/.skill file types — custom icons + double-click import (SUP-307)

iddogino:
## Summary

Gives agent-template and skill packages a first-class branded identity on disk and in the OS.

**Phase A — extensions.** The three export endpoints (`export-template`, `export-full`, `skills/:dir/export`) now serve `.agent`/`.skill` files as `application/octet-stream` (so browsers don't "correct" the name to `.zip`), with human-readable display-name filenames in `Content-Disposition` (quoted + RFC 5987 `filename*`, matching the workspace-file convention; slugs are opaque minted ids so they no longer leak into filenames). `downloadBlob` prefers and decodes the server filename. Import dialogs accept the branded extensions plus legacy `.zip`. Extension literals live in one shared module (`package-extensions.ts`); a unit test mechanically enforces sync with the `fileAssociations` block below.

**Phase B — OS integration.** electron-builder `fileAssociations` registers both types as Owner with custom document icons on macOS/Windows (generated by `build/file-icons/make-icons.js`; top-level associations verified to merge with — not clobber — the quick-dispatch "Any File" mac entry and Windows generics). Double-clicked/`Open With` files flow through the existing race-free pull model: main queues the path, classifies it **from disk by zip content** (root `CLAUDE.md` → agent template, root `SKILL.md` → skill; the extension only *orders* the validators, so renamed files import as what they actually are and broken packages get the right diagnostic), and the renderer's `PackageImportHandler` shows a confirm dialog (agent, with optional full import + upload progress) or an agent picker (skill). Per-kind size caps are enforced pre-upload, before the user picks a target.

Bytes cross to the renderer exactly once, at import time, via the pre-authorized `read-local-file` path — classification sends only the verdict.

**Review hardening (multi-angle review, all findings fixed):**
- Synchronous re-entrancy guard on both import handlers — a double-click during the disk-read round-trip can no longer duplicate an import or silently drop the next queued package
- Skill size cap checked before zip validation (as a diagnostic, so an oversized template renamed `.skill` still classifies by content); over-cap `.skill` files get the 100MB message up front instead of a misleading 500MB one
- Skill picker shows a loading spinner / error state while agents load instead of "No agents"
- `exportSkill` returns `{ zipBuffer, skillName }` so downloads use the SKILL.md title, not the directory name
- Shared `readLocalFileAsFile()` replaces four inline copies (the new one had dropped the MIME type); `ClassifiedImportPackage` declared once and reused across main/preload/renderer; repaired `download.test.ts`

## Test plan

- [x] `npx tsc --noEmit` clean; eslint clean on touched files
- [x] Unit: classifier (8), package-extensions + fileAssociations sync (3), download (4), opened-files (7), agents routes (170), skillset-service (161) — all green
- [x] E2E (mock): `agent-import.spec.ts` incl. new branded-.agent import test, `skill-export-import.spec.ts` incl. `.skill` round-trip — 8/8
- [x] Packaged build (`electron-builder --mac dir`, driven over CDP): Info.plist doc types + preserved "Any File" entry; cold-start open → Import Agent dialog → real import; warm `.skill` open → picker → import; corrupt file → error toast; Finder icons verified via lsregister + NSWorkspace
- [ ] True cold-start double-click on a real (installed/signed) build — needs a distributed artifact

Linear: SUP-307


iddogino:
Follow-up `4e6f5266`: skill exports now wrap entries in a top-level `<skill-dir>/` folder (the Anthropic-convention layout `detectZipPrefix` already strips), and import naming falls back **frontmatter `name` → wrapper folder name → `imported-skill`** (in both `validateSkillZip`, which feeds the double-click dialog, and `importSkillFromZip`).

This fixes a round-trip asymmetry: a skill whose SKILL.md has no frontmatter exported with the right *filename* but re-imported as "Imported Skill", because the name previously existed only in the download filename — which import deliberately never trusts. Now the name travels inside the package bytes, so it survives renames/re-zips, and frontmatter-less Cowork-style packages import under their wrapper name too.

Back-compat both ways: old versions importing new wrapped exports already strip the prefix; new versions still import old flat exports (last-resort fallback unchanged).

Tests: wrapped-layout export assertions, wrapper-name fallback in validate + import, flat-package fallback preserved, and a frontmatter-less export→import round-trip regression test. Unit 166/166, `skill-export-import` E2E 4/4 through the real UI.

PR #418, opened 2026-07-08
feat: Gamut platform web vendor for search + fetch (SUP-367)

JeremyBischoff:
## What

Adds a Gamut-provided "Platform" web vendor covering both web search and web fetch, alongside the existing Exa (BYOK) and native options. Platform routes through the platform proxy on the user's Gamut login (`Authorization: Bearer <platform token>` to `{proxy}/v1/exa/search` and `/v1/exa/contents`), so there is no per-user API key. The proxy is a transparent passthrough, so both responses parse with the existing Exa schemas and reuse Exa's mappers.

## Why

Web vendors beat native tools, and native is the only thing a user gets if they never open Settings. When unset and a Gamut login is present, Platform is the default - once the user picks a vendor, that choice sticks (same as LLM/STT). No silent fallback to another vendor when a credential breaks.

## Vendor resolution (sticky)

Same shape as LLM/STT providers:

| stored pin | GET `webProvider` | `webProviderIsDefault` |
|---|---|---|
| unset | Platform if Gamut login, else native | `true` |
| `platform` / `exa` / `native` | that pin | `false` |

GET `webProvider` is always the concrete vendor the agent runs. PUT still writes the stored pin (`null` clears). A pin is a contract: broken credential = broken service. Exa is never auto-selected.

## A billing bug this turned up, in two places

Building the vendor exposed a defect in how container-facing routes talk to the platform proxy. Two commits here exist only to fix it.

**Container requests didn't tell the proxy which member was acting, so it couldn't find a seat to bill and rejected the request as unpaid** - already true for the host browser, and it would have been true for web search/fetch too.

The global fetch interceptor rewrites `Authorization` to `<token>::<memberId>` for proxy calls, but only when an attribution scope is open. `Authenticated()` opens one for browser requests. Container-facing routes have no session, so nothing opened one, and an org-scoped bearer reached the proxy naming no acting member.

The proxy accepts that (`auth.ts` returns `memberId: null`, role `org_runtime`). Billing does not: `/v1/exa` and `/v1/browserbase` are both gated routes, so `billingGate()` runs with `seatId = null`, finds no per-seat subscription (`billing/kv.ts`), and denies a seat-subscribed org holding no pool credits with **402 insufficient_balance**.

- **Web search / fetch:** introduced by this PR. Fixing it is part of making the feature correct.
- **The host browser:** already live on `main`. Fixed here too (one root cause, one helper).

The proxy token names an *agent*, not a user, so both gates now resolve the agent's owner and run the handler under their attribution scope. Single-user installs have no `agent_acl` row, resolve to `null`, open no scope, and keep sending their already member-scoped access key: unchanged.

With both gates doing the same work, `RequireProxyToken()` was deleted as a duplicate; the two web routes use `IsAgent`.

**AUTH_MODE live attribution is proven** on staging with an org-scoped `PLATFORM_TOKEN`: container web search/fetch returned 200 and outbound Authorization was `token::memberId`.

## What changed

- `PlatformWebProvider` - one class exposing both `search` and `fetch`, swapping only URL + auth against `ExaWebProvider` (Bearer instead of `x-api-key`).
- Sticky `resolveEffectiveWebVendor()` - pin stands; unset defaults to Platform-if-login / native.
- `IsAgent()` opens the agent owner's attribution scope; `RequireProxyToken()` deleted. `getAgentOwnerUserId()` lifted out of `timezone-resolver`, ordered by `createdAt`.
- `GET /settings` returns the active `webProvider` plus `webProviderIsDefault`. Web tab pre-selects that vendor and marks `(default)` when unset.
- `PUT /settings` validates the id at the boundary; accepts `null` to clear back to unset default.
- `validate-web-key` rejects `platform` before dispatch (login-based).
- Runtime proxy failures map to actionable copy by status: 401 / 403 / 402.
- Exa request bodies built once and shared by both vendors.

## How verified

**Live, against the Gamut staging proxy** (real `PlatformWebProvider`):

| probe | result |
|---|---|
| bogus token -> `POST /v1/exa/search` | `401` |
| `POST /v1/exa/answer` (non-allowlisted) | `404` |
| `provider.search(...)` | 2 hits |
| `provider.fetch(...)` | title + content + host-stamped `fetchedAt` |

**Live AUTH_MODE (org-JWT `PLATFORM_TOKEN`, throwaway data dir):**

| probe | result |
|---|---|
| container `POST /v1/exa/search` | `200`, Authorization `token::memberId` |
| container `POST /v1/exa/contents` | `200`, Authorization `token::memberId` |

**Unit:** request shape, response mapping, remedy table, sticky pin / unset→Platform / unset→native (no Exa auto-pick), GET active vendor + isDefault, PUT null-clear, Web tab default marker / login gate, and org-scoped bearer gaining `::memberId` under real ALS from inside the `IsAgent` handler.


iddogino (changes requested):
Pending change - remove automated escalation

PR #419, opened 2026-07-08
fix: evict idle session processes in agent container to stop OOM accumulation

yiw190:
## Bug

Every session in the agent container keeps its `claude` subprocess (~250MB RSS) parked forever after its turn ends: the SDK query's for-await loop waits for the next user message, and nothing ever stops it. `SessionManager` has no idle reaper, and no caller ever hits `DELETE /sessions/:id` in normal operation.

Recurring scheduled tasks make this fatal: `TaskScheduler.executeTaskInner` creates a **new session per execution**, so the container gains one immortal ~250-340MB process per run until the cgroup OOM-kills it. Observed in two production deployments:

- `rug-s-personal`: hourly task → **46 claude processes**, total RSS pinned exactly at the 4GB container limit → repeated OOM kills, collateral session corruption on a second agent.
- `adegoke-s-personal`: 3 tasks every ~15 min → **34 processes / ~10GB demand** against an 8g limit → OOM crash-restart loop, surfacing to the user as "The assistant hit an error and stopped" via the iMessage integration.

## Repro

Against an unpatched `agent-container` image: create N sessions via `POST /sessions` (the scheduler's exact call), let each turn complete (`result: success`), and watch `ps`: process count and RSS grow strictly linearly (1 proc / ~250MB per session) and never shrink — still resident minutes after all work finished.

## Fix

Add an idle reaper to `SessionManager` with **split thresholds** by session class:

| Class | Signal | Default | Env override |
|---|---|---|---|
| Interactive (chat / iMessage / etc.) | no `metadata.isAutomated` | **5 min** | `SESSION_IDLE_EVICTION_MINUTES` |
| Automated (cron / webhook) | `metadata.isAutomated: true` | **0** (next sweep once idle) | `SESSION_AUTOMATED_IDLE_EVICTION_MINUTES` |

- A **30s** sweep stops the subprocess of any session that is idle past its class threshold. `< 0` disables that class; `0` means "evict on the next sweep after idle" (not mid-turn) — worst case ~30s of parked RSS after an automated turn ends.
- Idleness is driven by the CLI's `session_state_changed` events, **gated on a `result` for this turn** (same stale-idle race guard as `message-persister`). `result` alone is the fallback when state events are absent. Sessions with running background tasks are never evicted mid-job.
- Eviction stops **only the subprocess**: the `SessionData` entry, WebSocket subscribers, and the process's `claudeSessionId` all survive, so the next `sendMessage` restarts the query with `--resume` transparently. No UI/API behavior change.
- `sendMessage` clears `lastResultSubtype`, sets busy, and awaits any in-flight eviction so a racing message can't hit a closing queue.
- Browser lock is released on eviction. Chat integrations intentionally use the interactive threshold (long-lived conversations).
- **Sessions resumed via a bare `getSession()` are born-`idle`, not born-`busy`** (second commit): a `GET /sessions/:id` on a cold session spawns a process that never gets a message, never emits a `result`, and under born-`busy` would sit beyond the reaper forever — the same immortal-process class, through the read path. `sendMessage` sets busy itself right after resuming, so the send path is unaffected.

## Verification

- Unit tests in `session-manager-idle-eviction.test.ts` (interactive / automated thresholds, stale-idle ignore, background-task holdoff, resume-after-evict, getSession-resume evictability); agent-container suite green.
- Local image with interactive override 1m: idle session reaped; mem dropped ~450MB/2 procs → ~220MB/1 proc.

## Notes

- Workflow/Agent subagents do not spawn per-subagent OS processes on CLI 2.1.197 — leak is per-session main processes only.
- Env semantics: `0` = next sweep, `< 0` = disabled. These env vars are new.
- Sweep is 30s so dense webhook bursts reclaim sooner; finest cron cadence is still ~60s.

## Next steps (follow-up PRs, deliberately out of scope here)

1. **Lazy resume** — `resumeSession` should not start the subprocess at all; a bare `GET /sessions/:id` currently pays ~250MB for a read (the reaper now cleans it up, but the spawn is pure waste — a freshly resumed session's in-memory `messages` is empty either way). First `sendMessage` already cold-starts with `--resume`, the same path an evicted session takes, so reads can become free. Prepared on the local branch `refactor/lazy-session-resume` (stacked on this one).
2. **Run-to-completion mode for automated sessions** — cron/webhook runs could invoke the CLI in a mode that exits after its `result` instead of parking, eliminating the incident class by construction and shrinking the reaper's job to interactive parks only. Needs verification that the pinned SDK supports a single-turn mode with the same streaming surface.


yiw190:
Follow-up: background-task holdoff now matches `#438` / `adf83d5`.

- Merged `main` so this branch sits on SDK 0.3.206
- Replaced blanket `task_started` tracking with `background_tasks_changed` snapshot ∪ incremental liveness (`agent-container/src/background-task-liveness.ts`)
- Incremental fallback only registers real bg paths (`backgroundTaskId`, `async_launched`, `local_workflow`) — foreground `local_agent` `task_started` no longer blocks eviction
- Snapshot self-heals sticky incremental ids; malformed frames are ignored

Verified: `background-task-liveness` + `session-manager-idle-eviction` — 22 tests green.

iddogino:
Thanks @yiw190 — this PR correctly identified the idle-session OOM accumulation and the right shape of the fix (reap the subprocess, keep session state, resume transparently), and that approach ships in #440.

During review we found the idleness detection needed to sit on the SDK's newer settlement signals (`session_state_changed` + `background_tasks_changed`) to avoid re-deriving busy/idle from raw stream traffic, plus a few teardown races and a transcript-flush race on eviction, so the implementation was rebuilt on those rather than iterated here. Closing in favor of #440 — the reaper concept, per-class thresholds, and evict-don't-delete semantics from this PR all carried over.

PR #424, opened 2026-07-08
test: positive-signal coverage for silent background paths

iddogino:
Stacked on #420 (base = `chore/node-22`; retarget to main after it merges — the webhook-nudge lines build on the `.catch` hardening there).

## Why
The Node-22 CI investigation showed three background paths whose failure produces no signal anywhere:
- `ChatIntegrationManager.start()` — startup.ts logs only failures; success was unobservable (unlike every other bg service, it had no "started" marker).
- The fire-and-forget `import('trigger-manager')` in `createWebhookTrigger` — the existing unit suite mocks the db and never observes the nudge; the module cycle it crosses is exactly where CI saw an incomplete-namespace TypeError die as an unattributed unhandled rejection.
- The fire-and-forget `import('platform-service')` in platform-auth-service — same shape, zero coverage.

## What
Each path gets a one-line **success log** (positive signal, greppable by a future boot-smoke CI job) plus a test:
| Path | Test | Approach |
|---|---|---|
| ChatIntegrationManager.start | `chat-integration-manager-start.test.ts` | real manager + real db, temp data dir |
| webhook cold-start nudge | `webhook-trigger-service.coldstart.test.ts` | **unmocked** real module cycle; waits for completion log, asserts the failure-warn did not fire |
| platform-service notify | `platform-auth-service.notify.test.ts` | via public `revokePlatformToken()` (no stored token → no network) |

## Why not full UI-level E2E
Both `createWebhookTrigger` entry points (`setup_trigger`, `create_webhook_endpoint` agent tools) are hard-gated on platform auth (`getPlatformAccessToken()` / `isPlatformComposioActive()`), which the E2E mock environment doesn't fake — reaching them from Playwright requires a platform-proxy mock harness (separate work item). These tests close the actual gap at the tightest seam that exercises the real modules.

## Verification
- All 3 new tests pass; full suite 6403 passed (338 files)
- **Mutation-tested**: breaking the dynamic-import path makes the coldstart test fail (timeout + failure-warn asserted)
- Booted the production tsup bundle: `[ChatIntegrationManager] Started (0 startup integration(s))` appears
- typecheck + lint clean


iddogino:
Extended per review discussion: **all nine** startup-launched services now have positive start markers — added the missing PlatformService log (was fully silent) and a real completion log for TriggerManager (its "Starting..." fired *before* the initial poll). New `startup.boot-signals.test.ts` boots the real `initializeServices()` and asserts every marker + zero `Failed to start` + the TriggerManager token-gate holding. Mutation-verified. Full suite 6404 green.

iddogino:
Reworked the boot test per Iddo's review: it no longer runs inside vitest's module runner — it **builds the production tsup bundle and spawns `node dist/web/server.mjs` as a real child process**, then asserts every service's started marker in actual process stdout (+ HTTP readiness, zero `Failed to start`, TriggerManager token-gate). This also makes it sensitive to missing chunks / load-time module-cycle breaks — the artifact-level failure class the vitest runner can't see. Dropped the now-subsumed `chat-integration-manager-start.test.ts`. Mutation-verified against the real process (silenced marker → `timed out waiting for marker: [PlatformService] Started`). Runs in ~3.5s incl. the build. Full suite 6406 green.

PR #425, opened 2026-07-09
feat(SUP-343): platform notifications — realtime OS notifications, merged inbox, detail route

iddogino:
## Summary

App half of **SUP-341** (platform notifications), consuming the platform PR SkillfulAgents/platform#158 (SUP-342/344). Notifications are authored in Customer.io, land in the platform's `notifications` table one row per recipient, and reach the app two ways: **proxy-live reads** (inbox, badge — no local mirror, one code path for both auth modes) and a **direct Supabase Realtime subscription** (desktop OS notifications, instant inbox updates) using a proxy-minted user-scoped JWT where the RLS SELECT policy is the security boundary.

## What's in here

**Manager (`platform-notifications-manager.ts`, desktop/non-auth only)**
- Subscribes to Realtime INSERTs via the existing `SupabaseRealtimeClient`, which gains an optional `table` param (defaults to `webhook_events`; the trigger flow is byte-identical)
- Every INSERT broadcasts an SSE inbox invalidation; OS notifications additionally pass a dedup gate: persisted `lastNotifiedAt` watermark (survives restarts, suppresses reconnect replays/backdated inserts) + in-session id set. OS notifs only ever fire from live INSERTs, never from backlog
- 50-minute tick refreshes the 1h Realtime JWT or reconnects a dead subscription; a **generation token** bumped on stop/start keeps a stale in-flight connect from resubscribing with an old identity or leaking intervals
- `SupabaseRealtimeClient.connect()` now **rejects when the socket closes before ever opening** (previously the promise never settled — a connection-refused at launch would silently strand the manager without its refresh interval and kill realtime at the 1h token expiry)

**UI**
- Platform rows merged into the existing Notifications inbox (client-side merged pagination over both sources), markdown-stripped previews, Megaphone/Platform tag
- New `/notifications/$id` detail route: title as page title, `MarkdownBlock` body, `action_url` through the shared URL-scheme allowlist; opening write-through-marks the row read
- OS-notification click (Electron + web paths) navigates to the detail route; sidebar badge and dock badge combine local + platform unread counts
- Settings toggle ("Platform notifications", default on) gates the popup in renderer, main-process fallback, and manager

**API routes (`/api/platform-notifications`)**
- Thin relays for list/unread-count/read/read-all; polled GETs degrade upstream failures to the disconnected shape (200) so an **offline desktop doesn't error-loop the 30s/60s polling into Sentry captures**; pagination params sanitized/clamped

## Test plan

- ✅ 6,457 unit tests (340 files) — ~60 new across manager (fake-timer JWT refresh, dead-subscription reconnect, failed-initial-connect recovery, stop/start race, watermark equal-timestamp restart replay), realtime client (table threading in all three join spots, close-before-open rejection), API routes (degradation, validation, read-all fan-out), notification handler (platform SSE invalidation, click → detail route, settings toggle), merged inbox (page-2 walk with mixed sources, no dropped/duplicated rows)
- ✅ Playwright mock E2E: 230/231 passed; the 1 failure (`connection-account-toggle.spec.ts`, untouched by this diff) re-passed 10/10 on `--repeat-each=2` — flake
- ✅ `tsc --noEmit` + eslint clean
- ✅ Full loop proven live against local Supabase + wrangler dev: curl → cio receiver → Realtime INSERT → OS notification broadcast + open inbox updating in ~100ms without reload; mark-read write-through and cross-user isolation verified against real RLS

## Notes

- Requires the platform PR deployed for anything to light up; until then the inbox degrades to local agent notifications only (`connected: false`)
- Known v1 gaps (deliberate): no offline inbox (proxy-live reads), closed-window OS-notification clicks don't deep-link to the detail route, auth-mode web clients rely on the 60s refetch instead of realtime push

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -11,6 +11,7 @@ import scheduledTasks from './routes/scheduled-tasks'
 import webhookTriggers from './routes/webhook-triggers'
 import chatIntegrationsRouter from './routes/chat-integrations'
 import notifications from './routes/notifications'
+import platformNotifications from './routes/platform-notifications'
 import proxy from './routes/proxy'
 import mcpProxy from './routes/mcp-proxy'
 import browser from './routes/browser'
@@ -177,6 +178,7 @@ app.route('/api/scheduled-tasks', scheduledTasks)
 app.route('/api/webhook-triggers', webhookTriggers)
 app.route('/api/chat-integrations', chatIntegrationsRouter)
 app.route('/api/notifications', notifications)
+app.route('/api/platform-notifications', platformNotifications)
 app.route('/api/proxy', proxy)
 app.route('/api/agent-bootstrap', agentBootstrap)
 app.route('/api/mcp-proxy', mcpProxy)
--- a/src/api/routes/platform-notifications.test.ts
+++ b/src/api/routes/platform-notifications.test.ts
@@ -1,0 +1,223 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import { Hono } from 'hono'
+
+// ---------------------------------------------------------------------------
+// Mock dependencies
+// ---------------------------------------------------------------------------
+
+let mockProxyUrl: string | null = 'https://proxy.test.example'
+vi.mock('@shared/lib/platform-auth/config', () => ({
+  getPlatformProxyBaseUrl: () => mockProxyUrl,
+}))
+
+let mockToken: string | null = 'pk_test'
+vi.mock('@shared/lib/services/platform-auth-service', () => ({
+  getPlatformAccessToken: () => mockToken,
+}))
+
+vi.mock('@shared/lib/config/settings', () => ({
+  getSettings: () => ({ platformAuth: { memberId: 'sub_member_1' } }),
+}))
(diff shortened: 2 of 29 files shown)


iddogino:
**Review findings addressed in cc7af1a3:**

1. **Pagination past the platform fetch cap** — fixed. `totalPages` now clamps the platform contribution to `PLATFORM_FETCH_CAP`, so every advertised page is backed by fetched rows (on the last page the `(page+1)*PAGE_SIZE` fetch window always covers the clamped total, so no page can render empty). Regression test walks all 7 pages of a 120-row platform-only inbox and asserts the last page has content with next disabled. The trade-off is the label reads "100 total" when 120 exist — advertising only reachable rows seemed strictly better than advertising unreachable pages.

2. **"Mark all" caps at 100 platform rows** — left as-is per the review annotation (accepted). It's documented at the call site; the button converges on repeated clicks, and >100 unread platform announcements shouldn't occur at realistic volumes. Happy to add a bounded pagination loop if it ever bites.

Suite after the fix: 6,458 unit tests green, typecheck + lint clean.

PR #442, opened 2026-07-10
feat: agent graph interactions — elbow connectors, drawing, and editing

westaskew:
Builds on #436 (targets its branch, not main) with the interaction layer for the connections graph.

## Nodes
- Circular agent badges: icon-only status at the top of the circle, regular-weight `text-xs` name wrapping to two lines
- Resource chips are bare icons; hover reveals a frosted pill with the name, humanized cron schedule, a tone-colored status badge (dot + matching text), and aggregated usage ("23 API calls", "never fired")
- Hover lifts nodes with a soft shadow (no more border ring); hairline 0.5px borders; faint gray canvas via inline style (xyflow's stylesheet silently overrides `bg-*` utilities — every earlier attempt was a no-op)

## Connectors
- Orthogonal elbow routing (FigJam-style) between perimeter anchors on the nodes' **visible** shapes — resource edges dock on the 40px chip instead of stopping short at the label column
- Solid blue lines carry a traveling comet tracer: speed is log-scaled to usage, direction follows the data flow (in/out), bidirectional links run opposing tracers on one line; idle stays gray dashed, broken endpoints red dashed
- Click to select: pill grab handles on every segment (dragging an end segment spawns a stub bend), endpoint dots slide around the node perimeter and magnetize to the four cardinal ports, and a floating toolbar offers delete / edit-permissions
- Delete (toolbar or ⌫) unlinks accounts/MCPs or revokes invoke permissions both ways; trigger edges aren't deletable. Edit jumps to the caller agent's policy tab via a nav-transient one-shot
- User-dragged geometry (segment coords + anchor angles) persists per edge in user settings (`graphEdgeGeometry`), cleared by Reset layout

## Drawing new connections
- Hovering a node reveals its four ports with staged FigJam-style affordances (dot → circled arrow → filled while grabbing) and a "click and drag to connect" hint
- Dragging out and dropping creates the real relationship: agent→agent invoke policy, agent↔account link, agent↔MCP link; invalid targets fade during the drag

## Panel
- Right-hand directory of accounts/MCP servers with linked-agent counts; hovering a row spotlights that resource's edges

## Also
- `fix`: Slack Bolt's constructor-time token verification could crash the whole app on an invalid token (unhandled rejection → main-process shutdown); now deferred and awaited inside the connector's error handling

## Testing
- typecheck + lint clean; `home-graph` e2e specs pass (3/3)


westaskew:
Rebased onto `main` (0.4.13) and retargeted the base from `feat/agent-graph-spike` to `main` — #436 hasn't merged yet, so this PR now **includes** its two commits plus the interaction work on top. The only rebase conflict was the migration numbering: main had minted 0027–0030, so the audit-log index migration was regenerated as `0031_misty_sumo` (same two indexes, snapshot now correct against main's chain), and the audit-log index sets from both branches were unioned in `schema.ts`.

PR #445, opened 2026-07-11
fix: sweep stale agent-browser screenshots (pinned dir + 24h janitor)

iddogino:
Item 3 of the container-lifecycle audit (follow-up to #443, #444).

## Problem
Every `browser_screenshot` and `browser_get_state` call makes agent-browser write a uniquely named ~500KB PNG that nothing ever deletes — proven by months-old files accumulating in `~/.agent-browser/tmp/screenshots/`. `browser_get_state` shoots on every call, so heavy browsing sessions leak disk fast.

## Fix
- **Pin the output directory** via `AGENT_BROWSER_SCREENSHOT_DIR` at server startup (new `screenshot-janitor.ts`), making the janitor's target deterministic instead of guessing the CLI's default-path convention across versions. Verified against the agent-browser 0.27.2 binary that the env var is honored. The pinned default mirrors agent-browser's own convention (`$HOME/.agent-browser/tmp/screenshots`) so files accumulated *before* this change get swept too; a pre-set env var wins.
- **Janitor sweep**: delete files older than 24h, at boot + hourly (unref'd interval), wired next to the InputManager sweep in `server.ts`.

## Why not unlink-after-read
The tool result advertises the file path and agents re-read it (same turn or shortly after) — eager deletion would break that contract. 24h is far beyond any legitimate reuse.

## Details
- Age measured by **mtime**, not birthtime (epoch-0 on some mounts, e.g. S3-backed NFS); screenshots are written once so mtime = creation time.
- Subdirectories are never touched; per-file stat/unlink errors (concurrent-delete races) are skipped.
- Missing directory (no screenshot taken yet) is a no-op.
- `execBrowser` spreads `process.env`, so the pinned var reaches every spawned CLI/daemon; the pin runs in server startup code, before any request can trigger a browser command.

## Tests
- 8 new unit tests: old-deleted / fresh-kept, just-under-boundary kept, subdirs untouched, missing dir no-op, custom max-age, env pinning (default, pre-set wins, HOME fallback).
- Full agent-container suite: 619 passed, 8 skipped.
- tsc + eslint clean (remaining warnings pre-exist on untouched server.ts lines).

### DIFF
--- a/agent-container/src/screenshot-janitor.test.ts
+++ b/agent-container/src/screenshot-janitor.test.ts
@@ -1,0 +1,98 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import fs from 'fs/promises'
+import os from 'os'
+import path from 'path'
+import {
+  pinScreenshotDir,
+  sweepStaleScreenshots,
+  SCREENSHOT_MAX_AGE_MS,
+} from './screenshot-janitor'
+
+describe('pinScreenshotDir', () => {
+  it('sets the default under HOME when unset', () => {
+    const env: NodeJS.ProcessEnv = { HOME: '/home/claude' }
+    const dir = pinScreenshotDir(env)
+    expect(dir).toBe('/home/claude/.agent-browser/tmp/screenshots')
+    expect(env.AGENT_BROWSER_SCREENSHOT_DIR).toBe(dir)
+  })
+
+  it('respects a pre-set AGENT_BROWSER_SCREENSHOT_DIR', () => {
+    const env: NodeJS.ProcessEnv = {
+      HOME: '/home/claude',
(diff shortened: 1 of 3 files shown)


iddogino:
Live verification on a real agent container (Docker Desktop dev runtime, same image as Lima prod): ran an isolated agent-browser probe (throwaway session + profile) inside a running `superagent-*` container — the Linux binary saved to `/home/claude/.agent-browser/tmp/screenshots/screenshot-<ts>.png`, byte-for-byte the janitor's pinned default. Also confirmed only `/workspace` is bind-mounted, so screenshots accumulate in the container's writable layer, and containers run with `AGENT_BROWSER_USE_HOST=1` — Chrome on the host, but the CLI (and its screenshot files) stay in-container, which is where the janitor runs.

PR #459, opened 2026-07-13
feat: detect firewalled host-browser CDP proxy at launch

iddogino:
## Problem

When Browser Host = Google Chrome on Windows, the browser tool can fail with an opaque error while **nothing reaches Sentry**. Root cause chain (from a support investigation of a Windows user whose browser tool couldn't even reach google.com):

- Chrome launches fine on the host and the CDP proxy binds the WSL2 gateway IP — every failure on *that* side is already instrumented.
- The container then connects to `host.docker.internal:<proxyPort>`. Windows Defender Firewall (or third-party AV) commonly drops WSL2→host traffic on the vEthernet (WSL) interface, so this hop times out.
- That failure surfaced only as a raw `fetch failed` inside the agent container: no Sentry event, no actionable message for the agent or user.

## Fix

**1. Probe the proxy from the runner's network side at launch** (`ChromeProvider.launch`)
- New `ContainerClient.probeHostPortFromRunner(host, port)` → `'reachable' | 'unreachable' | 'unknown'`. Base default is `'unknown'`; WSL2 implements it by running `curl --max-time 4 http://<bridgeIp>:<proxyPort>/json/version` inside the distro — containers NAT through the distro's eth0, so this crosses exactly the firewall boundary the container will hit.
- Only a **proven** block fails the launch: curl exit 7 (refused) or 28 (timed out — the firewall-drop signature). It then reports to Sentry (`component:browser`, `operation:cdp-proxy-reachability`), tears down Chrome + proxy, and throws a message naming Windows Firewall on the vEthernet (WSL) interface with the remediation.
- Anything inconclusive (curl missing, wsl broken, TCP reset) is `'unknown'` and proceeds — a broken probe can never fail a working setup.

**2. Clear agent-facing error as backstop** (`agent-container/src/server.ts`)
- A network-level failure fetching `/json/version` now throws a firewall diagnosis with next steps (allow the app through the firewall / switch Browser Host to the built-in browser) instead of `fetch failed`. Covers non-WSL2 runners and anything the probe can't prove.

The launch error flows through the existing `/api/browser/launch-host-browser` 503 → container → agent path, so the agent sees the actionable message either way.

## Testing

- 9 new tests: curl-exit classifier + WSL2 probe behavior (`wsl2-container-client.test.ts`), and the ChromeProvider gate — unreachable fails with teardown (no orphan Chrome), unknown/reachable proceed, loopback-direct runners never probe (`chrome-provider.sup217.test.ts`).
- All 64 tests in the touched test files pass; 118 related container-client tests pass; typecheck + lint clean on touched files.

## Note

The probe assumes `curl` exists in the bundled WSL2 Alpine distro — if absent it degrades to `'unknown'` (fail-safe, behavior unchanged) but the probe is inert. Worth a one-time check that the rootfs ships curl.

### DIFF
--- a/agent-container/src/server.ts
+++ b/agent-container/src/server.ts
@@ -764,7 +764,22 @@ async function launchHostBrowserIfNeeded(): Promise<HostBrowserInfo | undefined>
   // (ws://host:port/devtools/browser/<id>), not just ws://host:port.
   // Query Chrome's /json/version endpoint to discover it.
   const cdpHost = `${cdpIp}:${data.port}`;
-  const versionRes = await fetch(`http://${cdpHost}/json/version`);
+  // A network-level failure here (vs. an HTTP error) means the host browser
+  // launched but this container can't reach its debugging port — on Windows
+  // that is almost always the host firewall dropping container→host traffic.
+  // Surface that diagnosis instead of an opaque "fetch failed".
+  let versionRes: Response;
+  try {
+    versionRes = await fetch(`http://${cdpHost}/json/version`);
+  } catch (err) {
+    const cause = err instanceof Error ? err.message : String(err);
+    throw new Error(
+      `The host browser launched, but its debugging endpoint at ${cdpHost} is unreachable from inside the agent container (${cause}). ` +
+      `This usually means a firewall on the host machine is blocking container-to-host connections ` +
+      `(on Windows: Windows Defender Firewall or antivirus blocking the app on the "vEthernet (WSL)" network). ` +
+      `Ask the user to allow the app through their firewall, or to switch Browser Host to the built-in browser in Settings.`
+    );
+  }
   if (!versionRes.ok) {
     throw new Error(`Failed to query CDP /json/version: ${versionRes.status}`);
   }
--- a/agent-container/src/server.ts
+++ b/agent-container/src/server.ts
@@ -723,11 +723,28 @@ async function launchHostBrowserIfNeeded(): Promise<HostBrowserInfo | undefined>
   const browserAuthHeaders: Record<string, string> = { 'Content-Type': 'application/json' };
   if (proxyToken) browserAuthHeaders['Authorization'] = `Bearer ${proxyToken}`;
 
-  const response = await fetch(`${hostAppUrl}/api/browser/launch-host-browser`, {
-    method: 'POST',
-    headers: browserAuthHeaders,
-    body: JSON.stringify({ agentId: agentId || 'default' }),
-  });
+  // A network-level failure here means the launch request never reached the
+  // host app at all — the host never launches Chrome and never reports to
+  // Sentry, so this bare 'fetch failed' used to be the only trace. On Windows
+  // it is almost always Windows Firewall blocking the app's API port for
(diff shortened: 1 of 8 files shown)


iddogino:
## Forcing this failure mode on a Windows machine (for testing)

Block only the proxy's port range, not the whole WSL interface: the container also reaches the host app's API over the same path, and the CDP proxy binds a random **ephemeral** port (49152–65535, from `findFreePort`) while the app's API port is fixed and lower. That's also exactly the wild failure shape — the app's main port got an "Allow access" rule at first run, but a fresh random-port listener didn't.

In an **admin PowerShell**:

```powershell
# Force the block (ephemeral ports only, WSL interface only)
New-NetFirewallRule -DisplayName "TEST WSL CDP block" -Direction Inbound -Protocol TCP `
  -LocalPort 49152-65535 -InterfaceAlias "vEthernet (WSL)" -Action Block
```

Then, with Browser Host = Google Chrome, ask an agent to open any page. Expected with this PR: launch fails fast with the "blocked by Windows Firewall on the vEthernet (WSL) interface…" message, Chrome gets torn down (no orphan window lingering), and a Sentry event lands with `component:browser` / `operation:cdp-proxy-reachability`. On current main you instead get an opaque in-container retry loop — so this doubles as a before/after demo.

Cleanup:

```powershell
Remove-NetFirewallRule -DisplayName "TEST WSL CDP block"
```

**Notes:**

- **Sanity-check the block independently** before blaming the app: launch the browser once, note the proxy port from the app log line `[ChromeProvider] Chrome CDP on 127.0.0.1:…, proxy on <ip>:<port>`, then `wsl -d superagent -- curl -v --max-time 4 http://<ip>:<port>/json/version` — it should hang and exit 28 while the rule is active.
- **Windows 11 24H2 caveat:** newer WSL routes through the Hyper-V firewall and interface-alias rules may not bite. The blunt alternative is `Set-NetFirewallHyperVVMSetting -Name '{40E0AC32-46A5-438A-A0B2-2B479E8F2E90}' -DefaultInboundAction Block` (that GUID is the WSL VM creator), but that blocks the host API too, so the failure surfaces earlier — as a failed launch request rather than the probe message. Prefer the port-range rule where it works.
- **To exercise the container-side backstop message** (the clearer `/json/version` error in `agent-container/src/server.ts`): make the host probe inconclusive by renaming curl inside the distro — `wsl -d superagent -- sh -c 'mv /usr/bin/curl /usr/bin/curl.bak'` — with the firewall rule still active. Launch then succeeds host-side (probe returns `unknown`) and the container's own connect failure produces the new firewall-diagnosis error. Restore curl afterwards.

iddogino:
**Heads-up + handover:** validation on a real Windows machine confirmed the bundled Alpine distro ships **no curl**, so this PR's runner-side probe currently always returns `unknown` (fail-safe holds — it can never break a launch — but it detects nothing). It needs a busybox-wget fallback with stderr-based classification before merge. Full spec, plus the machine-state context and validation matrix, is in `FIREWALL-HANDOVER.md` on the `feat/firewall-detect-remediate` branch (PR #465).

PR #465, opened 2026-07-13
feat: detect Windows Firewall blocks with one-click UAC fix

iddogino:
## Problem

Follow-up to #459, from a live-debugged support case: when the first-run Windows Firewall "Allow access?" prompt is cancelled, Windows writes **persistent Inbound Block rules** for the app exe. Nothing visible breaks — chat traffic is loopback/outbound — but every connection from the agent container back into the host (browser launch, Composio/MCP tool proxies) is silently dropped, with **zero Sentry signal on either side**: the host never sees the request, and the container has no telemetry. On the affected machine the only symptom was the agent reporting `fetch failed`.

## What this adds

**Detection** (`src/main/windows-firewall/`) — Windows-only, cheap elsewhere:
- Non-admin PowerShell read of enabled `Inbound + Block` rules whose application filter targets `process.execPath` (reading rules needs no elevation), plus the Win11 24H2+ Hyper-V/WSL firewall layer (`Get-NetFirewallHyperVVMSetting`, scoped to the WSL VM-creator GUID).
- Output is Zod-validated at the process boundary (handles PowerShell 5.1's single-element JSON collapse). 5-minute cache; `?refresh=1` bypass.
- Detection **reports to Sentry** (`component:firewall`) — this closes the observability gap and will tell us the real incidence rate of the cancelled-prompt population.
- A broken probe can never show a false banner: any probe failure → not-blocked + the probe error itself captured.

**Banner + one-click fix**:
- New `FirewallBlockedSidebarBanner` in the existing runtime-banner family, mounted in the sidebar banner stack, driven by React Query against `GET /api/firewall/status`.
- **Fix** button → `POST /api/firewall/fix` → one elevated PowerShell run behind a standard UAC prompt: removes Block rules scoped strictly to our exe, adds a program-scoped Allow rule (`Profile Any`), applies the WSL-scoped Hyper-V fix only when that layer was the blocker, then **re-detects and only reports success if the block is actually gone**. Declined UAC (`exit 223` sentinel) gets distinct copy ("an administrator must click Yes") vs. hard failure (manual instructions).

**Installer pre-seed** (`build/installer.nsh`):
- `customInstall`/`customUnInstall` add/remove the same Allow rule via `netsh`, **gated on the installer actually running elevated** — our NSIS installer is per-user (`perMachine: false`) and normally runs without elevation, and electron-updater's silent updates must never trigger UAC. So this covers admin installs for free; the banner is the primary path for everyone else. Rule name kept in sync with `ruleDisplayName()` so the two paths never duplicate rules.

**Dev/test mode**: `SUPERAGENT_FAKE_FIREWALL_BLOCK=1` fakes a block on any platform, gated the same way as `SUPERAGENT_TEST_UPDATES` (`!app.isPackaged` in Electron; non-production in web) so a stray env var can't fake a block in shipped builds.

## Testing

- **13 unit tests** (`windows-firewall.test.ts`): rule detection incl. the single-object JSON quirk, Hyper-V-only block, probe-failure safety, caching/refresh, fix short-circuit, UAC-declined vs failed classification, persisted-block honesty, Hyper-V line inclusion, fake mode.
- **Gated E2E** (`e2e/specs/firewall-banner.spec.ts`, skips unless the fake flag is set): drives the full production wiring — status route → hook → banner → fix mutation → banner clears without reload. Verified passing locally:
  `SUPERAGENT_FAKE_FIREWALL_BLOCK=1 E2E_MOCK=true npx playwright test firewall-banner` → 1 passed.
- Navigation E2E suite (25 specs, heaviest sidebar coverage) passes with the changes; the gated spec skips cleanly in normal runs.
- Typecheck + lint clean on all touched files.

## Not covered / known limits

- Real UAC elevation and `netsh`/`New-NetFirewallRule` behavior need a manual pass on a real Windows machine before release (I can supply the force-a-block repro commands from the #459 thread).
- Third-party AV firewalls that don't register rules in WFP won't be detected by the rule read — but they *will* still be caught by #459's connection-time diagnosis messages.
- The NSIS macro assumes the packaged exe name `Gamut.exe`; revisit if the artifact name changes (rebrand freeze applies).


iddogino:
## Windows validation steps (end-to-end)

**Precondition:** a Windows machine with the bundled WSL2 runner set up, running a build of THIS branch. ⚠ Detection targets `process.execPath`, so what you run decides what you block in step 1: a packaged install = `Gamut.exe`; `npm run dev:electron` = `node_modules\electron\dist\electron.exe`. Steps below assume a packaged install.

### 1. Force the exact cancelled-prompt state (admin PowerShell)

```powershell
New-NetFirewallRule -DisplayName "gamut.exe" -Direction Inbound -Program "$env:LOCALAPPDATA\Programs\Gamut\Gamut.exe" -Action Block -Profile Private,Public
```

(The real cancelled prompt writes a TCP+UDP pair; one rule is enough — detection matches any enabled Inbound Block on the exe. An existing Allow rule can stay: Block beats Allow, so the state is still broken.)

**Confirm the block is real** before trusting the app (non-admin, app running):

```powershell
wsl -d superagent -- ip route show default          # note the IP after 'via'
wsl -d superagent -- curl -sv --max-time 5 http://<GATEWAY_IP>:47891/api/auth-config
```

Expected: hangs ~5s then curl exits 28. If this responds, the rule didn't bite — stop and investigate before testing the app.

### 2. Launch the app → banner appears

Start (or restart — detection is cached 5 min, a fresh launch queries immediately) and look at the left sidebar under the wordmark:

> 🛡 **Windows Firewall is blocking agent connections.**
> Browser and connected tools can't reach this app.
> **Fix now →**

Also check Sentry: an event tagged `component:firewall` / `operation:detect` should arrive with the block rule names.

### 3. Negative path first: decline the UAC prompt

Click **Fix now** → button flips to *Waiting for Windows approval…* → click **No** on the UAC dialog.

Expected: banner stays, adds *"Approval was declined — an administrator must click Yes."* and the button reads *Try again →*. No Sentry `captureException` for this (declined ≠ failure).

### 4. Happy path: click Try again → Yes on UAC

Expected: banner disappears **without an app restart** (the fix re-detects before claiming success and pushes the fresh status into the query cache).

### 5. Validate the actual firewall state (non-admin)

```powershell
Get-NetFirewallApplicationFilter -Program "*Gamut*" | Get-NetFirewallRule | Format-Table DisplayName, Direction, Action, Enabled, Profile
```

Expected: **zero Block rows**; exactly one `Gamut agent connections | Inbound | Allow | True | Any`.

Ground truth again:

```powershell
wsl -d superagent -- curl -s --max-time 5 http://<GATEWAY_IP>:47891/api/auth-config
```

Expected: JSON response instead of a hang. Then the real feature test: set Browser Host = Google Chrome and ask an agent to open google.com — a Chrome window should appear and the page content should come back.

### 6. Cleanup

```powershell
Remove-NetFirewallRule -DisplayName "Gamut agent connections"
```

### Any-platform smoke (no Windows needed)

```bash
SUPERAGENT_FAKE_FIREWALL_BLOCK=1 npm run dev:electron   # banner shows on launch; Fix clears it (no UAC)
SUPERAGENT_FAKE_FIREWALL_BLOCK=1 E2E_MOCK=true npx playwright test firewall-banner   # same loop, automated
```

iddogino:
**False-positive found and fixed** (reported by @iddogino running this branch on a healthy machine): the banner keyed on `Get-NetFirewallHyperVVMSetting DefaultInboundAction: Block` as a secondary signal. That was wrong on two axes — it's a **stock default** on Win11 24H2+, and it governs traffic **into the WSL VM**, not WSL→host (our container→host path, which only the regular host-firewall rules affect; the live-debugged case was fixed purely by removing exe Block rules, Hyper-V untouched).

As of the latest commit:
- `blocked` keys **strictly on explicit enabled Inbound Block rules targeting our exe** — the deterministic cancelled-prompt signature.
- The Hyper-V read is retained as telemetry only (Sentry extra + status field), never gates the banner.
- The elevated fix script no longer contains `Set-NetFirewallHyperVVMSetting` at all — the previous version would have weakened a stock protection to clear a phantom banner.
- Tests updated to pin all three (14 passing).

Validation-steps addendum: on a healthy Win11 24H2 machine, `Get-NetFirewallHyperVVMSetting -PolicyStore ActiveStore` showing `Block` for `{40E0AC32-46A5-438A-A0B2-2B479E8F2E90}` with **no banner** is now the expected PASS state.

iddogino:
**Handover for on-machine validation:** `FIREWALL-HANDOVER.md` at the repo root of this branch (temporary, delete before merge). It covers both this PR and #459: current state, the two open mysteries from today's Windows validation (banner absent despite Block rules; `Connection refused` ≠ firewall — likely port/instance ground-truth issue), ranked hypotheses with exact commands, the confirmed missing-curl defect in #459's probe, and the full validation matrix.

PR #468, opened 2026-07-14
Session long sleep: schedule_resume wakes the same session (SUP-30)

iddogino:
## Summary

Lets a session do extended waits (hours → multi-day) and resume **the same session with full context**, surviving container restarts, app restarts, and GC. Closes SUP-30 — the ticket description is the canonical spec.

A "wake" is a one-shot `scheduled_tasks` row with a new nullable `resumeSessionId` (migration `0028`). When it fires, the scheduler branches: instead of `createSession`, it sends a `[SYSTEM]` resume message (echoing the agent's own wake note) into the existing session with `shouldQuery: true`. Cold-container revival is inherited from the existing lazy-resume path, so restart survival comes for free.

### What's included

- **Service**: `createSessionWake` with replace semantics (one pending wake per session — a new one cancels and returns the old), `getPendingWakeForSession`, `listPendingWakesByAgent`, `listSessionIdsWithPendingWakes`, `cancelPendingWakeForSession`
- **Scheduler**: `executeSessionWake` with session-exists guard, duplicate-fire guard (`lastWake` slot in session metadata, mirroring `getSessionForScheduledExecution`), agent-exists guard, and `cancelAwaitingInput` before send. Transient failures leave the wake `pending` for poll retry; it only fails after 6h overdue
- **Agent tool**: `schedule_resume` (container blocking tool + host interceptor beside `handleScheduleTaskTool`) — deliberately separate from `schedule_task` so agents distinguish "resume THIS conversation" from "spawn an independent session". Wakes are labeled `session wake` in `list_scheduled_tasks`
- **Protections**: auto-delete monitor skips sessions with pending wakes; deleting a session cancels its wake; wakes are filtered out of agent-level automation lists and summary counts
- **Routes**: `/run-now` gained a wake branch (resumes the session — "Wake now" exercises the same delivery as the scheduler via shared `buildWakeMessage`); `DELETE` broadcasts `session_updated` for wakes
- **UI**: `pendingWakeAt/TaskId/Note` on `ApiSession`; MoonStar indicator in sidebar (alongside the unread dot), session tabs, and related-sessions rows; `PendingWakeBanner` above the composer with the wake note + **Wake now** / **Cancel**; `schedule_resume` tool renderer; `session_updated` global SSE invalidation

## Validation

All tests written TDD (red first):
- Host unit suite **6,872 passed** · container suite **637 passed** · typecheck + lint clean
- Full mock E2E suite **245 passed, 0 failed**, including new `session-long-sleep.spec.ts` (banner, sidebar badge, Wake now, Cancel, home-triggers exclusion) + a `schedule resume` MockContainerClient scenario

**Live end-to-end validation** (real dev server, real Docker container, real LLM, isolated data dir):
1. **Warm wake** — agent was told a secret word, called `schedule_resume` for +2 min, went idle; scheduler fired 26s after due; the same session replied *"The secret word is PINEAPPLE. WAKE-OK"*
2. **Re-sleep + replace** — on wake, agent scheduled a 45-min wake then replaced it with a 2-min one; DB confirmed the placeholder `cancelled` with exactly one `pending` wake
3. **Cold wake** — dev server killed and the container `docker rm -f`'d with the wake pending; wake time passed while fully offline; on restart the catch-up scan fired it within seconds, cold-started a fresh container, SDK-resumed, and the agent replied *"The secret word is PINEAPPLE. COLD-WAKE-OK"* — exactly one session existed throughout

### DIFF
--- a/agent-container/src/input-manager.ts
+++ b/agent-container/src/input-manager.ts
@@ -34,6 +34,7 @@ type EarlyResult =
 // means the host handler died mid-request; nobody is coming back for it.
 const AUTOMATED_INPUT_TYPES: ReadonlySet<string> = new Set([
   'schedule_task',
+  'schedule_resume',
   'list_scheduled_tasks',
   'cancel_scheduled_task',
   'pause_scheduled_task',
--- a/agent-container/src/mcp-server.ts
+++ b/agent-container/src/mcp-server.ts
@@ -20,6 +20,7 @@ import {
   pauseScheduledTaskTool,
   resumeScheduledTaskTool,
 } from './tools/schedule-task'
+import { scheduleResumeTool } from './tools/schedule-resume'
 import {
   getAvailableTriggersTool,
   listTriggersTool,
@@ -76,7 +77,7 @@ export function createUserInputMcpServer() {
     tools: [
       requestSecretTool, requestConnectedAccountTool, searchConnectedAccountServicesTool,
       requestRemoteMcpTool, searchRemoteMcpServicesTool,
-      scheduleTaskTool, listScheduledTasksTool, cancelScheduledTaskTool,
+      scheduleTaskTool, scheduleResumeTool, listScheduledTasksTool, cancelScheduledTaskTool,
       pauseScheduledTaskTool, resumeScheduledTaskTool,
       deliverFileTool, deliverSessionTool, requestFileTool, requestBrowserInputTool,
       ...(includeScriptRun ? [requestScriptRunTool] : []),
--- a/agent-container/src/system-prompt-vars.test.ts
+++ b/agent-container/src/system-prompt-vars.test.ts
@@ -40,6 +40,18 @@ describe('generateSystemPrompt rendering', () => {
     expect(out.includes('platform-dependent')).toBe(!composio && !webhook)    // disconnected fallback
   })
 
+  // The pause/resume guidance must always render: without it agents reach for
+  // schedule_task ("new session") when they mean "continue THIS conversation
+  // later", which loses the context the wait was for.
+  it('always teaches schedule_resume and how it differs from schedule_task', () => {
+    const out = generateSystemPrompt()
+    expect(out).toContain('## Pausing and Resuming This Session')
+    expect(out).toContain('mcp__user-input__schedule_resume')
+    // The decision rule both directions
+    expect(out).toContain('THIS SAME conversation')
+    expect(out).toContain('Use `schedule_task` only for genuinely independent work')
+  })
+
   // A heading whose body is entirely gated renders as a title with the next
   // heading directly beneath it. Some headings (`## File Handling`) are static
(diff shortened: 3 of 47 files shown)


iddogino:
Addressed all three review findings in ad800ef1 — all confirmed real:

**1. Non-atomic wake replacement** — Confirmed, and slightly worse than reported: `parseAtSyntax` threw *after* the old wake was cancelled, so an invalid/past `wakeTime` destroyed the session's valid wake and left none. Fixed three ways: the expression is validated before any mutation, the read→cancel→insert now runs in a single synchronous better-sqlite3 transaction, and migration `0029` adds a partial unique index (`resume_session_id WHERE status='pending'`) as a storage-layer backstop. New tests: invalid/past wakeTime preserves the existing wake; 5 concurrent creations end with exactly one pending wake.

**2. Wake-now / scheduler double delivery** — Confirmed. Both paths now funnel through a new shared `deliverSessionWake` (`src/shared/lib/scheduler/wake-delivery.ts`) with a per-task in-process claim (scheduler singleton and HTTP routes share one process) and a fresh task-status re-read under the claim. Send-then-record crash-window semantics are preserved deliberately — flipping status before the send would lose the wake on a crash-after-claim; the `lastWake` metadata guard still covers crash-between-send-and-record. A concurrent second caller gets `409` (route) or a logged skip (scheduler). Unit test holds one delivery inside `sendMessage` while a second arrives: exactly one send.

**3. Phantom active session on failed delivery** — Confirmed. Added `messagePersister.markSessionIdle()` (routes to the existing `finalizeIdle`, no died-mid-turn error semantics) and the delivery path reverts the optimistic active flag before rethrowing for retry.

Validation: full host suite 6,886 passed, container 637 passed, tsc + lint clean, wake E2E specs green (fresh DB applies `0029`), plus a live re-run against a real container: index migrated onto the existing validation DB, agent re-slept, "Wake now" resumed the same session through the new claimed path ("MANUAL-WAKE-OK PINEAPPLE" — context intact), and an immediate second "Wake now" was correctly rejected with 400.

iddogino:
Added system-prompt guidance in fcfb8edb (this was open question 4 in the spec) — a new "Pausing and Resuming This Session" section right after "Scheduling Tasks" teaching the decision rule: `schedule_resume` for same-conversation continuity (waiting on a reply/review/external process), `schedule_task` only for genuinely independent or recurring work, plus the end-your-turn / one-pending-wake / re-sleep mechanics.

A/B validated live with a real container and the same organic prompt ("I sent the contract to Dana — check in with me tomorrow at 9am whether she replied", no tool named):
- **Without the section**: the agent used `schedule_task` — a context-less new session, exactly the anti-pattern this feature replaces
- **With it**: the agent used `schedule_resume` ("I'll wake back up in this same conversation tomorrow at 9am with full context intact"), wake correctly scheduled at 9am in the owner's timezone

Covered by a rendering test in system-prompt-vars.test.ts; container suite 638 passed, tsc clean.

PR #471, opened 2026-07-14
Send agent ID + name as attribution headers on container LLM requests (SUP-396)

iddogino:
## Summary

App side of per-agent LLM cost attribution (SUP-395 / SUP-396): every container-originated request to the platform proxy now carries the agent's identity as custom headers, so the proxy can record per-agent usage (SUP-397).

```
x-superagent-agent-id: tk0mligft9
x-superagent-agent-name: Attribution%20T%C3%ABst%20Agent
```

## Design

One deviation from the ticket: the host does **not** set `ANTHROPIC_CUSTOM_HEADERS` directly. Container env travels via a Docker `--env-file`, and `writeEnvFile` strips `\r\n` from values — the multi-header format needs a newline between pairs and would be silently flattened into one garbage header. Instead:

- **Host** (platform provider only): `getContainerEnvVars(agent?)` gains an optional `AgentIdentity` param and injects plain `SUPERAGENT_AGENT_ID` / `SUPERAGENT_AGENT_NAME` env vars. Other providers ignore the param, so direct-Anthropic/generic endpoints never receive agent names. Identity resolves at env-build time (container start) via a new cycle-free `readAgentDisplayNameSync` in `file-storage.ts`; renames apply on the next container restart (platform side groups by ID with latest-observed name).
- **Container**: new `attribution-headers.ts` composes `ANTHROPIC_CUSTOM_HEADERS` from those vars, appending after any user-set value. The display name is **always percent-encoded** — undici rejects non-Latin-1 header values, which would fail every LLM request for a non-ASCII agent name. Decode contract documented on SUP-397.
- Both keys added to `RESERVED_ENV_VAR_KEYS` so custom env config can't spoof attribution.
- All runtimes covered: docker/k8s/lambda via `buildAgentEnv`, plus the lima/wsl2 `buildEnvFile` overrides.

Deferred: host-originated calls (session-naming summarizer etc. via `getConfiguredLlmClient()`) send no headers and will show as "Unattributed" — they're app overhead, not agent traffic.

## Testing

- Unit: header composition (10 tests — encoding, capping, lone surrogates, user-header append), provider injection/sanitization (6), `buildAgentEnv` identity passthrough.
- Full suites green: agent-container 642, host container 738, llm-provider 297; `tsc` + eslint clean in both packages.
- **Live E2E**: built the container image from this branch, ran a scratch app instance against a logging MITM proxy in front of the real platform proxy; verified both headers on `POST /v1/messages` (values above), 200 responses, and agent reply — and that the host summarizer path is header-less as expected.

### DIFF
--- a/agent-container/src/attribution-headers.test.ts
+++ b/agent-container/src/attribution-headers.test.ts
@@ -1,47 +1,44 @@
 import { describe, it, expect } from 'vitest'
-import { withAgentAttributionHeaders } from './attribution-headers'
+import {
+  withAgentAttributionHeaders,
+  captureAgentIdentity,
+  isAgentIdentityEnvKey,
+} from './attribution-headers'
 
 describe('withAgentAttributionHeaders', () => {
-  it('returns env unchanged when no agent id is present', () => {
+  it('adds no headers when the identity has no agent id', () => {
     const env = { PATH: '/usr/bin', ANTHROPIC_AUTH_TOKEN: 'tok' }
-    expect(withAgentAttributionHeaders(env)).toBe(env)
+    expect(withAgentAttributionHeaders(env, {})).toEqual(env)
   })
 
-  it('returns env unchanged when the id sanitizes to empty', () => {
-    const env = { SUPERAGENT_AGENT_ID: '!!!' }
-    expect(withAgentAttributionHeaders(env)).toBe(env)
+  it('adds no headers when the id sanitizes to empty', () => {
+    const out = withAgentAttributionHeaders({}, { id: '!!!' })
+    expect(out).not.toHaveProperty('ANTHROPIC_CUSTOM_HEADERS')
   })
 
-  it('composes id and name headers', () => {
-    const out = withAgentAttributionHeaders({
-      SUPERAGENT_AGENT_ID: 'abc123',
-      SUPERAGENT_AGENT_NAME: 'My Agent',
-    })
+  it('composes id and name headers from the identity', () => {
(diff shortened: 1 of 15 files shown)


iddogino:
Addressed the three review findings in f1e84e34:

- **Identity immutability (P2)**: headers are now composed from a boot-time identity snapshot captured at container-server module load (Docker env only), so neither per-session `customEnvVars` (spread over `process.env` in `createQuery`) nor the runtime `POST /env` endpoint can change which agent the container reports as. `withAgentAttributionHeaders` also forces `SUPERAGENT_AGENT_ID`/`NAME` back to boot values after all merges, and `POST /env` rejects both keys with 403. Live-verified in a real container: `POST /env {SUPERAGENT_AGENT_ID: victim}` → 403 with host auth (401 without), normal keys still 200, headers unchanged.
- **Case-insensitive header collision (P2)**: any pre-existing `X-Superagent-Agent-*` line in a user-configured `ANTHROPIC_CUSTOM_HEADERS` is stripped case-insensitively before the trusted values are appended (the namespace is reserved), so a lowercase forgery can no longer merge into a `victim, real` combined value. Other user headers are preserved.
- **Numeric/boolean names (P3)**: `readAgentDisplayNameSync` now coerces YAML-ambiguous scalars (`123`, `true`) back to strings instead of dropping the name.

Re-validated end-to-end on a rebuilt image against the real platform proxy (headers present, completion streamed). Container suite 678 passed, host suites green, tsc + lint clean in both packages.

PR #473, opened 2026-07-14
fix: self-heal corrupt skillset caches via nuke-and-reclone, keep offline refreshes quiet

iddogino:
## Problem

A half-cloned skillset cache — `.git` present but no resolvable `origin/HEAD`, or an empty tree with no `index.json` — breaks every skills refresh **forever**:

1. `gitPull` can't resolve the default branch, captures to Sentry, and silently no-ops (ELECTRON-4K, 968 events / 9 users)
2. `readIndexJson` then throws on the missing `index.json` (ELECTRON-4M, 662 events; ELECTRON-AF on Windows)

Both fire on every refresh for the same user (4K and 4M share trace IDs), and nothing ever repairs the directory. Probing real git shows the likely origin of the state: cloning an **empty** platform skillset repo leaves exactly this shape — `git remote set-head origin --auto` fails with `error: Cannot determine remote HEAD`, so `origin/HEAD` never resolves.

## Fix

- **`resolveDefaultBranch` distinguishes offline from unresolvable.** The `set-head --auto` stderr is no longer swallowed — it's logged and classified against real git messages (verified live: offline → `Could not resolve host` / `unable to access …`; empty/corrupt repo → `Cannot determine remote HEAD`).
- **Unresolvable → nuke and re-clone once.** `refreshSkillsetCache` removes the repo dir and re-runs `ensureSkillsetCached(ref)`. Same recovery when a successful pull still leaves no `index.json` (git-cached providers only).
- **Offline never nukes.** Intact cache → serve stale quietly. Incomplete cache → defer recovery to the next refresh with connectivity. Remote-unavailable errors (network and auth) are no longer captured to Sentry in `gitPull` or either `refreshAgentSkills` capture site — they're environmental, not actionable.
- **Re-clone still empty → distinct error** ("Skillset repository has no index.json at its root"). That's a broken-remote-content signal (e.g. an empty platform skillset repo), a new fingerprint separate from cache corruption, and it does not loop into further re-clones.

Addresses ELECTRON-4K, ELECTRON-4M, ELECTRON-AF (to be resolved in Sentry once the carrying release shows zero events, per the sweep process — intentionally not using auto-close keywords).

## Validation

- 5 new recovery tests: half-clone heal (asserts one `git clone` + set-head warning logged), offline with intact cache (stale served, no clone, no capture), offline with incomplete cache (no nuke, no capture, retried later), missing-index-after-pull re-clone, and re-clone-still-empty distinct error (no re-clone loop).
- Reworked the fetch-error tests: the old "unexpected fetch error" fixture was actually a network timeout — now split into an offline test asserting **no** Sentry capture and a genuine-error test asserting capture.
- Error classification cross-checked against **real git** output for both the empty-remote and unreachable-host cases.
- 172 tests green in the service suite, 490 across dependent suites (skillsets/agents routes, providers, templates); tsc + eslint clean on top of latest main.

### DIFF
--- a/src/shared/lib/services/skillset-service.test.ts
+++ b/src/shared/lib/services/skillset-service.test.ts
@@ -13,6 +13,20 @@ import type {
 // Hoisted Mocks
 // ============================================================================
 
+const mockCaptureException = vi.hoisted(() => vi.fn())
+
+vi.mock('@shared/lib/error-reporting', () => ({
+  captureException: mockCaptureException,
+  captureMessage: vi.fn(),
+  addErrorBreadcrumb: vi.fn(),
+  setErrorTag: vi.fn(),
+  setErrorContext: vi.fn(),
+  setErrorReportingUser: vi.fn(),
+  initErrorReporting: vi.fn(),
+  getErrorReporter: vi.fn(() => null),
+  flushErrorReporting: vi.fn(async () => true),
+}))
+
 const { mockExecFile, mockAnthropicCreate, mockGetApiKey, mockGetModels } = vi.hoisted(() => {
   const mockExecFile = vi.fn<
     (cmd: string, args: string[], opts?: unknown) =>
@@ -845,7 +859,7 @@ Instructions here`
       expect(updated.originalContentHash).toBe(hashTestSkillPackage({ 'SKILL.md': skillContent }))
     })
 
-    it('gitPull reports unexpected fetch errors to Sentry', async () => {
+    it('gitPull keeps the stale cache quietly when the fetch fails offline', async () => {
       const skillContent = '# Test Skill\nOriginal content'
       const meta = buildMetadata({ originalContentHash: contentHash(skillContent) })
       const config = buildSkillsetConfig()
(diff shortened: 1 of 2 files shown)


iddogino:
Addressed all three review findings in 171884bc — each was validated empirically before fixing:

1. **[P1] Timeout kill nuking healthy caches — confirmed & fixed.** Probed Node directly: an execFile timeout yields `killed: true, signal: 'SIGTERM'` with a bare `"Command failed: git …"` message, so the string classifier called it unresolvable. New `isRemoteUnavailableError()` inspects the error object — `killed`/`signal` metadata, plus a walk of `cause`/`AggregateError.errors` chains for network errnos (ENOTFOUND, ECONNREFUSED, ETIMEDOUT, undici codes, …) — before falling back to message patterns. Used at every classification point (set-head, fetch/reset, both `refreshAgentSkills` capture sites). Regression test: set-head killed by timeout with an intact cache → stale index served, no clone, no capture.

2. **[P2] Missing `origin` remote bypassing recovery — confirmed & fixed.** Verified: `git remote set-url origin …` on a repo without the remote exits 2 with `error: No such remote 'origin'`. The set-url call now lives inside the recovery try, and "No such remote" routes through the same nuke-and-reclone as an unresolvable origin/HEAD. Regression test included.

3. **[P2] Platform offline/auth noise — confirmed & fixed.** `resolveCloneUrl` failures (undici `fetch failed` with ENOTFOUND only on `cause`; `Failed to get platform Git URL: 401/403/…`) are now classified remote-unavailable via the object-aware helper, as are the archive-path equivalents for git-less machines (`Not authorized to download platform skillset`, `Platform not connected`, `Platform server error while downloading`). Two regression tests exercise the real platform git path end-to-end.

Bonus fix surfaced by #3: the suite's `child_process` mock lacked `execFileSync`, so the platform provider's git probe silently failed and `usesGitCache` was `false` in tests — the existing platform tests stubbed `/git-url` (a git-path endpoint) but were actually exercising the archive path. The mock now provides `execFileSync`, so all platform tests hit the git path they were written for.

176 tests in the service suite, 543 across dependent suites, tsc + eslint clean.

PR #485, opened 2026-07-16
feat: in-app Apple Container first-install (SUP-413)

JeremyBischoff:
## Summary
- Add in-app first-install for Apple Container on eligible Macs (darwin + arm64 + macOS ≥ 26): pin signed `.pkg` → SHA256 verify → admin elevate with elevated rehash → `container system start` → refresh availability.
- Gate install behind explicit Start/Install (`allowInstall`); no silent provision, no GitHub install fallthrough for Apple.
- Keep setup UI honest: download % then phase labels, progress scoped to the active runner, persist selected runner on success, clear CHECKING on failure, drop stale cancel errors once the runner is available.

## Test plan
- [x] Unit: eligibility, `allowInstall`, digest mismatch never elevates, elevate rehash script, cancel mapping, mutex, startRunner persistence / failure refresh
- [x] Manual smoke (eligible Mac): Install → password → start → send agent message
- [x] Manual: cancel password / cancel mid-download recover without wedging UI; progress only on the Apple card
- [ ] CI green on this draft
- [ ] Confirm Apple option absent on ineligible machines (Windows / Intel / macOS < 26) if a reviewer has one handy

PR #487, opened 2026-07-16
feat: in-app Apple Container first-install (SUP-413)

JeremyBischoff:
## What
In-app first-install for Apple Container on eligible Macs, plus dogfood-week parity so agents can actually talk to the host proxy and recover from a stopped runtime.

**Footprint:** 17 files · ~1280 net (src ~690 / tests ~590). Kept as one PR (Iddo approved).

## Why
Apple was installable but not dogfood-ready: `host.docker.internal` is NXDOMAIN inside Apple containers, `--tail` logs exit 64, and a stopped apiserver hard-failed the next message. Backport the Lima/Docker/WSL2 hardening so the team can run Apple for a week.

### Install / Start (Install or Start on Apple)
```
eligible Mac? (Apple silicon + macOS 26+)
 ├─ no → refuse
 └─ yes → supported CLI installed? (1.1 <= v < 2.0; absent/old/unparseable all count as no)
      ├─ no + Install allowed → download pinned pkg → hash check
      │    ├─ hash mismatch → never elevate, fail
      │    └─ hash ok → admin password
      │         ├─ cancel → recoverable error, clear “checking”
      │         └─ ok → root re-hash → installer → start runtime (bounded)
      └─ yes / after install → start runtime services (bounded)
           ├─ fails (except “already running”) → report + clear checking
           └─ ok → wait briefly → refresh availability
                └─ background: check agent image
```

### Agent container (parity)
```
run fails
 ├─ “kernel not configured” → set recommended kernel → retry once
 └─ else → is the runtime actually up? (status + real list probe)
      ├─ down → restart runtime → retry once
      └─ up → surface the original error (no false heal)

host talk-back URL = default-network gateway IP
 └─ if gateway unknown → refuse start (do not use host.docker.internal)
```

Stop escalation: stop → kill → force-delete container VM (bounded).

### What it can’t do
Hard-TCC mounts; huge dataless iCloud trees; cancel mid-download; Apple live e2e in CI. Clean-machine Install should be smoked once before merge (unit-mocked today).

## Parity carried in this PR
1. Probe-based run recovery when the runtime is down (no error-text heuristics)
2. Force-delete last-resort stop (`container delete --force`, 10s bound)
3. `isRunning` requires `system status` + `container list` (both 10s bound)
4. Bounded stop/setup diagnostics attached to Sentry on escalation/setup failure
5. Corrupt image recovery: failed container record deleted first, then `image delete --force` (live-verified 1.1.0 semantics: `--force` only quiets not-found)
6. Memory stats via guest `/proc/meminfo` (CLI has `stats`, but meminfo is the proven dogfood path; `cpuPercent: 0` on purpose)
7. Logs use `-n` (not `--tail`)
8. Host API / bridge IP via gateway from `container network inspect default` - fail closed if missing
9. Version gate: CLI must be 1.1 <= v < 2.0; unsupported installs route through the pinned installer (upgrade in place, still password-gated). Known compromise: the readiness path reads an old install as "not installed" (no third availability state)
10. iCloud hang guard: `container run` bounded at 2 min (base hook, Apple-only override); timeout surfaces "download the folder or remove it" instead of hanging on dataless Desktop & Documents trees

Also: native `--timeout 60` on `container system start`, and bounded `system stop` on shutdown/restart.

## How it was verified
- Unit: eligibility, allowInstall, digest mismatch never elevates, elevate rehash + mktemp shape, cancel mapping, mutex, startRunner persistence / failure refresh, gateway fail-closed, bounded start command
- Live CLI 1.1.0 smoke (already-installed): start → healthy → agent LLM via gateway-routed proxy; staged self-heal after `container system stop`
- typecheck + lint + vitest green locally
- [x] Manual: clean-machine / CLI-uninstalled Install once (password → start → send message) - done 2026-07-16, full smoke matrix passed (cold start, mid-session self-heal, clean-machine install, admin cancel)

## Notes for the reviewer
- No silent provision: Install only via explicit Start/Install (`allowInstall`).
- No double Sentry on run recovery (base `start` / ensure already capture).
- Follow-ups (not this PR): Hard-TCC mount errors, dataless iCloud hang, pull-stall watchdog, optional `container stats` shape probe.
- Replaces fork-head draft #485.


src/shared/lib/container/apple-container-client.ts
```diff
@@ -187,3 +255,144 @@ export class AppleContainerClient extends BaseContainerClient {
...
+      )
+    }
+
+    // Elevate: copy + re-hash under root before installer (closes same-UID TOCTOU).
+    reportProgress(onProgress, 'Installing - enter your password if prompted...', null)
+    const elevateScript = [
+      'set -e',
+      `SRC=${shellEscape(pkgPath)}`,
+      `EXPECTED=${shellEscape(APPLE_CONTAINER_PKG_SHA256)}`,
+      'TMP=$(/usr/bin/mktemp /tmp/superagent-container-XXXXXX.pkg)',
```
iddogino: In Mac, filename has to **end** in XXXX to become random. See experiment below:

**Current code - collision:** 
```
$ mktemp /tmp/superagent-container-XXXXXX.pkg
/tmp/superagent-container-XXXXXX.pkg
```

**Correct:**
```
$ mktemp /tmp/superagent-container-XXXXXX
/tmp/superagent-container-pupOCa
```

iddogino (changes requested):

JeremyBischoff:
Re the mktemp template comment: done in `874a2c85` - the elevate script now mktemps a random dir (`/tmp/superagent-container-XXXXXX`) and copies the pkg inside it, so the template ends the filename.

iddogino (changes requested):
### 1 - Existing Apple Container installations are accepted regardless of version

[`isAvailable()` only checks whether `container --version` succeeds](https://github.com/SkillfulAgents/SuperAgent/blob/b23277f390b3535501d26dd242c30a945cfe62a4/src/shared/lib/container/apple-container-client.ts#L420-L422). The pinned 1.1 installer is invoked [[only when the command is completely absent](https://github.com/SkillfulAgents/SuperAgent/blob/b23277f390b3535501d26dd242c30a945cfe62a4/src/shared/lib/container/apple-container-client.ts#L530-L540)](https://github.com/SkillfulAgents/SuperAgent/blob/b23277f390b3535501d26dd242c30a945cfe62a4/src/shared/lib/container/apple-container-client.ts#L530-L540).

Immediately afterward, Superagent always runs:

```text
container system start --enable-kernel-install --timeout 60
```

`--timeout` is supported in Apple Container 1.1, but not in the [[0.12.3 command reference](https://github.com/apple/container/blob/0.12.3/docs/command-reference.md)](https://github.com/apple/container/blob/0.12.3/docs/command-reference.md). Existing dogfood users can realistically have a 0.x installation because Superagent supported Apple Containers before 1.0.

For those users:

- The UI says “Start,” not “Install” or “Update.”
- Superagent never installs the pinned version.
- Every start attempt fails on the unsupported argument.
- Future incompatible major versions would also be accepted.

Recommended fix:

- Parse `container system version --format json` or `container --version`.
- Define a supported range, such as `>=1.0,<2.0`, or require 1.1 explicitly.
- Surface an “Update Apple Container” action when installed but incompatible.
- After starting, verify both client and API server versions.
- Add tests for absent, old, supported, and future-major installations.

This is the clearest difference from Lima: Lima’s binary is bundled and its VM configuration/version is reconciled; Apple currently treats “installed” as “compatible.”

### 2 - The corrupt-image recovery does not have the semantics claimed

[`removeCorruptImage()` uses `container image delete --force`](https://github.com/SkillfulAgents/SuperAgent/blob/b23277f390b3535501d26dd242c30a945cfe62a4/src/shared/lib/container/apple-container-client.ts#L410-L415), with a comment saying this handles images still referenced by a container.

Apple’s [[1.1 command reference](https://github.com/apple/container/blob/1.1.0/docs/command-reference.md)](https://github.com/apple/container/blob/1.1.0/docs/command-reference.md) describes `--force` as ignoring images that are not found; it does not detach an image referenced by an existing container. If a failed `container run` leaves a container record, image deletion can still fail before the repull.

Delete the failed container first, or add a live integration test demonstrating the referenced-image recovery case.

### iCloud 

I think we should solve this (tree protection) - it was a persistent user bug

iddogino (approved):

JeremyBischoff:
**1 (version gate)** - Done in 7f77ab00.

**2 (corrupt-image recovery)** - Done in 7f77ab00 - the failed container record is deleted (best-effort, 10s bound) before `image delete --force`, and the comment no longer claims detach semantics.

**iCloud** - Agreed it needed solving in this PR.

PR #492, opened 2026-07-16
feat: home graph connection editing — elbow connectors, drawing, geometry persistence

westaskew:
## Summary

Interaction layer for the home graph — **part 3 of the stack**, based on `feat/home-graph-view` (#491). West's work, building on Iddo's base from parts 1–2.

- **Elbow connectors**: orthogonal routes (waypoint model: per-segment cross-coords, port-anchored ends, rounded corners) between perimeter anchors on the nodes' visible shapes. Active links carry a traveling comet tracer (speed = log-scaled usage, direction = data flow, two-way = opposing tracers); idle gray / broken red stay static dashes.
- **Editing**: click selects — per-segment pill handles (end-segment drags spawn stub bends), endpoint dots slide the perimeter and magnetize to cardinal ports, floating toolbar deletes the underlying link/permission or jumps to the caller's policy editor. Geometry persists per edge (`graphEdgeGeometry` in user settings) alongside node positions.
- **Drawing**: FigJam-style — ports surface on hover with staged anchor affordances; dragging out draws a connection and creates the real relationship (invoke policy / account / MCP link); invalid targets fade while dragging. Webhooks/crons/chats are owned by their agent and created through forms, so their ports stay decorative.
- **Panel**: right-hand accounts/MCP directory; hovering a row spotlights that resource's edges.
- Supporting changes: `humanizeCron()` extracted as a non-hook form; one-shot `openAgentSettings` in NavTransientContext so "edit permissions" on an edge can navigate to the agent's settings tab without replay on later visits.

## Review notes

- `resource-panel.tsx` (added here) is **removed in part 4**, superseded by on-canvas details — skim it accordingly. It's kept here so each PR matches a real, working stage of the feature.

## Stack

1. `feat/home-graph-endpoint` (#490)
2. `feat/home-graph-view` (#491)
3. 👉 `feat/home-graph-connection-editing` (#492) (this PR)
4. `feat/home-graph-polish` (#493)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #491, opened 2026-07-16
feat: home graph view — force layout, usage-weighted edges, URL-driven home toggle

westaskew:
## Summary

Frontend half of the agent connections graph — **part 2 of the stack**, based on `feat/home-graph-endpoint` (#490). This portion was authored by Iddo; the commits preserve his authorship.

- The home page gains a **list/graph view toggle**, URL-driven via a `view` search param — deep-linkable and back/forward-safe (covered in `home-page.test.tsx`).
- **`@xyflow/react` canvas**: agents and their resources (accounts, MCPs, chats, webhooks, crons) as nodes, mind-map style.
- **d3-force layout** seeds positions; user drags persist per node (`graphNodePositions` in user settings), with a reset-layout action.
- **Edges weighted by usage** from the batch topology endpoint; the response is `homeGraphSchema.parse()`d at the fetch boundary so a drifting server shape fails loudly.
- Node identity still flows from the existing global endpoints, so agent status stays live via SSE.

New deps: `@xyflow/react`, `d3-force` (+ `@types/d3-force`).

## Test plan

- Unit: `use-graph-data.test.ts` (graph assembly), `home-page.test.tsx` (URL-driven toggle).
- E2E (`e2e/specs/home-graph.spec.ts`): toggle deep-linking/back-forward, topology edge rendering + node-click navigation, drag persistence + layout reset.

## Stack

1. `feat/home-graph-endpoint` (#490)
2. 👉 `feat/home-graph-view` (#491) (this PR)
3. `feat/home-graph-connection-editing` (#492)
4. `feat/home-graph-polish` (#493)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #493, opened 2026-07-16
feat: home graph polish — cards, count chips, hover details, spark charts

westaskew:
## Summary

Polish pass on the home graph — **part 4 (final) of the stack**, based on `feat/home-graph-connection-editing` (#492).

- **Cards**: agent nodes become standard cards (name top-left, status icon top-right); resource chips go rounded-square with a corner tone dot to match.
- **Hover details**: chip hover opens the detail card with a 150ms grace period so the cursor can cross onto the card (Settings stays reachable); pinned cards pop in with a deterministic per-resource stagger, so details-mode ripples across the canvas. Cron/webhook cards embed the same activity spark charts the agent homepage sidebar uses — keyed per owning agent so react-query dedupes the fetch across an agent's trigger nodes.
- **Count chips** ride the connectors (replacing the old node-level usage fine print), default to a fixed distance in from the target node so long routes don't strand them mid-air, and are draggable along the route (`chipT` persisted per edge).
- **Staged ports** and a global **details view** toggle.
- Removes `resource-panel.tsx` from part 3 — superseded by the on-canvas details.

After this merges, the stack's tip is byte-for-byte identical to `feat/agent-graph-interactions` (plus the standalone Slack fix).

## Stack

1. `feat/home-graph-endpoint` (#490)
2. `feat/home-graph-view` (#491)
3. `feat/home-graph-connection-editing` (#492)
4. 👉 `feat/home-graph-polish` (#493) (this PR)


iddogino:
Consolidated into #490 — the whole stack is now one squashed commit there.

PR #495, opened 2026-07-16
fix: chat integrations permanently died after one failed reconnect

iddogino:
## Problem

Field report (Tomer / TYP Buyers): Slack chat integration "occasionally stops responding" on the desktop app; iMessage integration stuck flagged in error state. Sentry corroborates fleet-wide: ELECTRON-3T (`SlackConnector.connect` via `reconnectAll`, 150 events / 11 users since May 26), plus the ELECTRON-3V `code=4000` bursts and ELECTRON-3G gateway timeouts on iMessage.

Root cause: **one failed reconnect attempt permanently orphaned the integration.** Both reconnect paths (`powerMonitor.resume` → `reconnectAll`, and the health-check) did `removeIntegration` → `connectIntegration`; if the connect threw (typical right after wake, while Wi-Fi is still re-associating), the integration vanished from the in-memory `connections` map — and every recovery mechanism iterated only that map. Compounding it:

- `SlackConnector.isConnected()` always returned `true` until an explicit `disconnect()`, so the health check was blind to dead Slack sockets → silent death with an `active` badge.
- No path ever wrote status `active` back after recovery → iMessage wore a permanent stale `error` badge even when its own reconnect loop had recovered.
- `removeIntegration` wiped the failure counter, making the 15-strike auto-pause unreachable.
- Fire-and-forget disconnect let old and new sockets fight over one-connection-per-identity gateways (the iMessage `code=4000` "replaced by another connection" bursts).
- `@slack/socket-mode` 2.0.6's internal reconnect leaks unrecoverable failures as unhandled promise rejections (`delayReconnectAttempt` has no `.catch`) — fatal at the app level (SUP-305).

## Fix

1. **DB-driven reconcile** — the health check's work list is now every startup-eligible integration (status `active`/`error`) from the DB, not the map. Orphans rebuild immediately; present-but-disconnected connectors get a grace window so their own faster loop goes first; failed reconnects retry every tick; the failure counter survives teardown so auto-pause actually fires (once per outage notification, not per tick).
2. **Slack connector owns its socket** — `SocketModeReceiver` built with `autoReconnectEnabled: false` (the leaky internal reconnect path never runs), own 1s→60s backoff restart loop, event-driven honest `isConnected()`, unrecoverable auth codes (`slack-error.ts`) stop the loop and surface via `onError`. Fail-fast `retryConfig` so the manager loop owns pacing.
3. **Status truth** — successful manager reconnects and connector self-recovery write `active` back.
4. **Race hardening** — `reconnectAll` coalesces overlapping resume bursts, force-reconciles, then retries at 15/30/60s instead of leaving wake failures to the 5-minute tick; `removeIntegration` awaits old-socket teardown (5s cap) before rebuild; a post-connect identity check tears down sockets whose owner was removed mid-connect.

## Testing

- **TDD throughout** — 33 new unit tests written red-first: reconcile (11), resume (6), Slack lifecycle (7), slack-error (9). Headline regressions pinned: "failed reconnect retried next tick", "wake-time failure recovers on next tick".
- **Regression** — full unit suite green; 650 tests across all files touching these modules; tsc + eslint clean.
- **Live against real Slack** (gated `*.live.test.ts`, `SLACK_LIVE=1`, never in CI):
  - Connector: raw websocket kill → drop detected in ~50ms → self-reconnect in ~1.5s → events flow post-recovery → user-authored message through the full inbound path → outbound OK.
  - Manager: orphan recovery (DB row + empty map → tick → connected + `active` writeback), takeover of a dead socket past grace (fresh connected connector), `reconnectAll` force-rebuild in ~600ms with a real inbound user message reaching the manager queue 1.8s later.

Not covered live: the one-line `powerMonitor.on('resume')` wiring (unchanged, needs a physical lid-close), real DB status writes (service fn unchanged; call contract unit-tested), iMessage connector internals (unchanged; benefits via the manager fixes).

After release, ELECTRON-3T / ELECTRON-3V / ELECTRON-3G going quiet is the field signal to watch.

https://claude.ai/code/session_01Tba7Jg6eGvxzCWvBS92HxY

### DIFF
--- a/src/shared/lib/chat-integrations/chat-integration-e2e.test.ts
+++ b/src/shared/lib/chat-integrations/chat-integration-e2e.test.ts
@@ -159,6 +159,10 @@ describe('Chat integration E2E', () => {
     (containerManager.ensureRunning as any).mockResolvedValue(mockContainerClient)
 
     registeredSessions.clear()
+
+    // connectIntegration cancels itself on a stopped manager; this harness
+    // drives addIntegration directly (no start()), so mark the manager running.
+    ;(chatIntegrationManager as unknown as { isRunning: boolean }).isRunning = true
   })
 
   afterEach(async () => {
--- a/src/shared/lib/chat-integrations/chat-integration-manager-archived-restore.sup233.test.ts
+++ b/src/shared/lib/chat-integrations/chat-integration-manager-archived-restore.sup233.test.ts
@@ -89,6 +89,9 @@ describe('SUP-233 reconnect restore ignores archived sessions', () => {
     testSqlite = new Database(':memory:')
     testDb = drizzle(testSqlite, { schema })
     migrate(testDb, { migrationsFolder: path.join(process.cwd(), 'src/shared/lib/db/migrations') })
+    // connectIntegration cancels itself on a stopped manager; this harness
+    // drives it directly (no start()), so mark the manager running.
+    ;(chatIntegrationManager as unknown as { isRunning: boolean }).isRunning = true
   })
 
   afterEach(async () => {
--- a/src/shared/lib/chat-integrations/chat-integration-manager-reconcile.test.ts
+++ b/src/shared/lib/chat-integrations/chat-integration-manager-reconcile.test.ts
@@ -1,0 +1,333 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+
+// ---------------------------------------------------------------------------
+// DB-driven health-check reconcile.
+//
+// The old runHealthChecks iterated the in-memory `connections` map, and both
+// reconnect paths did removeIntegration → connectIntegration. A single failed
+// connect left the integration absent from the map, so no later tick ever saw
+// it again — permanently dead until app restart, with whatever status happened
+// to be in the DB at the time ("stops responding").
+//
+// The reconcile loop instead treats the DB as the work list: every startup-
+// eligible integration (status active/error) is checked each tick, so a failed
+// reconnect is simply retried on the next tick. These tests drive
+// runHealthChecks() directly with the service layer mocked and connectors
+// stubbed, and assert:
+//   1. an orphaned integration (in DB, not in the map) is reconnected,
+//   2. a FAILED reconnect is retried on the next tick (the regression),
+//   3. the failure counter survives ticks so auto-pause is reachable,
+//   4. a present-but-disconnected connector gets a grace window before the
+//      manager tears it down (its own reconnect loop goes first),
+//   5. recovery writes status 'active' back (both via manager reconnect and
+//      via connector self-recovery), so no stale error badge remains,
+//   6. healthy integrations cause zero writes and zero rebuilds,
+//   7. a failed attempt records status 'error' with the attempt count,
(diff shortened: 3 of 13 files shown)


iddogino:
## Deep-review follow-up (commit 1a87ad6a)

A multi-lens review of the original commit (own pass + concurrency lens + Slack/Bolt API-contract lens, cross-checked against an independent external review) found **two P1 lifecycle races and several smaller gaps**. All fixed here, TDD (10 new/rewritten unit tests, red→green).

### P1 — stale rebuild overrides user lifecycle ops
A rebuild spans two await gaps (teardown up to 5s, then connect). A user **pause / delete / config update** landing in either gap could:
- resurrect a paused integration (live socket on a `paused` row, or `error` written over `paused` making it startup-eligible again),
- restore **pre-update credentials** (the rebuild reconnects from its stale row snapshot, tearing down the new-credential connector),
- report the user's own interference as a reconnect failure.

**Fix:** per-integration **lifecycle generation**. Public ops bump it; rebuilds capture it up front, re-read after teardown, verify after connect, and treat any change as *cancellation* — no reconnect, no status writes, any socket already opened is torn down. Failure cleanup only touches map entries the failing connect owns. `stop()` mid-rebuild also cancels.

### P1 — in-flight Slack restart survives `disconnect()`
`client.start()` awaiting `apps.connections.open` has no websocket, so socket-mode's `disconnect()` resolves without touching it — the pending start then **opens a socket after teardown finished**. That ownerless socket keeps acking its round-robin share of events → intermittent silent message loss on the replacement connector (the exact symptom this PR fixes). Trigger is realistic: manager takeover fires precisely when the connector's own loop has been retrying, so a start is often mid-flight.

**Fix:** the restart promise is tracked; once `disconnecting` is set its success handler closes the late socket, `disconnect()` waits for it, and a late `'connected'` can't flip a torn-down connector live. `connect()` now uses locals + a post-start check, so a disconnect racing it rejects cleanly instead of the ELECTRON-3T `Cannot read properties of null` TypeError.

### Smaller
- Coalesced resumes now queue **one more force pass** (force:false follow-ups can't be trusted after a second sleep — `isConnected()` can read stale-true on half-open sockets).
- WebClient calls bounded at 30s (library default is *unbounded*; a blackholed `apps.connections.open` could wedge an attempt indefinitely).
- Websocket-stage restart failures rejected with `undefined` — log now says so; stray 5s teardown race timer cleared.

### Validation
- 7,132 unit tests green (full suite), tsc + eslint clean.
- Both gated live suites re-run against the real workspace **on the fixed code**: kill→detect 51ms, self-heal 1.5s, orphan recovery / manager takeover / resume rebuild, real user-authored inbound message end-to-end — all attempt-1.

Verified-clean along the way (installed-source traced): the SUP-305 unhandled-rejection leak path is structurally dead with `autoReconnectEnabled:false`; `installerOptions.clientOptions.retryConfig` is the correct (and load-bearing) plumbing in bolt 4.7.0; the 5-code unrecoverable set exactly matches socket-mode's own; `'connected'` fires on every re-connect so the backoff reset is sound.

https://claude.ai/code/session_01Tba7Jg6eGvxzCWvBS92HxY

PR #499, opened 2026-07-17
Fix AUTH MODE model selection for workspace members

iddogino:
## Summary

- add an authenticated, least-privilege model configuration endpoint for workspace members
- add a sanitized client configuration endpoint for member-visible onboarding, analytics, retention, and Composio state
- migrate member-visible consumers away from the workspace-admin-only settings endpoint
- preserve agent-owner access to agent-scoped default model preferences
- add AUTH MODE E2E coverage for owner default-model editing, composer model selection, and the absence of full-settings 403s
- clean model configuration mocks and use visible model-family rows in E2E interactions

## Root cause

Shared model pickers loaded their catalog from `/api/settings`, which is workspace-admin-only in AUTH MODE. Members therefore received a 403 and rendered an empty model list. Several other member-visible surfaces also subscribed to that same privileged endpoint.

The initial UI fix also hid the agent default-model card from every non-workspace-admin, even though `AgentAdmin()` authorizes agent owners to update their own agent preferences. This update aligns the UI with that agent-scoped backend authorization.

## Impact

Workspace members can see and select available models, and agent owners can edit their agents' default models. Member-visible surfaces receive only the small, sanitized configuration payload they require, while the full settings API remains workspace-admin-only.

## Validation

- AUTH MODE Playwright E2E: 1 passed
- full Vitest suite: 7,322 passed, 31 skipped, 1 todo
- TypeScript typecheck passed
- ESLint passed with 0 errors
- staged diff check passed


iddogino:
superseded by PR 559

PR #505, opened 2026-07-17
fix: keep pending-request actions reachable on tall cards (SUP-419)

JeremyBischoff:
Re-applies the SUP-419 fix from #500, which was reverted in #504 because it merged without review. Same change, now up for review.

## The bug

When the agent asks a question and you pick "Other" then paste a long answer, the "Other" box grew unbounded and pushed Submit below the viewport with no way to scroll to it - the answer couldn't be submitted. The pending-request card lives in a footer slot that sizes to its content, so only the message list scrolled, never the card.

## The fix (at the shared request-card shell, not per-card)

- Clamp the card to half the viewport height and let it scroll internally.
- Pin the action row to the card bottom with a full-width divider so Allow/Submit stay reachable across every pending-request kind (question, script, connected-account, MCP, etc.), not just the "Other" case reported.
- Give the content above the divider the same 16px breathing room the buttons already have below it, so the divider doesn't crowd inputs.

The remote-MCP card has a second layout where the buttons sit beside a service picker; there it keeps a plain button row, since the full-width divider only makes sense where the row spans the card.

## Verification

Exercised in the E2E mock: with a 60-line paste the card clamps to 360px and scrolls internally while Submit stays on-screen and submits. Question / multi / script / secret / connected-account / default-MCP cards checked the same way. The MCP side-by-side layout isn't reachable in the mock (needs a live registered MCP server); it reverts to the pre-existing plain button row, so no visual change there.

### DIFF
--- a/src/renderer/components/messages/remote-mcp-request-item.tsx
+++ b/src/renderer/components/messages/remote-mcp-request-item.tsx
@@ -12,6 +12,7 @@ import { Button } from '@renderer/components/ui/button'
 import { Input } from '@renderer/components/ui/input'
 import { ToolPolicyEditor } from '@renderer/components/settings/tool-policy-editor'
 import { DeclineButton } from './decline-button'
+import { RequestError } from './request-error'
 import { RequestItemShell } from './request-item-shell'
 import { RequestItemActions } from './request-item-actions'
 import { cn } from '@shared/lib/utils/cn'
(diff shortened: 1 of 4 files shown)


src/renderer/components/messages/remote-mcp-request-item.tsx
```diff
@@ -650,7 +650,7 @@ export function RemoteMcpRequestItem({
               />
             ) : null}
           </div>
-          <RequestItemActions>
+          <div className="flex justify-end gap-2 pt-4">
```
iddogino: Can we keep using `RequestItemActions` and add an `inline={true}` prop ?

src/renderer/components/messages/request-item-shell.tsx
```diff
@@ -161,7 +161,10 @@ export function RequestItemShell({
   const showStopButton = !!(sessionId && agentSlug)
 
   return (
-    <div className="border rounded-[12px] bg-muted/30 shadow-md text-sm" {...dataAttrs}>
+    <div
+      className="max-h-[50vh] overflow-y-auto border rounded-[12px] bg-muted/30 shadow-md text-sm"
```
iddogino, line 165: Submission errors can be hidden below the sticky footer -> can you verify errors in submission are actually visible ?

iddogino (changes requested):

JeremyBischoff:
Re the `inline={true}` ask: Done in `6c308ecb` - side-by-side MCP actions use `RequestItemActions` again with `inline`, so that row stays a plain end-aligned layout while everyone else keeps the sticky footer.

Re submission errors under the sticky footer: verified - with the sticky opaque bar, the error after the actions could sit under it on a tall scrolled card. Done in `6c308ecb` - the shell hands the error into the sticky action bar (below the buttons, same order as before), so it stays visible with Skip/Submit.

PR #506, opened 2026-07-17
feat(microvm): pass agentId to lifecycle /microvm/run

yiw190:
## Summary
- Include `agentId` on the lifecycle `/microvm/run` body so Phase C can bind the client cert SPIFFE URI SAN to `cell/org/agent/launch`
- No behavior change until gamut-infra Phase C enables cert issuance (`AGENT_GATEWAY_URL` + workload CA)

## Test plan
- [ ] Typecheck / existing microvm runtime tests
- [ ] Deploy before or with gamut-infra Phase C terraform (issuance requires `agentId` once enabled)
- [ ] Staging session launch still succeeds after both land

### DIFF
--- a/src/shared/lib/container/lambda-microvm-runtime.ts
+++ b/src/shared/lib/container/lambda-microvm-runtime.ts
@@ -539,7 +539,7 @@ async function serviceFetch<T>(
 
 async function runMicrovm(
   config: MicrovmRuntimeConfig,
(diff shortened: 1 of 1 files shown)


yiw190:
Not needed: workload SPIFFE URI is cell/org/launch only; agentId stays host-app PROXY_TOKEN scope (gamut-infra#83).

Issue #366, opened 2026-07-01
Subagent-originated secret/file/AskUserQuestion input requests never surface (silent hang)


JeremyBischoff:
## Summary

When a **subagent** (spawned via the Task tool) calls a "top-level" input tool - `mcp__user-input__request_secret`, `mcp__user-input__request_file`, or `AskUserQuestion` - the request is never surfaced to the host. It appears in **neither the app UI nor any chat connector** (Telegram/Slack/iMessage). The subagent blocks in the container awaiting input that can never be provided, and the whole turn hangs silently.

## Root cause

Subagent stream messages carry a non-null `parent_tool_use_id`, so `src/shared/lib/container/message-persister.ts:761` diverts them to `handleSidechainMessage` and returns before the main tool-routing switch. The sidechain paths only dispatch three input tools (`request_browser_input`, `request_script_run`, `computer_use`) - in two places: the complete-message block loop (`:1417-1433`) and the subagent streaming path (`:1674-1692`). `request_secret` / `request_file` / `AskUserQuestion` are only handled on the **main-agent** path (`:1809` / `:1903` / `:1913`), which a subagent never reaches.

Consequences:
- No `secret_request` / `file_request` / `user_question_request` event -> the app renderer (which has the handlers, e.g. `src/renderer/hooks/use-message-stream.ts:865`) shows nothing.
- `isAwaitingInput` is never set and `pendingInputRequests` stays empty -> the host doesn't know the session is blocked, so cancel-on-message can't rescue it (it no-ops when not awaiting). Only an interrupt/restart unblocks it.

## Reproduce

Dispatch a subagent and have it call `request_secret` (e.g. ask it to fetch a GitHub token from within a subagent). The subagent hangs; nothing surfaces in the app or Telegram.

## Fix direction

Surface these three tools from the subagent paths, mirroring the existing `request_browser_input` handling, in **both** spots (`:1417-1433` complete-message and `:1674-1692` streaming), each calling its existing handler (`handleSecretRequestTool` / `handleFileRequestTool` / `handleAskUserQuestionTool`). Consider extracting a shared dispatch helper so main + sidechain can't drift again.

## Open design question

Should subagents be able to call these top-level tools at all, or be restricted to the subagent-native input tools? Surfacing (so the user can respond) matches how `browser_input` already works and removes the silent hang; restricting is the alternative. Decide before implementing.

## Validation

[...]

PR #510, opened 2026-07-18
fix: surface subagent user-input asks on the host (SUP-424)

JeremyBischoff:
## What

Surface child-agent (Task/Agent) blocking user-input asks on the host the same way the main agent does: emit the existing `*_request` SSE and mark the conversation awaiting so the UI / chat indicator leave "Working…".

## Why

Sidechain messages early-return before the main tool dispatch. The subagent handlers only wired browser_input / script_run / computer-use (+ capability review on stream), so tools like `request_file` (and AskUserQuestion / secret / account / remote_mcp if a child ever calls them) never got a card, and browser_input often never flipped awaiting - silent hang. GitHub #366 / Linear SUP-424.

## How

Private `dispatchBlockingUserInputTool` shared by main stream, subagent stream, and complete-assistant sidechain. First delivery wins when stream stop and complete-assistant both carry the same `toolUseId` (reuse `pendingInputRequests`). script_run / computer-use / capability-review stay on their own conditional-await paths.

## How it was verified

- Unit: `message-persister.test.ts` - stream matrix for the six dispatcher tools + complete-assistant AskUserQuestion + stream-then-complete dedupe (260/260)
- typecheck + lint clean on the change
- Live smoke: `web-browser` child called `request_browser_input` - card shown on desktop; awaiting behaved. Telegram correctly showed the desktop-only fallback for browser_input (expected - that type is unsupported in chat)

## Notes for the reviewer

- Live value today is tools children already have (`request_browser_input`, `request_file` on `web-browser`) plus awaiting. AskUserQuestion is on the host dispatcher and covered by unit tests, but `general-purpose` Agent does not currently expose that tool (ToolSearch: no match) - interactive Telegram AskUserQuestion from a child is a follow-up (expose the tool), not unlocked by this PR alone.
- Proxy-review awaiting is intentionally out of scope (separate PR).
- Footprint: 2 files; prod net-negative.


src/shared/lib/container/message-persister.ts
```diff
@@ -1875,29 +1907,25 @@ class MessagePersister {
...
-            if (block.type === 'tool_use' && block.name === 'mcp__user-input__request_script_run') {
-              this.handleScriptRunRequestTool(
-                sessionId,
-                block.id,
-                JSON.stringify(block.input || {}),
-                state.agentSlug
-              )
+            if (block.type !== 'tool_use') continue
+            const input = JSON.stringify(block.input || {})
+            this.dispatchBlockingUserInputTool(
```
iddogino, line 1928: **[P2] Clear resolved sidechain input requests**

This dispatcher stores the child request in `pendingInputRequests` and marks the host session as awaiting, but a child response arrives as a sidechain `user` message. `handleMessage` returns through `handleSidechainMessage` before the main `user` branch can run `handleToolResults`, so the matching request is never removed and `isAwaitingInput` stays `true` until the top-level turn settles.

That leaves a resolved card eligible for reconnect replay and keeps the session marked awaiting while the child has resumed. A subsequent user message can also enter `cancelAwaitingInput` and interrupt that resumed work. The cleanup should be count-aware in case multiple child requests are pending.

I verified the failure on this PR head. Adding the following test under `describe('subagent input request surfacing', ...)` gives **260 passed / 1 failed**; it fails at `expected true to be false` for the awaiting-state assertion:

```ts
it('clears awaiting state and replay entry when a subagent input is resolved', () => {
  const globalEvents: any[] = []
  const cleanup = messagePersister.addGlobalNotificationClient((event) => {
    globalEvents.push(event)
  })

  try {
    setupTaskTool()

    simulateSidechainToolUse('AskUserQuestion', 'sub-q-resolved', {
      questions: [{
        question: 'Pick DB?',
        header: 'DB',
        options: [{ label: 'Postgres', description: 'pg' }],
        multiSelect: false,
      }],
    })

    expect(messagePersister.isSessionAwaitingInput(SESSION_ID)).toBe(true)
    expect(messagePersister.getPendingInputRequests(SESSION_ID)).toHaveLength(1)

    globalEvents.length = 0

    mockClient._sendMessage({
      type: 'user',
      parent_tool_use_id: 'task-tool-1',
      message: {
        role: 'user',
        content: [{
          type: 'tool_result',
          tool_use_id: 'sub-q-resolved',
          content: 'Postgres',
        }],
      },
    })

    expect(messagePersister.isSessionAwaitingInput(SESSION_ID)).toBe(false)
    expect(messagePersister.getPendingInputRequests(SESSION_ID)).toHaveLength(0)
    expect(globalEvents).toContainEqual(expect.objectContaining({
      type: 'session_input_provided',
      sessionId: SESSION_ID,
    }))
  } finally {
    cleanup()
  }
})
```

iddogino (changes requested):
Requesting changes because the new sidechain request dispatch does not have a matching sidechain tool-result cleanup path. I reproduced this against head `7c577f5d88205cf3b437358efc436644f49a5a62`: the complete `message-persister.test.ts` run produced **260 passed / 1 failed**, with `isSessionAwaitingInput()` still returning `true` after the child received its matching `tool_result`.

Please clear the resolved sidechain request from the reconnect-replay store, transition out of awaiting state when no other requests remain, broadcast `session_input_provided`, and cover the request → sidechain response lifecycle with the regression test in the inline comment.

JeremyBischoff:
Re the sidechain cleanup comment: good catch. Done in `9b988a83` - when a child answer matches a pending ask, we clear the replay entry, leave awaiting only when nothing else is open, and broadcast `session_input_provided`. Added your test plus a guard that two open asks stay awaiting until the last one resolves.

src/shared/lib/container/message-persister.ts
```diff
@@ -1875,34 +1907,58 @@ class MessagePersister {
...
+            state.pendingInputRequests.delete(block.tool_use_id)
+            clearedPending = true
+            this.broadcastToSSE(sessionId, {
+              type: 'tool_result',
+              toolUseId: block.tool_use_id,
+              result: block.content,
+              isError: block.is_error || false,
+            })
+          }
+          if (clearedPending && state.isAwaitingInput && state.pendingInputRequests.size === 0) {
```
iddogino: Awaiting state ignores other request stores. Clearing the final `pendingInputRequests` entry sets `isAwaitingInput = false` without checking `pendingComputerUseRequests`. This hides a still-pending computer-use approval and makes subsequent messages bypass `cancelAwaitingInput`. The inverse also occurs: an auto-approved script remains in the replay map and can keep the session incorrectly awaiting. The condition should count all actually blocking requests, excluding auto-approved entries.

iddogino (changes requested):

JeremyBischoff:
Good catch. Done in `4ff03f38` - the waiting-light clear now counts real pocket-A asks (skips auto-approved replay leftovers) and still-open computer-use approvals. Approving the last computer-use ask uses the same check so awaiting does not get stuck. Added tests for both failure modes plus the CU-last-blocker exit.

JeremyBischoff:
Reconciled with main to pick up #527 (`b9b7e202`). Since #527 shipped its own sidechain clear, I dropped my inline one and kept `resolveSidechainInputRequests` - upgraded to the both-shelves `hasBlockingPendingRequests` rule so both your comments still hold (resolved sidechain requests cleared + computer-use / auto-approved counted). 269 green. Re-requesting review.

iddogino:
This PR's substance has been re-implemented on current main in #562 (it conflicts with the merged #527/#528, which landed overlapping pieces — notably `hasBlockingPendingRequests` and the sidechain resolve path). #562 carries the unified dispatcher, most of this PR's test matrix (ported against the post-#527/#528 APIs), and an additional external-blocker guard on the two rewritten clear sites. Once #562 merges this one can be closed. Context: #562 is Phase 1 of the pending-user-input rearchitecture plan; #560 (merged) is its characterization/acceptance suite.

JeremyBischoff:
Closing in favor of #562, which merged and carries this PR's substance on current main.

Verified on main rather than taken from the description: the characterization matrix has zero `surfacesToday: false` rows, so every blocking kind now surfaces from every delivery path. All seven behaviors this PR tested have equivalents there - matrix rows (`:741`, `:749`), first-delivery-wins dedupe (`:805`), sidechain resolve (`:814`), the both-shelves rule (`:832`), the auto-approved-script skip (`:851`) - plus an external-blocker deferral guard (`:888`) this PR could not have known about, since it pre-dates #528. Lifecycle suite runs 50/50 green on main locally.

Keeping the branch for now.

PR #518, opened 2026-07-20
fix(mcp): close DNS-rebind SSRF on remote-MCP discovery (SUP-434)

JeremyBischoff:
## Summary

- Closes DNS-rebind SSRF on remote-MCP / OAuth discovery: a public hostname that resolves to a private/link-local IP (e.g. metadata) is rejected before any outbound fetch.
- Shared `mcpSafeFetch` (own module) resolves DNS, pins the socket to a vetted address (undici `connect.lookup`), and follows redirects only after the same resolve/validate/pin on each Location (max 5 hops).
- Wired through discovery, OAuth metadata/token flows, remote-MCP routes, and the MCP proxy. Also rejects URL-canonical IPv4-mapped loopback (`::ffff:7f00:1`) and the full IPv6 link-local range (`fe80::/10`).

## Why we refuse private IPs but allow public ones

SuperAgent is the one making the request - not the user's browser.

If we allow a private IP, the app can reach things that only the **host** can see: localhost services, cloud metadata (`169.254.169.254`), internal VPC stuff. In multi-user / AUTH_MODE that means one person can aim the server's network privilege at those targets.

A public IP is out on the internet - same class of place a browser could hit anyway. The danger we're closing is using SuperAgent as a proxy into the machine's private network, not "never talk to the internet."

```
User picks a remote MCP URL
        │
        ▼
  SuperAgent server fetches it
  (server's network, not the browser's)
        │
        ├──────────────────────────────┐
        ▼                              ▼
   Public IP                      Private IP
   (internet)                     (host-only)
        │                              │
        ▼                              ▼
   Normal MCP / OAuth          localhost services
   on the open internet        cloud metadata 169.254.169.254
                               internal VPC / link-local
                                      │
                                      ▼
                               Attacker reads secrets /
                               hits internal APIs via
                               SuperAgent as a proxy
```

**Before (name-only check)**

```
URL hostname text
  "looks public?" ──yes──► fetch
       │
       │  (misses: evil.nip.io → 169.254.169.254)
       └──no──► refuse
```

**After (resolve + pin + safe redirects)**

```
URL hostname text
  │
  ▼
DNS: what IPs does this name have?
  │
  ├─ any private? ──yes──► refuse
  │
  └─ all public ──► connect pinned to a vetted IP
                         │
                         └─ on 3xx: re-check Location the same way
                            (max 5; no private hop; no cross-origin
                             Authorization / 307 body replay)
```

**Why the split**

```
Browser → private IP     usually blocked / useless
SuperAgent → private IP  can reach the host's insides

So: public OK (internet peers)
    private refuse (don't rent out the host's network)
```

## Test plan

- [x] `url-safety-dns-ssrf.test.ts` — DNS-rebind reject, multi-A, hex-mapped IPv6, E2E localhost exception
- [x] `mcp-safe-fetch.test.ts` — public redirect + Auth strip, redirect→private refuse, 307 body, live pin connect
- [x] `fea0::1` in private-host table (`fe80::/10`)
- [x] Related MCP suites green with DNS mocks (oauth, oauth-ssrf, remote-mcps, mcp-proxy)
- [x] `npm run typecheck`
- [x] Live DNS rebind check locally: nip.io / localtest.me rejected; example.com accepted

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -90,6 +90,7 @@
         "tailwind-merge": "^2.6.0",
         "tailwindcss-animate": "^1.0.7",
         "tar": "^7.5.15",
+        "undici": "^7.28.0",
         "uuid": "^14.0.1",
         "ws": "^8.18.0",
         "yauzl": "^3.3.0",
@@ -22831,7 +22832,6 @@
       "version": "7.28.0",
       "resolved": "https://registry.npmjs.org/undici/-/undici-7.28.0.tgz",
       "integrity": "sha512-cRZYrTDwWznlnRiPjggAGxZXanty6M8RV1ff8Wm4LWXBp7/IG8v5DnOm74DtUBp9OONpK75YlPnIjQqX0dBDtA==",
-      "devOptional": true,
       "license": "MIT",
       "engines": {
         "node": ">=20.18.1"
--- a/package.json
+++ b/package.json
@@ -128,6 +128,7 @@
     "tailwind-merge": "^2.6.0",
     "tailwindcss-animate": "^1.0.7",
     "tar": "^7.5.15",
+    "undici": "^7.28.0",
     "uuid": "^14.0.1",
     "ws": "^8.18.0",
     "yauzl": "^3.3.0",
--- a/src/api/routes/mcp-proxy.test.ts
+++ b/src/api/routes/mcp-proxy.test.ts
@@ -1,10 +1,19 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
 import { Hono } from 'hono'
+import * as dnsPromises from 'node:dns/promises'
 
 // ---------------------------------------------------------------------------
 // Mock dependencies
 // ---------------------------------------------------------------------------
 const mockValidateProxyToken = vi.fn()
+
+vi.mock('node:dns/promises', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('node:dns/promises')>()
+  return {
+    ...actual,
+    lookup: vi.fn(),
+  }
+})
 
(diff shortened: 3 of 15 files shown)


iddogino (changes requested):
Good catch on the source issue, a few issues with the PR:

1. Not sure I see the case for the broad refactoring of most of `url-safety.ts` in the scope for this PR. Adds unnecessary risk for this PR. I'd leave it as is
2. `mcpSafeFetch` to be introduced in a separate `mcp-safe-fetch.ts`.
3.  `isPrivateIPv6` is wrong - misses IPs like `fea0::1` which are still private. IPv6 link-local space is `fe80::/10`, extending through `febf::`.
4. Forcing `redirect: 'manual'` prevents redirect-based SSRF, but every caller now receives an unhandled 3xx. **This breaks previously supported MCP and OAuth endpoints** using HTTP→HTTPS, canonical-host, or trailing-slash redirects—including token refresh for existing integrations. Implement a bounded redirect loop that resolves, validates, and pins every Location target before following it.

JeremyBischoff:
Re the review:

1/2. Done in `9d31b1f0` - left DNS resolve in url-safety; moved `mcpSafeFetch` into its own `mcp-safe-fetch.ts` so that file isn't carrying the fetch machinery.

3. Done in `9d31b1f0` - link-local check now covers the full `fe80::/10` range (`fea0::` included).

4. Done in `9d31b1f0` - bounded redirect loop (max 5, same shape as the Slack download path): each Location is resolve + validate + pin before follow. Also strips Authorization/Cookie on cross-origin hops, and won't replay a 307/308 body onto a new origin.

PR #525, opened 2026-07-21
Fix silent no-op grant when approving a remote MCP that needs re-auth

iddogino:
## Problem

Approving a `request_remote_mcp` request against an **existing** MCP server whose OAuth has expired (`status: 'auth_required'`) silently does nothing:

1. `provide-remote-mcp` rebuilds `REMOTE_MCPS` with `.filter(mcp.status === 'active')` — the stale server is dropped — but **still resolves the pending input as a success**.
2. The container tool can't find the resolved ID in `REMOTE_MCPS`, so it skips the query restart / tool injection, yet returns the bare "Access to the remote MCP server has been granted."
3. The model is told "granted" twice, hunts for tools that never register, and gives up.

Seen in the wild on a sales agent whose Granola MCP token had expired: two approvals both no-oped; it only recovered after the stale registration was deleted and a fresh OAuth connect ran. Fresh connects always work, which is why this doesn't repro locally without a stale record.

## Fix (three layers)

- **Approval UI** (`remote-mcp-request-item.tsx`, `mcp-server-card.tsx`): port the connected-account pattern. Non-active servers are never auto-selected and can't be toggled; their card shows the `McpStatusPill` ("Re-auth needed") and a **Reconnect** button that drives the existing `initiate-oauth` `mcpId` re-auth flow. Once the server comes back `active`, it's selected and Allow Access enables. Also tightened the card's `status` type from `string` to the `'active' | 'error' | 'auth_required'` union.
- **Backend guard** (`agents.ts` provide-remote-mcp): if any requested server isn't `active`, return 409 with `needsReauth: true` + server names instead of resolving the pending input.
- **Honest container message** (`agent-container/src/tools/request-remote-mcp.ts`): if an approval resolves but the server never lands in `REMOTE_MCPS`, return an explicit error saying the tools were NOT registered and the connection needs re-authentication, instead of the bare success.

## Testing

- `npx tsc --noEmit` clean in root and `agent-container`
- eslint clean on changed files
- New tests: 409 guard in `provide-remote-mcp.test.ts`; stale-server pill/Reconnect/re-auth-then-allow flow in `remote-mcp-request-item.test.tsx`
- Adjacent suites (`agents`, `remote-mcps`, `mcp-proxy`, `security-repro`, `connected-account-request-item`) all pass
- Not exercised end-to-end against a real expired-token server (needs a seeded stale registration)

https://claude.ai/code/session_01CZ3s61NJW8vkhTpZ6yxnc8

### DIFF
--- a/agent-container/src/tools/request-remote-mcp.test.ts
+++ b/agent-container/src/tools/request-remote-mcp.test.ts
@@ -1,0 +1,95 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { inputManager } from '../input-manager'
+
+const mockAddRemoteMcpServer = vi.fn()
+vi.mock('../claude-code', () => ({
+  getCurrentProcess: () => ({
+    addRemoteMcpServer: (...args: unknown[]) => mockAddRemoteMcpServer(...args),
+  }),
+}))
+
+const GRANOLA_MCP = {
+  id: 'mcp-granola-1',
+  name: 'Granola',
+  proxyUrl: 'http://host/api/mcp-proxy/agent/mcp-granola-1',
+  tools: [{ name: 'list_meetings' }, { name: 'get_meetings' }],
+}
+
+describe('requestRemoteMcpTool', () => {
+  let originalRemoteMcps: string | undefined
+
+  beforeEach(() => {
+    vi.clearAllMocks()
+    originalRemoteMcps = process.env.REMOTE_MCPS
+  })
+
+  afterEach(() => {
+    if (originalRemoteMcps === undefined) {
+      delete process.env.REMOTE_MCPS
+    } else {
+      process.env.REMOTE_MCPS = originalRemoteMcps
+    }
+  })
+
+  async function invokeTool() {
+    const { requestRemoteMcpTool } = await import('./request-remote-mcp')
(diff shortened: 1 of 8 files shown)


iddogino:
Both review findings were valid — addressed in bae4808d:

**P1 (wrong account selected after reconnect):** confirmed. `handleOAuthComplete` now prefers the reconnected server by id, carried through the flow in a ref (`reconnectMcpIdRef`), and only falls back to the URL match for the new-server flow where no id exists yet. Selection is also additive now (mirrors `connected-account-request-item`), so a sibling account the user had checked isn't silently dropped. Regression test: two servers sharing a URL, reconnect the stale one, assert the provide payload contains its id.

**P2 (all unhealthy servers forced through OAuth):** confirmed — `auth_required` is set by the proxy on any 401, including bearer servers with revoked tokens, and `/initiate-oauth` 400s for them. `handleReconnect` now branches on `authType`:
- `oauth` → existing re-auth flow
- `bearer` → inline token input; `PATCH /api/remote-mcps/:id` with the new token, then `POST /:id/discover-tools` (which flips status back to `active` on success)
- `none` → re-probe via `discover-tools` alone (covers transient `error` states)

Tests added for all three recovery paths.

PR #527, opened 2026-07-21
fix: cancel pending browser-input requests on browser close + mark subagent input requests as awaiting

iddogino:
## Incident

A scheduled session sat pinned on **"working"** overnight with a pending browser-input card — and no open browser to act on it. The main agent had launched a **background** `web-browser` subagent that parked on `request_browser_input` at a GitHub login wall; the main agent then closed the browser out from under it, summarized, and ended its turn. The subagent could never settle, so turn-end kept emitting `session_waiting_background` forever, while the header contradicted the in-chat "Waiting for input…" chip.

## Two defects

1. **Status never flipped for subagent-originated requests.** Only the main-stream `content_block_stop` path called `markSessionAwaitingInput`; the sidechain delivery paths (complete assistant message + subagent stream events) only broadcast the card. The in-chat chip is renderer-local state, but the sidebar/header orange dot reads the persister's `isAwaitingInput` — so the agent showed "working" while blocked on the user.
2. **Browser teardown orphaned pending requests.** `/browser/close` and `/browser/notify-closed` cleaned up tabs/screencast/state but never touched the container `InputManager`, leaving `browser_input` pendings alive for their 24h human-input TTL — with the requesting (possibly different) session blocked behind an unanswerable card.

## Fix

- `handleBrowserInputRequestTool` now marks the session awaiting input itself (idempotent with the existing main-path marking), covering every delivery path.
- New `inputManager.rejectByType(type, reason)`, called from both browser-close routes. A blocked awaiter gets `Browser input request cancelled: the browser was closed…`, finishes its turn, the session settles to idle, and the card clears at the turn boundary.

## Tests (all failed before the fix)

- **Persister unit ×2** — sidechain `request_browser_input` (stream-event and complete-message paths) must set `isAwaitingInput` + emit `session_awaiting_input`.
- **Container unit ×3** — `rejectByType` rejects only `browser_input` pendings; unblocks the real `request_browser_input` tool handler with the cancelled error result.
- **E2E ×2** — new `SubagentBrowserInputScenario` (`subagent browser input` trigger) replicates the incident: card visible + status must read `awaiting_input` (previously stuck at `working`), and declining returns the agent to idle.

Validated: full persister file (254), chat-integrations + container suites (1453), agent-container suite (717), status/input/browser E2E suites (11), `tsc --noEmit` + eslint clean in both packages.

## Follow-up (not in this PR)

A background subagent that *dies* (process exit, TaskStop) still leaves its card and awaiting status behind — cleanup needs the host to track request→subagent origin and dismiss on terminal `task_updated`/`task_notification`.

https://claude.ai/code/session_01XawqeKZeJV9pT9Fjy7MFgA

### DIFF
--- a/agent-container/src/input-manager.test.ts
+++ b/agent-container/src/input-manager.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest'
+import { describe, it, expect, vi } from 'vitest'
 import {
   inputManager,
   AUTOMATED_INPUT_TTL_MS,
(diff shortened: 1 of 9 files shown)


iddogino:
Addressed both review findings in 21c11231 (failing tests written first for each):

**P1 — sidechain tool_results now clear awaiting state.** `handleSidechainMessage` routes 'user' messages through a new `resolveSidechainInputRequests`: for each tool_result matching a tracked `pendingInputRequests` entry it drops the replayable card, broadcasts the same `tool_result` event the main path emits (other tabs drop their copy), and clears `isAwaitingInput` — gated on no other request remaining open, so a parallel main-agent question keeps the orange status. Scoped to tracked requests only: ordinary subagent Bash/Read results don't broadcast or touch the flag. The mock now preserves `parent_tool_use_id` on subagent input results (the top-level shape was masking the production path) and holds the turn open ~2.5s after resolve so the new E2E asserts the Complete → `working` (not `awaiting_input`) → `idle` sequence. Unit coverage: sidechain resolve clears flag+card, unrelated sidechain results don't, parallel-request gating.

**P2 — close-before-registration race closed.** `requestBrowserInputTool` now synchronously checks `getBrowserState().active` immediately before `createPendingWithType` and returns a cancelled error result when no browser is open. The check+register and the close path's state-flip+`rejectByType` are each synchronous blocks on the event loop, so they cannot interleave — a close is either seen by the guard or its `rejectByType` sees the pending. New `request-browser-input.test.ts` covers close-before-registration (this test hung on a fresh 24h pending before the fix) and the normal active-browser flow.

Validation: persister file 257/257, agent-container 719 passed, chat-integrations+container 1456 passed, the status/input/browser E2E suites 12/12, tsc + eslint clean in both packages.

https://claude.ai/code/session_01XawqeKZeJV9pT9Fjy7MFgA

PR #535, opened 2026-07-21
Fix capability reviews silently auto-denying after 10 minutes on unattended sessions

iddogino:
## Problem

The SEO Agent's scheduled run this morning (session `6399fe0b`, 13:01 UTC) called `Workflow` under the default `review` policy. The approval card, session promotion, and Action-Required notification all fired correctly — but at exactly **+10:00** the launch was denied with the CLI's generic *"The user doesn't want to take this action right now"*, and the turn ended reporting success. Nobody had declined anything.

Root cause: the capability gate is a PreToolUse hook (canUseTool is shadowed under `bypassPermissions`), and the CLI enforces a **default 600s per-hook timeout** on SDK hook callbacks when the matcher config carries no `timeout`. The "park indefinitely awaiting a human" design was only proven for canUseTool (AskUserQuestion), which has no such ceiling. The container's 24h human-input TTL never got a chance.

## Fix

- **Extend the hook timeout**: the gate matcher now passes `timeout` (seconds) = 25h — one hour past `HUMAN_INPUT_TTL_MS`, so the container's TTL sweep always rejects first with the real `reviewDeclinedMessage` instead of the CLI's blunt abort.
- **Detect abandonment**: the CLI signals any abandonment of a parked hook (timeout, interrupt, parent abort) via `control_cancel_request`, which the SDK surfaces as the hook callback's `AbortSignal` — previously ignored. The gate (extracted to `capability-gate-hook.ts` for testability) now listens: on abort it rejects the parked pending entry (no more 24h zombie in the input manager) and emits a `capability_review_cancelled` frame.
- **Close the card**: the host handles the new frame with the existing `completeCapabilityReview` → `capability_review_resolved` SSE, so an unanswerable approval card is closed everywhere instead of dangling. This also covers the interrupt path, which previously stranded the container-side pending entry the same way.

## Tests

- New `capability-gate-hook.test.ts`: pass-through / block / park+approve (once & session scope) / user decline / **abort → pending wiped + cancellation reported** / late abort after settle is a no-op / hook timeout stays above the human TTL.
- `message-persister.test.ts`: a container cancellation frame clears the replay store and broadcasts `capability_review_resolved` like a decision does.
- `tsc`, `eslint`, and the adjacent container suites (claude-code, capability-policies, input-manager, session-manager-settings) are green. Two partial `vi.mock('./input-manager')` factories gained the newly imported `HUMAN_INPUT_TTL_MS` export.

Notification-copy follow-up ("wants to launch agents" for workflow launches) is split into its own PR.

https://claude.ai/code/session_01D1cNaBPEb5VemoSwSPAFJT

### DIFF
--- a/agent-container/src/capability-gate-hook.test.ts
+++ b/agent-container/src/capability-gate-hook.test.ts
@@ -1,0 +1,129 @@
+import { describe, it, expect, vi } from 'vitest'
+import {
+  createCapabilityGateHook,
+  CAPABILITY_REVIEW_HOOK_TIMEOUT_S,
+  REVIEW_CANCELLED_REASON,
+  type CapabilityGateContext,
+} from './capability-gate-hook'
+import { inputManager, HUMAN_INPUT_TTL_MS } from './input-manager'
+import type { Capability } from './capability-policies'
+
+function makeContext(overrides: Partial<CapabilityGateContext> = {}): {
+  ctx: CapabilityGateContext
+  grants: Set<Capability>
+  onSessionGrant: ReturnType<typeof vi.fn>
+  onReviewCancelled: ReturnType<typeof vi.fn>
+} {
+  const grants = new Set<Capability>()
+  const onSessionGrant = vi.fn((capability: Capability) => grants.add(capability))
+  const onReviewCancelled = vi.fn()
+  const ctx: CapabilityGateContext = {
+    sessionId: 'session-1',
+    getPolicies: () => ({ subagents: 'review', workflows: 'review' }),
+    getSessionGrants: () => grants,
+    onSessionGrant,
+    onReviewCancelled,
+    ...overrides,
+  }
+  return { ctx, grants, onSessionGrant, onReviewCancelled }
+}
+
+function invoke(ctx: CapabilityGateContext, toolName: string, toolUseId: string, signal?: AbortSignal) {
+  const hook = createCapabilityGateHook(ctx)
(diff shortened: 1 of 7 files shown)


iddogino:
Addressed the P2 race in b629f8a1 with the tombstone approach:

- `capability_review_cancelled` now checks whether the card is already stored. If yes → `completeCapabilityReview` as before; if not (the handler is still awaiting its grant lookup) → the toolUseId is tombstoned in `state.cancelledCapabilityReviews`.
- `handleCapabilityReviewTool` consumes the tombstone (`Set.delete` as check-and-consume) after its awaits and returns before broadcasting — so neither the card, nor `markSessionAwaitingInput`, nor the Action-Required notification fire for a cancelled review.
- Tombstones deliberately survive turn boundaries rather than being cleared on `session_active`/`session_idle`: a hung grant lookup can outlive the turn (interrupts tear down the very container the fetch is talking to), and since tool_use ids never recur, a lingering tombstone can only ever suppress its own stale card. They're consumed on match and die with the streaming state.
- Added the requested delayed-lookup race test (deferred `mockClient.fetch` promise, cancel lands mid-await, then the lookup resolves): asserts no `capability_review_request` broadcast, empty replay store, and that a subsequent review with a fresh id still renders. Verified the test fails without the consume check.

https://claude.ai/code/session_01D1cNaBPEb5VemoSwSPAFJT

PR #538, opened 2026-07-21
fix: skip host auto-sleep scans for Lambda MicroVM

yiw190:
## Bug

For Lambda MicroVM agents, the host `AutoSleepMonitor` still walks every running agent once a minute and calls `listSessions()` — session metadata lives on the S3 Files-backed `/data` mount (network filesystem I/O). That scan is wasted work: background auto-sleep already ends in `stop({ escalateToForceStop: false })`, which is a no-op on this runtime because AWS `idlePolicy` performs suspend/resume. So we pay network I/O every minute and reclaim nothing.

Staging already showed how expensive sync work on that mount is: SQLite on `/data` blocked Node for 5.268s and tripped the same 4s ECS health-probe `AbortError` at only ~10.5% CPU. This PR removes one redundant source of `/data` pressure; it does not fix broader SQLite isolation.

## Reproduction

1. Run host-app with `containerRunner=lambda-microvm` and one or more running agents.
2. Wait for the one-minute `AutoSleepMonitor` cycle.
3. Observe `listSessions()` (and related `/data` reads) for each agent even though the eventual MicroVM auto-sleep `stop()` does not suspend or terminate.

## What changed

- Add polymorphic `shouldRunHostAutoSleep()` (default `true` for container runtimes).
- Lambda MicroVM returns `false` so the monitor skips session / keep-alive scans when they cannot reclaim anything.
- Gate that check **before** active-session checks or session filesystem reads.
- Tests: default runtimes stay on host sweep; MicroVM opts out; monitor path does no session I/O when opted out.

## Why this shape (not “host must never Suspend”)

The goal of this PR is **less session / `/data` I/O on S3 Files**, not a claim that host-driven `Suspend` is impossible. Auto-resume on the next inbound request works either way. We skip the host sweep here because, with today’s no-op auto-sleep `stop()`, the scan is pure cost.

## Test plan

- [x] Focused tests: 142 passed
- [x] `npm run lint`: 0 errors (40 existing warnings)
- [ ] `npm run typecheck`: existing missing optional `e2b` / `@daytonaio/sdk` tool dependencies
- [ ] Full `npm run test:run`: 7,516 passed; one unrelated existing failure in `webhook-trigger-service.sup226.test.ts`

### DIFF
--- a/src/shared/lib/container/base-container-client.test.ts
+++ b/src/shared/lib/container/base-container-client.test.ts
@@ -31,6 +31,12 @@ class TestContainerClient extends BaseContainerClient {
   }
 }
 
+describe('host auto-sleep', () => {
+  it('is enabled by default for container runtimes', () => {
+    expect(new TestContainerClient({ agentId: 'a' }).shouldRunHostAutoSleep()).toBe(true)
+  })
+})
+
 describe('buildAgentEnv', () => {
   afterEach(() => enableToolSearch.mockReturnValue(true))
 
--- a/src/shared/lib/container/base-container-client.ts
+++ b/src/shared/lib/container/base-container-client.ts
@@ -289,6 +289,10 @@ export abstract class BaseContainerClient extends EventEmitter implements Contai
     this.config = config
   }
 
+  shouldRunHostAutoSleep(): boolean {
+    return true
+  }
+
   /**
    * Emit an 'error' event without crashing the process.
    *
--- a/src/shared/lib/container/container-manager.ts
+++ b/src/shared/lib/container/container-manager.ts
@@ -113,6 +113,10 @@ class ContainerManager {
     return client
   }
 
+  shouldRunHostAutoSleep(agentId: string): boolean {
+    return this.getClient(agentId).shouldRunHostAutoSleep()
+  }
+
   /**
    * Get cached container info. Returns cached status if available,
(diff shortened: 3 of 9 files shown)


yiw190:
## Staging canary evidence

Deployed commit `1f62e133` to the `agi-ddfc46` staging canary using immutable image digest `sha256:35dda9324885cf717cded53391de892c0da2b2a0a7f124780f06524f9c31d6da` (task definition revision 9).

Verified:
- ECS task is `RUNNING` / `HEALTHY`; service rollout is `COMPLETED`
- Public route returns HTTP 200
- No `host_health_failed` event after API readiness during the initial observation window

Controlled A/B used the same agent and S3 Files access point for one AutoSleep cycle:

**Before (`:8`)**
- Logged `Agent atl9ybww8u idle for >30m, stopping...`
- Task network RX: 1,634,675 Bytes/Second
- Task network TX: 5,485 Bytes/Second

**After (`:9`)**
- Logged `Runtime owns idle sleep; skipping host session scans`
- No idle-stop attempt
- Task network RX: 376 Bytes/Second
- Task network TX: 367 Bytes/Second

That controlled cycle reduced task RX by 99.98% and TX by 93.31%. Task network includes all traffic, so the direct behavioral proof is the logged branch: the Lambda MicroVM runtime bypassed `listSessions()`.

PR #539, opened 2026-07-21
fix: local-LLM loopback endpoints unreachable on the Apple Container runtime (SUP-447)

JeremyBischoff:
## What

Local-LLM agents reach the model on the Apple Container runtime when the model server is bound to all interfaces, and get a named, actionable failure when it is not.

Two things had to be true and only one of them was:

| Host model server | Before | After |
|---|---|---|
| bound to `0.0.0.0` | dead at DNS | works |
| bound to `127.0.0.1` (Ollama/LM Studio/llama.cpp default) | dead at DNS | start fails, message names the bind and the fix |
| not running | dead at DNS | start fails, message says nothing is listening |

Docker, Lima, WSL2 and Podman are unchanged.

## Why

`rewriteLoopbackForContainer` hardcoded `host.docker.internal`, which is NXDOMAIN inside Apple Container guests - Apple has no `--add-host` equivalent. The agent baked an unreachable `ANTHROPIC_BASE_URL`, so every model call died at DNS.

Threading the runtime's real host address in fixes the address, but not the whole problem. Docker Desktop's `host.docker.internal` is a *proxy* that forwards to the host's loopback. Apple Container's vmnet gateway IP is a *real network interface*, so it only reaches services actually listening on that interface. Same address, same code, different semantics - and loopback-only is the default bind for every common local-model server.

So the corrected address alone still produces a dead agent for the most likely configuration. The failure just moves from `ENOTFOUND` to `ECONNREFUSED`. This PR fixes the address and makes the remaining cases explain themselves instead of failing silently.

## How it works

Before starting a container, if the runtime's host address is one of our own interfaces, probe it. A refusal has two causes with two different fixes, so `127.0.0.1` is probed as well to tell them apart:

```
probe gateway:port
  ├─ connects            → start normally
  ├─ refused → probe 127.0.0.1:port
  │              ├─ connects  → wrong bind    → "bound to 127.0.0.1 only", how to rebind
  │              └─ refused   → not running   → "nothing is listening", start it
  └─ timeout/DNS/other   → unknown → never blocks a start
```

The probe runs on the host, not in a container. A service bound only to `127.0.0.1` is absent from every other interface, so connecting to the gateway IP from the host is refused for exactly the reason it is refused inside the guest. One TCP connect, ~20ms, no container spawn.

## How it was verified

Measured on macOS 26.5.1, Apple Container CLI 1.1.0, Docker 29.6.1, real `superagent-container` image, real Ollama.

Reachability of the baked address, probed from inside a real guest:

| Host service bound to | Apple guest to gateway IP | Docker guest to `host.docker.internal` |
|---|---|---|
| `127.0.0.1` | ECONNREFUSED | reachable |
| `0.0.0.0` | reachable | reachable |

End to end in the running app, generic provider pointed at `http://localhost:11434` on the Apple runtime:

- **Ollama stopped** - start blocked, banner reads "nothing is listening on port 11434 on this Mac at all", no bind advice.
- **Ollama at default bind** (confirmed `127.0.0.1` via `lsof`, not docs) - start blocked, banner names the bind and gives the per-server fix.
- **`OLLAMA_HOST=0.0.0.0`** - agent starts, container's `ANTHROPIC_BASE_URL` is `http://192.168.64.1:11434`, and the guest reaches real Ollama at it.

Suites: full unit run green (7535 passed, 428 files), typecheck and lint clean. Both new guards were mutation-checked - removing the forwarding-name gate and converting the deferral into a rethrow each fail exactly one test.

## Notes for the reviewer

- **New failure mode.** An agent that previously started and then failed on every model call now refuses to start. That is the intended trade, but it is a behavior change beyond the original bug.
- The gate is `isLocalInterfaceAddress(hostAddress)`. Only Apple returns a numeric interface IP; Docker/Lima/WSL2 return `host.docker.internal` and Podman returns `host.containers.internal`, so all four return before the provider is consulted or the network is touched. That is what keeps them byte-identical.
- Only a definitive `ECONNREFUSED` fails a start. Timeouts, DNS and permission errors are `unknown` and never block. Same rule the CDP proxy probe follows in `chrome-provider.ts`.
- Apple still fails closed on an unknown gateway. The check deliberately swallows that error so the identical throw surfaces from the env build a few lines later, where it carries the right context. There is a test pinning this.
- `agent-banners.tsx` gains `whitespace-pre-line` and top-aligns its icon, so multi-line remediation renders as written. Single-line start errors are unaffected. Without it the message collapsed into one unreadable `text-xs` run.
- Podman now uses `host.containers.internal` (via the shared `getContainerHostUrl` helper) instead of the old hardcoded Docker name, matching the address it already uses for host-API talk-back.
- `isLocalInterfaceAddress` is duplicated from `chrome-provider.ts` rather than extracted. SUP-459 consolidates host-address resolution and is stacked on this branch; extracting now would collide with it.

## Known gap, not addressed here

Lima and WSL2 map `host.docker.internal` to a **real** bridge IP via `--add-host`, so they have the same loopback-bind problem as Apple. They are out of scope here to keep this branch byte-identical for them and the SUP-459 rebase clean. Worth a follow-up once SUP-459 lands the unified resolver, since the gate would then be one predicate instead of four.

### DIFF
--- a/src/renderer/components/layout/agent-banners.tsx
+++ b/src/renderer/components/layout/agent-banners.tsx
@@ -55,9 +55,15 @@ export function AgentBanners({ slug, startAgent }: AgentBannersProps) {
       {/* Start error banner */}
       {startAgent.isError && (
         <div className="shrink-0 border-b bg-destructive/10 px-4 py-2">
-          <div className="flex items-center gap-2 text-xs text-destructive select-text">
-            <AlertCircle className="h-3 w-3 shrink-0" />
-            <span>Failed to start agent: {startAgent.error.message}</span>
+          {/* Start failures may carry multi-line remediation steps (e.g. an
+              unreachable local model naming the bind to change), so keep the
+              newlines and top-align the icon against the first line. Single-line
+              messages are unaffected. */}
+          <div className="flex items-start gap-2 text-xs text-destructive select-text">
+            <AlertCircle className="mt-0.5 h-3 w-3 shrink-0" />
+            <span className="whitespace-pre-line">
+              Failed to start agent: {startAgent.error.message}
+            </span>
           </div>
         </div>
       )}
--- a/src/shared/lib/container/apple-container-client.test.ts
+++ b/src/shared/lib/container/apple-container-client.test.ts
@@ -168,7 +168,8 @@ describe('AppleContainerClient host talk-back (host.docker.internal is NXDOMAIN 
     const client = new AppleContainerClient({ agentId: 'abc123' })
     expect(client.getHostApiBaseUrl()).toBe('http://192.168.64.1:47891')
     expect(client.getHostBridgeIp()).toBe('192.168.64.1')
-    // Cached: one inspect for both calls.
+    expect(client.getContainerHostAddress()).toBe('192.168.64.1')
+    // Cached: one inspect for all three calls.
     expect(mockExecSyncWithPath).toHaveBeenCalledOnce()
   })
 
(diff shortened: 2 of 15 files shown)


JeremyBischoff:
Closing. Local-model support on the Apple Container runtime is out of scope while we focus on cloud environments. The cloud runtimes (Kubernetes, Lambda MicroVM) reach the host through `HOST_PUBLIC_URL` and never hit this bug. The branch stays on GitHub if this comes back.

PR #541, opened 2026-07-22
Apply the both-shelves awaiting-clear rule at every resolution site (SUP-451)

JeremyBischoff:
## What

Apply the both-shelves awaiting-clear rule at three remaining resolution sites, so the "needs input" pill turns off exactly when the human's turn ends - not while the machine is still working.

**Stacked on #510** (base `fix/subagent-input-surfacing`): reuses its `hasBlockingPendingRequests` helper, so it can't build off `main` alone. Retarget to `main` once #510 lands. Linear: SUP-451.

## The bug class

`isAwaitingInput` should be true exactly when a human must act. #510 introduced `hasBlockingPendingRequests` (both request shelves - pending input asks + pending computer-use, skipping auto-approved) and used it at the sidechain clear. Three other resolution sites still cleared on the wrong condition:

| Site | Was | Now |
|---|---|---|
| capability-review approve (`completeCapabilityReview`) | never cleared - waited for the Task's tool_result, minutes later | clears on approval, both-shelves rule |
| script_run approve (`clearPendingScriptRun`, new) | never cleared - waited for the script's tool_result after a ≤30s host exec | clears on approval, before the script runs |
| main-path tool_result | cleared unconditionally on ANY tool_result | delete resolved id first, then clear only if no blocking request remains |

## Why each matters

- **capability-review / script_run**: after you approve, the wait belongs to the machine (a Task running for minutes / a host script for ≤30s), but the session stayed orange "needs input" the whole time - tray, app menu, sidebar, home, `/api/agents`.
- **main-path over-clear**: with a background subagent parked on the user, its ask lives on the parent session's shelves while the main agent keeps working. The main agent finishing an ordinary tool cleared the pill even though the subagent still needed you. Found by an adversarial review pass over the branch.

All five awaiting-clear sites now share the one `hasBlockingPendingRequests` rule.

## Verification

- Three repro tests, each red-before / green-after, driving the real stream + resolution paths:
  - `awaiting-input-subagent-review.repro.test.ts`
  - `awaiting-input-script-run.repro.test.ts`
  - `awaiting-input-background-subagent.repro.test.ts`
- Full `message-persister.test.ts` suite green (267); typecheck + lint clean.

## Footprint

5 files, +307/-5 (2 source, 3 repros).

### DIFF
--- a/src/api/routes/agents.ts
+++ b/src/api/routes/agents.ts
@@ -2669,6 +2669,11 @@ agents.post('/:id/sessions/:sessionId/run-script', AgentUser(), async (c) => {
       return c.json({ error: `Script type "${scriptType}" is not supported on ${platform}` }, 400)
     }
 
+    // The user approved — the host now runs the script (up to 30s below). The wait has
+    // flipped from the human to the machine, so clear the awaiting light now instead of
+    // leaving the session stuck "needs input" until the script's tool_result lands.
+    messagePersister.clearPendingScriptRun(c.req.param('sessionId'), toolUseId)
+
     // Execute the script with a 30s timeout
     const { exec, execFile } = await import('child_process')
     const { promisify } = await import('util')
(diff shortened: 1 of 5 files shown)


JeremyBischoff:
Closing. Base #510 closed unmerged and SUP-424 was solved another way. SUP-451 stays open.

PR #547, opened 2026-07-22
feat: give the agent desktop vs web environment awareness (SUP-455)

JeremyBischoff:
## What

Gives the agent a host-resolved fact about its runtime surface (native desktop app vs browser-reached web) and injects it into the system prompt, so the agent's own free-form prose is surface-correct. Today the prompt assumes desktop unconditionally, so a web-hosted agent tells the user to "open the app on your machine" when there is no desktop app to open.

The surface fact's type is exported so any later host-side consumer that needs the surface (a deep-link or chat-message builder, say) reuses one resolver and one type rather than re-resolving it - but nothing depends on that yet, and this change stands on its own.

## How it works

Extends the existing host-fact -> gated-prompt rail (the same shape as `HOST_PLATFORM` -> `computerUse`), one parallel hop:

- **Host** (`container-manager`) resolves the surface once at container start via `resolveAgentEnvironment()`: `process.type === 'browser'` is the Electron main process (desktop), otherwise web plus an optional public URL. It is stringified into a reserved `AGENT_ENVIRONMENT` env var that custom env cannot clobber.
- **Container** (`claude-code`) reads it, Zod-parses against a lenient mirror schema, and falls back to `desktop` on absent or malformed input (never throws). One gated block in `system-prompt.md` renders the surface-correct line.

Design choices worth knowing:

- Two surfaces (`desktop` / `web`) with an optional `publicUrl`, not a three-way split. No consumer distinguishes self-hosted web from hosted cloud.
- `publicUrl` is optional so a self-hosted web deploy with no public URL still renders as web rather than silently falling back to desktop.
- Host schema is strict, container mirror is lenient, so a newer host image that adds a field is not rejected mid-rollout.
- `computerUse` stays a separate axis. Desktop-Linux is desktop-yet-not-computer-use, so surface and computer-use capability are genuinely orthogonal.

## Verified

- Unit tests: resolver (both arms + fallback + illegal-state rejection), container parse (desktop / web-url / web-no-url / absent / malformed, never throws), prompt render for both surfaces (correct block, no leaked template tags), reserved-key guard.
- Rendered the actual prompt block for every surface and read the copy end to end.
- typecheck + lint green in both packages (host and agent-container).

Note for reviewers: the agent-container prompt is baked into the ghcr image, so a live agent reflects this only after an image rebuild + container recreation.

### DIFF
--- a/agent-container/src/claude-code.ts
+++ b/agent-container/src/claude-code.ts
@@ -128,6 +128,29 @@ function parseConnectedAccounts(): Map<string, Array<{ name: string; id: string 
   return accounts;
 }
 
+// Wire mirror of the host's AgentEnvironment (src/shared/lib/container/agent-environment.ts).
+// agent-container is an isolated package, so it re-declares the shape it expects from the
+// AGENT_ENVIRONMENT env var rather than importing the host type - same contract pattern as
+// parseRemoteMcps / parseConnectedAccounts. Kept LENIENT (no .strict()) on purpose: during a
+// rollout the host image can be newer than this container image and send an added field; a
+// strict parse would reject it and wrongly fall back to desktop. Extra keys are harmless here.
+const agentEnvironmentSchema = z.discriminatedUnion('surface', [
+  z.object({ surface: z.literal('desktop') }),
+  z.object({ surface: z.literal('web'), publicUrl: z.string().min(1).optional() }),
+]);
+type AgentEnvironment = z.infer<typeof agentEnvironmentSchema>;
+
+function parseAgentEnvironment(): AgentEnvironment {
+  const raw = process.env.AGENT_ENVIRONMENT;
+  if (!raw) return { surface: 'desktop' };
+  try {
+    const parsed = agentEnvironmentSchema.safeParse(JSON.parse(raw));
+    return parsed.success ? parsed.data : { surface: 'desktop' };
+  } catch {
+    return { surface: 'desktop' };
+  }
+}
+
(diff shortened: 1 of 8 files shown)


iddogino:
I'm not sure I understand how this PR relates to the linear issue?
The linear ticket is purely for the chat integration (when forwarding the request message, attach a link). No container involvement

The risk with adding this into the system prompt is that it can confuse the agent ("I am running in the browser? Guess I can't use bash then...", "The prompt says I am running in a Windows app, but when I try to run poweshell commands in my bash tools they fail and the file system looks like linux").

Remember - agents run in a Linux Container always, regardless of where the App runs. We had a whole class of these bugs when prompting around the Computer Use feature (b/c there the agent can use host shell).

JeremyBischoff:
I think I misinterpreted the first part of the ticket "it said Desktop → this is actually running in the cloud - it should know that" and inferred that this meant the agent should be aware of its environment so it can word things and link correctly - and then split it out based on this. The risk makes sense and we can fix the original bug/ticket without this change anyways since the message is host-emitted.

JeremyBischoff:
Closing.

The premise did not survive review: injecting a desktop-vs-web fact into the system prompt risks the agent reasoning from it wrongly, and the original ticket is host-emitted so it needs no container involvement at all. SUP-453 shipped that way in #563 (merged 2026-07-25) without any of this.

Nothing here is salvageable as a smaller change - all 154 lines exist to carry that one fact into the prompt, so there is no remainder once the premise goes. Not reopening in another shape.

PR #549, opened 2026-07-23
feat(scheduler): classifier cron storage + guards (SUP-368)

JeremyBischoff:
Storage + guards groundwork for classifier crons (SUP-368), ahead of the fire path that wires them. Nothing fires, classifies, or escalates yet: this PR lands the data model, the config validation, and the guards so the fire-path PR can be pure logic.

**On the diff size:** GitHub shows ~+3,300, but ~2,560 of that is a single generated file - the Drizzle migration snapshot (`meta/0032_snapshot.json`, a full-schema dump, not hand-edited). The reviewable change is ~770 source+test lines across 1 commit. A `.gitattributes` marks the snapshots generated; it won't shrink this PR's count (GitHub reads attributes from `main`) but will for future migration PRs.

## The concept this prepares for

```
Today  every scheduled cron → spawn a full agent session (expensive, always)

Goal   classifier cron → cheap model judges the situation
                              ├─ settle   → do the easy version cheaply, stop
                              └─ escalate → spawn the big-model session
```

Most scheduled runs are "nothing to do here." A small model (Haiku) makes that call for pennies instead of spawning Opus every time. This PR ships none of the judging - only the place to store such a cron and the guards that keep it inert until the fire path exists.

## Two layers, only one is live

```
LIVE / WIRED
  createScheduledTask({ executionMode:'classifier', config })
       Zod-validates config at the write boundary
       stores a scheduled_tasks row (execution_mode='classifier', classifier_config=JSON)
       agents can't see it, can't cancel/pause it, can't run-now it

GROUNDWORK / PRESENT BUT UNWIRED
  classifier_runs table + classifier-run-service
  the durable per-fire state machine the fire PR will drive
  shipped now so the fire PR doesn't stack a migration on new logic
```

## Data model (migration 0032)

On the existing `scheduled_tasks` table:
```
execution_mode    'session' | 'classifier'   NOT NULL default 'session'
classifier_config text (JSON) | null         null for session rows
```
`default 'session'` is the backfill: every existing cron keeps today's behavior with no data change.

New `classifier_runs` table, one row per fire of a classifier cron:
```
id                 PK
scheduled_task_id ┐
fire_at           ┘ UNIQUE together     fire identity: dedup + crash recovery
status             classifying | resolved   (indexed)
classify_session_id / verdict / reason / escalate_session_id
deadline_at        stall watchdog cutoff
resolved_at / created_at
```
The `(scheduled_task_id, fire_at)` unique index is load-bearing: two scheduler ticks firing the same cron at the same instant can't create two runs.

## Creating a classifier cron (wired)

```
createScheduledTask({ executionMode, classifierConfig })
   ├─ 'session' (default)     store as today, classifier_config = null
   └─ 'classifier'
         validateClassifierConfig(config):
           - criteria non-empty
           - escalateModel / escalateEffort present
           - gather.sources EMPTY  (PR1 rejects any non-empty gather)
           - classifyModel/Effort default to haiku/low
           invalid → THROW, nothing stored
           valid   → JSON.stringify → store in classifier_config
```

Config is validated on the way into the DB (`validate*` throws) and on the way back out (`parse*` returns null and logs). The discriminated union on `executionMode` keeps a session row and a classifier row two distinct shapes validated the same way.

## The guards (wired - the safety of this PR)

Invariant during soak: agents cannot see or touch classifier rows.

```
isAgentVisibleScheduledTask(task) = parseExecutionMode(task.execution_mode) === 'session'

agent schedule tools:
   list schedules    .filter(isAgentVisibleScheduledTask)   classifier hidden
   schedule summary  .filter(isAgentVisibleScheduledTask)   classifier hidden
   cancel by id      reject unless visible                  "not found"
   pause / resume    reject unless visible                  "not found"

manual fire:
   POST /:taskId/run-now   execution_mode==='classifier' → 400
```

A classifier row lives in the same table agents already manage, so without these filters an agent could list, cancel, or hand-fire a cron it was never meant to see.

## classifier_runs lifecycle (groundwork - nothing calls it yet)

The state machine the fire PR drives. Shipped now to lock the shape.

```
createClassifierRun(task, fireAt)          idempotent claim
   row exists for (task,fireAt)? → return it        (crash-safe re-claim)
   else INSERT; unique race? → catch, re-read, return winner
      status = 'classifying'
         setClassifySessionId / storeClassifierVerdict (pre-escalate durability)
      markClassifierRunResolved → status = 'resolved'
   (all updates guarded on status='classifying')
```

## Not in this PR

```
- the fire loop        classifier crons don't run yet
- the seed script      no way to materialize a schedulable classifier row outside tests
- gather runner        sources must be empty (kinds forward-declared for SUP-456)
- agent authoring      agents create session crons only; classifier is privileged
- classify / escalate  no model calls, no session spawning
```

The seed was intentionally left out: the scheduler fire loop (`getDueTasks` / `executeTaskInner`) has no `execution_mode` filter yet, so a seeded classifier cron would auto-fire as a normal session at its next cron time. The fire PR owns that scheduler guard and the create surface together.

## How verified

- Unit + route + persister tests: config validation at the write boundary, the run-now 400 guard, the agent-visibility gate, the idempotent per-fire claim.
- Migration 0032 applied against a seeded copy of a pre-existing `scheduled_tasks` table: existing rows survive with `execution_mode='session'`, and the unique `(task, fire_at)` constraint fires.
- typecheck + lint clean.


JeremyBischoff:
Closing. Parked under SUP-368 with the fire path and model cascade work. Rebuild the storage on current main when classifier crons are picked up.

PR #553, opened 2026-07-23
fix: skip host auto-sleep scans for Lambda MicroVM (reapply of #538)

iddogino:
Re-opens the changes from #538 (authored by @yiw190), which was merged in error and reverted in #552. This PR restores the exact same change for proper review.

> **Note:** the "Files changed" diff will appear empty until revert PR #552 is merged into `main`; it will then show the full change set.

Original PR description follows.

---

## Bug

For Lambda MicroVM agents, the host `AutoSleepMonitor` still walks every running agent once a minute and calls `listSessions()` — session metadata lives on the S3 Files-backed `/data` mount (network filesystem I/O). That scan is wasted work: background auto-sleep already ends in `stop({ escalateToForceStop: false })`, which is a no-op on this runtime because AWS `idlePolicy` performs suspend/resume. So we pay network I/O every minute and reclaim nothing.

Staging already showed how expensive sync work on that mount is: SQLite on `/data` blocked Node for 5.268s and tripped the same 4s ECS health-probe `AbortError` at only ~10.5% CPU. This PR removes one redundant source of `/data` pressure; it does not fix broader SQLite isolation.

## Reproduction

1. Run host-app with `containerRunner=lambda-microvm` and one or more running agents.
2. Wait for the one-minute `AutoSleepMonitor` cycle.
3. Observe `listSessions()` (and related `/data` reads) for each agent even though the eventual MicroVM auto-sleep `stop()` does not suspend or terminate.

## What changed

- Add polymorphic `shouldRunHostAutoSleep()` (default `true` for container runtimes).
- Lambda MicroVM returns `false` so the monitor skips session / keep-alive scans when they cannot reclaim anything.
- Gate that check **before** active-session checks or session filesystem reads.
- Tests: default runtimes stay on host sweep; MicroVM opts out; monitor path does no session I/O when opted out.

## Why this shape (not "host must never Suspend")

The goal of this PR is **less session / `/data` I/O on S3 Files**, not a claim that host-driven `Suspend` is impossible. Auto-resume on the next inbound request works either way. We skip the host sweep here because, with today's no-op auto-sleep `stop()`, the scan is pure cost.

## Test plan

- [x] Focused tests: 142 passed
- [x] `npm run lint`: 0 errors (40 existing warnings)
- [ ] `npm run typecheck`: existing missing optional `e2b` / `@daytonaio/sdk` tool dependencies
- [ ] Full `npm run test:run`: 7,516 passed; one unrelated existing failure in `webhook-trigger-service.sup226.test.ts`

https://claude.ai/code/session_011KXR4aYGTLAmTSd4cn2Dkt

### DIFF
--- a/src/shared/lib/container/base-container-client.test.ts
+++ b/src/shared/lib/container/base-container-client.test.ts
@@ -31,6 +31,12 @@ class TestContainerClient extends BaseContainerClient {
   }
 }
 
+describe('host auto-sleep', () => {
+  it('is enabled by default for container runtimes', () => {
+    expect(new TestContainerClient({ agentId: 'a' }).shouldRunHostAutoSleep()).toBe(true)
+  })
+})
+
 describe('buildAgentEnv', () => {
   afterEach(() => enableToolSearch.mockReturnValue(true))
 
--- a/src/shared/lib/container/base-container-client.test.ts
+++ b/src/shared/lib/container/base-container-client.test.ts
@@ -31,12 +31,6 @@ class TestContainerClient extends BaseContainerClient {
   }
 }
 
-describe('host auto-sleep', () => {
-  it('is enabled by default for container runtimes', () => {
-    expect(new TestContainerClient({ agentId: 'a' }).shouldRunHostAutoSleep()).toBe(true)
-  })
-})
-
 describe('buildAgentEnv', () => {
   afterEach(() => enableToolSearch.mockReturnValue(true))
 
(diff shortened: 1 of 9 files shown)


yiw190:
Superseded by https://github.com/SkillfulAgents/SuperAgent/pull/568.

Instead of skipping AutoSleepMonitor for MicroVM, #568 makes host AutoSleep cheap (in-memory activity, no S3 Files `listSessions`) and restores host-owned suspend. AWS `idlePolicy` stays as orphan GC only.

PR #564, opened 2026-07-23
fix(chat): stop Telegram and Slack working indicators stranding at turn end

JeremyBischoff:
8 files, +392/-27. First of five splits out of #405, which is being closed in favor of contained PRs.

## What

Two chat surfaces strand a "Working…" indicator permanently at turn end. Both are fixed here; neither touches session state, the container protocol, or the stream layer.

**Telegram draft.** The working-indicator draft was only cleared authoritatively on session removal, so a turn that ended without the session being torn down left "Working…" showing until the ~30s draft expiry. `stopWorking` now blanks the un-committed draft in place. A `yieldingToStream` flag on `stopWorking`/`clearIndicator` skips the blank on the first-reply-token clear, since the streamed reply reuses that same draft and a hard teardown would wipe it. Independent surfaces (Slack reaction, iMessage typing) ignore the flag and always clear.

**Slack reaction.** A Slack indicator is a reaction, which never expires, so any teardown that loses track of a live one strands it forever. Three fixes: only untrack a reaction once removal is confirmed gone, so a transient `reactions.remove` failure is retried rather than dropped; track the reaction before the add call, so an add that lands server-side but errors client-side is still swept by teardown; serialize add/remove per chat so a paint from one indicator tick cannot land after the clear from the next and re-add the reaction.

## Why this is its own PR

#405 bundled these with a container wire-protocol change behind an image rebuild. They cherry-picked out of it with zero conflicts, which is the evidence they were never coupled: no shared files with the stream work, no shared failure mode.

## Testing

- `src/shared/lib/chat-integrations/`: 663 passed, 7 skipped
- `tsc --noEmit` clean, eslint 0 errors
- Coverage added: manager-level `yieldingToStream` wiring in both regression directions (the first-token clear must yield the surface to the stream, terminal clears must not - previously invisible to every test); a Slack sweep over two tracked reactions dropping only the confirmed-gone one while retrying the transient failure; the per-key serializer resolving a value after a prior rejection on the same key; a failed Telegram blank-draft teardown being swallowed rather than retried.

Not live-smoked against real Telegram/Slack in this pass - the changes are unit-covered and were CI-green as part of #405, but a reviewer should know the manual pass hasn't been re-run since the split.

### DIFF
--- a/src/shared/lib/chat-integrations/base-connector.ts
+++ b/src/shared/lib/chat-integrations/base-connector.ts
@@ -171,8 +171,14 @@ export abstract class ChatClientConnector {
    * Stop the working indicator as the response takes over. Idempotent — safe to
    * call repeatedly. Default no-op for connectors whose indicator is ephemeral
    * and self-expires.
+   *
+   * `yieldingToStream` marks the clear that fires on the first reply token: a
+   * connector whose indicator SHARES the reply surface (Telegram streams the reply
+   * into the same draft the placeholder used) must leave that surface for the
+   * stream rather than tear it down. Connectors with an independent surface (a
+   * Slack reaction, an iMessage typing bubble) ignore the flag and always clear.
    */
-  async stopWorking(_chatId: string): Promise<void> {}
+  async stopWorking(_chatId: string, _opts?: { yieldingToStream?: boolean }): Promise<void> {}
 
   /**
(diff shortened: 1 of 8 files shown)


JeremyBischoff:
Closing. Parked under SUP-486 with the other Telegram indicator work. The Slack reaction fixes are written up there too.

PR #565, opened 2026-07-23
fix(chat): clear session state when a turn-start send fails

JeremyBischoff:
4 files, +80/-1. Second of five splits out of #405.

## What

`markSessionActive` ran before a throwable container send on the app send route and the x-agent invoke route, with no cleanup on failure. Since #339 removed every time-based backstop, a failed send left `isActive` stuck true - a permanent "Working…" on the chat indicator and the desktop sidebar, with nothing running that could ever clear it.

Both routes now clear via `markSessionInterrupted` when the send throws. Host-only: nothing is running in the container to interrupt. The app route guards on `wasQueued` so it never settles an in-flight turn.

## Why this is its own PR

Smallest independent slice of #405 and the only one confined to the API routes. It shares no file with the other four splits and cherry-picked out with zero conflicts.

## Testing

- `src/api/routes/agents.test.ts` + `x-agent.test.ts`: 276 passed
- `tsc --noEmit` clean, eslint 0 errors
- Both routes covered for the throw path: send fails, session settles rather than pinning.

Not live-smoked since the split; unit-covered and was CI-green as part of #405.

### DIFF
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -123,6 +123,7 @@ vi.mock('@shared/lib/container/message-persister', () => ({
     subscribeToSession: vi.fn(),
     unsubscribeFromSession: vi.fn(),
     markSessionActive: vi.fn(),
+    markSessionInterrupted: vi.fn(),
     cancelAwaitingInput: vi.fn(),
     broadcastSessionEvent: vi.fn(),
   },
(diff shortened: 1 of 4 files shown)


JeremyBischoff:
Closing to redo against current main. The send-failure cleanup is still missing there.

PR #566, opened 2026-07-23
Shadow registry for pending user-input requests (Phase 2)

iddogino:
Phase 2 of the user-input request rearchitecture (follows #560 characterization and #562 unified dispatch). Introduces the typed pending-request model and `UserInputRequestManager` as a **write-through shadow** of the existing stores: the legacy shelves stay authoritative, every mutation mirrors into the registry, and **nothing reads it for behavior yet** — zero behavior change by construction, enforced by assertions rather than by promise.

## What's new

**`src/shared/lib/user-input/request-schema.ts`** — one typed envelope for "the agent is blocked on a human": a Zod discriminated union over all 11 kinds (question, secret, connected_account, file, remote_mcp, browser_input, script_run, capability_review, computer_use, proxy_review, x_agent_review) with `id`, `scope {agentSlug, sessionId?}` (no sessionId ⇒ agent-scoped, i.e. reviews), `blocking`, `autoApproved`, and a per-kind payload. The envelope is strict; payloads are deliberately lenient (`.catch`) so a malformed tool input can never make shadow registration break a real delivery path.

**`src/shared/lib/user-input/request-manager.ts`** — the registry:
- `register` is first-delivery-wins (stream stop and complete-assistant can both carry the same tool_use — same rule the Phase 1 dispatcher enforces).
- `resolve` is idempotent and outcome-tagged (`answered | declined | cancelled | superseded | timeout | invalidated`).
- Resolves are **shelf-scoped** where the legacy delete is: a main-path `tool_result` deletes blindly from the stream shelf today, so `resolveIfShelf` refuses to let it evict a computer-use or review entry whose own store still holds it.
- `isSessionAwaiting` / `isAgentAwaiting` are the derived projections (any open blocking non-auto-approved request in scope; agent-scoped reviews block every session of the agent). Phase 3 promotes these to authoritative; this PR only compares them.

## Write-through wiring

- **Persister (8 hook points)** — nearly everything funnels through the single `broadcastToSSE` interceptor (set on the 8 request event types, clear on turn boundaries), plus the direct-delete doors: `completeCapabilityReview`, `clearPendingComputerUseRequest`, `cancelAwaitingInput`, `handleComputerUseRequestTool`, `resolveSidechainInputRequests`, `handleToolResults`, and `unsubscribeFromSession` (session death invalidates its entries).
- **ReviewManager (all 9 entry/exit paths)** — reviews register as agent-scoped envelopes (`proxy_review` / `x_agent_review` by `details.xAgent`); every exit (decision, sweeps, deny-all, 5-min timeout, abort, rejectAll) settles the envelope with a faithful outcome.

## Equivalence checks (the point of the phase)

- **Strict shelf parity**: at every mutation point (dev/test only, skipped in production) the registry's per-shelf view must equal the legacy shelf exactly — throws under vitest, logs in dev.
- **Awaiting projection vs the imperative bit**: compared a microtask after each mutation (the imperative mark/clear lands later in the same synchronous stretch), warn-once-per-episode + counter, never throws — because Phase 0 proved the bit is genuinely wrong in pinned cases, and this counter is the telemetry that sizes the Phase 3 flip.

## Validation

- `tsc --noEmit` and eslint clean.
- **Full unit suite: 7,718 passed / 0 failed** — with the strict parity assertion live under every existing persister/review test (352 in the directly affected suites, unmodified).
- ~30 new tests: manager unit suite (dedupe, shelf-scoped resolves, projections, diagnostics), an equivalence describe in the lifecycle characterization suite (per-kind mirror + the split-brain pins showing projection right where the bit is wrong), and a ReviewManager write-through describe (all exits → correct outcomes).
- **E2E: 40/40** across the seven request-flow specs (`user-input-requests`, `mixed-pending-requests`, `proxy-review`, `proxy-review-xagent`, `capability-review`, `awaiting-input-status`, `pending-request-reconnect`), with the dev server logging **zero shelf mismatches** and exactly **6 awaiting-divergence episodes — every one `bit=false, projection=true`**: the known split-brains observed live, with the registry on the correct side each time.

## Next (not in this PR)

Phase 3 flips awaiting to the derived projection and deletes `registerAwaitingBlockerSource`; the persister's awaiting-status test matrix is the acceptance suite.

### DIFF
--- a/src/api/routes/agents.ts
+++ b/src/api/routes/agents.ts
@@ -2679,7 +2679,7 @@ agents.post('/:id/sessions/:sessionId/capability-review', AgentUser(), async (c)
         return c.json({ error: 'Failed to reject capability launch' }, 500)
       }
 
-      messagePersister.completeCapabilityReview(sessionId, toolUseId)
+      messagePersister.completeCapabilityReview(sessionId, toolUseId, 'declined')
       trackServerEvent('request_declined', { type: 'capability_review', capability, withReason: !!declineReason })
       return c.json({ success: true, declined: true })
     }
@@ -2710,7 +2710,7 @@ agents.post('/:id/sessions/:sessionId/capability-review', AgentUser(), async (c)
     if (scope === 'session') {
       messagePersister.grantSessionCapability(sessionId, capability)
     }
-    messagePersister.completeCapabilityReview(sessionId, toolUseId)
+    messagePersister.completeCapabilityReview(sessionId, toolUseId, 'answered')
 
     trackServerEvent('capability_launch_approved', { capability, scope })
     return c.json({ success: true })
@@ -2976,7 +2976,7 @@ agents.post('/:id/sessions/:sessionId/computer-use', AgentUser(), async (c) => {
         return c.json({ error: 'Failed to reject computer use request' }, 500)
       }
 
-      messagePersister.clearPendingComputerUseRequest(sessionId, toolUseId)
+      messagePersister.clearPendingComputerUseRequest(sessionId, toolUseId, 'declined')
       trackServerEvent('request_declined', { type: 'computer_use', method, withReason: !!declineReason })
       return c.json({ success: true, declined: true })
     }
@@ -2999,7 +2999,7 @@ agents.post('/:id/sessions/:sessionId/computer-use', AgentUser(), async (c) => {
       if (!resolveResponse.ok) {
         return c.json({ error: 'Failed to resolve computer use request' }, 500)
       }
-      messagePersister.clearPendingComputerUseRequest(sessionId, toolUseId)
+      messagePersister.clearPendingComputerUseRequest(sessionId, toolUseId, 'answered')
       return c.json({ success: true })
     }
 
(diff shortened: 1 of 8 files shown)


iddogino:
Both review items addressed in 1bde42fe:

**Real outcomes at the clear doors** — `completeCapabilityReview` and `clearPendingComputerUseRequest` now accept the caller's outcome (defaulting to `answered` for the approve path). All six route call sites and the container's `capability_review_cancelled` frame pass explicit outcomes: `declined` on the decline routes, `cancelled` for the abandoned-hook frame, `invalidated` when computer-use execution fails after approval, `answered` on success. Regression tests pin the non-success paths (`declined` capability review, `declined` + `invalidated` computer-use).

**Review-shelf parity** — the strict parity assertion now covers the third store: after every `ReviewManager.pending` mutation (all 9 entry/exit paths, including the timeout/abort closures), the registry's agent-scoped review view is compared against the pending store — throws under vitest, logs in dev. Unit tests cover both the match case (session-scoped noise excluded) and the missed-write-through mismatch case; the existing 38 ReviewManager tests plus the write-through describe now run under the assertion.

Validation: tsc + lint clean, full unit suite 7,721 green, review-flow E2E green with zero shelf mismatches. (One capability-review E2E failure during validation reproduced hopping between unrelated tests under parallel contention and passes 6/6 on clean runs — pre-existing flake, not related: the failing tests don't create reviews.)

PR #572, opened 2026-07-24
Fix stuck Working after lost settle signals (SUP-468)

JeremyBischoff:
Closes https://linear.app/datawizz/issue/SUP-468/fix-stuck-working-settle-signal-loss-class-liveness-replay-backstop

## Why

Telegram shows Working while the host considers a session active.

A normal turn ends with a result followed by an idle signal.
If the idle signal is lost because the socket attached late, disconnected, or the query process died, the reply can be complete while the host remains active forever.
Reconnect recovery could also replay the previous turn after a successor turn had started, or preserve background tasks from a dead process.

## What

Normal and fallback settlement now follow the same rule:

```text
normal:     running → result → idle → settled
lost idle:  running → result → 5s grace → clean gate → settled
                                         └─ blocked → stay active
```

Post-result activity prevents an old result from settling new work:

```text
post-result frame
├─ output only        → cancel grace, keep result for a later real idle
├─ running/requesting → cancel grace, forget the previous result
└─ no activity        → settle after grace only when every blocker is clear
```

### Cases fixed

1. **Idle is lost after a result**
   - The host starts a single five-second grace period after observing the result.
   - When it expires, the host settles only if no background task, input request, pending blocking request, or external review is open.

2. **The subscriber attaches after an instant turn already finished**
   - The container replays the latest result, authoritative background-task snapshot, and idle state.
   - Replayed frames are marked so a host that already saw the live frames does not process them twice.

3. **The query exits abnormally**
   - The container emits a synthetic error result followed by idle, and records both for late-join replay.
   - New-turn evidence resets prior error suppression, so consecutive failed turns each receive their own terminal frames.

4. **A dead query leaves background tasks behind**
   - Starting a replacement query emits an empty task snapshot.
   - The host cannot remain pinned by task IDs owned by the dead process.

5. **A successor turn starts before a reconnect**
   - The container replays running or requesting evidence for the successor turn instead of the previous turn's result and idle.
   - The host keeps the live turn active.

## Safety rules

- Never infer completion without first observing a result.
- Never settle while background work, user input, a pending blocking request, or external review is open.
- Any output cancels the fallback timer.
- Running or requesting proves a new turn and invalidates the previous result.
- Completion state and notifications finalize once.

## Scope

- 4 files
- Implementation: +185 / -23 lines
- Regression tests: +510 lines

## Verification

- [x] Root TypeScript typecheck
- [x] Agent-container TypeScript typecheck
- [x] Lint with zero errors
- [x] Root unit suite: 7,741 passed
- [x] Agent-container unit suite: 746 passed
- [x] Telegram normal completion: reply delivered and Working cleared
- [x] Telegram forced container stop: Working cleared within one second, with no leaked `done` output

### DIFF
--- a/agent-container/src/claude-code.ts
+++ b/agent-container/src/claude-code.ts
@@ -1,6 +1,6 @@
 import { query, Query, SDKMessage, SDKUserMessage } from '@anthropic-ai/claude-agent-sdk';
 import { z } from 'zod';
-import type { UUID } from 'crypto';
+import { randomUUID, type UUID } from 'crypto';
 import { EventEmitter } from 'events';
 import * as fs from 'fs';
 import * as path from 'path';
@@ -29,6 +29,14 @@ import { fileHooks, resolveToolFilePath } from './file-hooks';
 type QueryWithAsyncCancel = Query & {
   cancelAsyncMessage(messageUuid: string): Promise<boolean>;
 };
+
+function isNewTurnEvidence(message: SDKMessage): boolean {
+  if (message.type !== 'system' || !('subtype' in message)) return false;
+  if (message.subtype === 'session_state_changed') {
+    return 'state' in message && message.state === 'running';
+  }
+  return message.subtype === 'status' && 'status' in message && message.status === 'requesting';
+}
 
 /** Generate prefixed MCP tool names from a tools array, optionally excluding some by name. */
 function mcpToolNames(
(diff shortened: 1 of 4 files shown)


JeremyBischoff:
Superseded by #627.

#572 covered only the half-open-after-result case via a result-armed grace and no longer typechecks against main. #627 covers that case plus discarded idle on reconnect, silence reattach when result and idle are both lost, late-join background snapshot replay, and the muted-socket ownership races around reconnect.

PR #578, opened 2026-07-24
Derive session awaiting-input from the request registry (Phase 3)

iddogino:
Phase 3 of the user-input request rearchitecture (follows #560, #562, #566): the session "awaiting input" status stops being an imperative per-path bit and becomes a **derived projection** of the `userInputRequestManager` registry.

## What changed

- **One flip point.** `state.isAwaitingInput` is now an edge-detection cache of `isSessionAwaiting(sessionId, agentSlug)`, gated on an active turn. A new private `syncSessionAwaiting` recomputes it after every registry transition and broadcasts on the edges (`session_awaiting_input` + automated-session promotion on rise, `session_input_provided` on fall). Turn-boundary teardowns still reset silently alongside `isActive`, exactly as before.
- **Deleted the imperative machinery:** `markSessionAwaitingInput`/`markAwaitingInput`, `clearAwaitingInputForAgentIfUnblocked`, `hasBlockingPendingRequests`, the `registerAwaitingBlockerSource` predicate set (the #528 hack), the now-tautological `compareAwaitingProjection` divergence telemetry, and both per-session `|| (isActive && hasAgentLevelReviews)` hacks in agents.ts (the agent-list one stays — it covers dashboard reviews with no session).
- **ReviewManager** now calls `messagePersister.syncAgentSessionsAwaiting(agentSlug)` after registering/settling a review — the agent-scoped registry entry is what blocks the agent's sessions. A session that starts *while* a review is parked now reads as awaiting immediately (previously only sessions active at request time were marked).
- **Split-brains fixed** (the Phase 0 pins now assert the correct behavior): answering one of several parallel requests keeps the waiting light on; `completeCapabilityReview`/`clearPendingComputerUseRequest` flip the bit and broadcast together; a stray tool_result on a parked computer-use approval no longer drops the light; an agent-wide sync cannot clear awaiting under a live review.
- **`session_active` supersedes stale computer-use entries** (partial SUP-213). Required here: the projection would read a pre-idle leftover approval as a live wait on the fresh turn. `session_idle` still keeps them for SSE replay of a genuinely parked card.
- **`recoverSessionAwaitingInput` changed signature** — with no forced bit, recovery re-establishes the actual missed requests (store + registry, marked `recovered` and excluded from SSE replay since they carry no payload; clients render the card from the transcript). agents.ts passes the unresolved tool calls it already scans for.

## Validation

- Full unit suite green (the 245KB persister matrix is the phase's acceptance suite; ~25 harness fixes, all the same shape — mark the session active before simulating request events, as production always does). Store-parity assertions stayed armed throughout.
- 78/78 mock E2E across the 14 awaiting-adjacent specs (request lifecycles, parallel requests, capability/proxy reviews, computer use, reconnect replay, message queueing, mixed pending, status indicators).
- 3 scripted live-container repros against a real model: park → answer → keep working, two-turn with background bash, and a wire tap proving both awaiting edges fire while sampling route truth every 2s.

## New gap-closure tests

- E2E: a parked proxy review flags a session that starts after it (asserted via the sessions route inside the slow-work window).
- E2E: an abandoned computer-use approval does not survive into the next turn — reload replay must not resurrect it (red on main, green here).
- Route tests (×4): awaiting-input recovery — trailing-turn extraction, script_run exclusion, queued-message handling, inactive-session no-op.

### DIFF
--- a/e2e/specs/awaiting-input-status.spec.ts
+++ b/e2e/specs/awaiting-input-status.spec.ts
@@ -8,6 +8,8 @@ import {
   gotoAgentHome,
   uniqueName,
   uniqueSuffix,
+  waitForCurrentSessionId,
+  waitForPendingProxyReview,
   type TestAgent,
   type TestSession,
 } from '../helpers/agents'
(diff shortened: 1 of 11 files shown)


iddogino:
All four findings confirmed and fixed in 7476af0f, plus one adjacent hole found while validating:

1. **Recovered requests skipped during cleanup** — `cancelAwaitingInput` now snapshots the raw pending store instead of the replay-filtered getter, so recovered entries get their container-side pending rejected.
2. **Result blind-clearing awaiting on a still-active session** — the silent clear now happens only on the settling paths (error/interrupt, or the result-driven finalize); when the session stays active (state-event authority or open background work) the projection is re-derived. The blind clear was also eating the falling edge: the review's later settle saw cache == derived and never broadcast `session_input_provided`.
3. **Runtime-started turns** — the `running` self-heal now ends with the same trailing `syncSessionAwaiting` as `markSessionActive`.
4. **Re-subscription** — `doSubscribeToSession` now detaches only the transport; the pending stores carry over and the registry entries survive, so the preserved cache stays derived and replay data is kept. `unsubscribeFromSession` remains the full teardown for its real callers (session delete, x-agent cleanup, auto-delete monitor — all verified).

Adjacent fix: `finalizeIdle` never reset the cache — `markSessionActive`'s trailing sync can set it under an open agent review, and the `markSessionIdle` revert would then strand "needs input" on an idle session. It now resets the cache alongside `isActive`.

Each fix has a lifecycle-crossing test ("turn-boundary and transport crossings" in the characterization suite), proven red on ca34b728 (5/5 fail) and green now. 618 unit tests across the six awaiting-adjacent suites, 54 E2E tests across nine specs, typecheck + lint all green.

PR #583, opened 2026-07-25
Unified pending-request wire and client store (Phase 4)

iddogino:
Phase 4 of the pending user-input request rearchitecture (`user-input-requests-rearchitecture.md` §3.4). Phases 0–3: #560, #562, #566, #578.

## Server: one wire for every request kind

- `UserInputRequestManager.onTransition()` — the registry's single funnel emits exactly one `created` per accepted registration and one `resolved` per settlement, no matter which mutation path drove it. Emitting from the funnel (instead of per-callsite) is what makes silent settles structurally impossible on this wire. Duplicate registrations, idempotent re-resolves, and malformed drops emit nothing; listener errors are swallowed.
- The persister broadcasts `user_request_created` / `user_request_resolved` — global notification stream always, session SSE stream when session-scoped. Both events carry a **top-level `agentSlug`** for the global stream's per-user ACL filter. Recovery synthetics (`payload.recovered === true`) never hit the wire (`isReplayableUserInputRequest`).
- New `GET /api/agents/:id/pending-requests?sessionId=` snapshot behind `AgentRead()`. A session's view is its own requests ∪ the agent-scoped reviews; the agent view is everything in scope. The filter checks the agent **before** matching sessionId — the session param is caller input.
- Review envelopes now carry the full details plus the derived `displayText`, so review cards render from the envelope alone.

## Client: events invalidate, the snapshot is the truth

- New `usePendingUserRequests` store (React Query, canonical-slug key, 30s safety poll). The wire events are **invalidation triggers only** — the refetch reads the registry snapshot, so event bursts collapse into one consistent read, cross-stream ordering is irrelevant, and a stale in-flight response can never overwrite a newer one. A failed fetch throws (keeps last snapshot + retries) rather than caching an empty success.
- `use-pending-requests` now projects unified envelopes onto the legacy per-kind card shapes — descriptors, cards, and the renderer are untouched. The projection re-validates the lenient payloads (questions go through the same normalizer as the recovery path). Two properties are pinned by tests:
  - session-scoped entries are display-gated on `isActive` (an abandoned computer-use approval survives idle server-side for replay and must not gate the composer behind a dead card; agent-scoped reviews render regardless)
  - `dismissedRequestIds` is React state, not a ref — answering a card removes it synchronously, not a refetch round-trip later
- The activity indicator derives awaiting from the server's own predicate (`blocking && !autoApproved`), fixing the kinds the old per-type list omitted: script_run, computer_use, and capability_review used to read "Working…" while a card was parked.
- The global handler also nudges the dashboard's legacy `['proxy-reviews']` poll, so sessionless dashboard-triggered reviews appear/settle on push instead of the next 30s tick — previously they had no push signal at all.

## Kept live / deferred

Legacy per-type SSE events, the stream arrays, and the dashboard review poll all stay during the migration window. Deferred to Phase 5: decision-endpoint consolidation, ReviewManager→adapter, dashboard panel store migration, dead-array cleanup (the browser tray still reads `pendingBrowserInputRequests` — that array must survive the cleanup), container-emitted `pending_request_opened/settled` protocol, and `register()` upgrading a recovered synthetic when the real registration replays.

## Pre-PR review round

A 4-reviewer fan-out (server wire, client store, parity/consumers, test quality) ran before this PR; every fix below was proven red first:

- **P1** — global-stream ACL bypass: the notifications filter reads only a top-level `agentSlug` and forwards slug-less events to every authenticated user in auth mode; the events nested it inside `scope`. Now top-level on both frames.
- **P1** — cross-agent snapshot leak: `getSnapshotForScope` matched `sessionId` without an agent constraint, so `/api/agents/A/pending-requests?sessionId=<agent-B-session>` returned B's payloads. Agent check now comes first, unconditionally.
- **P2** — question projection raw-cast `payload.questions`; a question without `options` crashed the card with no reload escape (the unified bucket wins the dedupe against the normalized recovery copy). Now normalized.
- **P2** — `!res.ok → []` cached failures as success-empty, blanking every card and flipping the indicator to "Working…" until the next poll. Now throws.

Accepted transients (parity with or better than legacy): card appearance rides one invalidate→refetch round-trip; ~1-RTT indicator flash at turn start if a computer-use entry survived idle; strict snapshot parsing means a version-skewed envelope errors the query (keeps last data) rather than degrading per-kind.

## Validation

- Full unit suite: 7,775 passed (new: manager transitions/snapshot, persister wire incl. top-level-slug + cross-agent tests, route snapshot ×4, review displayText, hook test file, projection/dismissal/indicator pinning tests)
- E2E: 74/74 across the 13-spec acceptance set (user-input, mixed-pending, computer-use, reconnect, concurrent-tabs, proxy-review ×2, awaiting-status, capability, subagent-browser-input, connected-accounts, mcp-policy, message-queueing)
- `tsc --noEmit` clean; eslint clean (pre-existing warnings only)

### DIFF
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -3240,13 +3240,13 @@ describe('pending-requests snapshot — GET /:id/pending-requests', () => {
     expect(body.requests.map((r) => r.id).sort()).toEqual(['req-mine', 'req-review'])
   })
 
-  it('recovery synthetics are excluded from the snapshot', async () => {
+  it('recovery synthetics stay in the snapshot — payload-less, but still blocking waits', async () => {
     park('req-live', 'sess-1')
     park('req-recovered', 'sess-1', { recovered: true })
 
     const res = await getReq(app, '/api/agents/test-agent/pending-requests?sessionId=sess-1')
     const body = (await res.json()) as { requests: Array<{ id: string }> }
-    expect(body.requests.map((r) => r.id)).toEqual(['req-live'])
+    expect(body.requests.map((r) => r.id).sort()).toEqual(['req-live', 'req-recovered'])
   })
 
   it("a sessionId belonging to a DIFFERENT agent leaks nothing through this agent's gate", async () => {
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -457,6 +457,7 @@ import { listPendingScheduledTasks, listPendingScheduledTasksByAgents } from '@s
 import { listArtifactsFromFilesystem } from '@shared/lib/services/artifact-service'
 import { deleteNotificationsBySessionIds, getSessionIdsWithUnreadNotifications, getUnreadNotificationsByAgents } from '@shared/lib/services/notification-service'
 import { messagePersister } from '@shared/lib/container/message-persister'
+import { userInputRequestManager } from '@shared/lib/user-input/request-manager'
 import { containerManager } from '@shared/lib/container/container-manager'
 import { listUserSecrets, setSecret, getSecret, keyToEnvVar, getSecretEnvVars } from '@shared/lib/services/secrets-service'
 import { readJsonFileStrict, writeJsonFileAtomic } from '@shared/lib/utils/file-storage'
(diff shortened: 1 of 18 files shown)


iddogino:
All three review findings addressed in 41b1da7a — each fix proven red first (5 failing tests pre-fix):

**1. Fail-closed global broadcast (P1).** Confirmed: the scope schema types `agentSlug` as optional, `streamEventToPendingRequest` maps a non-string slug to `undefined`, and the ACL filter's truthiness check also fails open on `''`. `broadcastRequestTransition` now only broadcasts to the global stream when the scope carries a verified non-empty slug; the session stream still receives the event (its subscribers are AgentRead-gated per session). Test covers both the missing-slug and empty-string cases on both event types.

**2. Review fallbacks until the first snapshot (P1).** `undefined` (never fetched) is now distinguished from a successful empty `[]`: while the snapshot has never succeeded, proxy reviews render from the legacy poll and capability reviews from the stream arrays — both still live during the migration window — so a blocked agent always has an actionable card. The moment a snapshot exists it is authoritative. Tests pin the fallback for both kinds plus the authoritative-empty case.

**3. Recovered waits in the snapshot (P2).** One note: this was parity with old main, not a Phase 4 regression — the legacy SSE replay also filtered recovered entries, so the old indicator read "Working…" against a recovered card too. Fixed properly now: recovery synthetics are included in the snapshot (they are real blocking waits, so the indicator counts them), remain excluded from the wire events, and the per-kind card guards drop the payload-less entries so no broken card renders — the transcript keeps rendering those cards. Tests updated at the manager, route, and projection layers.

Validation: full unit suite 7,780 green; the 13-spec E2E acceptance set 74/74; typecheck and lint clean. (One E2E run mid-round showed 17 failures that reproduced then vanished — traced to the Vite cold-start window after an `npm rebuild better-sqlite3` invalidated the dep cache, first-scheduled tests hitting a not-ready runtime; three consecutive runs later: 21/21, then 74/74.)

PR #585, opened 2026-07-25
Pre-warm the agent CLI subprocess so session creation doesn't wait on its boot

iddogino:
Starting a session took a few seconds before anything appeared, and it had been getting worse. This removes the largest fixed piece of that wait.

## Where the time went

`POST /api/agents/:id/sessions` is synchronous end to end and the UI is blocked for all of it: the composer stays disabled and navigation only happens once the mutation resolves (`agent-home.tsx:192` → `agent-shell.tsx:129`). The host's own work is ~30ms — `getAgent`, `ensureRunning` (cached), secrets, prefs and `registerSession` all measured 1–37ms. Effectively all of the wait was `client.createSession`, which blocks until the container's CLI has spawned and emitted `system init`, because that message carries the canonical session id.

Ruled out as causes, with measurements: the CLI/SDK version (benched 0.2.118 / 0.3.181 / 0.3.197 / 0.3.206 / 0.3.219 — all ~1.4–2.0s to init, no trend), our `query()` options (cumulative bisect over settingSources, hooks, subagents, the 63k system prompt, 5–6 SDK MCP servers, 92 tools — flat throughout), and time-to-first-token (median 2.5–3.6s per month across 745 real transcripts, Feb→Jul, flat). What does vary is machine load: the same probe swung between 1.1s and 7.5s depending on what else was running.

## What this does

**Pre-warm a CLI subprocess** using the SDK's `startup()`/`WarmQuery`. This is the officially recommended approach for our deployment shape — the [hosting guide](https://code.claude.com/docs/en/agent-sdk/hosting), under "Long-running sessions", says to "use `streamInput()` to add turns to an active session and `startup()` to pre-warm subprocesses ahead of incoming traffic". `createQuery()` claims the parked handle instead of spawning; `buildQueryOptions()` was split out of `createQuery()` so both paths build identical options.

The pooling layer on top is ours — Anthropic documents the primitive but gives no guidance on matching or pool depth. It holds **whole pre-warmed processes**, not detached option sets: the options bake in `canUseTool`, every hook, and the browser/agents/chat MCP servers, all of which resolve `this.sessionId` at call time, so a handle warmed against one process could never be safely handed to another. A warm process is used only when the incoming request's profile key matches exactly; a mismatch discards it and starts cold.

**The host decides what to warm for.** `prewarmDefaults` carries the same preference chain minus the per-session pick — `agentPrefs.defaultModel ?? getEffectiveModels().agentModel`, plus default effort/speed — resolved through `resolveContainerModel` to a concrete wire id. This matters: the agent default is stored as the alias `"opus"` and only becomes `claude-opus-5` after resolution, so keying on the raw value would never match. The composer only puts a model on the wire when the user explicitly picks one (`composer-options.tsx:243`), so keying on last-used would let a single one-off pick cost the next session both a wasted spawn and a cold start. Callers without a hint (cron, chat, cross-agent) fall back to the session's own shape.

**Drop the CLI's boot-path telemetry and feature-flag fetch.** Measured on its own this is 2% — noise (see the table below). It is kept only as insurance for restricted-egress deployments, where that fetch has been reported to hang startup ([#16590](https://github.com/anthropics/claude-code/issues/16590), [#55623](https://github.com/anthropics/claude-code/issues/55623)). It does stop remote flags and killswitches from reaching our sessions, which is acceptable because the image pins the CLI version rather than self-updating.

Lifecycle: refill happens after the session is live, never in front of it; the profile is persisted so a booting container warms for the session about to arrive (kicked off before the dashboard scan — on 2 CPUs they compete); `POST /env` discards the parked process because it captured the old `REMOTE_MCPS`; `stopAll()` disposes it; and a warm-up that resolves after the wanted profile changed disposes itself rather than parking something nothing will claim.

**Pool depth is one.** A parked process holds ~232MB RSS (measured), so a deeper pool would reserve ~700MB per awake agent against a 4GB limit. A second session arriving mid-refill awaits the in-flight warm-up rather than racing it with a cold spawn, so bursts degrade to today's behaviour instead of contending for 2 CPUs. `SESSION_PREWARM=0` (or `prewarmEnabled: false`) disables it.

## Measurements

Three images built from the same commit, interleaved rounds, real containers and real LLM turns, 2 CPU / 4GB. Median `POST /sessions`:

| arm | idle container (n=12) | sleeping container (n=6) |
|---|---|---|
| baseline | 0.635s (p90 0.663) | 1.786s |
| telemetry flag only | 0.623s — **2%, noise** | 1.768s |
| pre-warm pool | **0.197s (p90 0.267) — 69%** | **1.427s — 20%** |

For the sleeping case the CLI cost is gone too — container logs show the boot warm-up landing and the request claiming it 0.3ms later, with `init` at +77ms. The remaining ~1.4s is `docker start` plus the health wait, which this doesn't touch.

One-off override behaviour, verified live: default cold 0.751s → default **0.188s** → one-off Haiku pick 0.521s (correctly not served a mismatched process) → default **0.183s** (pool stayed on the default) → default **0.197s**. Persisted profile after the Haiku session is `claude-opus-5`.

## Testing

- 761 container tests, 1862 host tests, both typechecks, eslint clean on changed files.
- 14 pool tests plus 5 profile-key tests and 2 process-binding tests: claim-on-match, discard-on-mismatch, parsed-default equivalence, warms-for-host-default-not-this-session's-pick, replaces-a-parked-stale-process, no-hint fallback, boot warm from persisted profile, env-change invalidation, stale warm-up self-disposal, shutdown disposal, kill switch.
- End to end through the real app: warm-claimed sessions return correct transcripts, and a model switch discards the warm process and still answers.
- Existing SessionManager suites pass `prewarmEnabled: false` — they mock the process class entirely and shouldn't have to model pre-warming.

## Review notes

`claude-code.ts` shows ~700 changed lines but only ~78 are real — the rest is re-indentation from unnesting the options object. Review with `git diff -w`.

Two bugs found by running against a real container that the unit tests missed, both fixed here: the staleness guard compared against the session's profile rather than the refill target (so every refill disposed itself), and a parked stale process was never replaced (so a boot warm-up built from the previous container's profile stranded every later session on a cold start).

Not addressed: a message into an idle-evicted session still pays a full cold boot, since a warm process can't serve it (`resume: <sessionId>` is baked into the options). That's the other half of the "existing sessions feel slower" report and wants its own change.

https://claude.ai/code/session_01FT8YFrqoNfy3wJSQ89UKFG

### DIFF
--- a/agent-container/src/claude-code.ts
+++ b/agent-container/src/claude-code.ts
@@ -1,4 +1,4 @@
-import { query, Query, SDKMessage, SDKUserMessage } from '@anthropic-ai/claude-agent-sdk';
+import { query, startup, Options, Query, WarmQuery, SDKMessage, SDKUserMessage } from '@anthropic-ai/claude-agent-sdk';
 import { z } from 'zod';
 import type { UUID } from 'crypto';
 import { EventEmitter } from 'events';
@@ -474,6 +474,10 @@ export class ClaudeCodeProcess extends EventEmitter {
   private lastTurnInformationals: SDKMessage[] = [];
   private lastResultMessage: SDKMessage | null = null;
   private lastSessionState: string | null = null;
+  // Pre-spawned CLI subprocess from prewarm(), waiting for a prompt. Claimed
+  // (once) by the next createQuery; see prewarm() for why the handle lives on
+  // the process rather than in a detached pool.
+  private warmHandle: WarmQuery | null = null;
   public slashCommands: { name: string; description: string; argumentHint: string }[] = [];
 
   constructor(options: ClaudeCodeProcessOptions) {
@@ -572,8 +576,49 @@ export class ClaudeCodeProcess extends EventEmitter {
   /**
    * Creates a new query instance with the standard configuration.
    * Used by start(), restart(), and interrupt() to avoid duplication.
+   *
+   * Claims the pre-warmed subprocess when prewarm() left one: its CLI is
+   * already spawned and past the initialize handshake, so the `init` message
+   * (and with it the canonical session id) lands in milliseconds instead of
+   * the ~1-3s a cold spawn costs. The handle is single-use — the SDK closes
+   * over the prompt stream we hand it here — so it is cleared on claim.
    */
   private createQuery(): Query {
+    const warm = this.warmHandle;
+    if (warm) {
+      this.warmHandle = null;
+      console.log(`[Session ${this.sessionId}] createQuery: claiming pre-warmed subprocess`);
+      return warm.query(this.messageQueue!);
+    }
+    return query({ prompt: this.messageQueue!, options: this.buildQueryOptions() });
+  }
+
+  /**
+   * Pre-spawn the CLI subprocess and complete its initialize handshake before
+   * a prompt exists, so the next start() pays no boot cost.
+   *
+   * The warm handle lives on the process (not in a detached pool of bare
+   * option sets) because the options bake in this instance's closures — the
+   * `canUseTool` callback, every hook, and the browser/agents/chat MCP servers
+   * all resolve `this.sessionId` at call time. A handle warmed against one
+   * process could therefore never be handed to another without misrouting
+   * those callbacks. The session manager pools whole pre-warmed processes
+   * instead, and only hands one to a request whose parameters match.
+   */
+  async prewarm(): Promise<void> {
+    if (this.disposed || this.warmHandle || this.queryInstance) return;
+    // Baked into the warm subprocess's options, so initializeQuery must not
+    // replace it on claim or the warm process would be unstoppable.
+    this.abortController = new AbortController();
+    this.warmHandle = await startup({ options: this.buildQueryOptions() });
+  }
+
+  /** Whether a pre-warmed subprocess is parked on this process, unclaimed. */
+  isPrewarmed(): boolean {
+    return this.warmHandle !== null;
+  }
+
+  private buildQueryOptions(): Options {
     const remoteMcpConfigs = this.buildRemoteMcpServers();
     const remoteMcpToolPatterns = Object.keys(remoteMcpConfigs).map(name => `mcp__${name}__*`);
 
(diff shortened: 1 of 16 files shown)


iddogino:
All three confirmed and fixed in 67a6013d. Each finding was reproduced before fixing, and each fix has a test that fails without it.

**[P1] Nested profile state didn't reach the key.** Confirmed — worse than key collision in general, it's a capability-policy bypass:

```
$ node -e "const key = p => JSON.stringify(p, Object.keys(p).sort()); ..."
A: {"capabilityPolicies":{},"customEnvVars":{},"model":"m"}
B: {"capabilityPolicies":{},"customEnvVars":{},"model":"m"}
COLLIDE: true
```

An `allow` profile and a `block` profile keyed identically, so a restricted session could claim a process with `Task`/`Agent`/`Workflow` baked into its query — block-tier policy strips those at query build time, not per call. Replaced with a recursive stable serialization. Arrays keep their order deliberately (`modelPromptHints` and `availableEnvVars` are concatenated into the prompt, so a different order is a different prompt), and `undefined` fields are dropped so an absent field and an explicitly-undefined one key the same. New `warm-profile.test.ts` covers nested policies, nested env vars, insertion-order stability, array-order sensitivity, and absent-vs-undefined.

**[P1] Parked processes registered as `currentProcess`.** Confirmed. `ClaudeCodeProcess`'s constructor assigned the module global, so the most recently *constructed* process won — which, once every session's refill constructs one, was the parked process from the moment a session started. An approved MCP was injected into a query nobody was talking to; the session that asked never got the tools, and if the parked process had since been discarded it was injected into a disposed one.

Took the binding approach rather than moving the assignment to `start()`: `createRequestRemoteMcpTool(getProcess)` is built against its owning process, threaded through `createUserInputMcpServer(() => this)` — the same getter pattern `createAgentsMcpServer` / `createChatMcpServer` / `createBrowserTools` already use. The global is deleted, so this is compile-enforced rather than convention. That also fixes the pre-existing case this predates: two live sessions in one container, where the newer one would swallow the older one's MCP injection.

**[P2] In-flight warm-ups weren't invalidated.** Confirmed, both halves. Added a warm generation, bumped synchronously at the top of `discardPrewarmed` whether or not anything is parked; a warm-up captures it at spawn and disposes itself if it no longer matches. Without that, the in-flight process holds a snapshot of the old environment and its profile is unchanged, so `desiredWarmKey` would happily park it. The `/env` handler now invalidates immediately after mutating `process.env`, before awaiting the `.env` write, closing the window you flagged — verified in container logs:

```
[ENV] Set environment variable: REMOTE_MCPS (2 chars)
[SessionManager] Discarding pre-warmed process: env var REMOTE_MCPS changed
[ENV] Updated .env file with REMOTE_MCPS
```

Next session after that came up cold (0.661s) and the one after was warm again (0.232s), so invalidation and recovery both work end to end.

Container suite is 761 passing; host suite 1862; both typechecks and eslint clean. On your note about local tests — `agent-container/` has its own `package.json`, so `cd agent-container && npm install && npx vitest run` is what runs this code's suite.

iddogino:
Confirmed and fixed in 0b6bfc19 — good catch, and it was self-inflicted by the P2 fix in the previous round.

`prewarm()` read `warmGeneration` at the top, before the `discardPrewarmed()` that replaces a parked process for a different profile — and that discard bumps the generation. So the replacement always failed its own generation check on arrival and disposed itself. The pool then stayed empty until some later refill happened to find nothing parked, and the next session paid a cold start.

Moved the capture to just after the discard, immediately before spawning.

Worth noting why the existing `replaces a parked process when the wanted profile changes` test didn't catch it: it asserted the replacement was **spawned and pre-warmed**, never that it **survived to be claimed** — both true even when the process disposes itself a tick later. It now also asserts `disposeCalls === 0`, and there's a new test for the path you described, which drives `prewarm(B)` while A is parked and then asserts the next matching session actually claims B rather than starting cold. Verified red before green by reverting the one-line move.

Also worth recording for anyone reading the diff later: the reachable trigger is narrower than "any profile switch". When a switch happens through `createSession`, the mismatched process is discarded inside `claimPrewarmed()`, which does **not** bump the generation, so that path was always fine — which is why the live 5-step override test passed earlier. It takes two overlapping session creations refilling for different profiles (or a direct `prewarm()` call) to reach the buggy branch.

Container suite 762 passing, typecheck and eslint clean.

PR #586, opened 2026-07-25
Registry consolidation: ReviewManager adapter, computer-use store, dead-subagent invalidation (Phase 5)

iddogino:
Phase 5 of the user-input request rearchitecture (follows #560, #562, #566, #578, #583). The registry (`userInputRequestManager`) becomes the single pending store for reviews and computer-use, and gains resource invalidation for dead subagents.

### 1. ReviewManager → registry adapter
The registry now owns the pending-review list (envelope payloads carry full `ReviewDetails` + `displayText` since Phase 4). ReviewManager keeps only a `settlers` map — the blocked proxied call's promise resolve/reject and the 5-minute auto-deny timer, keyed by id. All reads (`getPendingReviewsForAgent`, the three sweeps, `submitDecision`) re-source from the registry, so the dashboard poll endpoint and `hasAgentLevelReviews` gates work unchanged.

Two hardenings that fall out of the design:
- **`submitDecision` has a review-kind guard**: a caller-supplied id that resolves to a parked secret/question/computer-use entry returns `false` instead of settling a foreign wait.
- Registry-only review entries (from feeders other than `requestReview`) are now visible, decidable, and sweepable — previously they were invisible to every decision path.

The parity invariant flips to one-directional `verifyReviewSettlerParity`: a settler without a registry entry is a leak (a parked call no sweep can reach); an entry without a settler is legitimate.

### 2. Computer-use into the registry
`pendingComputerUseRequests` is deleted from `StreamingState`. `getPendingComputerUseRequests` rebuilds the legacy shape from registry envelopes for the SSE replay route and decision endpoints. The `session_active` supersede loop reads registry ids — closing a latent hole where registry entries could outlive the old map across a state recreation. The divergent clearing rules (survives idle for replay; cleared by decision routes / `cancelAwaitingInput` / new turn) are unchanged, now expressed only through `storeForKind`.

### 3. Dead-subagent invalidation (closes the #527 follow-up)
Envelopes gain an optional top-level `parentToolUseId`, populated on all sidechain dispatches (both delivery paths, all 6 blocking kinds + computer-use). `broadcastSubagentCompleted` settles every open request under the dead parent as `invalidated` — registry entry, replay store, container-side pending (`/inputs/:id/reject`), and awaiting status together. Previously a subagent that died mid-request orphaned its card until a coarser boundary or TTL.

### 4. Recovered-synthetic upgrade
`register()` now replaces a recovered stub when the real registration replays, re-emitting `created` — previously the stub blocked the real payload from ever entering the registry, so clients never received a renderable `user_request_created` for that id. The dispatch funnel's dedupe lets the real event through a recovered stub. Downgrades (recovered over real) are pinned impossible.

### 5. Container `connected_account` kind fix
`request-connected-account.ts` no longer stamps `inputType: 'secret'` via the legacy `createPending` wrapper. Verified no host code reads the container's `inputType`, so the change is contained to the container.

## Validation
- Red-proven where behavior changed: 7 tests observed red first across the three behavioral workstreams
- Typecheck + lint clean; full unit suite 7,833 green; container suite 743 green
- 13-spec request E2E set: 75/75 (incl. a new dead-subagent E2E — new mock scenario `dead subagent input`)
- **Live-validated** on an isolated dev instance (real Docker container, real Opus 5, real browser): (a) `request_secret` full cycle — card, awaiting indicator, reload-while-parked re-render, provide → resolve → idle; (b) the x-agent review chain through the adapter — parked list-review card re-rendered from the registry snapshot on fresh navigation, Allow Once unblocked the parked route via the settler, second invoke-review card allowed, cross-agent message delivered, idle

## Not in this PR (Phase 6 per plan §3.4)
Single decision endpoint, dashboard panel client migration, dead-array cleanup (`pendingBrowserInputRequests` must survive for the browser tray), legacy event removal, name normalization.

### DIFF
--- a/agent-container/src/tools/request-connected-account.test.ts
+++ b/agent-container/src/tools/request-connected-account.test.ts
@@ -1,0 +1,69 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { inputManager } from '../input-manager'
+
+describe('requestConnectedAccountTool', () => {
+  let originalAccounts: string | undefined
+
+  beforeEach(() => {
+    originalAccounts = process.env.CONNECTED_ACCOUNTS
+  })
+
+  afterEach(() => {
+    if (originalAccounts === undefined) {
+      delete process.env.CONNECTED_ACCOUNTS
+    } else {
+      process.env.CONNECTED_ACCOUNTS = originalAccounts
+    }
+  })
+
+  async function invokeTool(toolUseId: string) {
+    const { requestConnectedAccountTool } = await import('./request-connected-account')
+    const handler = (requestConnectedAccountTool as any).handler
+    inputManager.setCurrentToolUseId(toolUseId)
+    return handler({
+      toolkit: 'gmail',
+      reason: 'Allow access to Gmail to search for the shipping confirmation?',
+    }) as Promise<{ content: Array<{ type: string; text: string }>; isError?: boolean }>
+  }
+
+  it('parks the pending request as connected_account, not secret', async () => {
+    const toolUseId = `ca-test-${Date.now()}-1`
+    const resultPromise = invokeTool(toolUseId)
+
+    await vi.waitFor(() => expect(inputManager.hasPending(toolUseId)).toBe(true))
+    const entry = inputManager.getAllPending().find((p) => p.toolUseId === toolUseId)
+    expect(entry?.inputType).toBe('connected_account')
+
+    inputManager.resolve(toolUseId, 'granted')
+    const result = await resultPromise
+    expect(result.isError).toBeUndefined()
+    expect(result.content[0].text).toContain('gmail')
+  })
+
+  it('keeps the toolkit and reason on the pending metadata', async () => {
+    const toolUseId = `ca-test-${Date.now()}-2`
+    const resultPromise = invokeTool(toolUseId)
+
+    await vi.waitFor(() => expect(inputManager.hasPending(toolUseId)).toBe(true))
+    const entry = inputManager.getAllPending().find((p) => p.toolUseId === toolUseId)
+    expect(entry?.metadata).toMatchObject({
+      toolkit: 'gmail',
+      reason: 'Allow access to Gmail to search for the shipping confirmation?',
+    })
+
+    inputManager.resolve(toolUseId, 'granted')
+    await resultPromise
+  })
+
+  it('returns a declined message when the request is rejected', async () => {
+    const toolUseId = `ca-test-${Date.now()}-3`
+    const resultPromise = invokeTool(toolUseId)
+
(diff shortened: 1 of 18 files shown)


iddogino:
Review finding fixed in 6c467a5e.

**[P2] Sidechain `request_script_run` omitted `parentToolUseId`** — confirmed. script_run is dispatched adjacent to the blocking-tool funnel (its own handler call at both sidechain sites), so it missed the parent threading the six funnel kinds and computer-use got. `handleScriptRunRequestTool` now takes the parent, includes it in the `script_run_request` broadcast (the SSE-intercept feeder lifts it into the envelope), and both sidechain call sites pass `parentToolId`. Red-proven regression test added: sidechain script_run → subagent dies → entry invalidated, awaiting cleared.

Checked the remaining adjacent dispatch, capability reviews from a subagent's nested launch: those have a container-side cleanup channel (`capability_review_cancelled` fires when the CLI abandons the parked hook), so they don't share this leak.

Validation: typecheck + lint clean, full unit suite 7,834 green, script-run + subagent E2E specs 19/19.

iddogino:
Deep-E2E findings addressed.

**[P1] Child script_run missing `parentToolUseId`** — already fixed in 6c467a5e (the previous review round); the review snapshot predates it. Both sidechain dispatch sites pass the parent, with a red-proven regression test (`a dead subagent also invalidates its parked script_run approval`).

**[P1] Decisions don't settle under parallel tool calls** — confirmed and fixed in eb21a519, in two layers:

1. `messagePersister.completeInputRequest(sessionId, toolUseId, outcome)` — stream-store delete + registry resolve + `tool_result` broadcast + awaiting sync — called from every successful stream-kind decision path (secret / question / connected-account / remote-MCP / file / browser-input / script-run routes, plus the chat-connector resolves), mirroring what capability reviews and computer-use already did. Red-proven at the route layer (9 cases) and in the lifecycle suite (parallel arithmetic, no double-resolution when the real tool_result lands, snapshot/replay exclusion).
2. The reload-resurrection had a second source your live test exposed: with results held back, the transcript still shows the decided call as unresolved, so the client's history fallback re-rendered the card and GET-messages recovery re-asserted awaiting. The messages route now stamps settled outcomes onto held-back tool calls (tombstones on the persister, expiring at the turn boundary), so every history consumer sees a completed call. Chose transcript-stamping over a client-side filter deliberately: a client filter against the snapshot would race the recovery path after a host restart.

Mock fidelity gap closed too: `registerPendingInputs` gains `holdResultsUntilAll` (real-CLI parallel semantics — this is why mock E2E never caught it), with a new `ask parallel holdback` scenario and an E2E that declines one of two parallel requests, reloads, and asserts the declined card stays gone while the sibling survives — observed red against the route-settle layer alone, green with the stamping layer.

Validation: typecheck + lint clean, full unit suite 7,848 green, full 13-spec request E2E set 76/76.

PR #592, opened 2026-07-27
Honor SUPERAGENT_DB_PATH for SQLite outside the data dir

yiw190:
## Summary
- `getDatabasePath()` honors `SUPERAGENT_DB_PATH` when set; desktop/self-host default (`$SUPERAGENT_DATA_DIR/superagent.db` or `~/.superagent/superagent.db`) unchanged.
- `initDb()` mkdir’s the DB parent so a path like `/sqlite/superagent.db` cannot `SQLITE_CANTOPEN` on first open.
- Align `drizzle.config.ts` and replace dead `DATABASE_PATH` docs with `SUPERAGENT_DB_PATH` / `SUPERAGENT_DATA_DIR`.

Needed so cloud can keep settings/agents on the S3 Files mount while the live DB sits on Fargate ephemeral disk (Litestream track in gamut-infra).

## Test plan
- [x] `vitest` `data-dir.test.ts` + `db-path.test.ts`
- [ ] Desktop: unset env → DB still under userData / `~/.superagent`
- [ ] Set `SUPERAGENT_DB_PATH` to a nested path → parent created, DB opens

### DIFF
--- a/src/shared/lib/db/index.ts
+++ b/src/shared/lib/db/index.ts
@@ -20,8 +20,7 @@ function getMigrationsFolder(): string {
 }
 
 // Lazy initialization: defer DB creation until first access so that
-// SUPERAGENT_DATA_DIR (set by the Electron main process at startup)
-// is available when the path is resolved.
+// SUPERAGENT_DATA_DIR / SUPERAGENT_DB_PATH (set at startup) are available.
 let _sqlite: InstanceType<typeof Database> | null = null
 let _db: BetterSQLite3Database<typeof schema> | null = null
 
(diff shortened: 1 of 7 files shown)


src/shared/lib/db/index.ts
```diff
@@ -31,10 +30,9 @@ function initDb() {
   const dbPath = getDatabasePath()
   const dataDir = getDataDir()
 
-  // Ensure data directory exists
-  if (!fs.existsSync(dataDir)) {
-    fs.mkdirSync(dataDir, { recursive: true })
-  }
+  // Data dir (settings/agents) and DB parent may differ when SUPERAGENT_DB_PATH is set.
+  fs.mkdirSync(dataDir, { recursive: true })
```
iddogino: Why remove the exists check here?

iddogino (changes requested):

src/shared/lib/db/index.ts
yiw190: Restored — no functional reason to drop it; `recursive: true` made the guard redundant but the prior pattern is clearer. Kept the same for the DB parent when `SUPERAGENT_DB_PATH` differs from the data dir.

yiw190: agreed, unnecessary change. exist check restored.

PR #593, opened 2026-07-27
Add Kimi K3 to the Platform catalog with fast-mode pricing

iddogino:
Surfaces `kimi-k3` in the Platform provider's model picker. Pairs with SkillfulAgents/platform#229, which adds the proxy-side routing — **merge that first**, or the model appears in the picker and 400s.

## The catalog entry

- **Bare id** (`kimi-k3`, not a vendor-prefixed slug) — the proxy's routing and pricing key off bare ids.
- **1M context, image input** — both read from Fireworks' model metadata, not guessed.
- **normal/fast speeds.** Fireworks' fast path is a separate `-fast` router resource the proxy swaps in, billed as its own SKU at a flat **1.5x**, so one multiplier covers it. There's no flex/slow equivalent.
- **Web search and fetch off** — Fireworks' Anthropic-compatible endpoint takes function tools only, and the proxy strips Anthropic's server tools. Flagged so the picker warns rather than silently dropping them.

`kimi-k3` also goes into `model-pricing.json` so the cache rates are explicit rather than falling out of `deriveInputRelatedRate`'s defaults, and so transcript ids outside the catalog still price.

## Icon

Adds the Kimi brand mark — Moonshot AI's sphere glyph, from simple-icons (CC0) — and a "Moonshot AI" vendor tab. Renders correctly in light and dark (the existing `dark:invert` rule handles the flip).

Also maps OpenRouter's `moonshotai/*` listings onto the same icon, now that the asset exists.

## Verification

Driven through the real UI against a local proxy talking to real Fireworks — model picked from the picker at both tiers:

- transcript records `model: kimi-k3` with `usage.speed: fast`
- usage tab: **$0.016047 / 26,010 tokens**, attributed to `kimi-k3` — matching `(261×$3 + 149×$15 + 25,600×$0.30)/1M × 1.5` exactly

Full suite passes (8,028), typecheck clean, lint 0 errors.

## Note

The usage tab initially showed $0 for Kimi turns. Root cause was proxy-side (Fireworks streams OpenAI-shaped usage field names for K3, so the CLI wrote zero-token entries) and is fixed in the paired platform PR — no app change was needed, but it's why that PR is a hard prerequisite for the numbers here to be right.

https://claude.ai/code/session_01BcuyKPoE84HwPGTSQwbRTV


iddogino:
## Added the Kimi line to OpenRouter too (af58c5df)

Originally this PR touched OpenRouter only for the brand icon. Now it carries real catalog entries there as well — K3 as family latest, plus the two K2 versions still listed on OpenRouter:

| id | latest | context | $/Mtok in | $/Mtok out |
|---|---|---|---|---|
| `moonshotai/kimi-k3` | ✅ | 1,048,576 | $3 | $15 |
| `moonshotai/kimi-k2.7-code` | | 262,144 | $0.73 | $3.50 |
| `moonshotai/kimi-k2.6` | | 262,144 | $0.646 | $2.72 |

The K2 line earns its place by being an order of magnitude cheaper than K3, and it's the same pair the Platform proxy serves — so a pinned pick stays meaningful if an agent moves between providers.

**Prices are OpenRouter's own, fetched live from its model list — not mirrored from the Platform entries.** They genuinely differ: K2.6 is $0.646/$2.72 on OpenRouter against Fireworks' $0.95/$4.00. Same for context windows and vision, read from OpenRouter's advertised modalities.

No `supportedSpeeds` on any OpenRouter entry — OpenRouter ignores our speed header, so a speed knob there would be a lie. There's a test asserting that.

### Picker rendering

`kimi` is now a lineage family, so the three collapse into one **Kimi** row with `K2.6` / `K2.7 Code` / `K3` chips — the treatment Opus and Sonnet already get. Without it they'd splay into four rows, and the label splitter would have cut "Kimi K2.7 Code" into a `K2.7` row with a stray `Code` chip.

Platform, which has only K3, renders a single-version lineage row — exactly as Fable already does, so this isn't a new shape.

Verified in the running app against both providers:

- **OpenRouter** — `Kimi [K2.6][K2.7 Code][K3]`, no Speed control
- **Platform** — `Kimi [K3]`, with a Speed control

which is the correct split, since only the Platform proxy can honor a tier.

The two catalogs stay isolated: `openrouter` has no bare `kimi-k3`, `platform` has no `moonshotai/kimi-k3`, and the bare `kimi` alias resolves per-provider. Covered by tests.

8,030 tests pass, typecheck clean, lint 0 errors.

PR #595, opened 2026-07-27
Keep chat history reachable during long agent activity (SUP-484)

JeremyBischoff:
Closes https://linear.app/datawizz/issue/SUP-484/keep-chat-history-reachable-during-long-agent-activity

## What

The activity card now caps and scrolls inside itself, under a pinned status header, and the chat column guarantees the transcript a minimum height.

## Why

You can't scroll back into the conversation once the action list gets long.

The chat column is a two-row grid - transcript on top, pinned footer below. The footer row is `auto`, so it takes the height it wants first and the transcript gets the remainder. The footer holds the activity card, and that card had no cap: one row per subagent, and `subagent_completed` upserts rather than removes, so its height only ever grows within a turn. Once the card is as tall as the column, the remainder is zero.

Measured at 1280x720, 24 subagents with 4 still running:

| | main | this PR |
| --- | --- | --- |
| Chat history, long action list | 0px | 242px |
| Chat history, plus a pending question | 0px | 160px |
| Activity card | 530px | 262px (112px next to a question card) |
| Request card below the bottom of the window | 64px | none |

At 0px the transcript isn't merely hard to reach - there's no region left to scroll.

## How it works

Two bounds, because one doesn't hold:

- The card caps its own list at 30vh and scrolls it. The status row, timer and computer-use chip are pinned above the scroll area, so the label stays put while the rows move. Running subagents sort above completed ones, otherwise the capped card shows the oldest ✓ rows and pushes live work below the fold.
- The column floors the transcript with a grid row minimum. A per-card cap can't bound a sum: a card capped at 30vh plus a 376px request card is 710px of footer in a 656px column, which puts the transcript straight back to 0px. With the floor, the card is the thing that gives way - it shrinks to 112px and scrolls, and the request card stays whole and on screen.

The composer and the request card sit outside the scroll area, so nothing scrolls as a unit and the card the user has to act on is never scrolled away.

## How it was verified

`e2e/specs/activity-card-history-floor.spec.ts` drives a 24-subagent fan-out in a real browser and asserts both the heights and which element owns the overflow - the card's list scrolls, the footer does not. Two cases: activity card alone, and activity card plus a pending question. Layout needs a real browser here; jsdom reports zero heights for the broken and fixed states alike.

Also checked by hand at 1280x720 and 1280x600. At 600px the card degrades to just its status row rather than colliding with the question card.

## Repro

1. `E2E_MOCK=true npm run dev:web`
2. New agent, send `subagent fanout` - 24 subagents, 4 left running
3. Try to scroll up into the conversation
4. `subagent fanout with question` for the combined case

## Notes for the reviewer

- `agent-activity-indicator.tsx` reads as a large diff but it's about 12 real lines. The rest is the existing block being re-indented into the new scroll container - `?w=1` shows it.
- Flagged, not fixed: `subagent_completed` upserts instead of removing (`use-message-stream.ts`), so finished subagents keep their rows for the rest of the turn. That's what makes the card grow monotonically. Changing it is a product decision about what the card reports, and it wouldn't help a wide concurrent fan-out anyway.

### DIFF
--- a/e2e/specs/activity-card-history-floor.spec.ts
+++ b/e2e/specs/activity-card-history-floor.spec.ts
@@ -1,0 +1,64 @@
+import { test, expect } from '@playwright/test'
+import { SessionPage } from '../pages/session.page'
+import {
+  createAgent,
+  gotoAgentHome,
+  uniqueName,
+  uniqueSuffix,
+  type TestAgent,
+} from '../helpers/agents'
+
+// The live activity card sits in the chat column's pinned bottom strip, beside
+// the composer. It renders one row per subagent launched during the turn and
+// never drops a finished one, so the card grows for the whole turn. The strip
+// is sized to its content and the message list takes what is left — so past
+// enough rows the history has nothing left to take.
+//
+// Reported by a dogfooder as "can't scroll into the chat history when the
+// action list gets long". Asserted on geometry, not classes: a component test
+// cannot see this at all, because jsdom has no layout engine and reports every
+// height as 0 whether or not the bug is present.
+test.describe('Activity card must not squeeze out the chat history', () => {
+  let sessionPage: SessionPage
+  let agent: TestAgent
+
+  // The history must stay usable while a turn runs — enough for a few lines of
+  // transcript, not a sliver.
+  const MIN_HISTORY_HEIGHT = 160
+
+  test.describe.configure({ timeout: 45000 })
+
+  test.beforeEach(async ({ page, request }, testInfo) => {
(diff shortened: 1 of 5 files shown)


iddogino:
@JeremyBischoff not sure what issue this solves. Can you please provide a screenshot and/or steps to reproduce?

iddogino:
https://github.com/user-attachments/assets/d5474d8f-1c46-42f6-9235-c5e3308483bc

making the footer scrollable as an entire unit feels wrong (see attached video example). The scroll and max height should be internall to the working indicator

PR #596, opened 2026-07-27
Chat integrations consume the unified request wire and the gated settle path (Phase 7)

iddogino:
## Why

Phases 0–6 moved the renderer, notifications, Electron main, and the decision routes onto the registry. Chat integrations were the last consumer still reading the legacy per-type events off the in-process broadcast pipe and settling container inputs directly — two readers of "what is the agent blocked on", reading two different sources. Four mismatches were open today:

1. Recovered registrations emit no legacy event, so they were invisible to chat.
2. Chat had no resolved signal — a card settled elsewhere kept live-looking buttons, and pressing one buffered a phantom `earlyResult` in the container while the user believed they had answered.
3. Chat decisions bypassed the Phase-6 already-settled gate entirely.
4. A new request kind needed manual chat wiring or silently fell through `processSSEEvent` — the "kind #12 = one descriptor" promise didn't hold for chat.

## What

**One wire.** `processSSEEvent`'s eight per-type cases collapse to a single `user_request_created` case. The legacy events still fire (they die in Phase 8) and are deliberately *ignored* — handling both would double-send every card. The envelope→card mapping lives in a new `chat-integrations/request-card.ts`; connector `sendUserRequestCard` switches are untouched.

**Kind coverage is now structural.** `KIND_TO_CARD_TYPE` is a total `Record<UserInputRequestKind, …>`, so a twelfth kind is a compile error until someone decides what chat does with it. That's mismatch #4, fixed by the type system rather than by remembering.

**Reviews off the same wire.** `handleGlobalNotification` triggers on `user_request_created` filtered to `proxy_review`/`x_agent_review`, keeping the `review:<id>:<agent>` synthetic-card id byte-identical so decision routing is unchanged. This fixed their routing as a side effect: the old handler read a top-level `data.sessionId` that `broadcastReview` **never sets**, so it always fell through to the agent-wide fan-out and sent a session-less (agent-home) link. It now uses the agent's active session, matching what the review notification does.

**Chat decisions are gated.** `handleInteractiveResponse` checks `userInputRequestManager.getOpenRequest` *and* an agent match before touching the container, and passes `integration.agentSlug` to `reviewManager.submitDecision` (whose third parameter already existed for exactly this). A refused press gets an explicit "already handled" reply — silence was the failure mode being replaced.

**Text answers come from the registry.** `consumeOrCancelAwaitingInput` takes a `registry` dependency and reads open `question` entries from it, filtering recovered stubs (no card was ever posted for those, so typed text must not answer one). That removes the last chat dependency on the persister's replay mirror ahead of Phase 8.

### Two behaviour changes, both intended

| Change | Why |
|---|---|
| `capability_review` now surfaces in chat | Before, a workflow parked on an approval the chat user was never told about and the conversation just went quiet. Added to the `UserRequestEvent` union, `UNSUPPORTED_IN_CHAT`, and `describeUnsupportedRequest`. |
| Auto-approved asks send **nothing** | The registry already classes them as not-a-real-wait (`isRealWait` is false); "The agent wants to run a script, which needs your approval" pointed the user at a screen with nothing on it. |

### Deliberately not done

Card *dismissal* on `user_request_resolved`. No connector can dismiss a single card — `dismissOpenCards` is all-or-nothing, and Slack implements neither it nor `answerOpenQuestionWithText`. That needs connector work this phase promised not to touch; the interactive gate covers the correctness half.

## The test blind spot, closed

Every chat test around user-input requests handed `processSSEEvent` a hand-built legacy event and asserted what the connector did with it. That pins connector behaviour and nothing else: if the host stopped *emitting* those events, every one of them stayed green and the feature was dead.

New `unified wire contract` suite in `chat-integration-e2e.test.ts` drives a **real tool_use through the real `MessagePersister`** and asserts on the card that comes out of the chat manager — one row per kind, plus the review path off the global channel and three decision-gate rows. The `processSSEEvent`/resilience tests now build their events through the registry schema (shared `user-request-fixtures.ts`), and one asserts the legacy event is **ignored**; that pair is what pins the card came from the unified wire.

**Red-proof:** reverting the manager to pre-Phase-7 behaviour fails **36 tests**.

One harness change was needed: `subscribeGlobalNotifications()` extracted from `start()` (idempotent), because the e2e harness drives `addIntegration` directly and reviews therefore had no wire to arrive on.

## Validated

- `tsc --noEmit` clean; ESLint 0 errors (warnings all pre-existing)
- Full unit suite **8,089 passed** (455 files)
- Request-path mock E2E **53 passed** (`user-input-requests`, `awaiting-input-status`, `concurrent-tabs-request-sync`, `mixed-pending-requests`, `pending-request-reconnect`, `computer-use`, `proxy-review`, `capability-review`)
- Live Slack suite (real account → real Slack → real app instance → real container): **14/14 web host**, **12/12 electron host**

The live suite gained two checks:

- `question-answered-in-app` — the card Slack posted and the request the **app** answers are the same registry entry; answering through the app's decision route settles it and the continuation lands back in Slack. This is the cross-surface proof, and it's the check that fails if the two readers ever drift apart again.
- `unsupported-capability-review` — the newly-surfaced kind. `workflows: 'review'` is the shipped default, so it needs no settings tampering to reach.

## Two things found while validating

1. **The live `question-freetext-answer` check was passing via the cancel path.** Its title claimed "the same turn continues", but Slack implements neither `answerOpenQuestionWithText` nor `dismissOpenCards`, so free text there *always* cancels and starts a fresh turn. Both paths settle the request and both get a reply, so a check asserting only "the agent answered" couldn't tell them apart. Renamed `question-freetext-reply`; it now discriminates via the transcript and reports which path ran (measured on Slack: `cancelled → fresh turn`). Slack's missing free-text answer path is a real gap, not a Phase 7 regression — flagged, not built.
2. **The already-handled gate is not live-reachable on Slack.** It needs an interactive response, which means a human button tap; the free-text branch produces none. Covered by the wire-contract suite instead, and written into the skill's known-limits.

## Next

Phase 8 (final legacy removal) is unchanged and unblocked by this: consumers first (browser tray, auto-approved suppress-sets, cold-start fallbacks), then re-home registration out of the broadcast intercept, then delete the nine legacy broadcasts and collapse the persister's mirror store.

### DIFF
--- a/.claude/skills/slack-chat-validation/SKILL.md
+++ b/.claude/skills/slack-chat-validation/SKILL.md
@@ -117,11 +117,27 @@ install is only ever read.**
 |---|---|
 | `inbound-turn` | An inbound Slack message creates a chat session bound to a real container session whose transcript holds the turn |
 | `question-card` | `AskUserQuestion` renders as Block Kit with one button per option and connector-registered `cb_<n>` action ids |
-| `question-freetext-answer` | A plain-text reply resolves the open single-question card as the free-form answer and the **same turn continues** — then the registry is empty |
+| `question-answered-in-app` | The card posted to Slack and the request the **app** answers are the same registry entry: answering through the app's decision route settles it and the continuation lands back in Slack |
+| `question-freetext-reply` | A plain-text reply to an open single-question card settles it and the agent responds — and the report says **which** path ran (answer-in-turn vs cancel → fresh turn) |
 | `question-cancelled-by-unrelated-message` | An unrelated message cancels the parked question instead of deadlocking behind it, and the fresh turn runs |
 | `question-multi-is-not-consumed-by-text` | A multi-question ask posts one card per question, and text does **not** answer it (an answer would be attributed to the wrong question) |
 | `file-delivery-notice` | `deliver_file` reaches the conversation as real uploaded bytes with the description as caption (text fallback accepted and reported) |
-| `unsupported-*` (6) | Per kind: the exact refusal wording, the **exact host-aware link tail**, no interactive card, and the request staying parked in the registry |
+| `unsupported-*` (7) | Per kind: the exact refusal wording, the **exact host-aware link tail**, no interactive card, and the request staying parked in the registry |
+
+`unsupported-capability-review` is the newest row and the one with the least
+margin: `workflows: 'review'` is the shipped default, so asking the agent to run
+a workflow parks a capability review — which chat was silent about entirely
+before the connectors moved onto the unified request wire. It needs the agent to
+actually reach for the `Workflow` tool, so it is the row most likely to need its
+retry.
+
+`question-freetext-reply` reports its path rather than asserting one, because
+the two are connector-dependent: the free-text **answer** path needs
+`answerOpenQuestionWithText`, which only the Telegram connector implements. On
+Slack the same message legitimately takes the **cancel** path. Both settle the
+request and both get a reply, so a check that only asserted "the agent
+answered" would stay green if the answer path broke outright — which is what
+the previous version of this check did.
 
 The `unsupported-*` checks are the regression test for the host-aware notice
 (PR #563). They rebuild the expected tail from the host shape and assert the
@@ -141,8 +157,16 @@ what stops a queued notice from linking a rotated-away session.
 - **Button taps are not automated.** Slack has no API to simulate clicking a
   Block Kit button; interactions only arrive from a real client over Socket
   Mode. The suite asserts the card is posted with the right buttons and
-  `action_id`s, and covers the answer path via the free-text branch. The
-  `action_id` → decision round trip needs a human tap in Slack.
+  `action_id`s, and covers a real decision via `question-answered-in-app` (the
+  app's decision route settles the same registry entry the Slack card came
+  from). The `action_id` → decision round trip still needs a human tap.
+- **The already-handled gate is not live-reachable on Slack.** A press on a card
+  whose request was settled elsewhere must be refused rather than buffered in
+  the container. Reaching it needs an interactive response, which on Slack means
+  a button tap — and the free-text branch does not produce one, because Slack
+  has no `answerOpenQuestionWithText`. That gate is covered instead by the
+  wire-contract suite in `chat-integration-e2e.test.ts`, which drives the real
+  persister and the real registry.
 - **`browser_input` is not asserted as parked** (`requiresParked: false`).
   `mcp__user-input__request_browser_input` is only in the web-browser
   subagent's tool list, so the main agent reaches it indirectly and the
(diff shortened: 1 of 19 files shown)


iddogino:
Addressed the [P2] atomic-claim finding in `262252e2`.

**The fix.** New `userInputRequestManager.claimRequest(id)` / `releaseClaim(id)` — a synchronous check-and-mark, so on node's single thread exactly one of two concurrent deciders gets the request and the loser gets `null`. A claim is deliberately *not* a settlement: the request stays visible to the awaiting projection, the snapshot, and the wire, because if the claimer dies a human is still waiting on it. `resolve()` drops the claim with the entry so a re-registered `toolUseId` can't inherit a stale reservation.

The chat gate now claims before any await (releasing immediately on an agent mismatch), **re-checks with zero awaits between the check and `client.fetch`**, and releases in a `finally` — a leaked claim would leave the request permanently undecidable, which is worse than the race it guards.

**An honest limit.** The claim excludes another *chat* press. It does not exclude a decision on another surface, because the app's decision routes don't claim — that's SUP-483, the identical check-then-act shape in `gateRequestDecision`, already tracked and deliberately out of Phase 7's scope. What closes the scenario you described is the late re-check: it removes every host-side yield between "is it open" and the container call. What survives is the container round trip itself, which only the container can arbitrate — it is the one place that knows whether an input is still parked. That residual is written into the code comment rather than papered over. Wiring the nine app routes onto this same primitive is now mechanical.

**Tests.** The one you asked for (settle while `ensureRunning` is pending), a leaked-claim guard, and seven registry unit tests for the primitive.

Two earlier attempts were discarded rather than kept:

- The first versions failed under revert on a `waitForCondition` timeout, not on an assertion — they proved "no reply was sent" and died before evaluating the harm. Restructured to wait on a condition *both* the fixed and broken code satisfy, so the failure is now `expected [...] to have a length of +0 but got 1` — one `/resolve` posted to a container with nothing parked.
- A "two presses at once" e2e test looked right but the harness serializes the presses (instrumented: the second press sees the request already closed), so it never exercised the concurrent window. Deleted rather than kept as a test that passes for the wrong reason; the claim's atomicity is covered deterministically at the registry level, where it is actually decidable.

Also fixed real cross-test leakage found along the way: `mockResolvedValue` sets the default but does not drain queued `*Once` implementations, so an unconsumed one from a timed-out test fired in the next.

**Verification.** typecheck clean; lint 0 errors; full unit **8,098 green**; request-path mock E2E **53 passed**; red-proof — reverting only the claim fails both race tests on the harm assertion.

On the earlier CI red: `container-integration` failed on `curl https://deb.nodesource.com/setup_22.x` returning **403** during `docker build`. I re-ran that job on the identical commit and it passed, so it was the external dependency, not this diff. Worth noting separately that the image build pulls nodesource unpinned at build time, so any hiccup there reds an unrelated PR.

PR #598, opened 2026-07-27
Fix sessions stuck on "Working" from a stale background-task level set

iddogino:
## Symptom

A prod agent on 0.4.17-rc.2 (TV Downloader) ended **every** turn stuck on "Working…" with no background work outstanding, and Stop didn't clear it. Its container background tasks:

```
17:53:32  b0l9w1jgp  start (auto-backgrounded after a 120s foreground timeout)
17:55:32  b0l9w1jgp  killed
18:56:21  bhwi35mjn  start
19:28:23  bhwi35mjn  completed     ← last background task of the session
```

Three further turns ran after that with zero live background work (19:55, 19:56, 20:22), each ending pinned. At 19:55:18 the transcript shows `[SYSTEM] This session is resuming now — the user chose to wake it early`: a scheduled sleep and **cold resume**, which replaces the CLI process.

> **Correction (was in the first draft of this PR):** a second agent (SEO) was originally cited here as the headline evidence. Re-examination showed that session was *not* wedged — a nested background subagent (`a1652462544859ca5`, "Research MCP registry downstream consumers", spawned by another subagent and outliving it) was genuinely running and completed at 21:14. Its `isActive: true` with an empty `activeBackgroundTasks` is exactly what a live snapshot-only task looks like, and the union gate was behaving correctly. That incident is a separate observability bug — a nested background subagent pins the session while being invisible in the UI — and is **not** what this PR fixes.

## Cause

`openBackgroundWorkCount()` gates every settle path on the union of two sets: the incremental `activeBackgroundTasks` map, and `bgTasksSnapshot` — the SDK's `background_tasks_changed` level signal. The union is deliberate: either side alone can miss a transition.

But only the map was ever pruned. `bgTasksSnapshot` was assigned from SDK frames and read, never retired — not by a terminal per-task signal, not by Stop, not by a re-subscribe. And the SDK only re-emits the level on a membership *change*, with nothing at all at process startup:

> The level is per-process: nothing is emitted at startup, so consumers must reset to the empty set whenever the session's CLI process (re)starts.
> — `SDKBackgroundTasksChangedMessage`, agent-sdk 0.3.219

So any id that entered the snapshot and never left it was unretirable. Every later `session_state_changed: 'idle'` took the `session_waiting_background` branch, `finalizeIdle` never ran, and the session read "Working…" forever.

`SessionSettlementTracker` in agent-container already had both directions — `removeTask()` deletes from `snapshotTaskIds`, and `resetBackgroundTasks()` is wired to `query-start` with a comment that carried-over ids "would pin the session unevictable forever". That landed in #440, one day after the host got the union in #438, and was never ported across.

## Changes

- `clearBackgroundTask()` retires the id from `bgTasksSnapshot` too, and the two terminal call sites (`task_notification`, `task_updated`) no longer gate on map membership — the snapshot carries ids the map never had, since a subagent's own nested background agents notify onto the lead session.
- `markSessionInterrupted` / `markSessionInactive` reset the snapshot alongside the map clear they already did.
- New `process_restarted` frame: the container broadcasts it on `query-start` (idle eviction + `--resume`, crash, MCP-injection restart), the host retires tracked tasks and resets the level set. `query-start` is in-process, so the host previously had no way to learn the CLI was replaced.
- Transport reattach still carries the snapshot forward — that's the same live process with the same tasks running, and dropping it would under-report real work.

## Tests

Written first, all four red, each mirroring one observed failure:

| Test | Mirrors |
|---|---|
| `terminal task signal drains the snapshot when the removal frame never arrives` | SEO agent — mirror image of the existing self-heal test, which covers the other direction |
| `Stop drains the snapshot so later turns still settle` | TV Downloader — every turn after the interrupt |
| `process_restarted drops the snapshot from the dead process` | CLI process replacement |
| `announces process replacement to subscribers so the host can reset too` | container must emit the frame at all |

## Validation

- root suite: 8060 passed, 45 skipped, 0 failed
- container suite: 774 passed, 8 skipped, 0 failed
- `tsc --noEmit` clean on both projects; eslint clean on all four changed files
- The SDK-capture replay fixtures (`sdk206-fixtures-replay`, `subagent-task-events-replay`, `background-bash-busy-completion`) still pass — the check that mattered for relaxing the two call-site guards

## Notes

- The host-side changes take effect on app restart; the `process_restarted` path needs a container image rebuild.
- Dead-subagent cards lingering after a process restart remain the separate follow-up tracked from the earlier stuck-session incident — out of scope here.

https://claude.ai/code/session_01BRHQshiwhUSM64WNmuswhr

### DIFF
--- a/agent-container/src/server.ts
+++ b/agent-container/src/server.ts
@@ -1904,10 +1904,17 @@ async function handleWebSocketConnection(ws: WebSocket, sessionId: string) {
   // session_state_changed:'idle' as the idle authority from the first turn —
   // a 'result' alone must not end the session while queued messages keep the
   // runtime going.
+  // process_instance: identity of the CLI process backing the session right
+  // now. Background tasks are process-local and a fresh process emits no
+  // initial background_tasks_changed, so a client reconnecting after a restart
+  // it never observed (idle eviction + --resume, container restart — the live
+  // process_restarted broadcast fires before any subscriber exists) would
+  // otherwise keep bookkeeping for tasks that died with the old process.
   ws.send(JSON.stringify({
     type: 'system',
     subtype: 'capabilities',
     session_state_events: true,
+    process_instance: sessionManager.getProcessInstanceId(sessionId),
     timestamp: new Date(),
   }));
 
--- a/agent-container/src/session-manager-idle-eviction.test.ts
+++ b/agent-container/src/session-manager-idle-eviction.test.ts
@@ -707,6 +707,10 @@ describe('SessionManager idle eviction', () => {
     expect(received).toContainEqual(
       expect.objectContaining({ type: 'system', subtype: 'process_restarted' })
     )
+    // The live frame and the handshake must name the SAME process, or a client
+    // that sees both would read the second one as another restart.
+    const announced = received.find((m) => m.subtype === 'process_restarted')
+    expect(announced?.process_instance).toBe(manager.getProcessInstanceId(id))
   })
 
   it('records a fresh process instance id on a cold resume, before any subscriber exists', async () => {
--- a/agent-container/src/session-manager-idle-eviction.test.ts
+++ b/agent-container/src/session-manager-idle-eviction.test.ts
@@ -54,6 +54,9 @@ class MockClaudeProcess extends EventEmitter {
       throw err
     }
     this.running = true
+    // Faithful to ClaudeCodeProcess: initializeQuery() emits this from start(),
+    // i.e. before the caller can publish the session or attach a subscriber.
+    this.emit('query-start')
   }
 
   async sendMessage(content?: string): Promise<void> {
@@ -704,6 +707,37 @@ describe('SessionManager idle eviction', () => {
     expect(received).toContainEqual(
       expect.objectContaining({ type: 'system', subtype: 'process_restarted' })
     )
+  })
+
+  it('records a fresh process instance id on a cold resume, before any subscriber exists', async () => {
+    // Regression: process.start() emits query-start synchronously, BEFORE
+    // doResumeSession publishes the SessionData and long before the WebSocket
+    // subscriber attaches — so the live announcement reaches nobody. The host
+    // then reconnects still holding a bgTasksSnapshot from the dead process,
+    // the fresh CLI emits no initial snapshot, and the session is pinned
+    // "working" forever. The restart has to be readable at subscribe time, not
+    // only broadcast at restart time.
+    persistedSessions.set('cold-restart', {
+      sessionId: 'cold-restart',
+      claudeSessionId: 'cold-restart',
+      workingDirectory: workDir,
+      createdAt: new Date().toISOString(),
+      lastActivity: new Date().toISOString(),
+    })
+
+    await manager.getSession('cold-restart')
+
+    const afterResume = manager.getProcessInstanceId('cold-restart')
+    expect(afterResume).toBeTruthy()
+
+    // Same live process → the same id, so a plain reattach must not be
+    // mistaken for a restart (that would drop a genuinely running task).
+    expect(manager.getProcessInstanceId('cold-restart')).toBe(afterResume)
+
+    // A later replacement mints a new one.
+    const proc = spawnedProcesses[spawnedProcesses.length - 1]
+    proc.emit('query-start')
+    expect(manager.getProcessInstanceId('cold-restart')).not.toBe(afterResume)
   })
 
   it('a shouldQuery:false append does NOT promote an automated session', async () => {
--- a/agent-container/src/session-manager-idle-eviction.test.ts
+++ b/agent-container/src/session-manager-idle-eviction.test.ts
@@ -689,6 +689,23 @@ describe('SessionManager idle eviction', () => {
     expect(proc.stopCalls).toBe(1)
   })
 
+  it('announces process replacement to subscribers so the host can reset too', async () => {
+    // The host keeps its own copy of the background-task level set and has no
+    // other way to learn the CLI was replaced — `query-start` is an in-process
+    // event. Without this frame the host's snapshot keeps ids from the dead
+    // process forever and pins the session "working" at the end of every turn.
+    const { id, proc } = await createIdleSession()
+    const received: Array<Record<string, unknown>> = []
+    const unsubscribe = manager.subscribe(id, (msg) => received.push(msg as Record<string, unknown>))
+
+    proc.emit('query-start')
+    unsubscribe()
+
+    expect(received).toContainEqual(
+      expect.objectContaining({ type: 'system', subtype: 'process_restarted' })
+    )
+  })
+
   it('a shouldQuery:false append does NOT promote an automated session', async () => {
     const promoManager = new SessionManager(workDir, {
       prewarmEnabled: false,
--- a/agent-container/src/session-manager.ts
+++ b/agent-container/src/session-manager.ts
@@ -321,6 +321,10 @@ export class SessionManager extends EventEmitter {
         stateEventsAuthority: true,
       }),
       eviction: null,
+      // Seeded, not derived from query-start: this path starts the process
+      // before SessionData exists (see the await process.start() above), so the
+      // first query-start predates the listener registered below. Every later
+      // replacement re-mints, which is all the host needs to spot a change.
       processInstanceId: uuidv4(),
     };
 
--- a/agent-container/src/session-manager.ts
+++ b/agent-container/src/session-manager.ts
@@ -331,6 +331,7 @@ export class SessionManager extends EventEmitter {
     // initial snapshot — carried-over ids would pin the session forever.
     process.on('query-start', () => {
       sessionData.settlement.resetBackgroundTasks();
+      this.announceProcessRestart(sessionId);
     });
 
     process.on('stderr', (error: string) => {
@@ -618,6 +619,7 @@ export class SessionManager extends EventEmitter {
 
       process.on('query-start', () => {
         data.settlement.resetBackgroundTasks();
+        this.announceProcessRestart(sessionId);
       });
 
       process.on('stderr', (error: string) => {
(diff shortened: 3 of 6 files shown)


iddogino:
Confirmed and fixed in b7caa1f7 — the P1 was real, and worse than a dropped notification.

`this.broadcast()` starts with `this.sessions.get(sessionId)`, and on a cold resume `process.start()` runs at `session-manager.ts:641` while `this.sessions.set()` is at `:648`. So the frame wasn't just reaching an empty subscriber set — the session lookup missed entirely and it returned early. Same hazard in `createSession` (process starts inside the builder, `sessions.set` after). Exactly the eviction + `--resume` path this PR is meant to cover.

## Fix: durable process identity, replayed at subscribe time

- `SessionData.processInstanceId`, re-minted on every `query-start`. On a cold resume the listener stamps `data` before it's published, so the id is already correct by the time anything can read it — the ordering stops mattering.
- The WebSocket `capabilities` handshake carries it as `process_instance`. That's the path that actually covers restarts nobody was attached for.
- The host records it and drops its process-local background bookkeeping when the name changes.
- The live `process_restarted` frame stays, for restarts we *are* attached for, and resets unconditionally — there the container is asserting the replacement outright.

Identity rather than a generation counter, deliberately: the only question is "same process my snapshot came from?", and a counter would have to survive `SessionData` being rebuilt on resume — a fresh `ClaudeCodeProcess` restarts its own `queryGeneration` at 1, so post-resume it would collide with the pre-resume value and read as *no* change. Identity needs no persistence to stay correct.

The first-sighting guard matters in the other direction: a first handshake is not a restart, so a plain transport reattach (same id, same tasks still running) keeps the snapshot rather than dropping a live indicator and un-gating auto-sleep mid-job. Both directions are now tested.

## Tests

The container mock's `start()` now emits `query-start` the way `ClaudeCodeProcess.initializeQuery()` does, so the ordering hazard is genuinely modeled instead of assumed away — that alone made the new cold-resume test fail before the fix.

- `records a fresh process instance id on a cold resume, before any subscriber exists`
- `a capabilities handshake naming a new process instance drops the stale snapshot`
- `a reattach to the SAME process instance keeps running tasks tracked`

## Validation

- root suite: 8062 passed, 0 failed
- container suite: 775 passed, 0 failed
- `tsc --noEmit` clean on both projects; eslint clean on all changed files

iddogino:
Re-examined the whole PR through the lens of the cold-resume finding — that bug was "host mirrors process-local container state, and the signal that invalidates it can be dropped," so I went looking for the same shape elsewhere. Found one more, fixed in 80b7dd0a, plus one pre-existing issue outside this PR's scope that I'd rather report than silently bundle.

## Found: the restart signal is also dropped on interrupt (fixed)

`claude-code.ts:1649` restarts the query on the interrupt path ("Restarting query after interrupt"), which mints a new process instance. But `handleMessage` drops every non-`result` frame while `isInterrupted` is set — so that announcement was swallowed too.

The consequence is the **inverse** of the original bug, reached through the signal I just added:

1. Stop → `isInterrupted = true`, identity still `proc-A`
2. container restarts the CLI → `process_restarted(proc-B)` → **dropped**
3. next message clears `isInterrupted`; new turn registers real background work on `proc-B`
4. any transport reattach → handshake says `proc-B` ≠ recorded `proc-A` → read as a restart → **drops tasks that are genuinely running**

That's a live indicator lost and auto-sleep un-gated mid-job — precisely what the union gate exists to prevent. The new test fails on the old code with `background_task_completed` fired for a running task.

Fix: `process_restarted` passes the interrupt guard for the same reason `result` does — it's a fact about which runtime we're talking to, not turn content. It only touches background-task bookkeeping and process identity, both already cleared by the interrupt, so letting it through is idempotent.

## Also in this commit

- **`MockContainerClient` now sends a stable per-session `process_instance`.** Previously it sent none, so every E2E reattach no-opped and the contract had zero end-to-end coverage. Stable, not fresh-per-subscribe — a new id each time would make every reconnect look like a restart.
- **`createSession` seeds `processInstanceId` rather than deriving it from `query-start`**, now documented: that path starts the process before `SessionData` exists, so the first `query-start` predates the listener. Later replacements re-mint, which is all the host needs.
- The container test now asserts the live frame and the handshake name the *same* process, so a client seeing both can't read the second as another restart.

## Audit notes — things I checked that are fine

- **Mirror of the P1 on the host**: `streamingStates.set` precedes `client.subscribeToStream`, so state always exists when the handshake lands. No equivalent ordering hole.
- **Relaxed terminal-signal guards**: `task_id` + top-level `status` only ever co-occur on `task_notification`, whose status is terminal by type. Running before the sidechain filter is deliberate and correct — nested background agents notify onto the lead session, and those ids belong to the lead's process-level set.
- **Self-heal loop interaction**: `clearBackgroundTask`'s new snapshot delete is a no-op inside the `background_tasks_changed` handler, since the id is by definition absent from the snapshot just assigned.
- **Renderer**: `dropProcessLocalBackgroundState` retires through `clearBackgroundTask`, so `background_task_completed` still fires per task and `isWaitingBackground` clears when the list empties. No new SSE event type.
- **`activeSubagents` does not gate settling** (completion-detection only), so leftover subagent cards after a restart can't pin a session. Still the separate follow-up.
- **Version skew**: the agent image tag is pinned to the app version and auto-migrates on upgrade, so an old-container/new-host window is bounded by the image pull rather than open-ended.

## Reporting, not fixing: `lastResultSubtype` is dropped on re-subscribe

`subscribeToSession` carries `isActive: priorIsActive` forward but resets `lastResultSubtype: null` unconditionally. The authoritative-idle branch is gated on `state.isActive && state.lastResultSubtype !== null`, so a transport reattach landing between a turn's `result` and its `session_state_changed: 'idle'` permanently discards that turn's ability to finalize — the session reads "Working…" until the *next* turn completes.

Same family (state carried inconsistently across re-subscribe), same user-visible symptom, and it plausibly co-explains the SEO session I diagnosed — that one never got another turn, so it had nothing to self-heal on. It cannot explain the TV Downloader's every-turn pattern, which is why it isn't the main cause here.

It's a distinct mechanism from the level-set drain, and it self-heals on the next completed turn rather than pinning forever. Happy to fix it here or as a follow-up — say which.

## Validation

- root suite: 8063 passed, 0 failed
- container suite: 775 passed, 0 failed
- `tsc --noEmit` clean on both projects; eslint clean on all changed files
- E2E (mock, local): 277 passed / 8 failed — all 8 fail in `beforeEach` on agent-sidebar visibility, spread across unrelated specs, and all 21 tests in those files pass on a serial re-run. Local parallel-worker contention, not a regression; CI is the authority here.

iddogino:
**CI red is an external outage, not this branch** — leaving a note so nobody chases it.

`container-integration`, `happy-path`, and `deploy-report` all fail on the same thing:

```
agent-container/Dockerfile:38
RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash - && apt-get install -y nodejs
  → curl: (22) The requested URL returned error: 403
  → E: Unable to locate package nodejs
```

`deploy-report` is a cascade — it downloads the `e2e-results` artifact `happy-path` never produced.

**It is time-based, not branch-based.** Image-build outcome across recent runs:

| Time (UTC) | Branch | Result |
|---|---|---|
| 21:45 | **this branch** (b7caa1f7) | container-integration ✅ happy-path ✅ |
| 22:07 | fix/sup-454-skillset-safe-directory | ✅ |
| 22:37 | fix/dashboard-startup-states | ✅ |
| 22:37 | cloud-proxy-route | ✅ |
| 22:42 | **this branch** (80b7dd0a) | ❌ 403 |
| 22:58 | **this branch** (re-run, no code change) | ❌ 403 |

This branch passed both jobs an hour earlier with the same Dockerfile. The only change since is 80b7dd0a — six TypeScript files. The failing step is layer `[stage-0 3/23]`; the first `COPY` of app source is line 106 and `COPY src ./src` is line 138, so the build dies 68 lines before any code from this PR enters the image.

Every job that actually exercises the change is green: `unit-tests`, `container-tests`, `e2e-tests`, `e2e-auth-tests`, `e2e-a11y-audit`, `build-mac`, `build-win`.

Since a clean re-run reproduced it, this will block any push to main too once one lands. That step is an unpinned network fetch in the critical path of every image build — worth hardening separately (retry/backoff, or the official Node apt repo / a `node:22` base layer), but that's a repo-wide change and deliberately not part of this PR.

iddogino:
**Correction to the evidence in this PR** — the PR body has been updated.

The original write-up cited two prod incidents. On re-examination only one of them is this bug.

**SEO agent — not this bug.** I diagnosed that session live and found `isActive: true` with an empty `activeBackgroundTasks`, and read it as a stale level set. It wasn't. A nested background subagent — `a1652462544859ca5`, "Research MCP registry downstream consumers" — was genuinely running and completed at 21:14 with a full report. It appears in the lead transcript *only* as a completion notification, never as a launch, because it was spawned by the "Verify MCP registry publish flow" agent and outlived it.

`isActive: true` + empty map is precisely what a live snapshot-only task looks like: the lead never saw its `async_launched` ack, so it was only ever in the SDK snapshot. The union gate was working as designed. My analysis had stopped at line 851 of what is now an 1837-line transcript.

That incident is a real problem, but a different one: **a nested background subagent holds the session "Working" while being invisible in the UI** — the only card shown was a completed ✓ subagent. Observability, not settlement. Worth its own issue.

**TV Downloader — holds up.** Container background tasks:

```
17:53:32  b0l9w1jgp  start (auto-backgrounded after a 120s foreground timeout)
17:55:32  b0l9w1jgp  killed
18:56:21  bhwi35mjn  start
19:28:23  bhwi35mjn  completed     ← last background task of the session
```

Three turns ran after that with zero live background work (19:55, 19:56, 20:22), each reported stuck. And at 19:55:18: `[SYSTEM] This session is resuming now — the user chose to wake it early` — a scheduled sleep and cold resume, i.e. exactly the process-replacement path where the SDK emits no initial snapshot and the host carried a stale one. That is the case this PR fixes, and the same path the P1 review catch was about.

**What this does and doesn't change.** The defect itself is established independently of either incident — from the code, the SDK's documented "reset to the empty set whenever the CLI process (re)starts" contract, and `SessionSettlementTracker`, which already implements both drains container-side. The fix and its tests are unchanged. What changes is the strength of the field evidence: one confirmed reproduction rather than two.

Still not eliminated as a contributing path: `subscribeToSession` resets `lastResultSubtype` to `null` while carrying `isActive`, so a reattach landing between a turn's `result` and its `idle` discards that turn's finalization. It self-heals on the next turn, so it fits the every-turn pattern less well — but it's untested either way and remains a separate reported issue.

PR #600, opened 2026-07-27
Wait for dashboards through serialized startup (SUP-485)

JeremyBischoff:
https://linear.app/datawizz/issue/SUP-485/dashboard-view-treats-a-transient-startup-state-as-terminal

## Problem
Dashboards start serially inside the agent container.
While a dashboard is waiting behind its siblings, it reports `stopped`.
The renderer treated that transient state, and an unloaded artifact list, as a terminal failure.

## State behavior
| Input | User-facing result |
| --- | --- |
| Agent stopped and start allowed | Starting agent with spinner |
| Agent stopped and start forbidden | Ask an admin to start it |
| Agent start failed | Failure with retry |
| Artifact list not loaded | Waiting with spinner and 1-second polling |
| Loaded list does not contain dashboard | Dashboard not found |
| Dashboard stopped and queued | Waiting with spinner and 1-second polling |
| Dashboard starting | Waiting with spinner and 1-second polling |
| Dashboard crashed | Crashed state with restart action |
| Dashboard running, iframe loading | Waiting with spinner |
| Dashboard running, iframe loaded | Dashboard displayed |
| Any wait reaches 120 seconds | Slow-start message with restart action |

## Implementation
- Centralize the presentation policy in a pure `resolveDashboardViewState` function.
- Keep cached artifact data usable through background refetch failures.
- Poll artifacts quickly only while startup progress is expected.
- Keep the loading overlay visible until the iframe document loads.
- Reset dashboard-local state when navigating between dashboards.
- Stop pop-out polling immediately when a dashboard is missing or crashed.
- Prevent a deliberate restart from racing the view's automatic start.

The restart action reuses the existing agent stop and start APIs.
It restarts the entire agent container and therefore all dashboards for that agent.
No new backend restart route or container behavior was added.

## Reviewer guide
- State policy and boundary: `dashboard-view-state.ts`
- Renderer wiring and restart flow: `dashboard-view.tsx`
- Polling cadence: `use-artifacts.ts`
- Pop-out terminal handling: `agents.ts`
- Regression coverage: the three new co-located test files

## Scope
- No changes to dashboard launch order
- No new artifact statuses or API shapes
- No database or dependency changes

## Test plan
- [x] Dashboard state, polling, and restart-race unit tests
- [x] Typecheck
- [x] Lint
- [x] Headed visual state carousel
- [x] Real two-dashboard cold boot through Apple container
- [x] Real restart button stop/start recovery
- [x] GitHub CI: 10 passed, 0 failed

### DIFF
--- a/src/renderer/components/dashboards/dashboard-view.tsx
+++ b/src/renderer/components/dashboards/dashboard-view.tsx
@@ -31,6 +31,7 @@ export function DashboardView({ agentSlug, dashboardSlug }: DashboardViewProps) 
   const [restartError, setRestartError] = useState<string | null>(null)
   const iframeRef = useRef<HTMLIFrameElement>(null)
   const waitStartedAtRef = useRef<number | null>(null)
+  const autoStartedRef = useRef<string | null>(null)
   const { data: agent } = useAgent(agentSlug)
   const { data: artifacts } = useArtifacts(agentSlug, { pollFast })
   const startAgent = useStartAgent()
@@ -106,6 +107,8 @@ export function DashboardView({ agentSlug, dashboardSlug }: DashboardViewProps) 
   }, [startAgent, agentSlug])
 
   const handleRestartAgent = useCallback(async () => {
+    // The deliberate stop must not re-trigger this view's auto-start effect.
+    autoStartedRef.current = agentSlug
     setRestarting(true)
     setRestartError(null)
     setIframeLoaded(false)
@@ -123,7 +126,6 @@ export function DashboardView({ agentSlug, dashboardSlug }: DashboardViewProps) 
     }
   }, [isAgentRunning, stopAgent, startAgent, agentSlug])
 
-  const autoStartedRef = useRef<string | null>(null)
   useEffect(() => {
     if (autoStartedRef.current === agentSlug) return
     if (!agent || isAgentRunning || isAgentStarting || !canStart) return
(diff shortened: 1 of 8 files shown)


src/renderer/components/dashboards/dashboard-view.tsx
```diff
@@ -58,38 +134,24 @@ export function DashboardView({ agentSlug, dashboardSlug }: DashboardViewProps)
...
-          ? 'Agent is not running. Ask an admin to start it.'
-          : 'Starting up...'
-      : dashboard?.status === 'starting'
-        ? 'Dashboard is starting up...'
-        : 'Dashboard is not running. It will start automatically when the agent starts.'
+  const showFrame = isAgentRunning && dashboard?.status === 'running'
+  const showOverlay = viewState.kind !== 'ready'
+  const actionPending = restarting || stopAgent.isPending || startAgent.isPending
+
+  if (!showFrame) {
```
iddogino, line 147: iframeLoaded is only cleared for URL changes, refreshes, and this component's restart action. If the agent or dashboard leaves running through another control, the early return unmounts the iframe while iframeLoaded remains true. When it runs again, the resolver immediately returns ready and exposes the remounted iframe before its new document loads. Clear the flag whenever showFrame becomes false or key it to each iframe mount.

iddogino (changes requested):

src/renderer/components/dashboards/dashboard-view.tsx
JeremyBischoff: @iddogino done in `83a27cfe` - clearing the flag whenever showFrame goes false, so a remounted iframe waits for its own load.

PR #601, opened 2026-07-27
Delete the legacy per-type request wire and its mirror store (Phase 8)

iddogino:
Final phase of the user-input request rearchitecture. Phases 0–7 moved the renderer, notifications, Electron main, the decision routes and the chat connectors onto the registry. This deletes what they left behind: the legacy per-type SSE events, their verbatim replay, and the persister's mirror store.

The registry is now the only pending-request store, and `user_request_created` / `user_request_resolved` the only request wire.

Order is consumers → emission → store, so emission was provably dead before it was removed.

## Consumers

- **Browser tray.** Reads a new `usePendingBrowserInputRequests` — the same unified snapshot the in-chat cards project, filtered to the `browser_input` bucket, gated on `isActive` (exactly what `session_idle`/`session_error` did to the array it replaces) with a local dismissed-set so answering in the tray drops the overlay synchronously. A request the registry recovered, or one settled on another surface, now reaches the tray too. Accepted tradeoff, as planned: the overlay appears one snapshot refetch after the created event rather than synchronously with it.
- **Auto-approved suppress-sets.** Fed from `user_request_created` with `request.autoApproved`, keyed by kind. They stay outside React state for the original reason — the write has to be synchronous with the event to beat the fallback-card flash.
- **Cold-start fallbacks deleted.** `pendingCapabilityReviewRequests` + handler, the `capability_review_resolved` event + handler, the `hasSnapshot` fork, `usePendingProxyReviews`, `GET /:id/proxy-reviews`, and every `['proxy-reviews']` invalidation. The `PendingReview` type moved next to `reviewFromEnvelope`, now its only producer. Safe because a card's own created event invalidates the snapshot query — the event that announces a request is also what fetches it; the deleted path only covered the case where that refetch *also* fails.

## Emission and store

- **Registration re-homed.** Out of the `broadcastToSSE` intercept, into the eight per-kind handlers via a shared `registerStreamRequest`; `REQUEST_EVENT_TYPE_BY_TOOL_NAME` becomes `REQUEST_KIND_BY_TOOL_NAME` and recovery synthesizes envelopes through the same helper. What remains on the broadcast path is the turn-boundary settle and nothing else.
- **Legacy emission deleted.** The nine per-type broadcasts, both replay loops in the `/stream` route, `review-broadcast.ts` and all four `broadcastReview` callsites, and `ComputerUseRequestEvent`. `broadcastReview`'s only remaining job — nudging the sidebar for an agent with no active session — is covered by the global `user_request_created`, which the same handler already invalidates on. Note the `UserRequestEvent` union in `tool-definitions/types.ts` **stays**: since Phase 7 it is the chat *card* shape, not a wire shape.
- **Mirror store collapsed.** `pendingInputRequests`, `getPendingInputRequests`, `verifyStoreParity`, `shadowRegistryCheck`, `streamEventToPendingRequest`, `STREAM_EVENT_TYPE_TO_KIND`. The turn-cancel sweep reads `getStoreIdsForSession(…, 'stream')` — which keeps recovered synthetics, the exact property the old raw read existed for. The sidechain and tool_result paths key off `resolveIfInStore`'s return instead of a store-membership check; `capability_review_cancelled` asks the registry. `storeForKind` survives — it expresses the per-class clearing rules, not the mirror.

## Tests

Rewritten, not dropped, wherever the behaviour survived. The characterization suite reads `user_request_created` envelopes and the registry instead of per-type events and the replay store; `review-manager.test.ts` asserts on registry transitions instead of a `broadcastReview` spy. Only assertions about deleted mechanisms went (store parity, the cold-start fallbacks, the `proxy-reviews` invalidation). The E2E helper `waitForPendingProxyReview` moved onto the snapshot along with the route it polled.

## Validation

- `npx tsc --noEmit` clean; ESLint 0 errors (14 pre-existing warnings, none in touched files)
- Unit: **8,094 passed** across 450 files
- Request-path mock E2E gate: **64/64 green**, three separate runs — including `pending-request-reconnect` and `concurrent-tabs-request-sync`, the two specs written for exactly this wire, plus a `chat-flow` canary

Two full-suite runs on my machine failed 9–11 tests scattered across unrelated specs. That was load, not the diff: one failure timed out inside `page.waitForFunction(() => setTimeout(4000), { timeout: 5000 })` — a 4-second sleep missing a 5-second budget — with load average 21/37/37 from unrelated system churn. Re-running the same specs at `--workers=2` on that same loaded machine: 64/64 green. Unrelated but worth knowing: `.e2e-data` is never pruned between runs, and at ~244 accumulated agent directories `/api/agents` (which parses every agent's CLAUDE.md) gets slow enough to time out sidebar waits in `beforeEach`.

## New: live host-restart recovery probe

Transcript recovery is the one path with no mock-E2E coverage — it only fires when the host has no in-memory record of a request but the transcript shows an unresolved blocking tool call, which is what a restart produces and what a mock run never does. It matters more now that a reconnecting client has no per-type events to replay.

`e2e/live/host-restart-recovery.mjs` boots the real app against an isolated seeded data dir, parks a real `AskUserQuestion` on a real container, kills the host, boots it again.

**It fails, and it fails identically on `main` (3b447e43).** The transcript still shows the unresolved ask, but the restarted host neither recovers the request nor reports the session awaiting — it never resubscribes to the session, so `isSessionActive` is false and the `recoverSessionAwaitingInput` hook on the messages route never fires. Not a Phase 8 regression; a standing gap in restart handling that is now measurable. Committed red-on-purpose with that documented in its header. Deserves its own ticket.

## Follow-ups not in this PR

- Host-restart session recovery (the probe's finding)
- SUP-483 — wiring the nine app-side decision routes onto the `claimRequest` primitive added in #596


iddogino:
Both findings were real. Fixed in c38de1af.

## [P1] Auto-approved requests reappearing as approval cards

Confirmed, and reproduced exactly as you described — one descriptor instead of zero, for **both** kinds:

```
× suppresses an auto-approved computer_use from the snapshot alone (no live event seen)
× suppresses an auto-approved script_run from the snapshot alone (no live event seen)
AssertionError: expected [ { kind: 'computer_use', …(7) } ] to have a length of +0 but got 1
AssertionError: expected [ { kind: 'script_run',   …(6) } ] to have a length of +0 but got 1
```

The projection dropped auto-approved envelopes from the buckets (correct — no card is owed) but *also* discarded the fact that they were auto-approved, and suppression was fed only by the one-shot `user_request_created`. So a client that mounted or reconnected after that event had an empty suppression set while the message-history and streaming fallbacks happily rebuilt the card from the unresolved tool call.

Worth noting this is a Phase 8 regression for `script_run` specifically: the `/stream` route used to replay `script_run_request` verbatim on reconnect *including* `autoApproved: true`, which repopulated the set. Deleting that replay removed the only reconnect-safe feeder. (`computer_use` never had it — `getPendingComputerUseRequests` has always filtered `autoApproved` out of its replay — so that half was pre-existing.)

Fix is the one you prescribed: `projectUnifiedRequests` now returns the auto-approved ids it drops, and suppression is the union of the live set and the snapshot's.

```ts
const suppressedComputerUseIds = useMemo(
  () => new Set([...autoApprovedComputerUseIds, ...unified.autoApprovedComputerUseIds]),
  [autoApprovedComputerUseIds, unified.autoApprovedComputerUseIds],
)
```

Both regression tests are committed and fail without the change.

## [P1] Unvalidated recursive delete + leftover credentials

Also correct, and worse than clutter — the seed copies `settings.json` (live LLM/provider API keys) and `host-container-tokens.json`.

- Default target is now `mkdtempSync` under the OS temp dir; `--target` is still accepted but validated.
- `assertSafeTarget` refuses the source data dir, any ancestor of it, the home directory, any ancestor of the repo, and the filesystem root. An existing non-empty directory is only reusable if it carries the probe's own marker file.
- Nothing is recursively deleted unless that marker is present.
- The seeded copy is removed in `finally`, credentials with it. `--keep-data` opts out and says so in the log.

Verified against the exact paths you named:

```
/Users/iddogino                                 => REFUSED: ...is the source data dir or an ancestor of it
~/Library/Application Support/Superagent-dev    => REFUSED: ...is the source data dir or an ancestor of it
~/Library/Application Support                   => REFUSED: ...is the source data dir or an ancestor of it
~/Downloads                                     => REFUSED: not empty and carries no marker
```

I also deleted the two seeded directories my earlier runs had left in `~/Downloads`, both of which contained a copy of `settings.json`.

Re-ran the probe end-to-end afterwards: same 3/5 result (the pre-existing restart-recovery gap this PR documents), now ending with `removed seeded data dir /var/folders/.../host-restart-recovery-r9P4Ax`.

## Validation after the fixes

- `npx tsc --noEmit` clean; ESLint 0 errors (14 pre-existing warnings, none in touched files)
- Unit: **8,096 passed** across 450 files, 0 failures (+2 = the new regression tests)
- Request-path E2E gate at `--workers=2`: **39 passed**, 0 failures

## Unrelated: CI is red on `container-integration`

Not this diff — the container image build fails at `Dockerfile:38`:

```
#10 [stage-0 3/23] RUN curl -fsSL https://deb.nodesource.com/setup_22.x | bash - && apt-get install -y nodejs
#10 0.417 curl: (22) The requested URL returned error: 403
         E: Unable to locate package nodejs
```

NodeSource is returning 403 to the runners. It failed on the first run and again on a re-run, so it is not a transient blip this time. Same failure hit #596. Everything else is green: unit-tests, e2e-tests, e2e-auth-tests, e2e-a11y-audit, container-tests, happy-path, build-mac, build-win.

iddogino:
### [P2] Cleanup misses setup failures — fixed in `807ea810`

Confirmed, and reproduced exactly as described. Pointing `--source` at a directory with no `superagent.db`:

```
[restart-recovery] Error: ENOENT: no such file or directory, copyfile '…/fake-source/superagent.db' -> '…/T/host-restart-recovery-N2eAoi/superagent.db'
    at seed (…/host-restart-recovery.mjs:193:3)
    at main (…/host-restart-recovery.mjs:341:3)
```

and what survived:

```
-rw-r--r--  .host-restart-recovery-probe
-rw-r--r--  host-container-tokens.json
-rw-r--r--  settings.json      ← grep -c "sk-ant" → 1
drwxr-xr-x  agents/
```

`seed()` writes the marker at line 179 and copies `settings.json` at line 183, so *every* failure after that point was leaking credentials — the database copy is just the easiest one to trigger.

Three changes:

- **Setup moved inside the cleanup scope.** `seed()`, `removeProbeContainer()`, `freePort()` and the first `startApp()` now run in the `try`, so a throw from any of them reaches the `finally`.
- **`app` is nullable.** The `finally` distinguishes "never booted" (nothing to stop, but the seeded dir still has to go) from "booted and running". `--keep-app` only short-circuits cleanup when there is actually an app to keep — previously a null `app` would have thrown on `app.base` inside the `finally` and masked the original error.
- **A boot that never answers kills its own process group.** `startApp` spawns detached and `waitFor` swallows probe throws, so a vite that forked but never served would time out at 180s with the child alive — holding the port and reading a data dir the caller is about to delete. It now `SIGKILL`s the group before rethrowing.

Same repro against the fix:

```
[restart-recovery] removed seeded data dir /var/…/T/host-restart-recovery-EMwCRx
[restart-recovery] Error: ENOENT: … copyfile '…/superagent.db'
```

Cleanup runs first, the error still propagates and still sets a non-zero exit. Re-verified the `assertSafeTarget` guards all still refuse the source data dir, `$HOME`, `/`, and the repo root. The `RUN_DIR` log directories that remain in `$TMPDIR` hold vite logs only — scanned, no credentials.

### Tray-hook coverage

You flagged this as non-blocking, but `usePendingBrowserInputRequests` is new in this PR and had no direct test, so it should ship with one. Four added, each proven red against the specific behaviour it claims:

| test | mutation that fails it |
|---|---|
| projects `browser_input` out of the snapshot | bucket → `scriptRunRequests` |
| shows nothing while the session is inactive | delete the `isActive` gate |
| drops the overlay synchronously on dismiss | `dismiss` filter → `() => true` |
| forgets dismissals at the idle boundary | same |

The dismiss tests matter because the snapshot still lists the request when the tray answers it — the overlay has to go before the server round-trip settles it.

### Validation

`tsc` clean · ESLint 0 errors · focused suites **482 passed / 5 files** (`use-pending-requests` now 54, up from 50).

The host-restart probe stays a manual, intentionally-red probe — agreed it is not a CI regression test, and it can't be until the gap it measures is fixed. That is one of the two follow-ups listed in the description.

The earlier commit `9a9579ac` is unrelated to the review: it drops comments that narrated the Phase 8 edit rather than the code, including one on the pending-requests route that still claimed the legacy per-type events fire.

PR #576, opened 2026-07-24
fix: host-owned MicroVM idle stop and suspended UI status (SUP-474)

yiw190:
## Summary
- Host `stop()` / AutoSleep always Suspend (drop the auto-sleep no-op that left AWS as the only idle owner).
- Neutralize AWS product-idle: `maxIdleDurationSeconds` = lifetime cap (not `autoSleepTimeoutMinutes`); keep auto-resume + suspended retention (AWS IdlePolicy is all-or-nothing).
- Map `SUSPENDED` / `SUSPENDING` to UI `stopped`; warm-resume the same VM on `start()` / `/health` without a new `RunMicrovm`.

Stacked on #575 (activity clock) so AutoSleep’s RAM path actually drives MicroVM suspend.

**Supersedes** https://github.com/SkillfulAgents/SuperAgent/pull/568 (together with #575).

## Test plan
- [x] `vitest` — `lambda-microvm-runtime.test.ts` (suspend on auto-sleep stop, SUSPENDED→stopped, warm start)
- [ ] E2E (optional): suspend → UI stopped → `/health` auto-resumes without RunMicrovm
- [ ] Manual: auto-sleep stops a MicroVM agent; dashboard start warm-resumes

PR #613, opened 2026-07-29
fix: terminate MicroVMs on stop (no suspend) + dead-generation replace

yiw190:
## Summary
- **Stop / auto-sleep terminate** instead of suspend — MicroVM matches other runners (no warm resume). ~3s resume win was not worth suspend complexity / quota cost.
- AWS idlePolicy: `maxIdle` = lifetime, `autoResumeEnabled: false` (host owns idle via terminate).
- Keep **dead-generation replace** on `createSession` for TERMINATING/TERMINATED races (lifetime cap / concurrent teardown) — still useful without suspend.

Supersedes the suspend path that [#576](https://github.com/SkillfulAgents/SuperAgent/pull/576) targeted (closed). Staging canary verified on `yitian-superagent`.

## Bug / repro (original #613)
1. Agent MicroVM idle → eventually TERMINATING / gone while host still holds the old generation.
2. User creates a session → resume/connection failure → UI "Failed to create session".

## What changed
- `stop()` / auto-sleep → `teardown()` (terminate); drop `SuspendMicrovm` / `MICROVM_SUSPENDED_SECONDS`.
- `getInfoFromRuntime`: RUNNING only is running; SUSPENDED leftovers terminated (CAS-safe with #613 cleanup).
- `createSession`: ensure live generation; on unreachable post-create, replace once and retry.

## Test plan
- [x] `npx vitest run src/shared/lib/container/lambda-microvm-runtime.test.ts` (55)
- [x] Staging: `yitian-superagent` on `microvm-no-suspend-*` image — stop terminates (no suspend)
- [ ] Staging: force TERMINATING wake still replaces + creates session once

### DIFF
--- a/src/shared/lib/container/base-container-client.ts
+++ b/src/shared/lib/container/base-container-client.ts
@@ -1202,11 +1202,14 @@ export abstract class BaseContainerClient extends EventEmitter implements Contai
         )
       }
 
-      // Handle network errors with user-friendly messages
+      // Handle network errors with user-friendly messages.
+      // Keep the original as cause so callers can distinguish refused (never
+      // delivered) from reset/timeout (prompt may already be running).
       if (this.isConnectionError(err)) {
         this.handleConnectionError()
         throw new Error(
-          'Failed to start session - unable to connect to the agent. Please check that the agent is running and try again.'
+          'Failed to start session - unable to connect to the agent. Please check that the agent is running and try again.',
+          { cause: err },
         )
       }
 
--- a/src/shared/lib/container/container-manager.ts
+++ b/src/shared/lib/container/container-manager.ts
@@ -108,6 +108,9 @@ class ContainerManager {
             })
           })
         },
+        // MicroVM dead-generation replace (and similar) must restart through the
+        // manager so starts share startingAgents and rebuild env from the DB.
+        restartAgent: () => this.restartAgent(agentId),
       }
 
       client = createContainerClient(config)
@@ -115,6 +118,28 @@ class ContainerManager {
     }
 
     return client
+  }
+
+  // Single-flight restart used by runtime clients that tear down a dead generation.
+  private async restartAgent(agentId: string): Promise<void> {
+    if (this.stoppingAgents.has(agentId)) {
+      throw new Error(`Cannot restart agent ${agentId} while it is stopping`)
+    }
+    const inflight = this.startingAgents.get(agentId)
+    if (inflight) {
+      await inflight
+      return
+    }
+
+    this.markAsStopped(agentId)
+    const client = this.getClient(agentId)
+    const startPromise = this.doStartContainer(agentId, client)
+    this.startingAgents.set(agentId, startPromise)
+    try {
+      await startPromise
+    } finally {
+      this.startingAgents.delete(agentId)
+    }
   }
 
   /**
--- a/src/shared/lib/container/container-manager.ts
+++ b/src/shared/lib/container/container-manager.ts
@@ -132,6 +132,14 @@ class ContainerManager {
     }
 
     this.markAsStopped(agentId)
+    // Dead generation had live sessions; clear isActive before the new start
+    // (same as stopContainer — replace does not go through stopContainer).
+    messagePersister.markAllSessionsInactiveForAgent(agentId)
+    messagePersister.broadcastGlobal({
+      type: 'agent_status_changed',
+      agentSlug: agentId,
+      status: 'stopped',
+    })
     const client = this.getClient(agentId)
     const startPromise = this.doStartContainer(agentId, client)
     this.startingAgents.set(agentId, startPromise)
--- a/src/shared/lib/container/lambda-microvm-runtime.e2e.test.ts
+++ b/src/shared/lib/container/lambda-microvm-runtime.e2e.test.ts
@@ -21,7 +21,7 @@ import { LambdaMicroVmRuntimeClient } from './lambda-microvm-runtime'
 const enabled = process.env.RUN_MICROVM_E2E === '1'
 
 describe.skipIf(!enabled)('LambdaMicroVmRuntimeClient e2e (real AWS)', () => {
-  it('runs a real MicroVM, serves /health through the proxy, then suspends', async () => {
+  it('runs a real MicroVM, serves /health through the proxy, then terminates', async () => {
     const client = new LambdaMicroVmRuntimeClient({ agentId: `e2e-${Date.now()}` })
     try {
       await client.start()
@@ -63,26 +63,23 @@ describe.skipIf(!enabled)('LambdaMicroVmRuntimeClient e2e (real AWS)', () => {
     expect(max).toBeLessThan(30_000)
   }, 600_000)
 
-  // Stop means suspend; the next plain request resumes the same VM warmly.
-  it('auto-sleep suspends, then a plain request transparently auto-resumes (warm)', async () => {
-    const client = new LambdaMicroVmRuntimeClient({ agentId: `e2e-resume-${Date.now()}` })
+  // Stop means terminate; the next start is a cold boot (no warm resume).
+  it('stop terminates; a later start is a cold boot', async () => {
+    const client = new LambdaMicroVmRuntimeClient({ agentId: `e2e-terminate-${Date.now()}` })
     try {
       await client.start()
       expect((await client.getInfoFromRuntime()).status).toBe('running')
 
-      await client.stop() // plain stop -> suspend (default)
-      // Let the suspend settle (SuspendMicrovm completes in ~1s); the client maps
-      // SUSPENDED/SUSPENDING to 'running' since the VM auto-resumes on demand.
-      await new Promise((r) => setTimeout(r, 10_000))
-      expect((await client.getInfoFromRuntime()).status).toBe('running')
+      await client.stop()
+      expect((await client.getInfoFromRuntime()).status).toBe('stopped')
 
-      // No waitForHealthy kick: a normal request resumes + succeeds via the proxy.
       const t0 = Date.now()
+      await client.start()
+      const coldMs = Date.now() - t0
       const res = await client.fetch('/health')
-      const resumeMs = Date.now() - t0
       expect(res.ok).toBe(true)
-      console.log(`[E2E-RESUME] suspend -> /health ok (transparent auto-resume): ${resumeMs}ms`)
-      expect(resumeMs).toBeLessThan(15_000)
+      console.log(`[E2E-TERMINATE] stop -> cold start -> /health ok: ${coldMs}ms`)
+      expect(coldMs).toBeLessThan(60_000)
     } finally {
       await client.stop()
     }
--- a/src/shared/lib/container/lambda-microvm-runtime.test.ts
+++ b/src/shared/lib/container/lambda-microvm-runtime.test.ts
@@ -542,7 +542,7 @@ describe('LambdaMicroVmRuntimeClient lifecycle', () => {
     superCreate.mockRestore()
   })
 
-  it('createSession replaces after a create failure when the generation is TERMINATED', async () => {
+  it('createSession replaces after an unreachable create failure when the generation is TERMINATED', async () => {
     const { BaseContainerClient } = await import('./base-container-client')
     const client = newClient()
     await client.start()
@@ -593,6 +593,27 @@ describe('LambdaMicroVmRuntimeClient lifecycle', () => {
     superCreate.mockRestore()
   })
 
+  it('createSession does not replace on timeout even if the generation later looks dead', async () => {
+    const { BaseContainerClient } = await import('./base-container-client')
+    const client = newClient()
+    await client.start()
+    sendMock.mockClear()
+
+    const superCreate = vi.spyOn(BaseContainerClient.prototype, 'createSession')
+    superCreate.mockImplementationOnce(async () => {
+      responses.getState = 'TERMINATED'
+      throw new Error(
+        'Failed to start session - request timed out. This may be due to network issues or the AI service being slow. Please try again.',
+      )
+    })
+
+    await expect(client.createSession({ initialMessage: 'hi' })).rejects.toThrow(/timed out/)
+    expect(sendMock.mock.calls.some((c) => c[0].type === 'Run')).toBe(false)
+    expect(sendMock.mock.calls.some((c) => c[0].type === 'Terminate')).toBe(false)
+    expect(superCreate).toHaveBeenCalledTimes(1)
+    superCreate.mockRestore()
+  })
+
   it('concurrent createSession on a dead generation only RunMicrovm once', async () => {
     const { BaseContainerClient } = await import('./base-container-client')
     const client = newClient()
@@ -624,6 +645,141 @@ describe('LambdaMicroVmRuntimeClient lifecycle', () => {
     ])
     expect(runCount).toBe(1)
     expect(superCreate).toHaveBeenCalledTimes(2)
+    superCreate.mockRestore()
+  })
+
+  it('stale dead-generation observation does not terminate a newer healthy generation', async () => {
+    const { BaseContainerClient } = await import('./base-container-client')
+    let releaseGet: (() => void) | undefined
+    const getGate = new Promise<void>((resolve) => {
+      releaseGet = resolve
+    })
+    let observeGetStarted!: () => void
+    const observeGetStartedP = new Promise<void>((resolve) => {
+      observeGetStarted = resolve
+    })
+
+    const client = newClient()
+    await client.start()
+    sendMock.mockClear()
+
+    const terminateIds: string[] = []
+    let runCount = 0
+    let gated = false
+    sendMock.mockImplementation(async (cmd: { type: string; input?: { microvmIdentifier?: string } }) => {
+      if (cmd.type === 'Get') {
+        // Gate only the first Get (observeDeadGeneration on mvm-1).
+        if (!gated) {
+          gated = true
+          observeGetStarted()
+          await getGate
+          return { state: 'TERMINATING' }
+        }
+        return { state: responses.getState ?? 'RUNNING' }
+      }
+      if (cmd.type === 'Run') {
+        runCount++
+        responses.getState = 'RUNNING'
+        return { microvmId: `mvm-fresh-${runCount}`, endpoint: 'ep.lambda-microvm.aws' }
+      }
+      if (cmd.type === 'Terminate') {
+        terminateIds.push(String(cmd.input?.microvmIdentifier ?? ''))
+        return {}
+      }
+      if (cmd.type === 'Token') return { authToken: { 'X-aws-proxy-auth': 'tok' } }
+      return {}
+    })
+
+    const superCreate = vi
+      .spyOn(BaseContainerClient.prototype, 'createSession')
+      .mockResolvedValue({ id: 'sess-cas' } as never)
+
+    const createPromise = client.createSession({ initialMessage: 'hi' })
+    await observeGetStartedP
+    // Swap in a healthy generation while observe is still awaiting GetMicrovm.
+    client.stopSync()
+    responses.getState = 'RUNNING'
+    await client.start()
+    releaseGet?.()
+    await createPromise
+
+    expect(terminateIds).not.toContain('mvm-fresh-1')
+    expect(runCount).toBe(1)
+    superCreate.mockRestore()
+  })
+
+  it('restartAgent single-flight prevents duplicate RunMicrovm when replace races start', async () => {
+    const { BaseContainerClient } = await import('./base-container-client')
+    let inflight: Promise<void> | null = null
+    let client!: LambdaMicroVmRuntimeClient
+    const restartAgent = async () => {
+      if (inflight) return inflight
+      inflight = client.start().finally(() => {
+        inflight = null
+      })
+      return inflight
+    }
+    client = new LambdaMicroVmRuntimeClient({
+      agentId: 'agent-xyz',
+      envVars: { FOO: 'bar' },
+      restartAgent,
+    })
+    await client.start()
+    sendMock.mockClear()
+
+    let runCount = 0
+    sendMock.mockImplementation(async (cmd: { type: string }) => {
+      if (cmd.type === 'Run') {
+        runCount++
+        await new Promise((r) => setTimeout(r, 40))
+        responses.getState = 'RUNNING'
+        return { microvmId: `mvm-race-${runCount}`, endpoint: 'ep.lambda-microvm.aws' }
+      }
+      if (cmd.type === 'Get') return { state: responses.getState ?? 'RUNNING' }
+      if (cmd.type === 'Terminate') return {}
+      if (cmd.type === 'Token') return { authToken: { 'X-aws-proxy-auth': 'tok' } }
+      return {}
+    })
+    responses.getState = 'TERMINATING'
+
+    const superCreate = vi
+      .spyOn(BaseContainerClient.prototype, 'createSession')
+      .mockResolvedValue({ id: 'sess-race' } as never)
+
+    // createSession replaces then restartAgent; a concurrent ensureRunning-style
+    // restart shares the same single-flight lock (prod: ContainerManager.startingAgents).
+    await Promise.all([
+      client.createSession({ initialMessage: 'a' }),
+      (async () => {
+        await new Promise((r) => setTimeout(r, 5))
+        await restartAgent()
+      })(),
+    ])
+    expect(runCount).toBe(1)
+    superCreate.mockRestore()
+  })
+
+  it('observeDeadGeneration returning false on throttling does not replace', async () => {
+    const { BaseContainerClient } = await import('./base-container-client')
+    const client = newClient()
+    await client.start()
+    sendMock.mockClear()
+
+    sendMock.mockImplementation(async (cmd: { type: string }) => {
+      if (cmd.type === 'Get') throw new Error('ThrottlingException')
+      if (cmd.type === 'Run') return { microvmId: 'mvm-should-not', endpoint: 'ep.lambda-microvm.aws' }
+      if (cmd.type === 'Terminate') return {}
+      if (cmd.type === 'Token') return { authToken: { 'X-aws-proxy-auth': 'tok' } }
+      return {}
+    })
+
+    const superCreate = vi
+      .spyOn(BaseContainerClient.prototype, 'createSession')
+      .mockResolvedValue({ id: 'sess-throttle' } as never)
+
+    await expect(client.createSession({ initialMessage: 'hi' })).resolves.toEqual({ id: 'sess-throttle' })
+    expect(sendMock.mock.calls.some((c) => c[0].type === 'Run')).toBe(false)
+    expect(sendMock.mock.calls.some((c) => c[0].type === 'Terminate')).toBe(false)
     superCreate.mockRestore()
   })
 })
--- a/src/shared/lib/container/lambda-microvm-runtime.test.ts
+++ b/src/shared/lib/container/lambda-microvm-runtime.test.ts
@@ -74,7 +74,7 @@ const FULL_ENV = { ...REQUIRED_ENV, HOST_PUBLIC_URL: 'https://host.example' }
 const TOUCHED = [
   ...Object.keys(FULL_ENV),
   'AWS_REGION', 'AWS_DEFAULT_REGION', 'MICROVM_AGENT_IMAGE_VERSION', 'MICROVM_INGRESS_CONNECTOR_ARN',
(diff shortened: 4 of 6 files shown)


iddogino:
## Review: replace dead MicroVM generation on `createSession`

Consolidated from two independent review passes. **Requesting changes** — five findings, all of which I'd treat as blocking.

The diagnosis is right and the shape is right: recovery belongs inside the runtime client, not in callers branching on provider type. The `TERMINAL_MICROVM_STATES` extraction is a genuine dedup of three copies. The problems are all in the replacement path, which reimplements "start this agent" instead of delegating to `ContainerManager`, and which loses the identity of the generation it observed.

Local verification: `npx vitest run src/shared/lib/container/lambda-microvm-runtime.test.ts` → 51 passed; `npx tsc --noEmit` clean; `npx eslint` on both changed files clean. Findings below marked "reproduced" were confirmed with throwaway probe tests against this branch, not asserted from inspection.

---

### 1. The retry is not idempotent — the turn starts before the response exists

`lambda-microvm-runtime.ts:767` retries `super.createSession` after replacing the generation. The first `POST /sessions` may already have delivered the initial message and started the turn:

- `agent-container/src/session-manager.ts:284` calls `process.sendMessage(request.initialMessage, ...)` and *then* awaits `initCompletePromise` for up to 30s to learn Claude's session id. The prompt is running well before the caller can learn that it is.
- The host adds its own 60s `AbortController` on top (`base-container-client.ts:1099`), and the current `catch` retries on **any** throw — including `AbortError`, where the request provably may have landed.

`initialMessageUuid` does not make this safe:

- It is not a session-creation idempotency key. `agent-container/src/types.ts:79`: *"Optional UUID for message author attribution."*
- Four of six callers never pass it at all — `trigger-manager.ts:364`, `task-scheduler.ts:227`, `chat-integration-manager.ts:1132`, `scheduled-tasks.ts:316`. `x-agent.ts:690` passes it conditionally; only `agents.ts:1616` always does.

And the blast radius is larger than a duplicated message: `start()` mounts the same `agents/<id>/workspace` subPath on every generation, so whatever the first turn wrote before its VM died persists into the replacement. The retry then re-runs the prompt on top of those side effects.

A container-side idempotency key that persists and replays would be the complete fix, but that's an agent-container protocol change well beyond this PR. The proportionate fix is to **retry only on failures that prove the request never reached the container** (connect-refused, proxy 502 before forward) and never on ambiguous ones (abort/timeout, mid-response reset).

### 2. The dead-generation check loses the generation identity it observed (reproduced)

`isDeadGeneration()` (`:839`) inspects one MicroVM but returns a bare boolean. `replaceGenerationInner` then re-reads `agentStates` at `:861` and terminates *whatever is currently installed* — not the VM that was observed dead. If a healthy generation lands during the check's `GetMicrovm` await, the stale answer kills it.

Reproduced by holding the check's `GetMicrovm` open, installing a healthy generation during the await, then returning the stale `TERMINATING` answer:

```
terminated = ["mvm-2"]   <- the freshly installed, healthy generation
runs = 3
```

Note this is **not** closed by serializing `start()` and `replaceGeneration` (finding 3): `isDeadGeneration()` runs entirely outside any such lock, so its result is stale on return regardless. Thread the observed `microvmId` through replacement and make terminate + `cleanupLocal` compare-and-swap on it. The same generation-blind cleanup affects `getInfoFromRuntime`'s terminal branch (`:800`) and concurrent status sync.

### 3. `replaceGeneration` bypasses the manager's start dedupe → duplicate VMs, neither reclaimed (reproduced)

`replaceGenerationInner` calls `this.start()` directly (`:882`). `ContainerManager` dedupes starts through `startingAgents` (`container-manager.ts:466`, `:498`); this path doesn't participate. `start()`'s own guard — `if ((await this.getInfoFromRuntime()).status === 'running') return` (`:685`) — is check-then-act across an AWS round-trip.

One client instance, `createSession` racing a `start()` on a `TERMINATING` generation:

```
runCount = 2, terminateCount = 0
```

Two fresh VMs plus the original dead one. `agentStates` retains only one of the new pair, so the loser is orphaned — never terminated, billed until its own `maximumDurationInSeconds` (8h) expires.

This is reachable under exactly the conditions the PR targets. Request A replaces, stopping the local proxy; request B's `ensureRunning` sees cached `running` older than the 10s `RUNNING_STATUS_TTL_MS`, probes `isHealthy` against the closed port, fails, `markAsStopped`, then `doStartContainer` → `client.start()`. A dead VM normally has several triggers queued behind it, so concurrency here is the common case.

### 4. `start()`'s early return leaves `hasStarted` false → the replace terminates the VM, then throws (reproduced)

`lastStartOptions` / `hasStarted` are set at `:753-754`, *after* the early return at `:685`. Any instance whose `start()` took the already-running path has `hasStarted === false` and `lastStartOptions === undefined`.

Fresh instance for an agent whose module-level `agentStates` entry survives — which is what `containerManager.clearClients()` produces on a container-runner settings change (`settings.ts:645`) — with the generation `TERMINATING`:

```
terminateCount = 1, runCount = 0
Error: Cannot replace MicroVM generation before the first start
```

Two distinct defects. The precondition is checked at `:879`, *after* `teardown()` at `:873` has destroyed the VM — so a failed precondition is strictly worse than having no fix, and the internal message surfaces to the user as "Failed to create session". Guard before tearing down. Separately, the early-return path should record `hasStarted` / `lastStartOptions`, or both should be hoisted to the top of `start()`.

### 5. The replacement boots from stale `lastStartOptions`, silently dropping live-synced connections

`doStartContainer` (`container-manager.ts:510`) is where `CONNECTED_ACCOUNTS` and `REMOTE_MCPS` are rebuilt from the mapping tables. Between full starts, those projections are pushed into the *running* container by `syncAgentConnectionEnvironment` → `POST /env` (`connection-runtime-sync.ts:79`). Replaying a cached `StartOptions` reconstructs the snapshot as of the last full start, so **every account or remote MCP added since is lost** — it was written to the VM that was just terminated, and nothing re-pushes it. Nothing is logged when this happens.

The same call also does four things the replace path skips:

- `updateCachedStatus` — and the proxy port *does* change (`server.listen(0, ...)`, `:256`). `browser-stream-proxy.ts:128` builds its upstream WS URL from `getCachedInfo().port` without going through `ensureRunning`, so browser streaming targets a closed — and reusable — ephemeral port for up to the 300s `STATUS_SYNC_INTERVAL_MS`.
- `containerStartedAt` — the auto-sleep monitor's grace window.
- the `agent_status_changed` broadcast.
- fresh `TZ`, `customEnvVars`, mount health.

**Findings 3, 4, and 5 collapse into one fix.** `ContainerConfig` already carries `onConnectionError` as an injected-callback precedent (`types.ts:26`); adding a `restartAgent?: () => Promise<void>` that the manager wires to a `startingAgents`-guarded `doStartContainer` gives single-flight coverage across start-vs-replace, rebuilds env from the DB, keeps the status cache and broadcast coherent, and deletes `lastStartOptions` / `hasStarted` along with finding 4 entirely.

---

### Smaller items

- `if (failedId && agentStates.get(...)?.microvmId === failedId)` (`:872`) — `teardown()` re-reads `agentStates` and no-ops when empty, so both branches are equivalent to an unconditional `await this.teardown()`; the `else` is unreachable by effect. (Worth keeping the *shape* once finding 2's compare-and-swap makes the condition meaningful.)
- `if (agentStates.has(...)) return` (`:878`) with "another replace may have already started a fresh VM" — a replace can't: `replaceInFlight` serializes them on this instance and `cleanupLocal()` runs synchronously just above with no intervening await. What can set it is a concurrent plain `start()`, i.e. finding 3. The comment describes a protection that doesn't exist and misses the race that does.
- `isDeadGeneration` (`:839`) duplicates logic `getInfoFromRuntime` (`:788`) already has. A healthy `createSession` now costs **2 `GetMicrovm` calls** where it cost 1 (the other via `getPortOrThrow` → `getInfo`). `GetMicrovm` is account-throttled and this multiplies by agents × session creates.
- `stop()` and `stopSync()` both call `terminateWebSocketConnections()`; the replace path doesn't. `wsConnections` keeps sockets with listeners attached against the closed proxy.
- `captureException(..., level: 'warning')` on every replace (`:866`) — this is now an expected recovery, so it'll be steady Sentry volume. A breadcrumb, plus an event only when the replacement itself fails, would read better.
- If `start()` inside `replaceGenerationInner` throws, `teardown()` has already run: the agent is left with no VM, and `createSession` reports the start error rather than the original failure. Defensible, but untested.

### Test coverage

The four new tests are well-constructed for what they cover, and the RUNNING-failure negative case is the right instinct. But all four use a single client instance and stub `BaseContainerClient.prototype.createSession`, which is precisely why none of findings 1–4 surface. Worth adding:

- an instance that took `start()`'s early return (finding 4)
- a replace racing a `start()` (finding 3)
- a dead-generation check whose answer arrives after a new generation is installed (finding 2)
- `isDeadGeneration` returning false on a transient/throttling error, asserting no replace

The "no duplicate initial message" line in the test plan is still unchecked — finding 1 says it won't hold as written. I'd gate merge on findings 1 and 2, and want the staging run to count MicroVMs after a forced `TERMINATING` wake to confirm finding 3.

iddogino (changes requested):
Re-reviewed after the switch to terminate-on-stop. **The simplification is real and it's the right call** — dropping suspend removes the *cause* of this bug class rather than just handling it: the runtime no longer retains a handle across a sleep boundary, so it now matches every other runner. Trading ~3s of warm resume for that is a good deal.

Two earlier findings are fixed properly at the root, worth calling out: `{ cause: err }` in `base-container-client.ts` + `causeChainIncludes` makes `isUnreachableCreateSessionError` genuinely precise (ECONNRESET/ETIMEDOUT/`fetch failed` no longer retry), and `replaceGenerationInner` now avoids `teardown()` in favour of `terminateObserved` + `cleanupLocalIf`.

Verified locally: 57/57 unit tests pass, `tsc --noEmit` clean, `eslint src/shared/lib/container/*.ts` clean.

One blocking issue below, plus some smaller items.

---

## 🔴 The generation-wipe leak wasn't fixed — it moved to `stop()`, and got more reachable

`teardown()` (`src/shared/lib/container/lambda-microvm-runtime.ts:952-965`) still does an unconditional `cleanupLocal()` after `await terminateMicrovm(...)`. The replace path was rerouted around it; `stop()` now routes straight into it.

If a newer generation lands in `agentStates` while that Terminate is in flight, `cleanupLocal()` wipes it — dropping local state and stopping its proxy **without terminating the VM**. The result is a live MicroVM leaked in AWS, untracked, holding its S3 mount until the 8h lifetime cap.

Reachability went *up* with this PR, not down:

- Auto-sleep previously early-returned for microvm (`escalateToForceStop === false` → no-op). It now terminates on every idle sweep.
- `ContainerManager.ensureRunning` has no `stoppingAgents` guard, so "sweep decides idle, user sends a message" is a routine race: `ensureRunning` → stale cache → health probe fails → `doStartContainer` installs a new VM while `teardown`'s Terminate is still in flight.

Repro against this branch, unmodified:

```
PROBE freshId= mvm-probe-1  terminateIds= ["mvm-1"]
      infoAfter= {"status":"stopped","port":null}   => freshGenLeaked: true
AssertionError: expected 'stopped' to be 'running'
```

`mvm-probe-1` was wiped from local state and never terminated.

Fix — same CAS the rest of the file already uses. All tests pass with it, including the probe:

```ts
  private async teardown(): Promise<void> {
    const state = agentStates.get(this.config.agentId)
    if (state) {
      const config = getMicrovmRuntimeConfig()
      try {
        await terminateMicrovm(config.region, state.microvmId)
      } catch (error) {
        if (!isNotFound(error)) {
          captureException(error, { tags: { area: 'container', op: 'microvm.terminate' }, extra: { microvmId: state.microvmId } })
        }
      }
      // CAS: a newer generation may have been installed while Terminate was in
      // flight — wiping it here would leak that live VM.
      this.cleanupLocalIf(state.microvmId)
      return
    }
    this.cleanupLocal()
  }
```

Worth fixing the primitive rather than routing around it — `start()`'s error path calls `teardown()` too.

<details>
<summary>Regression test (fails on this branch, passes with the fix)</summary>

```ts
describe('teardown CAS on stop', () => {
  beforeEach(() => {
    Object.assign(process.env, FULL_ENV)
    resetMicrovmRuntimeForTests()
  })

  it('stop() teardown must not wipe a generation installed during TerminateMicrovm', async () => {
    const client = new LambdaMicroVmRuntimeClient({ agentId: 'agent-xyz', envVars: { FOO: 'bar' } })
    await client.start()
    sendMock.mockClear()

    let releaseTerminate!: () => void
    const gate = new Promise<void>((r) => { releaseTerminate = r })
    let terminateStarted!: () => void
    const terminateStartedP = new Promise<void>((r) => { terminateStarted = r })

    const terminateIds: string[] = []
    let gatedOnce = false
    let runCount = 0
    sendMock.mockImplementation(async (cmd: { type: string; input?: { microvmIdentifier?: string } }) => {
      if (cmd.type === 'Run') {
        runCount++
        responses.getState = 'RUNNING'
        return { microvmId: `mvm-probe-${runCount}`, endpoint: 'ep.lambda-microvm.aws' }
      }
      if (cmd.type === 'Get') return { state: responses.getState ?? 'RUNNING' }
      if (cmd.type === 'Terminate') {
        terminateIds.push(String(cmd.input?.microvmIdentifier ?? ''))
        if (!gatedOnce) { gatedOnce = true; terminateStarted(); await gate }
        return {}
      }
      if (cmd.type === 'Token') return { authToken: { 'X-aws-proxy-auth': 'tok' } }
      return {}
    })

    // Auto-sleep sweep terminates mvm-1 ...
    const stopPromise = client.stop()
    await terminateStartedP
    // ... while a concurrent ensureRunning observes the dying gen and starts a fresh one.
    responses.getState = 'TERMINATING'
    await client.getInfoFromRuntime()   // CAS-drops mvm-1
    responses.getState = 'RUNNING'
    await client.start()                // installs mvm-probe-1
    releaseTerminate()
    await stopPromise

    responses.getState = 'RUNNING'
    // The fresh generation must still be installed (not silently wiped + leaked).
    expect((await client.getInfoFromRuntime()).status).toBe('running')
  })
})
```

</details>

## 🟠 Stale live E2E asserts the behavior this PR deletes

`src/shared/lib/container/lambda-microvm-runtime.e2e.test.ts:66-88` — *"auto-sleep suspends, then a plain request transparently auto-resumes (warm)"* — calls `client.stop()` then asserts `getInfoFromRuntime()` is `'running'`. That's now false by design. Line 24's title says "then suspends" too.

It's `describe.skipIf(!enabled)` behind `RUN_MICROVM_E2E=1` so CI stays green, but the live suite is exactly what you'd reach for to validate this change, and it now encodes the old semantics.

## 🟠 The replace path still doesn't mark its sessions inactive

Auto-sleep now routes through `stopContainer`, which does `markAllSessionsInactiveForAgent` + broadcast — that comes for free with this change, nice.

But `replaceGeneration` still terminates a VM without telling the persister, so sessions that lived on it stay `isActive` with orphaned subscriptions (`terminateWebSocketConnections` calls `removeAllListeners()` first, so no close event fires either). Users see a session stuck "working" forever.

Related: the reason given for not reusing `containerManager.restartContainer` (= `stopContainer` + `ensureRunning`) was that microvm `stop()` only suspended. That's no longer true — reusing it would fix this for free.

## 🟡 Smaller items

- **`createSession` auto-boots a never-started agent.** `deadId === null` + no local state → `replaceGeneration(reason, null)` → `restartAgent()`. All prod callers do `ensureRunning` first so it's mostly benign, but a stop racing a create resurrects the agent, or throws `Cannot restart agent X while it is stopping` and masks the real error.
- **`else { await this.start() }`** in `replaceGenerationInner` boots an env-less VM (no `PROXY_TOKEN`, no connected accounts, no MCPs) for any client not built by `containerManager.getClient`. Better to throw than to boot a crippled agent.
- **`resolveIdleSeconds` is now dead in production** — only its own test block imports it. The comment says "kept for tests/callers"; there are no callers. Delete it plus `lambda-microvm-runtime.test.ts:167-187`, or wire it up.
- **`suspendedDurationSeconds: 1`** is an undocumented sentinel. If AWS's minimum is above 1, *every* `RunMicrovm` fails — that's a total outage of the runtime, so worth stating explicitly that the staging canary covered it.
- **`getInfoFromRuntime` now performs a Terminate** for SUSPENDED leftovers — a destructive side effect in a read-shaped method called from the periodic `syncAgentStatus` sweep. Reasonable as a one-time migration for pre-policy VMs, but worth time-boxing and removing after rollout.
- `SuspendMicrovmCommand` is still declared in the test's AWS SDK mock (`lambda-microvm-runtime.test.ts:26,33`).

## ❓ Open question

Now that dead generations are rare rather than the steady state after every idle period, does `ensureAliveGeneration` still earn an AWS round trip on *every* `createSession`?

The post-create replace-and-retry is the high-value half — it catches the lifetime-cap and stop-race cases. The pre-check is the expensive half, and `ensureRunning`'s health probe already backstops it a moment earlier. Dropping it would shave latency off the common path with little apparent loss. Was there a case you saw where the pre-check caught something the post-create path wouldn't have?

yiw190:
> Re-reviewed after the switch to terminate-on-stop. **The simplification is real and it's the right call** — dropping suspend removes the _cause_ of this bug class rather than just handling it: the runtime no longer retains a handle across a sleep boundary, so it now matches every other runner. Trading ~3s of warm resume for that is a good deal.
> 
> Two earlier findings are fixed properly at the root, worth calling out: `{ cause: err }` in `base-container-client.ts` + `causeChainIncludes` makes `isUnreachableCreateSessionError` genuinely precise (ECONNRESET/ETIMEDOUT/`fetch failed` no longer retry), and `replaceGenerationInner` now avoids `teardown()` in favour of `terminateObserved` + `cleanupLocalIf`.
> 
> Verified locally: 57/57 unit tests pass, `tsc --noEmit` clean, `eslint src/shared/lib/container/*.ts` clean.
> 
> One blocking issue below, plus some smaller items.
> 
> ## 🔴 The generation-wipe leak wasn't fixed — it moved to `stop()`, and got more reachable
> `teardown()` (`src/shared/lib/container/lambda-microvm-runtime.ts:952-965`) still does an unconditional `cleanupLocal()` after `await terminateMicrovm(...)`. The replace path was rerouted around it; `stop()` now routes straight into it.
> 
> If a newer generation lands in `agentStates` while that Terminate is in flight, `cleanupLocal()` wipes it — dropping local state and stopping its proxy **without terminating the VM**. The result is a live MicroVM leaked in AWS, untracked, holding its S3 mount until the 8h lifetime cap.
> 
> Reachability went _up_ with this PR, not down:
> 
> * Auto-sleep previously early-returned for microvm (`escalateToForceStop === false` → no-op). It now terminates on every idle sweep.
> * `ContainerManager.ensureRunning` has no `stoppingAgents` guard, so "sweep decides idle, user sends a message" is a routine race: `ensureRunning` → stale cache → health probe fails → `doStartContainer` installs a new VM while `teardown`'s Terminate is still in flight.
> 
> Repro against this branch, unmodified:
> 
> ```
> PROBE freshId= mvm-probe-1  terminateIds= ["mvm-1"]
>       infoAfter= {"status":"stopped","port":null}   => freshGenLeaked: true
> AssertionError: expected 'stopped' to be 'running'
> ```
> 
> `mvm-probe-1` was wiped from local state and never terminated.
> 
> Fix — same CAS the rest of the file already uses. All tests pass with it, including the probe:
> 
> ```ts
>   private async teardown(): Promise<void> {
>     const state = agentStates.get(this.config.agentId)
>     if (state) {
>       const config = getMicrovmRuntimeConfig()
>       try {
>         await terminateMicrovm(config.region, state.microvmId)
>       } catch (error) {
>         if (!isNotFound(error)) {
>           captureException(error, { tags: { area: 'container', op: 'microvm.terminate' }, extra: { microvmId: state.microvmId } })
>         }
>       }
>       // CAS: a newer generation may have been installed while Terminate was in
>       // flight — wiping it here would leak that live VM.
>       this.cleanupLocalIf(state.microvmId)
>       return
>     }
>     this.cleanupLocal()
>   }
> ```
> 
> Worth fixing the primitive rather than routing around it — `start()`'s error path calls `teardown()` too.
> 
> Regression test (fails on this branch, passes with the fix)
> ```ts
> describe('teardown CAS on stop', () => {
>   beforeEach(() => {
>     Object.assign(process.env, FULL_ENV)
>     resetMicrovmRuntimeForTests()
>   })
> 
>   it('stop() teardown must not wipe a generation installed during TerminateMicrovm', async () => {
>     const client = new LambdaMicroVmRuntimeClient({ agentId: 'agent-xyz', envVars: { FOO: 'bar' } })
>     await client.start()
>     sendMock.mockClear()
> 
>     let releaseTerminate!: () => void
>     const gate = new Promise<void>((r) => { releaseTerminate = r })
>     let terminateStarted!: () => void
>     const terminateStartedP = new Promise<void>((r) => { terminateStarted = r })
> 
>     const terminateIds: string[] = []
>     let gatedOnce = false
>     let runCount = 0
>     sendMock.mockImplementation(async (cmd: { type: string; input?: { microvmIdentifier?: string } }) => {
>       if (cmd.type === 'Run') {
>         runCount++
>         responses.getState = 'RUNNING'
>         return { microvmId: `mvm-probe-${runCount}`, endpoint: 'ep.lambda-microvm.aws' }
>       }
>       if (cmd.type === 'Get') return { state: responses.getState ?? 'RUNNING' }
>       if (cmd.type === 'Terminate') {
>         terminateIds.push(String(cmd.input?.microvmIdentifier ?? ''))
>         if (!gatedOnce) { gatedOnce = true; terminateStarted(); await gate }
>         return {}
>       }
>       if (cmd.type === 'Token') return { authToken: { 'X-aws-proxy-auth': 'tok' } }
>       return {}
>     })
> 
>     // Auto-sleep sweep terminates mvm-1 ...
>     const stopPromise = client.stop()
>     await terminateStartedP
>     // ... while a concurrent ensureRunning observes the dying gen and starts a fresh one.
>     responses.getState = 'TERMINATING'
>     await client.getInfoFromRuntime()   // CAS-drops mvm-1
>     responses.getState = 'RUNNING'
>     await client.start()                // installs mvm-probe-1
>     releaseTerminate()
>     await stopPromise
> 
>     responses.getState = 'RUNNING'
>     // The fresh generation must still be installed (not silently wiped + leaked).
>     expect((await client.getInfoFromRuntime()).status).toBe('running')
>   })
> })
> ```
> 
> ## 🟠 Stale live E2E asserts the behavior this PR deletes
> `src/shared/lib/container/lambda-microvm-runtime.e2e.test.ts:66-88` — _"auto-sleep suspends, then a plain request transparently auto-resumes (warm)"_ — calls `client.stop()` then asserts `getInfoFromRuntime()` is `'running'`. That's now false by design. Line 24's title says "then suspends" too.
> 
> It's `describe.skipIf(!enabled)` behind `RUN_MICROVM_E2E=1` so CI stays green, but the live suite is exactly what you'd reach for to validate this change, and it now encodes the old semantics.
> 
> ## 🟠 The replace path still doesn't mark its sessions inactive
> Auto-sleep now routes through `stopContainer`, which does `markAllSessionsInactiveForAgent` + broadcast — that comes for free with this change, nice.
> 
> But `replaceGeneration` still terminates a VM without telling the persister, so sessions that lived on it stay `isActive` with orphaned subscriptions (`terminateWebSocketConnections` calls `removeAllListeners()` first, so no close event fires either). Users see a session stuck "working" forever.
> 
> Related: the reason given for not reusing `containerManager.restartContainer` (= `stopContainer` + `ensureRunning`) was that microvm `stop()` only suspended. That's no longer true — reusing it would fix this for free.
> 
> ## 🟡 Smaller items
> * **`createSession` auto-boots a never-started agent.** `deadId === null` + no local state → `replaceGeneration(reason, null)` → `restartAgent()`. All prod callers do `ensureRunning` first so it's mostly benign, but a stop racing a create resurrects the agent, or throws `Cannot restart agent X while it is stopping` and masks the real error.
> * **`else { await this.start() }`** in `replaceGenerationInner` boots an env-less VM (no `PROXY_TOKEN`, no connected accounts, no MCPs) for any client not built by `containerManager.getClient`. Better to throw than to boot a crippled agent.
> * **`resolveIdleSeconds` is now dead in production** — only its own test block imports it. The comment says "kept for tests/callers"; there are no callers. Delete it plus `lambda-microvm-runtime.test.ts:167-187`, or wire it up.
> * **`suspendedDurationSeconds: 1`** is an undocumented sentinel. If AWS's minimum is above 1, _every_ `RunMicrovm` fails — that's a total outage of the runtime, so worth stating explicitly that the staging canary covered it.
> * **`getInfoFromRuntime` now performs a Terminate** for SUSPENDED leftovers — a destructive side effect in a read-shaped method called from the periodic `syncAgentStatus` sweep. Reasonable as a one-time migration for pre-policy VMs, but worth time-boxing and removing after rollout.
> * `SuspendMicrovmCommand` is still declared in the test's AWS SDK mock (`lambda-microvm-runtime.test.ts:26,33`).
> 
> ## ❓ Open question
> Now that dead generations are rare rather than the steady state after every idle period, does `ensureAliveGeneration` still earn an AWS round trip on _every_ `createSession`?
> 
> The post-create replace-and-retry is the high-value half — it catches the lifetime-cap and stop-race cases. The pre-check is the expensive half, and `ensureRunning`'s health probe already backstops it a moment earlier. Dropping it would shave latency off the common path with little apparent loss. Was there a case you saw where the pre-check caught something the post-create path wouldn't have?

Thanks @iddogino 

— agreed on the blocking teardown() CAS issue, and on dropping ensureAliveGeneration.

Fixed in 630cebf6:

1 teardown() now terminates the observed id and cleanupLocalIfs it (same CAS as replace)
2 Removed ensureAliveGeneration — post-create replace-and-retry only
3 restartAgent marks sessions inactive + broadcasts stopped
4 Replace requires restartAgent (no env-less start() fallback); deadId === null does not resurrect
5 Removed dead resolveIdleSeconds; updated live e2e for terminate → cold start; documented suspendedDurationSeconds: 1

Also on the open question: agreed. After terminate-on-stop, dead gens are rare enough that a pre-create GetMicrovm on every createSession isn’t worth it. ensureRunning’s health probe covers the common path; post-create replace covers lifetime-cap / stop races.

On the open question: agreed — dropped the pre-check. Post-create replace is enough now that dead gens are rare.

Also test on staging (gamut-host-app:pr613-20260731013537)

iddogino (changes requested):
All previous items addressed, and the `teardown()` CAS + E2E rewrite look right. One regression from the last round.

## 🔴 Dropping the pre-check + tightening the null branch together kill the primary scenario

Each change is correct alone; combined they break the case this PR exists to fix.

The real `super.createSession` opens with `getPortOrThrow()` → `getInfo()` → the **overridden** `getInfoFromRuntime()`. On a TERMINATING generation that calls `cleanupLocalIf(observedId)` and deletes local state, then returns `stopped`. So by the time the catch runs `observeDeadGeneration()`, there is no state left → `null` → rethrow. `ensureAliveGeneration` used to run *before* that wipe, which is the only reason it worked.

Probe against `630cebf6`:

```
PROBE result= THREW: Container is not running | RunMicrovm calls= 0
```

Idle VM dies → user creates a session → "Failed to create session", no replacement attempted. `ensureRunning`'s health probe only covers it once the cached status is past `RUNNING_STATUS_TTL_MS`; inside that window nothing catches it. More generally `replaceGeneration` is now near-unreachable from `createSession`: it needs state present *and* `GetMicrovm` terminal, but `getPortOrThrow` just dropped the state in exactly those cases.

**Why CI is green:** every replace test mocks `BaseContainerClient.prototype.createSession` wholesale, so `getPortOrThrow` never runs and the state is never wiped. The mock hides the ordering that breaks it.

Fix — snapshot the generation before the call. No extra AWS round trip, and the no-resurrect guarantee holds (`installedId` is null for a never-started agent). 58/58 pass with it, `tsc --noEmit` and `eslint` clean:

```ts
  async createSession(options: CreateSessionOptions): Promise<ContainerSession> {
    // Snapshot the installed generation: super.createSession's getPortOrThrow
    // consults getInfoFromRuntime, which CAS-drops a terminal generation before
    // we ever reach the catch — leaving observeDeadGeneration nothing to see.
    const installedId = agentStates.get(this.config.agentId)?.microvmId ?? null
    try {
      return await super.createSession(options)
    } catch (error) {
      if (!isUnreachableCreateSessionError(error)) throw error
      let deadId = await this.observeDeadGeneration()
      if (deadId === null && installedId !== null && !agentStates.has(this.config.agentId)) {
        deadId = installedId
      }
      // Alive, unknown, or never had a generation — do not resurrect a stopped agent.
      if (deadId === null) throw error
      await this.replaceGeneration('post_create_unreachable', deadId)
      return await super.createSession(options)
    }
  }
```

Worth adding one test that lets the real `getPortOrThrow` ordering run — otherwise this regresses again. A faithful stand-in is enough:

```ts
vi.spyOn(BaseContainerClient.prototype, 'createSession')
  .mockImplementation(async function (this: LambdaMicroVmRuntimeClient) {
    const info = await this.getInfoFromRuntime()
    if (info.status !== 'running' || !info.port) throw new Error('Container is not running')
    return { id: 'sess-real' } as never
  })
```

PR #614, opened 2026-07-30
fix: align Users/Auth settings with Platform in cloud AUTH_MODE (SUP-509)

yiw190:
## Summary
- In env-managed (platform-controlled) AUTH_MODE, Settings → **Users** opens `{platform}/dashboard/organizations/{orgId}?tab=team` instead of the local invite UI, with an external-link icon in the sidebar.
- Hides **Signup & Access**, **Authentication Methods**, and **Password Policy** on the Auth tab when platform-managed; keeps Session & Lockout and Trusted Origins.
- Self-hosted AUTH_MODE without an env platform token keeps the existing local Users/Auth UI (e2e auth suite unaffected).

Fixes https://linear.app/datawizz/issue/SUP-509/web-deployment-align-user-management-with-platform

## Test plan
- [ ] Cloud AUTH_MODE admin: Settings → Users shows external icon and opens Platform Team page
- [ ] Cloud AUTH_MODE admin: Auth tab has no Signup & Access / Authentication Methods / Password Policy
- [ ] Cloud AUTH_MODE admin: Session & Lockout + Trusted Origins still present
- [ ] Self-hosted AUTH_MODE (no env platform token): Users tab still lists/invites local users; full Auth sections visible
- [ ] Electron / non-AUTH_MODE: Users/Auth nav items still absent
- [ ] Deep link `/settings/users` in cloud mode falls back to another settings tab (Users is external-only)


src/shared/lib/auth/auth-settings.ts
```diff
@@ -0,0 +1,21 @@
+import { getSettings, DEFAULT_AUTH_SETTINGS, type AuthSettings } from '@shared/lib/config/settings'
+import { isAuthMode } from './mode'
+
+/** AUTH_MODE + PLATFORM_TOKEN = platform-controlled deployment (members managed on Platform). */
+export function isPlatformControlledAuth(): boolean {
+  return isAuthMode() && Boolean(process.env.PLATFORM_TOKEN?.trim())
```
iddogino: **The server-side and UI predicates disagree, and the gap is silent.**

This reads `PLATFORM_TOKEN` raw — no verification. The UI gates on `platformAuth?.source === 'env' && platformBaseUrl && orgId` (`global-settings-page.tsx:49`), where `orgId` comes from the **JWKS-verified** path: `buildEnvManagedStatus` leaves it `null` when the issuer is unconfigured, the token is expired, or JWKS is unreachable at startup (`platform-auth-service.ts:169-189`).

Provisioning keeps working in exactly that state, because the token-exchange org pin uses `decodeOrgIdFromToken` (`decode-org-id.ts:11` — base64 parse, no signature check).

So on a deployment with an expired token / unreachable JWKS / unconfigured issuer:
- users are still provisioned via exchange;
- the server **silently forces `requireAdminApproval` off** (line 14 below);
- the Auth tab is *not* hidden, and renders the "Require Admin Approval" switch as ON, read straight from settings.json.

An admin sees a security gate that is on and doing nothing. It's one-directional (the UI never hides more than the server forces), but a silently-ignored setting is worth closing.

Suggested fix: return `isPlatformControlledAuth()` as a `platformControlled` field on `/api/platform-auth` and branch the UI on that, keeping `orgId` only for building the link (falling back to `${platformBaseUrl}/dashboard` when it's null). Then the two gates are provably the same predicate.

src/shared/lib/auth/auth-settings.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+/** Merge defaults with persisted auth; force no local approval gate when platform-controlled. */
+export function resolveAuthSettings(auth?: AuthSettings | null): AuthSettings {
+  const resolved = { ...DEFAULT_AUTH_SETTINGS, ...auth }
+  // No local Users approve UI in env-managed mode — never ban pending approval.
+  if (isPlatformControlledAuth()) {
+    resolved.requireAdminApproval = false
```
iddogino, line 15: **Users already banned pending approval stay stuck.**

This prevents *new* bans, but a platform-controlled deployment that accumulated pending users before this ships keeps them banned: `exchangeDeploymentGrant` rejects them at the banned check (`token-exchange.ts:296-308`) → `invalid_grant`, and the Users tab that could unban them is now external-only. `users-tab.tsx` is the only UI in the app that can clear `banned` (I grepped — nothing else writes it).

A one-time clear of bans with `banReason = 'Pending admin approval'` when `isPlatformControlledAuth()` — at startup, or right here alongside the override — would close the upgrade path.

src/renderer/components/settings/global-settings-page.tsx
```diff
@@ -33,11 +34,23 @@ interface GlobalSettingsPageProps {
...
+
 export function GlobalSettingsPage({ onClose, onOpenWizard, initialSection, onSectionChange, sectionLinkProps }: GlobalSettingsPageProps) {
   const { isAuthMode, isAdmin } = useUser()
+  const { data: platformAuth } = usePlatformAuthStatus()
   const showAdminSettings = !isAuthMode || isAdmin
   const showAuthAdmin = isAuthMode && isAdmin
 
+  // Env-managed platform token = platform-controlled deployment; manage users there.
+  const usersExternalHref =
+    platformAuth?.source === 'env' && platformAuth.platformBaseUrl && platformAuth.orgId
```
iddogino: **Still no way to make a second admin** (carried over from the first pass, not addressed by the new commits).

In a platform-controlled deployment: only the first exchanged user is promoted to admin (`token-exchange.integration.test.ts:409`), everyone after that gets `defaultUserRole` (`member`) — a field this PR hides — and the role menu lives in `users-tab.tsx`, which this line makes external-only. Same for locally banning or removing a user.

Unless the Platform Team page can write the deployment's local Better Auth `role`/`banned` columns, these deployments are permanently single-admin: if that one person leaves, nobody can reach Settings → Admin/LLM/Runtime again.

Worth confirming against the platform side before merge. If the platform can't do it, keeping a trimmed local Users pane (role / ban / remove) *alongside* the external invite link would cover the actions that have no platform equivalent.

src/shared/lib/auth/token-exchange.integration.test.ts
```diff
@@ -470,30 +470,28 @@ describe('provisioning and identity mapping', () => {
...
-      .prepare(`SELECT count(*) AS n FROM session WHERE user_id = ?`)
-      .get(pending.id) as { n: number }
-    expect(sessions.n).toBe(0)
+    const second = dbModule.sqlite
+      .prepare(`SELECT banned, ban_reason FROM user WHERE email = 'second@example.com'`)
+      .get() as { banned: number; ban_reason: string | null }
+    expect(second.banned).toBe(0)
   })
 
+  // No PLATFORM_TOKEN → exchange fails closed at the org pin (covered above).
```
iddogino, line 534: Minor: this deleted the only test that the pending-approval ban actually blocks a session.

The rationale is right for the *exchange* path, but the ban still fires for self-hosted AUTH_MODE via email signup and `genericOAuth` social login (`auth/index.ts:145-157` is provider-agnostic), and nothing covers it now — `auth-settings.test.ts` only asserts settings resolution, not that the ban is applied and denies a session. A small self-hosted-mode case would restore it.

src/renderer/components/settings/settings-page.tsx
```diff
@@ -150,9 +158,31 @@ function SettingsPageContent({
...
+    // Deep link or stale active points at a non-content section (e.g. Users
+    // became external-only after platform-auth resolved) — fall back and sync URL.
+    setActive((current) => {
+      if (sectionIds.includes(current)) return current
+      return sectionIds[0] ?? ''
+    })
   }, [initialSection, sectionIds, isMobile])
 
+  const onSectionChangeRef = React.useRef(onSectionChange)
+  onSectionChangeRef.current = onSectionChange
```
iddogino: Minor: ref written during render. Harmless here and the intent is clear (keep the unstable `onSectionChange` identity out of the deps array), but the effect's own guards already make it idempotent — re-running on every render would early-return once `initialSection === active`. Plain deps would work and avoid the render-phase write.

src/renderer/components/settings/auth-tab.tsx
```diff
@@ -14,7 +14,12 @@ import { useSettings, useUpdateSettings } from '@renderer/hooks/use-settings'
 import { Plus, X, Globe } from 'lucide-react'
 import type { AuthSettings } from '@shared/lib/config/settings'
 
-export function AuthTab() {
+interface AuthTabProps {
+  /** Platform-controlled deployments manage signup/auth methods on Platform. */
+  hideLocalAuthSections?: boolean
```
iddogino, line 19: Minor, carried over: the actual SUP-509 behavior is still untested — `global-settings-page.tsx` (the link URL, and that `hideLocalAuthSections` is passed) and this file (that the three sections disappear) have no tests; only the generic external-section mechanism in `settings-page.test.tsx` does. `web-tab.test.tsx:20` already has the `usePlatformAuthStatus` mock pattern to copy.

Also still open: this file is +370/-211 for what is functionally a wrapper, and the `{/* ── Signup & Access ── */}` markers plus the two helper comments were dropped along the way. Re-indent-only (keeping the comments) would shrink the diff a lot and keep git blame useful.

iddogino (changes requested):
Re-reviewed after `4f891da9` + `0a8f3798`. Typecheck clean, ESLint clean on all changed files, 396 tests pass across `src/shared/lib/auth` + `src/renderer/components/settings` + `src/api/middleware` (incl. all 35 token-exchange integration tests, none skipped).

Both earlier blockers are genuinely fixed:

- **Deep-link blank pane** — the fallback + URL-sync effects in `settings-page.tsx` are correct, can't loop (after `change(active)`, `initialSection === active` short-circuits) and can't fight a user click. `AuthGate` blocks children until the session settles, so `isAdmin` is stable before mount and the new effect won't spuriously rewrite `/settings/users` for a slow session either. `settings-page.test.tsx` pins both the async and the warm-cache paths.
- **Approval lockout** — `resolveAuthSettings` centralizes the merge, all raw `DEFAULT_AUTH_SETTINGS` merge sites now route through it (none left outside `config/settings.ts`), and `/api/auth-config` serves the resolved value so the login page stays consistent.

Requesting changes on three remaining items — one new, two carried over. Details inline; the first is the one I'd most want resolved before merge.

src/shared/lib/auth/auth-settings.ts
yiw190: make sense -- will unified the judgment for platform env.

src/renderer/components/settings/global-settings-page.tsx
yiw190: True risk - will take another look and make sure

src/shared/lib/auth/auth-settings.ts
yiw190: suggestion makes sense - will add

yiw190:
Addressed the open review threads in `c7a49d53` (redeployed to staging `yitian-superagent`):

1. **UI/server predicate** — `/api/platform-auth` now returns `platformControlled: isPlatformControlledAuth()`. UI gates Auth hide + Platform invite link on that field (not JWKS `orgId`). Link falls back to `{base}/dashboard` when `orgId` is null.
2. **Stuck pending-approval bans** — `clearPendingApprovalBans()` runs at AUTH_MODE startup.
3. **Second admin** — Users tab stays **local** for role / ban / remove; Invite opens Platform Team when platform-controlled.
4. **Minors** — restored self-hosted pending-approval signup/ban session test; dropped render-phase ref write; added AuthTab / UsersTab / GlobalSettingsPage tests.

Staging canary: `gamut-host-app:sup-509-c7a49d53-auth` on https://yitian-superagent.ongamutstaging.com (release metadata pinned so controller does not reconverge to v0.4.19).

PR #620, opened 2026-07-30
Send agent contact card after iMessage setup

JeremyBischoff:
Send each newly connected iMessage agent as a saveable contact after setup.

Production: 8 files, +168/-4.
Tests: 3 files, +243.

## What changes

create integration → connect → build vCard → upload through the existing iMessage file gateway

- The card fills the contact name, `AI Agent` title, `Gamut` organization, optional description note, Gamut phone number, deployment link, and fixed Gamut photo.
- The link row uses the pretty `{name}-{id}` display slug. `UID` keeps the minted agent id, so renaming an agent cannot mint a second contact on the phone.
- An install with no public URL falls back to `https://gamut.so`, labeled `Gamut` rather than `Open in Gamut`.
- vCard values use CRLF endings, escaped text, and 75-octet folding.
- The iMessage connector now sends `.vcf` files as `text/vcard`.

## Worth reviewing

- **Creation trigger.** Both authenticated creation routes start the upload after the connect step and outside its `try/catch`, so a contact-card failure can never report the integration as failed to connect. When connect does fail, the upload finds no connector and returns.
- **Link source.** The card emits an https URL only on a public web host. Desktop gets the marketing row, because the desktop app link is a `superagent://` scheme that no phone has registered.
- **Gateway fallback.** The upload omits a chat ID so the gateway uses the phone stored during setup. This supports delivery before the user sends a message.

## Why not a connection hook

`connectIntegration` also runs during boot, reconnect backoff, and reconciliation.
A connection hook would resend the card unless this change added persistent delivery state.
The creation hook keeps the behavior create-only without adding migration or resend machinery.

## Deliberate limits

- This does not backfill existing integrations or add resend controls.
Both require persistent delivery state and a separate product decision.
- A create-and-reconnect overlap can still skip one upload.
Fixing that pre-existing lifecycle race would affect every provider.
- Multiple saved agents still share one Gamut phone number.
iOS contact identity behavior needs a separate product decision.
- Telegram, Slack, adaptive contact artwork, and agent avatars remain outside this change.

## Validation

- `npm run typecheck` passed.
- `npm run lint` passed with 0 errors and 40 existing warnings.
- Vitest over the `chat-integrations` and route suites: 74 files passed, 2 skipped, 1754 tests passed, 26 skipped.
- Manual iPhone check: the fallback card opened Gamut and saved the expected identity fields.
- Messages sent and received before and after saving the contact.
The unknown-sender banner disappeared after saving.

Two notes on what that validation no longer covers, after the review round:

- The earlier manual cloud check exercised the cloud-workspace link source, which this branch no longer emits.
The https row now ships only on a public web host.
- The `N:` field moved the display name into the family-name slot so iOS sorts on it.
`FN` drives what Contacts displays and is unchanged, but the card has not been re-checked on a device since that edit.


src/shared/lib/chat-integrations/contact-card.ts
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * Env and settings are read at call time — main-process env is assigned after
+ * this module is imported.
+ */
+export function resolveAgentWebUrl(agentSlug: string): string | null {
+  // `||`, not `??`: an empty value must fall back too, or the link degrades to a
+  // scheme-less dead row.
+  const base =
+    process.env.HOST_PUBLIC_URL?.trim().replace(/\/+$/, '')
+    || readCloudWorkspaceRecord()?.deploymentUrl?.trim().replace(/\/+$/, '')
```
iddogino: **Blocking.** These two sources are mutually exclusive by runtime, and the second one is a different machine.

- `HOST_PUBLIC_URL` is only ever set on a web/cloud host — `utils.ts:142` only reads it in the non-Electron branch, and nothing in the repo sets it for Electron.
- `cloudWorkspace` is only ever *written* from Electron main — `cloud-workspace-service.ts:319` hard-returns `NOT_AVAILABLE` off `isElectronMain()`. So a non-null record implies desktop, where `HOST_PUBLIC_URL` is always unset, so this branch is always the one taken.

And per `docs/cloud-workspace.md`, `deploymentUrl` is "another running instance of this same codebase" — a separate org deployment with its own agents dir and its own minted `[a-z0-9]{10}` ids. The doc is explicit that the desktop app is not a client of it yet ("not yet consumed by the UI; 'Open' just navigates to `deployment_url`").

So on a desktop install with an org cloud workspace, the card emits `https://<org-deployment>/agents/<local-agent-id>` — an id that doesn't exist over there — under the label **"Open in Gamut"**. That's exactly the dishonest-label case the `linkLabel` logic above was written to avoid.

The PR body says the resolver "follows the existing public-host and cloud-workspace configuration sources / does not introduce a second deployment URL policy." The first half is right; the second isn't — `deploymentUrl` has never been used as a self-URL before (its only other consumers are `platform-tab.tsx` display and `cloud-workspace-client.ts` token exchange).

Suggest dropping the cloud-workspace clause entirely. Desktop then falls to the honest `https://gamut.so` / `"Gamut"` row, which is already what ships for desktop-without-a-workspace.

Related: `resolveAgentWebUrl` has exactly one test (the null case). Neither branch is covered, which is why this didn't surface.

src/shared/lib/chat-integrations/contact-card.ts
```diff
@@ -0,0 +1,100 @@
...
+ * vCard for an agent, so iMessage users can save it as a real phone contact
+ * instead of a bare number. Server-side only — `resolveAgentWebUrl` reads
+ * settings, so this must never be imported by the renderer.
+ */
+
+import { readCloudWorkspaceRecord } from '@shared/lib/platform-auth/cloud-workspace-record'
+import { GAMUT_CONTACT_PHOTO_JPEG_BASE64 } from './gamut-contact-photo'
+
+/** The single Gamut iMessage line every agent is reached on. */
+const GAMUT_PHONE_E164 = '+12053967934'
```
iddogino: This duplicates `IMESSAGE_SETUP_NUMBER_RAW = '+12053967934'` at `chat-integration-setup-dialog.tsx:74`. If that line ever changes, the card silently ships a dead `TEL` while setup keeps working — a failure that only shows up on someone's phone.

Suggest exporting one constant (alongside `IMESSAGE_GATEWAY_URL` in `config-schema.ts`) and importing it in both places.

src/api/routes/chat-integrations.ts
```diff
@@ -228,6 +228,8 @@ chatIntegrationsRouter.post('/:id', ResolveAgent(), AgentUser(), async (c) => {
     // Start the integration
     try {
       await chatIntegrationManager.addIntegration(id)
+      // Not awaited: the upload has its own 30s timeout and setup must not block on it.
+      void chatIntegrationManager.sendContactCard(id)
```
iddogino: This sits inside the connect `try`, whose `catch` runs `updateChatIntegrationStatus(id, 'error', ...)`. A *synchronous* throw from `sendContactCard` would therefore report the integration as having failed to connect when it connected fine.

Today the method is `async` so it can't throw synchronously — but that makes the invariant depend on a keyword. Moving this below the `try/catch` costs nothing and matches the stated intent ("failures report without changing integration status").

Not hypothetical for tests: three files mock `chatIntegrationManager` without `sendContactCard` — `chat-integrations.sup229.test.ts:74`, `chat-session-scope.sup202.test.ts:74`, `x-agent-chat.test.ts:60`. They pass today only because none of them exercise a create route. The next test that does gets a `TypeError` surfacing as a bogus `status: 'error'`.

src/api/routes/x-agent-chat.ts
```diff
@@ -156,6 +156,7 @@ xAgentChat.post('/add', async (c) => {
 
     try {
       await chatIntegrationManager.addIntegration(id)
+      void chatIntegrationManager.sendContactCard(id)
```
iddogino: Same as the sibling route — this is inside the `try` whose `catch` flips the integration to `status: 'error'`. Worth moving below the `try/catch` so a cosmetic failure can never be reported as a connect failure.

src/shared/lib/chat-integrations/chat-integration-manager.ts
```diff
@@ -432,6 +434,44 @@ class ChatIntegrationManager {
...
+        slug: integration.agentSlug,
+        name: agent.frontmatter.name,
+        description: agent.frontmatter.description,
+        appUrl: resolveAgentWebUrl(integration.agentSlug),
+      })
+
+      // No chatId: the gateway falls back to the chat it already knows, else
+      // creates one from the phone it stored during `/setup`. That fallback is
+      // what lets this send before the user has ever messaged the agent.
+      await connector.sendFile('', card, `${agent.frontmatter.name}.vcf`)
```
iddogino: `agent.frontmatter.name` is free text and goes straight through as the upload filename. The only other `sendFile` call site (`:2248`) deliberately uses `path.basename` for exactly this reason.

Agent names can contain `/`, newlines, or quotes, which then flow into the gateway's `upload_attachment` filename and whatever it does with it downstream. A `path.basename` or slug pass would keep this consistent with the existing site.

Also worth considering here: `sendFile` accepts a `caption`, and nothing is passed. The user receives a bare `.vcf` attachment with no explanation, and if the upload fails the connector's fallback sends the literal text `[File: <name>.vcf]` (`imessage-connector.ts:359-362`). One caption line would improve both paths.

iddogino (changes requested):
Pulled the branch, installed deps, ran it locally: typecheck clean, lint clean, full `chat-integrations` + route suites pass (782 passed / 7 skipped). I also decoded the embedded photo (real 256x256 white logomark on `#111111`) and generated a sample vCard to check the wire format — folding is correct (max 75 octets/line, CRLF + leading space, no split multi-byte sequences), escaping is correct, and the `item1.URL` / `item1.X-ABLabel` grouping is right.

The vCard builder itself is solid. My concerns are all in the surrounding wiring, and one of them is blocking.

## Blocking

**The cloud-workspace URL fallback fires only in the case where it is wrong.** Details inline on `contact-card.ts`, but the short version: `HOST_PUBLIC_URL` and `cloudWorkspace` are mutually exclusive by runtime, and the second one points at a *different machine*. On a desktop install with an org cloud workspace, the card emits a link to another deployment's agent id — labeled "Open in Gamut", which is exactly the dishonest-label case the `linkLabel` split was written to avoid.

## Worth fixing before merge

- The fire-and-forget call sits inside the connect `try`, so a sync throw would misreport the integration as failed to connect (inline).
- The Gamut phone number is now duplicated in two files (inline).
- The `.vcf` filename is unsanitized user text, unlike the only other `sendFile` call site (inline).
- `sendContactCard` itself has no unit test — only the route-level "was it called" assertion. The provider gate, the two early returns, and the never-throws guarantee are the actual behavior-carrying logic, and a direct test would also pin the `sendFile('', ...)` empty-chatId contract that delivery-before-first-message depends on.

## Minor

- The card arrives as a bare `.vcf` with no caption. `sendFile` accepts one, and the error fallback path (`imessage-connector.ts:359-362`) sends the literal text `[File: <name>.vcf]`, which reads as noise. A one-line caption would cover both.
- `linkUrl` isn't run through `escapeValue`. `encodeURIComponent` covers the slug, but a comma in `base` would truncate the row. Same for the `UID:` slug — safe today given `[a-z0-9-]` ids, but free to harden.
- `N:;${name};;;` puts the whole display name in the *given name* field, so iOS sorts and greets on the full string. Cosmetic.
- The link uses the bare agent id rather than the pretty `{name}-{id}` display slug. The resolver tolerates bare ids so it works, but the URL the user lands on is uglier than the in-app one.

## What's good

The reasoning for *not* hooking `connectIntegration` (boot / reconnect backoff / reconcile would resend forever) is correct and well documented at `chat-integration-manager.ts:437-444`. Refusing to emit `superagent://` on a phone is the right call, and the label-honesty split is a nice touch. The RFC work — CRLF, TEXT escaping, 75-*octet* folding with the 74-octet continuation adjustment and the UTF-8 boundary walk-back — is genuinely correct rather than approximately correct, and the tests pin the non-obvious parts (multi-byte split, `NOTE` omission).

src/shared/lib/chat-integrations/contact-card.ts
JeremyBischoff: Done in `bd15cda4` - dropped the cloud-workspace clause, so `resolveAgentWebUrl` delegates to `resolveAppLinkContext` and returns null on desktop, and both branches are covered now.

src/shared/lib/chat-integrations/contact-card.ts
JeremyBischoff: Done in `bd15cda4` - one exported `IMESSAGE_PHONE_E164` in `config-schema.ts`, imported by the card and the setup dialog.

src/api/routes/chat-integrations.ts
JeremyBischoff: Done in `ee5ac00f` - moved below the try/catch in both routes. When connect fails the call still fires and no-ops, since `connectIntegration` deletes its map entry before rethrowing.

src/api/routes/x-agent-chat.ts
JeremyBischoff: Done in `ee5ac00f` - same move here.

src/shared/lib/chat-integrations/chat-integration-manager.ts
JeremyBischoff: Done in `bd15cda4` - the filename runs through `sanitizeUploadFilename` like the other call site, and the card sends with a caption now.

JeremyBischoff:
Also in `bd15cda4`: direct tests on `sendContactCard` (provider gate, both early returns, never-throws, and the empty-chatId call), escaped `linkUrl` and the `UID` slug, and the link now uses the `{name}-{id}` display slug while `UID` keeps the minted id. `N:` puts the whole name in the family-name slot.

Updated the PR body too. The link-source claim you flagged was wrong and is gone, and the validation section now says what the cloud check no longer covers.

iddogino (approved):

PR #627, opened 2026-07-30
fix: settle finished turns when idle is lost or discarded (SUP-468)

JeremyBischoff:
Closes https://linear.app/datawizz/issue/SUP-468/fix-stuck-working-settle-signal-loss-class-reply-memory-reattach-on

Settle a finished turn when the host loses or discards a settle frame, without inventing idle from a timer.

Footprint: 6 production files, +245 net. 4 test files, +696 net. Supersedes #572.

Likely match for the stuck Working / Thinking card Graham reported in `#champagne-gulping` on local Apple Container: flapping network plus no stream recovery.
Not claimed for his second symptom (reply invisible until container stop) - that is SUP-517.

## What each part does

Three independent fixes.
Together they settle a finished turn when a settle frame is lost or thrown away.

**Part 1 - stop forgetting "we already got a result"**

When the app reopens a conversation (or SSE reconnects), it reattaches the stream.
It used to wipe "last result" memory.

```
result arrives → host remembers "got a result"
external reattach → OLD: forget that memory
idle arrives     → host thinks "stale idle" → ignores it → spinner forever

NEW: reattach keeps the memory → idle settles
```

Nothing was lost on the wire.
The host threw away the note it needed.

`src/shared/lib/container/message-persister.ts`

**Part 2 - if the stream goes quiet mid-turn, ask the container whether the turn ended**

The host only recovered when the socket cleanly closed.
A half-open or wedged link never closes, so recovery never ran.

Quiet is not evidence a turn ended.
A long build is quiet. A session parked on a permission card is quiet for as long as the human takes.
So the host asks instead of assuming.

```
turn believed live
  + no frames for 30s
       → ask the container: is a turn actually in flight?
            ├─ still running → do nothing, socket untouched, ask again in 30s
            ├─ nothing running → we missed the ending: reattach once,
            │                    collect the replay
            └─ no answer / older container → unknown, never "finished"
```

The container already tracks this per turn - it is what gates its own idle-eviction reaper.
That answer is now `turnSettled` on `GET /sessions/:id`.
`isRunning` cannot answer it: that reports the CLI process and stays true for the life of the query loop.

Measured on a live container, same session across a turn boundary:

```
t+0.0s  isRunning=true  turnSettled=false   turn in flight
t+2.5s  isRunning=true  turnSettled=true    turn ended
```

Why it matters that the socket is left alone: the stream is unbuffered fan-out, so anything emitted
during a close/open gap is gone. Assistant text survives via the 15s transcript poll. Permission
requests, background-task updates and command lifecycle frames do not, and a lost input request
parks the session invisibly until the 24h TTL.

A/B on a rebuilt image - same container, same prompt, same five minutes, host code the only variable:

```
gate in place   1 WebSocket connection   (session start only)
gate removed    5 WebSocket connections  (one per probe window)
```

30s is a poll period, not a threshold. Silence is never itself treated as the turn ending, so no
measurement of "normal quiet" rides on the value.

Settling still needs a real container frame. A reattach that replays nothing actionable marks the
session inactive on the next probe rather than retrying forever - deliberately not a completion,
since nothing there proves the work succeeded.

Each stream replacement gets a unique token.
Late frames from the old socket cannot evict the replacement or start a second reconnect.
The container client also binds unsubscribe and map cleanup to the socket that owns them.

```
old muted socket closes late
  → must NOT steal the replacement's ownership
  → must NOT start a second reconnect
  → failed replacement still counts as a real close
```

Files:
- `src/shared/lib/container/message-persister.ts` - silence probe, attach/detach, stream tokens
- `src/shared/lib/container/base-container-client.ts` - socket owns its map entry / unsubscribe
- `agent-container/src/session-manager.ts` + `server.ts` - `turnSettled` on the session endpoint

**Part 3 - replay also says "no background jobs left"**

On late-join / reconnect replay, include the latest bg job list between result and idle.
The host already knows what to do with that frame.
(Usually the frame is absent, not empty: a session that never ran a background task has no snapshot
to replay, so it is simply omitted.)

Without that list, reconnect alone could say "turn over" while the host still held a ghost background job.
The session stayed parked.

```
OLD replay:  result → idle
NEW replay:  result → background snapshot → idle
                              └─ empty list clears the stuck job
```

```
host holding ghost job
  + replay says idle only     → stays "waiting on background"
  + replay says jobs=[] + idle → clears job → settles
```

Replay state is process-local.
A restarted process clears the previous process's result, background snapshot, and idle before accepting a new turn.

```
process restart → clear old result / bg / idle
                → only the new process can settle
```

`agent-container/src/claude-code.ts`

**Part 4 - a replay only settles the turn it belongs to**

Replayed terminal frames carry no turn identity, so an active host could not tell "the turn I am
waiting on ended" from "an earlier turn ended and is being replayed at me". Adopting the second as
the first reports a completion, and persists a success, for work that never ran.

The result uuid supplies that identity.

```
replayed result the host already processed  → dropped
replayed result the host has never seen     → processed, settles the turn
```

Only the result frame is dropped, deliberately not the batch. The idle behind it is still judged by
the existing result-gated guard, which settles only on a result seen for THIS turn - preserving the
case Part 1 and Part 2 exist for, a result delivered live whose idle was lost.

`src/shared/lib/container/message-persister.ts`

**How they fit**

```
Part 1  fixes: idle arrived, host discarded it
Part 2  fixes: idle/result never arrived (dead quiet socket)
Part 3  fixes: replay said "done" but not "no jobs left"
Part 4  fixes: replay said "done" about a different turn
```

Fast path unchanged: if result and idle arrive normally, settle immediately.
These only matter when that path breaks.

## Worth reviewing

1. **Turn identity on replay.** A replayed result the host already processed must not re-arm the turn's result memory, or the idle behind it settles the current turn. Dropping only the result, not the batch, is what keeps the lost-idle case working.
2. **The evidence gate, not the interval.** The probe never settles from elapsed time; it asks, and only a container reporting nothing in flight leads to a reattach. Idle still requires a container-emitted result + idle (and an empty background hold when one exists).
3. **Silence reattach ownership.** A muted old socket closes ~30s after replacement. The map delete and stream token must keep that close from stacking a second subscribe or orphaning the live socket.
4. **Process-local replay reset.** Clearing late-join state on query restart stops an old result/idle from settling a new process. Same-process late-join replay must still work.

## Why not the obvious alternatives

- **#572's 5s grace after result.** Measured green only when idle alone was lost. Red when result and idle were both lost, when idle was discarded on reconnect, and when a ghost background job remained. Also no longer typechecks against main.
- **False-idle / wall-clock watchdog.** Settles without a container frame and re-breaks the honesty rule from #339. A long tool call is also quiet.
- **Query a new `isSettled` field.** Larger, masks the wiped-memory bug, and does not repair a dead stream for the next turn.
- **Ping/pong alone.** Covers only socket death. Discarded idle and missing background snapshot happen on a healthy socket.

## What this does not fix

- **SUP-517** - reply invisible until the container is restarted. Messages come from the on-disk transcript and already poll every 15s, so a dead settle socket cannot hide a persisted reply. Separate defect.
- **Chat-surface indicator freezes** owned by other work (#564 / #565). This PR fixes the shared host settle copy those surfaces read.
- **Cloud ingress reaping itself.** Reattach recovers the host after a quiet stream dies. It does not change the cloud keepalive policy.
- **Containers older than `turnSettled`.** Silence recovery needs the container to answer. An older one omits the field, which the host reads as "unknown" and never as "finished", so it cannot false-complete - but it also will not self-recover. The image ships with the app, so this only spans a rolling upgrade.

## Validation

- `npm run typecheck` - pass
- `npm run lint` - 0 errors (40 pre-existing warnings)
- `npx vitest run src/shared/lib/container` - 949 passed, 5 skipped
- `cd agent-container && npx tsc --noEmit && npm test` - 778 passed, 8 skipped
- Live Apple Container smoke with rebuilt `superagent-container:settle-smoke`:
  - healthy turn settled in 4s with one exact assistant reply
  - SIGSTOP / CONT mid-turn: one replacement connect, one live socket, one exact final reply, settled
  - two real blackholed WebSocket cycles retained replacement ownership and closed the final socket
  - abrupt process loss left the app inactive
  - same-process interrupt / restart stayed active past the old-replay window and settled once from the new process
- Live A/B for the evidence gate, rebuilt image, same container / prompt / duration, host code the only variable:
  - `turnSettled` observed flipping false → true across a real turn boundary while `isRunning` stayed true
  - gate in place: 1 WebSocket connection over a five-minute quiet turn
  - gate removed: 5 WebSocket connections, one per probe window (the control, proving the probe fires)

### DIFF
--- a/agent-container/src/claude-code-late-join-replay.test.ts
+++ b/agent-container/src/claude-code-late-join-replay.test.ts
@@ -1,0 +1,86 @@
+/**
+ * Regression: late-join replay must carry the latest background_tasks_changed
+ * snapshot. Without it, a host holding a task whose terminal signal was lost
+ * parks in waiting-background after reattach announces "turn over".
+ */
+import { describe, it, expect } from 'vitest'
+import { z } from 'zod'
+import { ClaudeCodeProcess } from './claude-code'
+import { parseBackgroundTasksChanged } from './session-settlement'
+
+const backgroundTasksFrameSchema = z
+  .object({
+    type: z.literal('system'),
+    subtype: z.literal('background_tasks_changed'),
+    tasks: z.array(
+      z.object({
+        task_id: z.string(),
+        task_type: z.string().optional(),
+        description: z.string().optional(),
+      })
+    ),
+  })
+  .superRefine((frame, ctx) => {
+    if (!parseBackgroundTasksChanged(frame)) {
+      ctx.addIssue({ code: 'custom', message: 'rejected by parseBackgroundTasksChanged' })
+    }
+  })
+
+const resultFrameSchema = z.object({
+  type: z.literal('result'),
+  subtype: z.literal('success'),
+})
+
+const sessionStateFrameSchema = z.object({
+  type: z.literal('system'),
+  subtype: z.literal('session_state_changed'),
+  state: z.enum(['idle', 'running']),
+})
+
+describe('getLateJoinReplay — background snapshot', () => {
+  it('includes the latest background_tasks_changed between result and idle', () => {
+    const proc = new ClaudeCodeProcess({
+      sessionId: 'late-join-bg',
+      workingDirectory: '/tmp',
+    })
+
+    const track = (frame: unknown) =>
+      (proc as unknown as { trackForLateJoinReplay(message: unknown): void }).trackForLateJoinReplay(
+        frame
+      )
+
+    track(resultFrameSchema.parse({ type: 'result', subtype: 'success' }))
+    track(
+      backgroundTasksFrameSchema.parse({
+        type: 'system',
+        subtype: 'background_tasks_changed',
+        tasks: [],
+      })
+    )
+    track(
+      sessionStateFrameSchema.parse({
+        type: 'system',
+        subtype: 'session_state_changed',
+        state: 'idle',
+      })
+    )
+
+    const replay = proc.getLateJoinReplay() as Array<Record<string, unknown>>
+    const subtypes = replay.map((f) => f.subtype ?? f.type)
+
+    expect(subtypes).toContain('background_tasks_changed')
+    const bgIndex = replay.findIndex((f) => f.subtype === 'background_tasks_changed')
+    const resultIndex = replay.findIndex((f) => f.type === 'result')
+    const idleIndex = replay.findIndex(
+      (f) => f.subtype === 'session_state_changed' && f.state === 'idle'
+    )
+    expect(resultIndex).toBeGreaterThanOrEqual(0)
+    expect(bgIndex).toBeGreaterThan(resultIndex)
+    expect(idleIndex).toBeGreaterThan(bgIndex)
+    expect(replay[bgIndex]).toMatchObject({
+      subtype: 'background_tasks_changed',
+      tasks: [],
+      replayed: true,
+    })
+  })
+})
--- a/agent-container/src/claude-code-late-join-replay.test.ts
+++ b/agent-container/src/claude-code-late-join-replay.test.ts
@@ -54,6 +54,13 @@ describe('getLateJoinReplay — background snapshot', () => {
       backgroundTasksFrameSchema.parse({
         type: 'system',
         subtype: 'background_tasks_changed',
+        tasks: [{ task_id: 'finished-task' }],
+      })
+    )
+    track(
+      backgroundTasksFrameSchema.parse({
+        type: 'system',
+        subtype: 'background_tasks_changed',
         tasks: [],
       })
     )
--- a/agent-container/src/claude-code-stop-race.test.ts
+++ b/agent-container/src/claude-code-stop-race.test.ts
@@ -104,6 +104,36 @@ describe('ClaudeCodeProcess stop/restart teardown race', () => {
     await proc.stop()
   })
 
+  it('a restart clears process-local background replay state', async () => {
+    const proc = new ClaudeCodeProcess({ sessionId: 'replay-reset', workingDirectory: '/tmp' })
+    const track = (message: unknown) =>
+      (
+        proc as unknown as {
+          trackForLateJoinReplay(frame: unknown): void
+        }
+      ).trackForLateJoinReplay(message)
+
+    await proc.start()
+    track({ type: 'result', subtype: 'success' })
+    track({
+      type: 'system',
+      subtype: 'background_tasks_changed',
+      tasks: [{ task_id: 'old-process-task' }],
+    })
+    track({ type: 'system', subtype: 'session_state_changed', state: 'idle' })
+
+    await proc.stop()
+    await proc.sendMessage('follow-up')
+    expect(proc.getLateJoinReplay()).toEqual([])
+    track({ type: 'result', subtype: 'success' })
+    track({ type: 'system', subtype: 'session_state_changed', state: 'idle' })
+
+    const replay = proc.getLateJoinReplay() as Array<Record<string, unknown>>
+    expect(replay.some((frame) => frame.subtype === 'background_tasks_changed')).toBe(false)
+
+    await proc.stop()
+  })
+
   it('interrupt() landing during a stop() teardown must not revive the query', async () => {
     const proc = new ClaudeCodeProcess({ sessionId: 's4', workingDirectory: '/tmp' })
     await proc.start()
--- a/agent-container/src/claude-code.ts
+++ b/agent-container/src/claude-code.ts
@@ -487,6 +487,10 @@ export class ClaudeCodeProcess extends EventEmitter {
   private lastTurnInformationals: SDKMessage[] = [];
   private lastResultMessage: SDKMessage | null = null;
   private lastSessionState: string | null = null;
+  // Latest authoritative background-task snapshot, replayed to a late joiner:
+  // a host holding a task whose terminal signal was lost would otherwise stay
+  // pinned forever (the live consumer self-heals when a snapshot arrives).
+  private lastBackgroundTasksChanged: SDKMessage | null = null;
   // Pre-spawned CLI subprocess from prewarm(), waiting for a prompt. Claimed
   // (once) by the next createQuery; see prewarm() for why the handle lives on
   // the process rather than in a detached pool.
@@ -1528,6 +1532,8 @@ export class ClaudeCodeProcess extends EventEmitter {
       this.lastResultMessage = message;
       this.lastTurnInformationals = this.currentTurnInformationals;
       this.currentTurnInformationals = [];
+    } else if (msg.type === 'system' && msg.subtype === 'background_tasks_changed') {
+      this.lastBackgroundTasksChanged = message;
     } else if (msg.type === 'system' && msg.subtype === 'session_state_changed') {
       this.lastSessionState = msg.state ?? null;
     }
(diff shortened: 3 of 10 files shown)


iddogino (changes requested):
Pulled the branch, bootstrapped both packages, read the full diff plus the surrounding settle/transport machinery, and reproduced the checks.

**Validation reproduced:** typecheck clean (root + `agent-container`), lint 0 errors, `src/shared/lib/container` 940 passed / 5 skipped, `agent-container` 778 passed / 8 skipped. Merges cleanly with current `main` — nothing since `f17213d6` has touched these three files.

## What holds up

I went after the trickiest mechanics specifically, and they're correct:

- **Token ordering in `attachStreamTransport`** — the new token is installed *before* the old socket is closed, so the replaced socket's late `connection_closed` is dropped, and both the `.then` and `.catch` in `handleConnectionClosed` re-check the token. No stacked resubscribe, no orphaned live socket.
- **`base-container-client.ts`** — the `wsConnections.get(sessionId) === ws` identity checks, the `ownedSocket`-scoped `unsubscribe`, and the `unsubscribed` flag read after `await getPortOrThrow()` are all sound; there's no window where `ownedSocket` is assigned after the flag is read, and unsubscribing before the socket exists correctly suppresses creation.
- **`cancelSettleProbe` coverage** is complete: `finalizeIdle`, `markSessionInactive`, `markSessionInterrupted`, `unsubscribeFromSession`, and the error/interrupt result branch.
- **Part 1 is safe on the send path** — both `markSessionActive` call sites null `lastResultSubtype` regardless of ordering vs `subscribeToSession`, so preserving it across reattach can't leak a prior turn's result into a fresh one.
- **Part 3's reset is per-query, not per-turn** (`initializeQuery` only), so same-process late-join replay still works.

Parts 1 and 3 look right to me as written. The concerns below are all Part 2.

## 1. [High] Stale replay can settle a turn that never ran, and fires a false completion notification

`reattachAfterSilence` only fires while `isActive` is true — which is exactly the condition that bypasses the `content.replayed && !state.isActive` guard at `src/shared/lib/container/message-persister.ts:1491`. Replayed terminal frames carry no turn identity, so the host cannot distinguish "the turn I'm waiting on ended" from "some earlier turn ended."

Reachable path today:

- `src/api/routes/agents.ts:1955` calls `markSessionActive` **before** `await client.sendMessage(...)` (`:1998`)
- the catch at `:2009` returns 500 without reverting `isActive`
- `sendMessage`'s abort timeout is `30000`ms — exactly `STREAM_SILENCE_REATTACH_MS`

So: wedged container send → probe fires at the same instant → reattach → container (still idle from the *previous* turn) replays that turn's `result` + `idle`.

I probed this against the branch:

```
PROBE activity after stale replay: idle
PROBE completion notifications fired: 1
```

…for a message that never reached the agent. `persistAutomationStatus` records success too. That's arguably worse than the stuck spinner it replaces — the spinner is wrong but honest; this is affirmatively false.

The PR body's honesty rule ("settling still requires a frame the container actually emitted") holds literally, but the frame can be *stale*, and nothing on the wire distinguishes the two.

**Suggested fix:** the container already has `lastResultMessage.uuid` — either dedupe replayed results by uuid host-side, or (better, and it also addresses #3) put the container's current `session_state` on the `capabilities` handshake, which is already sent on every attach (`agent-container/src/server.ts:1911`). Then the host learns idle-vs-running authoritatively on reattach instead of inferring it from unlabeled replay. Reverting `markSessionActive` when the send fails is worth doing independently of this PR.

## 2. [Medium] The probe fires on any 30s quiet — including healthy turns — and each reattach is a lossy transport swap

Verified: a 5-minute quiet tool call produces **11 subscribes** (1 + 10 reattaches).

The container stream is pure fan-out with no buffering (`session-manager.ts:792` `subscribe`, consumed at `server.ts:1920`), so anything emitted in the close→open gap is gone permanently. Assistant text survives via the 15s transcript poll; `background_tasks_changed`, input/permission requests, and `command_lifecycle` do not. There's a `GET /inputs/pending` endpoint but nothing on the host calls it, so a lost input-request frame parks the session invisibly until the 24h TTL.

30s of mid-turn silence is routine in this product — long builds and test runs, plus any session sitting on a permission card waiting on a human. `it('a turn that starts on an already-quiet stream arms recovery')` enshrines exactly this.

**Suggested fix:** gate the reattach on positive evidence the turn ended (the handshake state from #1), and/or make-before-break — the container allows multiple subscribers per session; it's the host's 1-per-session `wsConnections` map plus the `existing.close()` in `setupWebSocket` that forces the gap. Failing both, a threshold well above typical tool duration.

## 3. [Low] `reattachAfterSilence` can loop forever with no path to settle

After a container-side process replacement, Part 3 correctly clears the replay state, so `getLateJoinReplay()` returns `[]` — a host stuck active then reattaches every 30s indefinitely with no way out.

Worth noting `getSession().isRunning` cannot rescue this: `isRunning()` is `isReady && isProcessing` (`claude-code.ts:1527`), which stays true for the life of the query loop, not per-turn. So there's currently no container-side answer to "is a turn in flight" — which is the same gap the handshake-state suggestion in #1 would close. A bounded retry plus a log would at least make it diagnosable in the meantime.

## 4. [Low] Repeated reattach re-replays informationals → repeated full-transcript reads

`appendInformationalEntry` is idempotent by uuid, but it `readFile`s the entire JSONL on every call (`src/shared/lib/services/session-transcript-append.ts:26`). Given the transcript-perf work in flight (#609–#611) and multi-hundred-MB transcripts, a stuck session re-reading on every 30s reattach is a real cost.

## 5. Nits

- `detachStreamTransport` writes a dead `streamTokens` entry that only `unsubscribeFromSession` deletes, so `markAllSessionsInactiveForAgent` (`message-persister.ts:963`) leaves one behind. Bounded by session count and harmless, but the asymmetry is easy to miss later.
- `initializeQuery`'s new comment replaced the one explaining why `emit('query-start')` exists ("the SessionManager listens for this to reset its settlement bookkeeping"). That rationale is now undocumented — worth keeping both.
- PR body says the replayed background snapshot is "usually empty"; it's usually *absent* — `lastBackgroundTasksChanged` stays null for sessions that never ran a background task, so the frame is simply omitted. Behaviorally fine, just a description mismatch.
- Stray blank-line removal before `constructor()` in `message-persister.ts`.

---

Requesting changes on **#1** specifically — it converts a visible-but-honest stuck state into a silent false completion (plus a persisted success outcome), and it's reachable through the ordinary send path. #2 is worth a decision before merge even if the answer is "accepted risk." The rest can follow up.

JeremyBischoff:
Done in `0ed497e0`.

1. Fixed with the result uuid: a replayed result the host already processed no longer re-arms the turn's result memory, so the idle behind it can't settle the current turn. Only the result frame is dropped, so a turn whose result landed live and whose idle was lost still settles.

2. Gated it on evidence instead of a threshold. `SessionSettlementTracker` already has the per-turn answer and it's what gates idle eviction, so it's now `turnSettled` on `GET /sessions/:id` and the probe asks before touching the socket. Same container and same five-minute turn: 1 subscribe with the gate, 5 without.

3. Bounded. Second consecutive settled-container probe marks the session inactive instead of reattaching again. Not a completion.

4. Mostly falls away with #2, since a healthy turn no longer reattaches.

5. Nits done.

On `isRunning`: it's process-level, but the settlement tracker has the per-turn answer. On a live container `isRunning` stayed true across a turn boundary while `turnSettled` flipped.

Recovery now needs a container new enough to send `turnSettled`. An older one reports nothing, which reads as unknown and never as finished, so it won't false-complete but it won't self-recover either.

Body's updated too - Part 2 describes the gate rather than the old silence-reconnect, and the background snapshot is usually absent rather than empty.

JeremyBischoff:
Closing. Parked under SUP-468. #1008 reworked the same settle code, so a fix for silent connection loss needs a rebuild on top of it.

PR #630, opened 2026-07-31
Autopilot mode: goal-contract engagement, watchdog review loop, approval auto-review (SUP-463)

iddogino:
## What

Autopilot mode: the user flips a composer switch to delegate a task end-to-end. The agent preflights (accounts, credentials, scope), declares a goal contract via a blocking `engage_autopilot` tool, and runs autonomously; a watchdog reviews every session stop against the contract with a small judge model and either lets the session rest (done), restarts it with a nudge (continue), or pauses and notifies (blocked / guardrails). While engaged, review-tier API/MCP calls are auto-judged on the user's behalf by an injection-resistant approval reviewer instead of parking approval cards the user isn't there to answer.

## Design highlights

- **State machine on session metadata** (`off → requested → engaged → paused`), every transition a guarded read-modify-write under the metadata file lock. Host-authoritative: the state rides every message to the container; changes regenerate the system prompt and re-query.
- **Watchdog** subscribes to global session events; judge verdicts are Zod-parsed and fail closed (unusable verdict ⇒ pause + notify, never loop blind). Mechanical guardrails: iteration cap and a no-progress fingerprint, applied atomically with the verdict. Restart reconciliation sweeps for sessions left engaged across an app restart.
- **Injection-resistant approval reviewer**: the judge sees ONLY user-typed messages (compaction summaries, sidechain prompts, cross-agent messages, and all harness traffic are excluded) plus the requested action with a capped body preview. Fail-closed on every error path. Decisions are audited (`approved_autopilot`/`denied_autopilot` + reason) and rendered as batched tool-call-style transcript cards.
- **Ask-the-user tools are denied with corrective feedback while engaged** (dual container seams: PreToolUse hook + canUseTool, covering the bypassPermissions shadow); computer-use/host-script grants and capability review-tier launches auto-reject with transcript cards.
- **Mid-turn semantics**: a queued message with the switch on is steering — the session stays engaged; a fresh-turn message drops engaged back to requested for re-preflight; flag-less surfaces disengage entirely. Cross-agent invocation into an engaged session is refused (409).

## Test plan

- Host suite 8,278 passed (463 files), container suite 794 passed; tsc + eslint clean in both packages
- ~70 new unit tests across the autopilot modules, routes, persister, notification manager, and renderer (grouping, windowing, toggle, stream resync)
- E2E (mock container): toggle round-trips off↔requested with reload persistence; seeded engaged session renders toggle state, batched approval decisions with expand, and the watchdog verdict card
- Validated live: full engage → continue ×N → done run, plus the approval reviewer approving/denying real Gmail proxy calls

https://claude.ai/code/session_01X1wxM8pTqYPvjR6Sy2kABX


iddogino:
Addressed all six findings in 3d24a745 (branch also rebased onto main — the merge conflicts in `agents.ts`, `agent-home.tsx`, and `message-input.tsx` are resolved and the PR is mergeable again).

**1. Approval context crossed session/task boundaries →** One session, one era. `requestedAt` is now stamped on every `requested` transition (and preserved through engagement); the judge sees only the single engaged session's user messages from that era onward. With more than one engaged session the origin is ambiguous and the request is denied without consulting the judge (the denial card lands in every engaged session so the requester sees why). Per-request session attribution on the proxy token would let us do better than deny-on-ambiguity — tracked as a follow-up.

**2. Untrusted transcript → nudge laundering →** The judge now returns `missing_criteria` (1-based indexes, validated against the stored contract), and the `[SYSTEM]` continuation is built exclusively from the contract's own criteria text — judge free text never enters agent-facing traffic (it survives only on the display card, which is host-JSONL/display-only). Invalid indexes degrade to a generic continuation; the no-progress fingerprint keys on the index set, which also makes the guardrail deterministic. The judge prompt additionally gets an explicit untrusted-data framing.

**3/P1 request representation →** The judged action now includes the full query string, and the body/arguments are shown to the judge COMPLETELY, with a 16,000-char reviewable cap that fails closed: an oversized or unreadable payload is denied without judging (`cannot be inspected in full`), audited as `denied_autopilot`. Applied to both the API proxy and the MCP proxy (`Arguments (complete)` replaces the 1,500-char preview).

**4. Queued transitions never reached the container →** The post-send special case is gone: transitions apply before every send, so the container always receives the new state with the message. The one queued send that must not restart the in-flight query — switch on while already engaged — is exactly the case with no state change; the remaining queued transitions (explicit toggle-off, paused→requested, foreign-surface disengage) are deliberate state changes where the interrupt/restart is the correct behavior (paused→requested now even gets the preflight reminder on that same message instead of dead-ending).

**5. Completion evidence unbounded →** The watchdog's transcript excerpt is filtered to the same engagement era (`requestedAt`, falling back to `engagedAt` for pre-existing blocks), so older work in a reused session can't satisfy the current contract's criteria.

Validation: host 8,349 tests, container 794, tsc + lint clean both packages, all three autopilot Playwright tests green. New regression tests cover each finding: multi-session ambiguity deny, era-bounded intent and evidence, injected-nudge exclusion (asserts the literal injection string never reaches the [SYSTEM] message), index-fingerprint no-progress, oversized body/args fail-closed, query-string visibility, and queued transition-before-send ordering.

https://claude.ai/code/session_01X1wxM8pTqYPvjR6Sy2kABX

iddogino:
Second review round addressed in `d41a70e5` (+ `e71bc411` for the CI flake). All five findings implemented:

**1. New sessions had no era boundary** — the create-session route now captures the era timestamp **before** `createSession` and persists it on the initial `requested` block, so `requestedAt` always precedes the initial prompt's transcript timestamp and the approval reviewer sees the task statement. Regression test asserts the stamp is taken before the container call (`agents.test.ts`, "opens the autopilot era BEFORE the session exists").

**2. Interrupts retained the previous era** — the watchdog's `engaged → requested` interrupt transition now restamps `requestedAt`. You're right that the next switch-on message is a no-op transition and can't restamp, so the interrupt itself opens the fresh era. Test ages the marker and asserts the restamp.

**3. Criterion identity ≠ no progress** — the guard now needs both signals: an unchanged missing-criteria fingerprint **and** the judge not affirming progress. The judge is shown a `PREVIOUS REVIEW` section (decoded from the stored criterion-index fingerprint back to contract text only — free-form fingerprints are deliberately never echoed into the trusted prompt section, tested) and returns a `made_progress` boolean, judged on the work done since that review. This uses the actual evidence (the era transcript) rather than criterion identity; the iteration cap remains the unconditional bound. Tests cover both directions plus the no-echo property.

**4. Forwarded headers now reviewed** — the API-proxy judge sees `Request headers (complete, as forwarded)` built from the exact skip-set both account providers apply when forwarding (nango now imports the shared `PROXY_SKIP_REQUEST_HEADERS` constant instead of a local copy, so review and forward cannot drift). `authorization` is in the skip-set, so credentials never reach the judge. A header block beyond the 16k reviewable cap fails closed (`denied_autopilot`, judge never consulted). Tests: Dropbox-API-Arg-style header visible to the judge, credential absent, oversize header fail-closed.

**5. MCP representation made canonical** — while engaged, the body must be exactly one well-formed UTF-8 JSON-RPC object with a string `method`; batched arrays, malformed JSON, and non-UTF-8 bodies are denied outright without judging. For canonical requests the judge now sees the real target URL including the query string, the complete forwarded header set (same skip-set as the forward path, hoisted to one shared constant), and the **complete JSON-RPC body** (replacing the extracted-arguments view, which also fixes non-`tools/call` review-tier methods whose params were previously invisible). Everything still under the 16k fail-closed cap. Tests: batch deny, malformed deny, full-body + query + headers visibility, oversize fail-closed.

**Red Tests workflow** — reproduced the failure shape: the two engaged-state auto-reject tests in `message-persister.test.ts` queued session metadata with `mockResolvedValueOnce`, which a stray async consumer from an earlier test can steal under slow coverage timing, downgrading the test to the not-engaged path. Couldn't reproduce in 3 local coverage runs, consistent with a timing race; both tests now pin the value for their duration and restore in a `finally` (`e71bc411`).

Validation: host 8,357 tests passed, focused suites 523 passed, tsc + eslint clean, autopilot Playwright 3/3.

iddogino:
Round-4 findings addressed in f5d18f13 (fixes) + 571f6c98 (merge of main, resolving the agents.test.ts / setup.ts conflicts).

**1. [P1] Approval executing after revocation.** Both proxy routes now capture an authorization snapshot before the model review — the engaged session set with each session's `requestedAt` era marker (`getAutopilotAuthorization`) — and revalidate it immediately before promoting an approve verdict (`isAutopilotAuthorizationCurrent`). Any change invalidates: a disengage, an interrupt's era restamp (so even same-session re-engagement for a *new* task doesn't inherit the approval), or a session appearing/disappearing. Cancelled requests (`c.req.raw.signal.aborted`) are also dropped instead of forwarded. Denials audit as `denied_autopilot` with an explicit reason. Tests: route-level revoked-mid-review tests in both proxies (judge approves → 403, upstream never called, revalidation compares the *same* pre-review snapshot), plus unit tests covering disengage, era-restamp-then-re-engage, and new-session invalidation.

**2. [P1] First-turn engagement race.** The initial-metadata write (including the `requested` era) is now awaited **before** `markSessionActive`/`subscribeToSession`. `engage_autopilot` reaches the host only through stream-event handling, so seeding the state before the subscribe closes the window completely — no fire-and-forget, no dependence on DB/registration timing. New test asserts via mock invocation order that the autopilot metadata write completes before the stream subscribe.

**3. [P1] Linux CI failure.** You were right and my flake diagnosis was wrong — the failure is deterministic on Ubuntu: `handleComputerUseRequestTool` exits through the unsupported-platform rejection before the autopilot branch, which satisfies the test's reject-waitFor while making the denial card unreachable. The test now enables the existing `E2E_MOCK` platform bypass (restored in `finally`), same as its auto-execute neighbor. The script-run twin is unaffected (`linux: ['shell']` passes that handler's platform guard).

**4. [P2] Lossy API body review.** The API proxy now decodes with `TextDecoder('utf-8', { fatal: true })` and no trimming, matching the MCP path — the judge sees exactly the bytes that forward. Non-UTF-8 bodies fail closed with `denied_autopilot` before the judge runs; test sends an invalid UTF-8 sequence and asserts the deny with no judge/upstream call.

**5. [P2] Watchdog test race.** The mechanical-block test now waits for the review entry itself (`vi.waitFor` around the file read) rather than reading the instant `paused` lands — `pauseAutopilot` persists before `appendReview` is awaited, exactly as you reproduced. The idle-review tests were audited too: they settle on the `finished` broadcast, which fires after the append, so they're not exposed.

**Mergeability:** merged `origin/main` (now a1618754) into the branch; conflicts in `agents.test.ts` (import union; dropped main's removed `listPendingScheduledTasksByAgents`) and `renderer/test/setup.ts` (main's `globalThis` ResizeObserver + canvas stubs supersede this branch's window-scoped stub).

Validation at 571f6c98: full host suite 8,446 passed / 45 skipped; focused proxy+autopilot+agents suites 532 passed; `tsc --noEmit` + eslint clean; autopilot Playwright E2E 3/3.

https://claude.ai/code/session_01X1wxM8pTqYPvjR6Sy2kABX

iddogino:
Round-5 findings addressed in 79ef4d39 (fixes) + 86010ba6 (merge of current main, now 64d0df06).

**1. [P1] Registration erasing the seeded state.** Confirmed — `registerSession` replaces the entry, so my round-4 ordering was self-defeating. Registration is now the ONE initial metadata write, carrying `initialMetadata` (autopilot era, model, effort, speed, creator) as its existing 4th parameter, awaited before the stream attaches, with the later registration removed. Two guards against regression: the route tests now assert on `registerSession` (with a no-second-autopilot-write check), and a new service-level test pins the replace semantics themselves — the exact repro you ran (seed via update, register, autopilot gone) plus the carried-in-registration round-trip.

**2. [P1] Stale watchdog verdicts.** Every verdict-driven transition now carries a `VerdictEraGuard` checked atomically inside the metadata lock: `disengageAutopilot('completed')`, `pauseAutopilot` (blocked + both escalation tails + mechanical blocks), `applyContinueVerdict`, and the nudge's pre-send recheck all no-op unless the current `requestedAt` still matches the era the verdict was computed under. `review()` captures the era with its goal snapshot; the failure-path escalation in `handleIdle` uses the era captured before the review started. I chose `requestedAt` alone as the identity: every path to re-engagement (interrupt restamp, or disengage → re-request) stamps a fresh era, so goal identity is subsumed. Covered by service-level guard tests plus two watchdog integration tests that hold the judge open, re-engage a new era mid-review, and assert a stale `done` cannot complete it and a stale `continue` cannot advance or nudge it.

**3. [P1] Revalidation at the forwarding boundary.** The guard now runs immediately before the actual outbound operation. API proxy: `makeApiCall` gained the suggested `beforeForward` callback — awaited by nango and by both composio modes (direct + proxy) after their internal credential/mode resolution, immediately before the fetch/`proxyExecute`; throwing `AutopilotAuthorizationError` aborts the forward and the route translates it into the standard 403 + `denied_autopilot` audit. MCP proxy: inline guard immediately before `mcpSafeFetch`, after token refresh and DB writes. Cancellation (`signal.aborted`) is checked at the same point. On the test gap you called out: the route tests now mirror the provider contract (the `makeApiCall` mock awaits `beforeForward`, and the revocation test asserts the outbound call is never reached), and a new `before-forward-contract.test.ts` pins the real providers — ordering (`token → guard → outbound`) and abort-on-throw for nango and both composio paths.

**4. [P2] Misleading approved card.** `reviewAutopilotApproval` now returns an idempotent `recordFinalDecision` handle: denials are final at review time and record immediately; approvals record nothing until the route reports its final decision — the executed approval (recorded when the boundary guard passes) or the deny substituted at the boundary. Reviewer tests cover deferred approval recording, the substituted-deny card, and once-only recording.

**Mergeability:** merged main through `64d0df06` (the SUP-478 cloud-agents merge included); the `agents.test.ts` conflict was main's new once-grant suite landing beside this branch's autopilot suites — all three kept.

**On the red broad Playwright job:** `gh run` history shows the identical 40s-timeout cascade on **main itself** — run 30672235775 at `8478f285` failed 141 tests with the same connections/MCP-toggle signature (a commit this PR didn't contain at the time), while the next main run at `d3a329ff` passed the job. Earlier runs of this PR passed the job (their only red was the Linux unit test fixed in round 4). So the cascade is a job-level flake that hits main and PRs alike; the new push re-runs it.

Validation at 86010ba6: full host suite 8,800 passed / 45 skipped; `tsc --noEmit` + eslint clean; autopilot Playwright E2E 3/3; ~15 new regression tests across the four findings.

https://claude.ai/code/session_01X1wxM8pTqYPvjR6Sy2kABX

PR #633, opened 2026-07-31
Narrow the browser stream during credential handoff (SUP-498)

JeremyBischoff:
Closes https://linear.app/datawizz/issue/SUP-498/narrow-the-browser-viewport-during-a-credential-handoff-so-sign-in

While `request_browser_input` is pending, narrow the streamed page to 450×900 so login forms reflow into the tray. Clear when nothing is pending. Per-handoff desktop escape in the handoff bar. Canvas `min-h-0` so Done/Decline stay on screen.

**Footprint:** prod 7 files / net +237 · test 3 files / net +351

## How the width is decided

```
pending browser handoff?
  no  → clear override, reset escape
  yes → user asked for desktop this handoff?
          yes → clear override
          no  → set 450×900 (mobile:false, deviceScaleFactor:1)
```

The input manager notifies on any change to its pending map and the viewport subscribes once, so every route that opens or clears a handoff re-derives without naming them: Done, Decline, browser close, session delete, the 60s sweep that retires abandoned requests, and handoff create. Screencast attach and the escape toggle derive directly. Tab switch and OAuth popup re-enter attach, so they pick up the current answer with no per-target bookkeeping.

## The container owns the mode

The container reports the mode it applied as a `viewport_mode` stream message and the tray renders that. An earlier revision had the renderer infer the mode from the width of arriving frames against two threshold constants; moving the answer to the side that already knows it removed both thresholds, three mirroring refs, a settling spinner and its timeout.

## Why not widen the drawer or zoom the frame

Measured on GitHub / Google / LinkedIn: the form is ~30% of a 1280px frame. Scaling that into a 450px tray leaves ~130px fields. Widening the drawer magnifies the empty background with the form and still cannot reach 1:1 under the 800px drawer cap. Zoom has the same problem and would rework click coordinate mapping on a password path. Reflow at phone width takes the same sites to ~86-93% form share.

`mobile: true` looks fine on those three sites and fails on a legacy page with no viewport meta (Chrome's 980px fallback scaled to fit → 37% usable). `mobile: false` holds at 81% there. `deviceScaleFactor` is required by CDP; omitting it rejects the call.

## Worth reviewing

1. **The pending-map subscription in `input-manager.ts`.** This started as seven hand-written `syncBrowserViewport()` calls, one per route that opens or clears a handoff. Three of them were missing: session delete, the 60s abandoned-request sweep, and (separately) a tab switch left the departed target narrowed for the rest of the session. Rather than add the three missing calls, the input manager now notifies on any pending-map change and the viewport subscribes once, so a route added later cannot forget. This is the one part of the PR that touches a primitive shared with secret, question and connected-account requests - the change is additive (fire a callback after mutation, one subscriber), but it is the diff to read hardest. It also buys the coverage below.
2. **Escape across viewer reconnect.** Cleanup clears the CDP override always, but keeps the escape flag while a handoff is still pending so reconnect re-derives desktop. Browser close rejects pendings then syncs, so the next handoff still starts narrow. The sticky-escape failure mode is the one to read for.
3. **`mobile: false` and `deviceScaleFactor: 1`.** Easy to "fix" toward a phone UA. Don't - both are load-bearing.
4. **Canvas `min-h-0`.** Unconditional. At today's 1280×720 the box never hits the floor; at 450×900 it stops starving Done/Decline. Same bound-the-slot shape as the activity-card overflow fix.
5. **The toggle does not flip on the report alone.** It flips locally and sends, because the next click derives from that value - letting only the report move it means one dropped report latches the toggle dead. It commits nothing if the socket was not open, so it cannot show a mode we never asked for.

## Known limits (not fixed here)

- UA-sniff sites ignore width → no reflow; escape recovers to today's desktop path (no regression, no gain).
- Centered fixed-width layouts may need pan inside the canvas.
- Overlapping handoffs share one escape flag, so a second handoff opening while the first is escaped inherits desktop width. Scoping it per request costs a wire field, and one shared browser showing one shared page makes the current behaviour defensible.
- Escape is shared across tabs for the whole handoff rather than per tab, so an OAuth popup keeps the width you chose instead of snapping back to narrow mid-login.
- `server.ts` attach/teardown wiring is not unit-tested (the module exports nothing, and refactoring it for testability is out of scope). What remains uncovered there is now three lines of CDP/socket wiring: the override clear on the target a tab switch leaves, the guard that stops a late-opening stale socket claiming the viewport channel, and handling the toggle ahead of the screencast guard. Those rest on code reading plus hand smoke. The pending-map invariant, which is where all three bugs actually were, moved into `input-manager.ts` and is covered.
- Escape is not persisted per site. Wait for a repetition signal before that.

## Validation

- `npm run typecheck` - green
- `npm run lint` - 0 errors
- `cd agent-container && npx tsc --noEmit` - green
- `npx vitest run src/renderer` - 1988 passed
- `npm --prefix agent-container test` - 791 passed, 8 skipped
- One table-driven test covers every route that clears the last handoff - resolve, decline, browser close, session delete, stale sweep - plus the open case. Deleting the notify from the sweep fails only the sweep row; deleting it from session delete fails only that row.
- Hand smoke on a live login: narrow, escape, toggle back, tab switch mid-handoff leaves no stuck-narrow tab, session delete mid-handoff does not leak escape into the next handoff.
- E2E left to CI. Locally it fails to boot on a `better-sqlite3` ABI mismatch after an Electron dev session, which is environmental.

Each new test was verified red against the defect it guards, then green.

Container half ships in the agent image. Existing agents need image rebuild + container recreate before live narrow applies. Renderer half is dormant until a narrowed frame arrives.

### DIFF
--- a/agent-container/src/browser-viewport.test.ts
+++ b/agent-container/src/browser-viewport.test.ts
@@ -1,0 +1,207 @@
+import { describe, it, expect, beforeEach, beforeAll } from 'vitest'
+import { inputManager } from './input-manager'
+import {
+  setViewportChannel,
+  syncBrowserViewport,
+  requestDesktopWidth,
+  detachViewportChannel,
+  startBrowserViewportSync,
+} from './browser-viewport'
+
+type Sent = { method: string; params?: Record<string, unknown> }
+
+/** Register a recording channel and return the CDP calls it writes into. */
+function recordCdp(): Sent[] {
+  const sent: Sent[] = []
+  setViewportChannel({
+    sendCdp: (method, params) => sent.push({ method, params }),
+    reportDesktopWidth: () => {},
+  })
+  return sent
+}
+
+/** Register a channel and return the modes reported to the viewer. */
+function recordModes(): boolean[] {
+  const modes: boolean[] = []
+  setViewportChannel({ sendCdp: () => {}, reportDesktopWidth: (v) => modes.push(v) })
+  return modes
+}
+
+/** Register a browser_input pending without awaiting it — the promise only
+ *  settles when the user answers, so awaiting here would hang the test. */
+function openHandoff(id: string, sessionId?: string): void {
+  void inputManager
+    .createPendingWithType(id, 'browser_input', { message: 'Log in.' }, sessionId)
+    .catch(() => {})
+}
+
+describe('browser viewport', () => {
+  beforeAll(() => {
+    startBrowserViewportSync()
+  })
+
+  beforeEach(() => {
+    // Drop the sender first so the reset sync below emits nothing, then let the
+    // no-pending sync clear any escape flag through the same path production uses.
+    setViewportChannel(null)
+    inputManager.rejectByType('browser_input', 'test reset')
+    inputManager.rejectByType('secret', 'test reset')
+    syncBrowserViewport()
+  })
+
+  // Literals, not the production constants: importing those would make this
+  // assertion follow any change to them instead of catching it.
+  it('narrows while a browser_input handoff is pending', () => {
+    const sent = recordCdp()
+    openHandoff('t-1')
+
+    expect(sent.at(-1)).toEqual({
+      method: 'Emulation.setDeviceMetricsOverride',
+      params: { width: 450, height: 900, deviceScaleFactor: 1, mobile: false },
+    })
+  })
+
+  it('clears when nothing is pending', () => {
+    const sent = recordCdp()
+
+    syncBrowserViewport()
+
+    expect(sent).toEqual([
+      { method: 'Emulation.clearDeviceMetricsOverride', params: undefined },
+    ])
+  })
+
+  // The tray reads only the latest request, but a second agent can open its own
+  // handoff. Resolving one must not un-narrow the page the user is still typing into.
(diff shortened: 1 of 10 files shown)


JeremyBischoff:
Closing. #1001 made full screen solve this.

The tray streams a 1280px page into a 450px drawer, so a sign-in form still arrives at about a third of native scale. But the expand button in the controls pill now takes the card to roughly 1:1. Done and Decline sit inside the card, so they survive full screen. The user can read the form and finish the handoff, one click into a canvas the handoff overlay already points them at.

This PR buys that click back, plus a phone layout instead of a scaled desktop one. It costs about 400 lines, a CDP viewport override, two new stream messages, and container-side viewport state with its own subscription.

It also collides with #1001. With the override active, frames arrive at 450x900, so the aspect ratio drops to 0.5 and the full-screen card measures to roughly 400px. Entering full screen during a handoff would give a narrow strip centered in an empty rail.

Rebased onto main and green before closing, so the branch is mergeable if this is worth revisiting. If the zero-click default is wanted later, the cheap version auto-enters full screen when a handoff opens, driving the `isExpanded` state `tray-manager` already owns.

PR #638, opened 2026-08-01
Fix host-browser profile disk growth (SUP-527)

iddogino:
Browser automation profiles under `host-browser-profiles/` grew to 10 GB on a dev machine (expected steady state ~500 MB). Four stacked causes, fixed as follows — see SUP-527 for the full forensics.

## Changes

**Prevention (Chrome launch flags, `chrome-provider.ts`)**
- Disable the optimization-guide features (`OptimizationGuideOnDeviceModel`, `OptimizationGuideModelDownloading`, `OptimizationHints`, `TextSafetyClassifier`) so Chrome stops downloading the 4 GB Gemini Nano model into automation profiles. Merged into the *existing* `--disable-features` flag — Chrome only honors the last occurrence, so a second flag would silently re-enable the occlusion-throttling workaround. A test pins the single-occurrence invariant.
- `--disable-component-update` stops browser-level components (`optimization_guide_model_store`, `component_crx_cache`, `WasmTtsEngine`, `Safe Browsing` — ~130 MB) from being re-downloaded into every per-agent user-data-dir.

**Reclamation (`profile-maintenance.ts`, new)**
- Sweep scheduled 3 minutes after startup (off the startup burst): removes the legacy singular `host-browser-profile` dir, deletes profile dirs whose agent no longer exists, strips regenerable caches from surviving profiles. `Default/` session state (cookies, logins, local storage) is never touched.
- Profiles whose browser is running when the sweep fires are skipped; launches await an in-flight sweep, so sweep and Chrome never touch a profile concurrently. Pending timer is unref'd + cancelled on shutdown.
- After sweeping, the directory is measured; >2 GiB reports to Sentry so GB-scale growth can't recur silently.

**GC on agent delete (`agents.ts`)**
- `DELETE /api/agents/:id` now removes the agent's profile dir (best-effort — the sweep is the safety net).

## Why not the ticket's proposed single user-data-dir + `--profile-directory`?

Chrome enforces one running process per user-data-dir (singleton lock), and we run concurrent per-agent Chrome instances, each with its own CDP port, headless setting, and download prefs. A shared dir would serialize all agents onto one process and break CDP target isolation. `--disable-component-update` removes the duplication vector with the same disk outcome and no architecture change.

## Validation
- 41 host-browser tests pass (15 new: real-filesystem sweep tests incl. session-state preservation, traversal/root-deletion guards, sparse-file telemetry threshold, in-use skip, delay/cancellation; launch-flag pinning via a real `launch()`)
- 310 agent-route tests pass (delete path)
- Typecheck + lint clean

Not live-validated: no real-Chrome launch or sweep against an actual 10 GB directory.

https://claude.ai/code/session_01FQsFpTFRiz9JafLHprwWwR

### DIFF
--- a/src/api/routes/agents.ts
+++ b/src/api/routes/agents.ts
@@ -161,6 +161,7 @@ import { AGENT_PACKAGE_EXTENSION, SKILL_PACKAGE_EXTENSION } from '@shared/lib/ut
 import { readAgentPreferences, updateAgentPreferences } from '@shared/lib/services/agent-preferences-service'
 import { agentPreferencesUpdateSchema } from '@shared/lib/types/agent-preferences'
 import { cleanupAgentData } from '@shared/lib/services/agent-cleanup-service'
+import { stopInstanceOnAllProviders } from '../../main/host-browser'
 import { deleteBrowserProfile } from '../../main/host-browser/profile-maintenance'
 import { logAuditEvent, logAuditEventOrThrow } from '@shared/lib/services/audit-log-service'
 import { loadSessionUsageTotals } from '@shared/lib/services/usage-service'
@@ -1076,9 +1077,13 @@ agents.delete('/:id', ResolveAgent(), AgentAdmin(), async (c) => {
     }
 
     // Remove the agent's dedicated host-browser Chrome profile. The container
-    // (and with it any host browser) is already stopped by deleteAgent.
-    // Best-effort: a leftover dir is reclaimed by the next startup sweep.
+    // teardown above only stops the browser on the ACTIVE provider — a Chrome
+    // launched before the user switched providers survives it — so stop this
+    // agent's browser on every provider first. deleteBrowserProfile itself
+    // refuses profiles claimed by an in-flight launch. Best-effort throughout:
+    // a leftover dir is reclaimed by a later startup sweep.
     try {
+      await stopInstanceOnAllProviders(slug)
       await deleteBrowserProfile(slug)
     } catch (error) {
       console.error('Failed to delete host-browser profile:', error)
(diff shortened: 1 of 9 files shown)


iddogino:
Both review findings fixed in 0e612f07:

**P1 (stale agent snapshot)** — `startBrowserProfileCleanup` now takes a supplier (`() => Promise<string[]>`) instead of an array; `startup.ts` passes `async () => (await listAgents()).map(a => a.slug)`, resolved when the timer fires. A supplier failure is captured and aborts the sweep — it is never treated as "no agents exist". New tests: agent created during the delay survives the sweep; failing supplier leaves everything intact.

**P2 (launch/sweep race)** — replaced the `isRunning()` callback with an in-use claim owned by `profile-maintenance`: `launch()` calls `markProfileInUse(instanceId)` *before* awaiting the in-flight sweep, so either the sweep sees the claim and skips the profile, or the launch sees the sweep's promise and waits. The claim is released in `stop()`, `handleExit`, and the four launch-failure paths that throw without reaching `stop()` (open-dispatch, pid-lookup, spawn-no-pid, proxy-bind). A claim leaked by a crash only makes the sweep skip that profile — the safe direction. New tests pin mark → wait → spawn ordering and release-on-stop; the maintenance-side test verifies a marked profile survives both orphan deletion and cache stripping until unmarked.

45 host-browser tests + 308 agent-route tests green; typecheck + lint clean.

iddogino:
Both residual P2s fixed in 4dcc4ab5:

**Claim release centralized** — `launch()` now marks the profile, then delegates the entire claimed phase to a new private `launchClaimed()` inside a single try/catch that releases the claim on any throw and rethrows. The four enumerated unmark sites are gone. This covers every pre-registration failure — `findFreePort`, `mkdirSync`, profile copy, preference reads/writes — including the disk-pressure case where leaking the claim would make the sweep skip exactly the profile that filled the disk. New test: an ENOSPC `mkdirSync` failure (before spawn/registration) releases the claim.

**Deletion coordinates with the claim, and stop is provider-independent** — new `stopInstanceOnAllProviders()` in `host-browser/index.ts` stops the instance's browser on every provider (not just `getActiveProvider()`), and the DELETE route calls it before `deleteBrowserProfile`. `deleteBrowserProfile` itself now refuses profiles claimed by an in-flight launch — a spawning Chrome isn't registered anywhere, so no stop call can reach it; the claim is the only signal it exists. The route's best-effort catch logs the refusal, and a later startup sweep reclaims the dir once the agent is gone and the claim released. New tests: provider-independent stop (chrome running while browserbase is the selected provider), and claimed-profile deletion refusal + success after release.

49 host-browser tests (4 new) + 308 agent-route tests green; typecheck + lint clean.

PR #652, opened 2026-08-03
fix: longer Better Auth sessions + crash reporting default on

yiw190:
## Summary
- Raise Better Auth defaults to **168h (7d)** max lifetime and **1440 min (24h)** updateAge
- Coalesce `shareErrorReports` to `true` when omitted (matches `shareAnalytics`)
- Align auth UI / enforcement fallbacks with the new defaults

Follow-up to Iddo's Phase 2 signup feedback: cloud onboarding skips PrivacyStep, so new workspaces should keep analytics/crash reporting on and use longer sessions.

## Test plan
- [x] `pnpm exec vitest run src/shared/lib/config/settings.test.ts src/api/middleware/auth-enforcement.test.ts`
- [ ] Fresh local settings.json omit path: Auth tab shows 168 / 1440
- [ ] With companion gamut-infra seed PR: new ECS org settings.json has the same values


yiw190:
Closing in favor of cloud-only defaults via the controller seed: https://github.com/SkillfulAgents/gamut-infra/pull/137

New ECS workspaces get analytics/crash reporting + longer sessions from first-boot settings.json without changing SuperAgent global defaults.

PR #663, opened 2026-08-04
Show Exa source favicons and metadata on web tool cards (SUP-539)

JeremyBischoff:
Closes https://linear.app/datawizz/issue/SUP-539/show-source-favicons-and-metadata-on-web-searchfetch-chat-cards

**Footprint:** 15 production files / +649 net · 7 test files / +502 net

A web search in chat rendered as blue links over a wall of text. A fetch rendered as a raw dump. Neither told you which sites an answer came from, or how old they were. The search vendor already returns each page's declared icon and publish date. We were dropping both.

## How a source row gets its icon and date

```mermaid
flowchart TB
  V["search vendor result<br/>icon · date · title · url · text"] --> M["provider mapping"]
  M --> T["tool-result text<br/>icon and date on the machine-readable line, kept out of<br/>the numbered hits the model reasons over"]
  N["built-in web tool<br/>title and url, usually nothing else"] --> T
  T --> P["renderer reads the text back"]
  P --> C{"icon supplied?"}
  C -->|"https, public host"| I["site icon"]
  C -->|"absent or refused"| G["globe"]
```

Both tool families land on the same cards. The built-in one supplies no icon, so it shows a globe and no date. Nothing is invented to fill the gap.

Guessing an icon from the domain root was the alternative. It misses a large share of real sites, and every result we can show an icon for already carries one. No third-party icon service.

## What you see

| | before | after |
|---|---|---|
| search, collapsed | tool name only | overlapping stack of site tiles, `+N` chip |
| search, expanded | list of blue links, then a text wall | one row per source: icon, linked title, domain, age, snippet |
| fetch, collapsed | tool name only | the page's icon |
| fetch, expanded | raw dump | icon, title, domain, age, then the page as formatted text |

The collapsed stack is the connections stack from agent home, extracted into a shared `TileStack` and rendered at 18px. Agent home consumes the same component at its existing sizes, output unchanged.

The fetch body is height-capped and scrolls. That inner scroller gets the same opt-in thumb the read-only message scroller already uses, so it does not fall back to the pale full-height platform scrollbar sitting over the text.

## Where the card decides to trust what it read

The search result text is written by the pages it came from, so the card re-anchors before it renders anything rich.

```
for each source the machine-readable line named
  is there a numbered block whose next line is that source's url?
    no  → render the plain link list, no dates, no snippets
  does the block count match the raw entry count?
    no  → render the plain link list
  is block N's header exactly the header we would write for source N?
    no  → render the plain link list
  yes to all → rich rows
```

Any single failure degrades the whole result to the old rendering. A page cannot plant text under another site's name, icon and link by emitting lines that look like a neighbouring block. The count check runs against the raw entry count, so an entry the schema drops degrades the card instead of letting the dropped hit's block shift into a neighbour's row.

The machine-readable line is parsed through a Zod schema, lenient per entry: a malformed entry drops alone, a malformed optional field drops without taking its entry.

The publish date rides that line too. Position alone could not tell the formatter's own date line from page text that opens with one, so a hit with no vendor date could show a date the page invented. The field is always written (empty when the vendor gave none), its presence marks the format, and the renderer only trusts the positional line on transcripts from before the field existed. As the trust marker it fails closed: a malformed value refuses the positional line rather than falling back to trusting it.

The fetch header is positional: title, url, then labelled metadata. The page writes its own title and declares its own icon url, so a newline in any of them used to shift that window and let the page plant a metadata line the card attributes to the fetch. The formatter now flattens every header field to one line, and the search list's date line gets the same treatment.

Page text is also rendered without images. Markdown image syntax becomes its alt text, so a page cannot fire an outbound request on card render. This closes a pre-existing hole: the old plain-markdown path rendered page-authored images, and those requests did fire.

## Where the icon request is allowed to go

The icon url is page-declared, so a hostile page can name anything, including an address on the reader's network.

```
not https                    → globe
literal loopback or LAN host → globe
image fails to load          → globe
otherwise                    → load it
```

A load failure is recorded against the url that failed, so a row that later receives a different icon url gets a fresh try. The collapsed fetch row renders nothing instead of a globe, because that row already shows one as its tool icon.

That host list is a hand-copy of the shared one, copied because the original pulls Node's DNS resolver and cannot enter the browser bundle. A test now compares the two over a table of hosts, so loosening either one fails the build.

Writing that test surfaced a gap in the shared list. A trailing dot is the root label, so `localhost.` is the same machine to any resolver, but both loopback checks read it as an ordinary public domain. One of them routes the cheap per-request re-check the cloud proxy runs against the deployment url it reads back from settings, which sent it to the plain-https branch:

| | shipped, before | shipped, after | dev |
|---|---|---|---|
| `https://localhost/` | blocked | blocked | allowed |
| `https://localhost./` | **allowed** | blocked | allowed |
| `https://foo.localhost./` | **allowed** | blocked | allowed |
| `https://127.0.0.1./` | blocked | blocked | allowed |
| `https://real.example./` | allowed | allowed | allowed |

The resolver-backed gate at discovery already rejected those two, so this closes a second layer rather than a live path. Public names are unaffected in both directions.

## What it can't do

- Transcripts written before this have no icon on the wire. They show globes, permanently.
- A running agent container keeps the old formatter until its image is rebuilt, so existing sessions emit no icon or date.
- The host check reads literal addresses. A public name that resolves to a private address still loads. Accepted: nothing can read the response back.
- The built-in fetch tool answers in prose with no structure to parse, so it shows plain input and output rather than a page card.
- A source the vendor has no icon for shows a globe. We do not guess one.

Typecheck and lint clean. Full unit suite passing (8899 + 808 container). The end-to-end mock now emits the real formatter text, so the rich card path renders in the E2E app too. Smoke-tested the search and fetch cards against the running app, light and dark.

### DIFF
--- a/agent-container/src/tools/web/format-results.ts
+++ b/agent-container/src/tools/web/format-results.ts
@@ -20,10 +20,14 @@ export interface WebSearchHostResult {
 
 export function formatWebSearchResults(data: WebSearchHostResult): string {
   // favicon rides on the Links line, not the numbered list: the renderer needs it and the model
-  // does not, so it stays out of the prose the model reasons over.
+  // does not, so it stays out of the prose the model reasons over. published is always present -
+  // the key's existence is what lets the renderer refuse a page-planted `Published:` line (a
+  // snippet can open with one), so "vendor gave no date" must stay distinguishable from "old
+  // formatter" and is written as ''.
   const links = data.hits.map((h) => ({
     title: h.title ?? h.url,
     url: h.url,
+    published: h.publishedDate ? oneLine(h.publishedDate) : '',
     ...(h.favicon ? { favicon: h.favicon } : {}),
   }))
   const lines: string[] = [`Links: ${JSON.stringify(links)}`, '']
@@ -34,7 +38,8 @@ export function formatWebSearchResults(data: WebSearchHostResult): string {
     data.hits.forEach((h, i) => {
       lines.push(`${i + 1}. ${h.title ?? h.url}`)
       lines.push(`   ${h.url}`)
-      if (h.publishedDate) lines.push(`   Published: ${h.publishedDate}`)
+      // Same one-field-one-line rule as the fetch header: the date is vendor-returned text.
+      if (h.publishedDate) lines.push(`   Published: ${oneLine(h.publishedDate)}`)
       if (h.snippet) lines.push(`   ${h.snippet}`)
       lines.push('')
     })
@@ -69,15 +74,16 @@ const oneLine = (value: string) => value.replace(/[\r\n]+/g, ' ')
  * (plus publish date when present) followed by the page's full content. Pure (no SDK / network).
  *
  * The header is positional - the reader takes line 0 as the title, line 1 as the url, and any
- * labelled lines after those as metadata. The page writes its own <title>, so a newline in it
- * would shift that window and let the page plant a `Favicon:`/`Published:` line the reader
- * attributes to the fetch. Flattening the two header fields keeps one field on one line.
+ * labelled lines after those as metadata. The page writes its own <title> and declares its own
+ * favicon URL, so a newline in either would shift that window and let the page plant a
+ * `Favicon:`/`Published:` line the reader attributes to the fetch. Flattening every header
+ * field keeps one field on one line.
  */
 export function formatWebFetchResult(data: WebFetchHostResult): string {
   const { result } = data
   const lines: string[] = [oneLine(result.title ?? result.url), oneLine(result.url)]
-  if (result.publishedDate) lines.push(`Published: ${result.publishedDate}`)
-  if (result.favicon) lines.push(`Favicon: ${result.favicon}`)
+  if (result.publishedDate) lines.push(`Published: ${oneLine(result.publishedDate)}`)
+  if (result.favicon) lines.push(`Favicon: ${oneLine(result.favicon)}`)
   lines.push('', result.content || '(no content returned)')
 
   if (data.warnings && data.warnings.length > 0) {
--- a/agent-container/src/tools/web/format-results.ts
+++ b/agent-container/src/tools/web/format-results.ts
@@ -10,6 +10,7 @@ export interface WebSearchHostHit {
   title: string | null
   snippet: string
   publishedDate?: string
+  favicon?: string
 }
 
 export interface WebSearchHostResult {
@@ -18,7 +19,13 @@ export interface WebSearchHostResult {
 }
 
 export function formatWebSearchResults(data: WebSearchHostResult): string {
-  const links = data.hits.map((h) => ({ title: h.title ?? h.url, url: h.url }))
+  // favicon rides on the Links line, not the numbered list: the renderer needs it and the model
+  // does not, so it stays out of the prose the model reasons over.
+  const links = data.hits.map((h) => ({
+    title: h.title ?? h.url,
+    url: h.url,
+    ...(h.favicon ? { favicon: h.favicon } : {}),
+  }))
   const lines: string[] = [`Links: ${JSON.stringify(links)}`, '']
 
   if (data.hits.length === 0) {
@@ -45,6 +52,7 @@ export interface WebFetchHostDoc {
   title: string | null
   content: string
   publishedDate?: string
+  favicon?: string
   fetchedAt: string
 }
 
@@ -61,6 +69,7 @@ export function formatWebFetchResult(data: WebFetchHostResult): string {
   const { result } = data
   const lines: string[] = [result.title ?? result.url, result.url]
   if (result.publishedDate) lines.push(`Published: ${result.publishedDate}`)
+  if (result.favicon) lines.push(`Favicon: ${result.favicon}`)
   lines.push('', result.content || '(no content returned)')
 
   if (data.warnings && data.warnings.length > 0) {
--- a/agent-container/src/tools/web/format-results.ts
+++ b/agent-container/src/tools/web/format-results.ts
@@ -61,13 +61,21 @@ export interface WebFetchHostResult {
   warnings?: string[]
 }
 
+/** The header's contract is one field per line, so a field can't be allowed to carry a newline. */
+const oneLine = (value: string) => value.replace(/[\r\n]+/g, ' ')
+
 /**
  * Format the host /web-fetch/fetch response into the text block the agent reads: a title/url header
  * (plus publish date when present) followed by the page's full content. Pure (no SDK / network).
+ *
+ * The header is positional - the reader takes line 0 as the title, line 1 as the url, and any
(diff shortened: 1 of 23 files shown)


agent-container/src/tools/web/format-results.ts
```diff
@@ -53,14 +61,23 @@ export interface WebFetchHostResult {
...
+ * labelled lines after those as metadata. The page writes its own <title>, so a newline in it
+ * would shift that window and let the page plant a `Favicon:`/`Published:` line the reader
+ * attributes to the fetch. Flattening the two header fields keeps one field on one line.
  */
 export function formatWebFetchResult(data: WebFetchHostResult): string {
   const { result } = data
-  const lines: string[] = [result.title ?? result.url, result.url]
+  const lines: string[] = [oneLine(result.title ?? result.url), oneLine(result.url)]
   if (result.publishedDate) lines.push(`Published: ${result.publishedDate}`)
+  if (result.favicon) lines.push(`Favicon: ${result.favicon}`)
```
iddogino: The header flattening stops one line short: `oneLine()` covers title and url, but `publishedDate` and `favicon` are pushed raw, and both are plain `z.string()` from the Exa schema — with the favicon URL being page-declared. Probed it locally:

```
favicon: 'https://evil.com/i.png\nPublished: 1999-01-01'
→ ["T", "https://evil.com/p", "Favicon: https://evil.com/i.png", "Published: 1999-01-01", "", "body"]
```

The parser attributes the spoofed `Published:` line to the fetch — the same header-shift class this PR closes for title/url. A newline could also prepend attacker-positioned body text ahead of the real content. Exposure is low (raw newlines in a vendor-returned URL are unusual), but the fix is the one-liner already applied two lines up: wrap all four header fields in `oneLine()`. The existing "flattens the header fields" test extends naturally.

src/renderer/components/ui/site-favicon.tsx
```diff
@@ -0,0 +1,116 @@
...
+      className={cn(
+        'inline-flex shrink-0 items-center justify-center overflow-hidden rounded-sm bg-white ring-1 ring-black/10',
+        className,
+      )}
+    >
+      <img
+        src={href}
+        alt=""
+        className="h-full w-full object-contain"
+        onError={() => setFailed(true)}
```
iddogino: Nit: `failed` is per-instance state and never resets when `src` changes — if a mounted instance receives a new `src`, a previous load failure sticks and the globe shows for a valid icon. Today's call sites key by URL or have immutable results, so this is latent, but `key={href}` on the `` makes it correct by construction.

src/renderer/components/messages/tool-renderers/web-fetch.tsx
```diff
@@ -1,10 +1,99 @@
...
+        <SourceMeta host={hostnameOf(headerUrl)} publishedDate={parsed.publishedDate} />
+      </div>
+      {/* pr-3 keeps the prose out from under the thumb, which overlays the content edge. */}
+      <div className="prose prose-sm max-w-none dark:prose-invert text-xs prose-p:text-xs prose-li:text-xs prose-headings:text-xs max-h-64 overflow-y-auto card-scrollbar pr-3">
+        <ReactMarkdown
+          remarkPlugins={[remarkGfm]}
+          urlTransform={markdownUrlTransform}
+          components={NO_MARKDOWN_IMAGES}
+        >
+          {stripLeadingTitle(parsed.body, parsed.title)}
```
iddogino: Nit: this runs un-memoized on every re-render of the expanded card — an O(n) split/join over a body that can be ~50k chars, and it returns a fresh string each time, guaranteeing ReactMarkdown re-parses. Worth folding into the `useMemo` that already holds `parsed`. Same shape in `web-search.tsx`, where `snippetOf(s)` is computed twice per source per render (guard + render) — compute once per source. Both are dwarfed by ReactMarkdown itself, but they're free wins.

src/renderer/components/messages/tool-renderers/web-result-parse.ts
```diff
@@ -0,0 +1,203 @@
...
+    const lines = result.split('\n')
+    const body = lines[0]?.startsWith('Web search results for query:')
+      ? lines.slice(1).join('\n')
+      : result
+    return { sources: [], leftover: body.trim() }
+  }
+
+  let links: unknown[] = []
+  try {
+    const parsed = JSON.parse(linksMatch[1])
```
iddogino, line 104: Convention question: CLAUDE.md says to use Zod `.parse()` when parsing JSON read back from file, and the Links line is JSON read out of a transcript. The hand-rolled `typeof` guards here are small and correct, and this boundary wants leniency rather than throw-on-mismatch — but a `z.array(...).catch([])`-style schema in a sibling file would match the house rule while keeping the degrade behavior. Your call whether this file counts as that boundary.

iddogino (commented):
Checked out the branch and went through it in depth: read every changed file plus the surrounding formatter/renderer plumbing, and verified locally — `tsc --noEmit` clean, ESLint clean on all touched files in both packages, all touched test files green (139 renderer/shared/provider tests + 12 agent-container formatter tests).

The anchoring design is sound — I tried to forge a block past it and couldn't — and `NO_MARKDOWN_IMAGES` quietly fixes a pre-existing hole (the old leftover-markdown path rendered page-authored images, which did fire outbound requests; worth a line in the description as a fixed bug). The `localhost.` root-label fix is real and has regression tests in both directions.

Four comments inline: one small security hardening in the fetch formatter (probed and confirmed), two nits, and one house-convention question. Happy to merge once the formatter one is addressed.

agent-container/src/tools/web/format-results.ts
JeremyBischoff: @iddogino Done in `7b628be0` - all four header fields go through oneLine() now, plus the search formatter's Published line, which is the same vendor text one function up.

src/renderer/components/ui/site-favicon.tsx
JeremyBischoff: @iddogino Done in `4b762bef` - keyed the failure to the href instead of a key on the img: once failed is set the img is not rendered at all, so remounting it cannot clear the state.

src/renderer/components/messages/tool-renderers/web-fetch.tsx
JeremyBischoff: @iddogino Done in `99f2fc45` for the fetch body and `e11618ad` for the search snippets - both derive once inside the parse memo.

src/renderer/components/messages/tool-renderers/web-result-parse.ts
JeremyBischoff: @iddogino Done in `7b628be0` - schema in a sibling file, lenient per entry so one malformed link degrades alone rather than dropping the batch.

PR #674, opened 2026-08-06
Replicate media generation with BYOK (SUP-554)

JeremyBischoff:
Closes https://linear.app/datawizz/issue/SUP-554/replicate-media-models-p1-byok

647 production lines across 23 files, plus 790 test lines across 9. Twelve of the production files are one-to-three-line settings wiring. The container-facing route and the whitelist are the parts worth reading.

## What agents can now do

Users paste a Replicate API key into Settings, Media Generation. Agents can then run approved media models and deliver the results as files.

```
container (agent territory)                 host                              vendor
  skill: media-generation ─── curl ──▶ /api/replicate/*  ──────────▶ api.replicate.com
  auth: $PROXY_TOKEN                     agent token check
                                         path/method table (6 ops)
                                         key present?
                                         model on the list?
                                         version belongs to model? (asks vendor)
                                         strip webhook, force Cancel-After, swap key
                                         relay the vendor's bytes back
  downloads outputs ──────────────────────────────────────────▶ replicate.delivery
```

The key stays host-side. The container never holds it, and a compromised or injected agent has no path to read it: it authenticates with the proxy token it already has, and the host swaps in the real key on the last hop. This follows `src/api/routes/web-search.ts`.

Enforcement is deny-by-default over six vendor operations. Run creation checks the model against a pinned list. Community-path creates must carry a qualified `owner/name:hash` version, and the route asks Replicate whether that hash belongs to that model before forwarding. Webhook fields are stripped and `Cancel-After: 600` is forced on both create forms. None of this lives in prompt text: an agent that ignores the skill entirely still cannot run an unapproved model or start an uncapped run.

## The model list is pinned, not generated from collections

26 models, named one at a time in `src/shared/lib/replicate/whitelist-source.ts`. The generator now only resolves each model's `official` flag from the vendor, since that decides which create path an agent must use and a wrong guess breaks the model silently.

The design called for seeding eight vendor collections and expanding them at release. That is reversed here, because expanding a collection hands membership to the vendor. Replicate files voice cloning under `text-to-speech` and frontier chat models under `image-to-text`, so a regeneration admitted 21 voice-cloning models and 5 chat models against two exclusions the design had already signed off. Collection expansion made the selection rules true only until the next regeneration.

Three rules decide what goes on the list: no identity or voice impersonation, no general-purpose language models, and nothing the agent already does with its own model. That last one removed captioning and embeddings. The agent reads images with its own model, so routing image-and-text-to-text through Replicate buys nothing and adds a billing path the app cannot see. Three image-to-prompt models survive under image generation, because recovering a generation prompt from an image has no in-app equivalent.

## Why the catalog is an endpoint and not part of the skill

`GET /api/replicate/catalog` serves the list. Putting the model list in the skill instead would have been simpler, and it would break: the skill ships with the container image, the whitelist ships with the app, and the two release on different cadences. They drift, and the agent starts naming models the gate refuses. Serving the catalog from the route itself makes that impossible, since both the catalog and the gate read the same array.

## Skill discovery

The Agent SDK reads skills from `CLAUDE_CONFIG_DIR` (`/workspace/.claude`), not from `$HOME/.claude/skills` where the Dockerfile bakes them, so a listing-only skill was never discoverable. `SessionManager` now copies baked skills into the workspace at startup.

That destination is not private. `skillset-service.ts` lists it, `agents.ts` edits and deletes from it, and `agent-template-service.ts` publishes it. So the copy takes a named list rather than globbing the image directory, and it drops each skill's `templates/` payload. Without that filter, publishing an agent template would ship the dashboard react-vite scaffold into the skillset repo.

## Worth reviewing

Three judgment calls.

**The version check is the whole whitelist.** Without it an agent pairs an approved model's name with any version hash and runs whatever it likes, so `verifyVersion` is load-bearing rather than defensive. It costs one free vendor read per community create.

**Poll and cancel take any prediction id on the owner's account**, not only ones this route created. Ids are unguessable UUIDs and the design accepted this. Scoping to route-created ids is P2 hardening.

**The upstream timeout is 120 seconds**, chosen to sit above the 55-second sync wait the skill instructs and below unbounded. It also bounds the streamed relay, so a very slow vendor response is cut at two minutes.

## What this does not fix

**Existing containers gain nothing until their image is rebuilt and the container recreated.** The route is live host-side regardless, so an agent told about the capability still works. It just lacks the skill that would have told it.

**Nothing is metered.** Runs bill to the user's own Replicate account. The platform-paid tier and its metering are P2, and they arrive together.

**Rate limiting is the vendor's.** Replicate allows 600 creates per minute, dropping to 6 while an account holds under $5 credit. Under BYOK that ceiling is per-user, so it needs nothing here. It becomes a shared ceiling at the P2 platform tier, which is where it should be handled.

## Validation

- `npm run typecheck` clean, `npm run lint` 0 errors
- full unit suite: 8986 passed
- `npm --prefix agent-container test`: 830 passed
- `E2E_MOCK=true npx playwright test -c playwright.wizard.config.ts e2e/specs/replicate-api-key.spec.ts`: 2 passed
- Live run against the real Replicate API, driving the shipped route with only the proxy-token check stubbed: all six allowed operations, both create forms, whitelist refusals on each, version verification against real and foreign hashes, the six body gates, path and method denials, and the response relay. An image was generated, polled to `succeeded`, and downloaded. Verified locally, not committed.
- Vendor contract confirmed live: `Cancel-After` and `Prefer: wait` are accepted, the qualified version form resolves, and the version endpoint answers 200 for a matching hash and 404 for a foreign one.
- Manual pass on a real install: settings entry and validation, key removal, an agent generating and delivering an image, the keyless refusal naming Settings, and refusals for an off-list model and for voice cloning.


iddogino:
Pushed 6643164a: your new MediaTab hit the same unit-test failure mode that MobileTab did (#675) — `global-settings-page.test.tsx` mocks every tab individually and eagerly renders all sections, so any new tab that uses react-query crashes it with "No QueryClient set". Added the one-line `vi.mock('./media-tab', ...)` and verified the 4 tests pass locally.

https://claude.ai/code/session_01CoBWzVQXeYrakg35yWiGhL

PR #682, opened 2026-08-06
PWA push notifications (iOS declarative Web Push) + notification channel abstraction (SUP-347)

iddogino:
## What

Notification delivery is now a **channel registry** behind the existing `triggerNotification()` choke point: a canonical `NotificationEvent` (including a new `navigatePath` deep-link computed once) is dispatched fire-and-forget to every registered `NotificationChannel`, with per-channel error isolation. Adding a future platform (APNs for the native companion app, platform relay per SUP-341) is one new file + a registry entry.

- **`ClientBroadcastChannel`** — the existing SSE `os_notification` broadcast, byte-identical frame (pure refactor; renderer display gating untouched).
- **`WebPushChannel`** — Declarative Web Push (`web_push: 8030`, title/body/navigate only; WebKit, iOS 18.4+) for the installed PWA, **no service worker**. Type allowlist: `session_complete` + `session_waiting`. Per-owner settings gate, auth-mode agent-access gate, prune on 401/403/404/410.

## Data & API

- Migration 0035: `push_subscriptions` (endpoint-unique, per-device `origin` so `navigate` URLs resolve on the address the phone actually uses, `user_id` owner in auth mode) + single-row `push_vapid_keys` (stable keypair; minting fresh keys drops orphaned rows). Both wiped by factory reset.
- `GET /api/push/vapid-public-key`, `POST/DELETE /api/push/subscriptions` (Zod-validated, owner-scoped delete, 404 on no-op).

## Client (iOS, no SW)

- `window.pushManager` capability gate (WebKit-only ⇒ scopes the feature to iOS/Safari); "Push to this device" card in Settings → Notifications; install-to-Home-Screen nudge on iOS browser tabs (incl. iPadOS desktop-UA disambiguation).
- Launch-time subscription re-upsert replaces the missing `pushsubscriptionchange` event, with **VAPID key-affinity checks** in both subscribe and revalidation so a factory-reset/restored server can't strand a device (no `InvalidStateError` loop, no resurrecting undeliverable rows).
- VAPID key fetch runs concurrently with the permission prompt to stay inside WebKit's transient user activation.

## Validation

- 71 new/updated unit tests (channel policy + payload shape, prune matrix, subscription service owner scoping, VAPID lifecycle); full suite green except 4 `global-settings-page` failures pre-existing on main since #614.
- **Live-validated on a real iPhone** over `tailscale serve` HTTPS: subscribe → session finished on desktop host → declarative push accepted by Apple → banner on locked phone → tap deep-linked into the session.
- `/code-review high`: 10 verified findings, 9 fixed; 1 explicit v1 scope skip (viewed-session suppression for push — needs server-side presence; documented).

## Follow-ups

- Presence-aware push suppression (don't buzz the device actively viewing the session)
- Classic service-worker fallback for Chrome/Firefox/Android (payload is wire-compatible)
- Tier B: platform relay sending, converging with SUP-341

https://claude.ai/code/session_01JEFVfhuRrjaPcrc4mvAocw


iddogino:
Addressed all four review findings in d017655b:

- **[P1] SSRF via arbitrary endpoints** — `pushSubscribeRequestSchema` now enforces an https-only allowlist of vendor push-service hosts (`web.push.apple.com`/`*.push.apple.com` for the v1 iOS target, plus FCM/Mozilla/WNS so the future SW fallback needs no schema change). Suffix matching is on the parsed hostname, so path tricks and `web.push.apple.com.evil.example` spoofs are covered — tests included.
- **[P1] Unbounded fan-out** — per-owner cap of 10 subscriptions (409-free refresh of an existing endpoint; new endpoint at cap → 429), sends batched 8-at-a-time, 10s socket timeout per request. No rate-limit middleware exists in the app today; the cap bounds both table growth and per-notification fan-out on this authenticated route.
- **[P2] 4-week default TTL** — event-appropriate TTLs (`session_waiting` 10min, `session_complete` 1h) plus a per-session RFC 8030 topic, so an undelivered stale "Action Required" is *replaced* by the session's completion push rather than queueing behind it.
- **[P2] Stale auth-era preferences** — in local mode the `local` settings sentinel now governs every subscription, ignoring a `userId` retained from a previous auth-mode life of the DB.

88 notification-suite tests green (23 new), typecheck + lint clean.

Also: the earlier CI failures are accounted for — e2e passed on rerun (home-graph edge-drag flake), unit-tests fail on the 4 `global-settings-page` tests that have been red on main since #614 (missing QueryClientProvider in mobile-tab tests, untouched by this branch).

PR #683, opened 2026-08-06
fix: complete MCP OAuth against active Cloud toggle target (SUP-560)

yiw190:
## Summary
- When Cloud toggle is on, MCP OAuth pending state lives in the cloud deployment — complete the deep-link callback against the **active** API base URL (cloud proxy), not always `localhost` (SUP-560).
- Pass the desktop deep-link scheme (`superagent` / `superagent-dev`) from Electron through initiate → pending flow → handoff HTML so cloud-hosted initiate still returns to the correct desktop app.
- Extract `planMcpOAuthCallback` / `parseMcpOAuthCompletionResponse` for testable main-process routing.

## Bug
With Cloud toggle on, finishing MCP OAuth completed against the local API (or used the cloud host’s default `superagent://` scheme). Result: missing pending state and/or the packaged Gamut.app opening instead of Dev Electron (`superagent-dev://`).

## Repro
1. Desktop Electron with Cloud toggle on (org cloud workspace reachable).
2. Connect a remote MCP that requires OAuth.
3. Complete OAuth in the browser.
4. Before: callback completed against local API / opened prod app. After: completes against cloud proxy; handoff uses the client’s scheme.

## Test plan
- [x] Unit: `mcp-oauth-callback.test.ts`, remote-mcps initiate/callback protocol tests, oauth flow `desktopProtocol` assertions
- [ ] Local toggle off: MCP OAuth handoff URL is `superagent-dev://…` and Dev Electron receives the callback
- [ ] Cloud toggle on (host with this build): OAuth completes successfully; pending state found on cloud
- [ ] Malformed `protocol` body rejected; falls back to default scheme
- [ ] Packaged app still uses `superagent://`

### DIFF
--- a/src/api/routes/remote-mcps.ts
+++ b/src/api/routes/remote-mcps.ts
@@ -105,10 +105,20 @@ function renderMcpOAuthCallbackHtml(payload: McpOAuthCallbackPayload, message: s
  * already completed server-side. It hands the result back into the app via the
  * custom scheme so the main process can notify the renderer over IPC.
  */
+// Only SuperAgent desktop deep-link schemes — never arbitrary URI schemes from the client.
+const DESKTOP_OAUTH_PROTOCOLS = new Set(['superagent', 'superagent-dev'])
+
+function resolveDesktopOAuthProtocol(candidate?: string): string {
+  if (candidate && DESKTOP_OAUTH_PROTOCOLS.has(candidate)) return candidate
+  const fromEnv = process.env.SUPERAGENT_PROTOCOL
+  if (fromEnv && DESKTOP_OAUTH_PROTOCOLS.has(fromEnv)) return fromEnv
+  return 'superagent'
+}
+
 function renderMcpOAuthHandoffHtml(payload: McpOAuthCallbackPayload, desktopProtocol?: string): string {
   // Prefer the scheme recorded on the flow: a cloud deployment serving this
   // callback has no SUPERAGENT_PROTOCOL of its own (SUP-560).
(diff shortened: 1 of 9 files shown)


src/api/routes/remote-mcps.ts
```diff
@@ -287,7 +290,15 @@ remoteMcps.post('/initiate-oauth', async (c) => {
   // not getAppBaseUrlFromRequest: the packaged renderer is served from file://, so its
   // fetches carry `Origin: null`. The AS redirects the external browser here to complete
   // the flow, so the URL must be one the local API server actually answers on.
-  const protocol = process.env.SUPERAGENT_PROTOCOL || 'superagent'
+  //
+  // The scheme comes from the client when it sends one (validated against the
+  // RFC 3986 scheme grammar): a cloud deployment serving a proxied Electron
+  // client has no SUPERAGENT_PROTOCOL of its own (SUP-560).
+  const clientProtocol =
```
iddogino: this bypasses url safety. anything is accepted (`file`, `vscode`, etc...)

iddogino (changes requested):

yiw190:
Addressed the review comment on `body.protocol`: replaced RFC 3986 grammar validation with an allowlist of the two desktop schemes main already registers (`superagent`, `superagent-dev`). Same resolver is used for initiate-oauth and the handoff HTML so `file` / `vscode` / etc. cannot become the deep-link scheme.

src/api/routes/remote-mcps.ts
yiw190: good catch - addressed in new commit

PR #721, opened 2026-08-12
feat(models): add grok-4.6 to platform and OpenRouter catalogs

yiw190:
## Summary
- SuperAgent's grok family still resolved to `grok-4.5` / `x-ai/grok-4.5`, so the new xAI model could not be selected (and a `grok` alias would keep routing to 4.5).
- Add `grok-4.6` (platform) and `x-ai/grok-4.6` (OpenRouter) as the family's latest/default. `grok-4.5` stays in both catalogs as a pin.
- Bill both models with xAI's 200k long-context cliff ($2/$6 → $4/$12). grok-4.5 cache read is now $0.30 / $0.60; grok-4.6 is $0.50 / $1.00. Priority/fast remains 2x.

Depends on platform proxy pricing/allowlist: https://github.com/SkillfulAgents/platform/pull/276

## Test plan
- [x] `npx vitest run src/shared/lib/llm-provider/model-catalog.test.ts src/shared/lib/services/usage-service.test.ts src/api/routes/llm.test.ts src/shared/lib/container/subagent-model-catalog.test.ts src/renderer/components/wizard/configure-model-step.test.tsx src/renderer/components/messages/model-family-list.test.tsx`
- [ ] Platform provider: picker shows Grok 4.6 as the grok default; Grok 4.5 remains selectable
- [ ] Selecting the bare `grok` alias sends `grok-4.6` to the proxy
- [ ] OpenRouter provider: same for `x-ai/grok-4.6`
- [ ] Usage tab bills a >200k grok-4.6 prompt at $4/$12

### DIFF
--- a/src/api/routes/llm.test.ts
+++ b/src/api/routes/llm.test.ts
@@ -91,7 +91,7 @@ describe('LLM proxy endpoint', () => {
     })
 
     it.each([
-      ['platform', 'grok', 'grok-4.5'],
+      ['platform', 'grok', 'grok-4.6'],
       ['bedrock', 'sonnet', 'us.anthropic.claude-sonnet-5'],
       ['generic', 'custom/qwen3', 'custom/qwen3'],
     ])('returns the resolved %s catalog default', async (providerId, browserModel, resolvedModel) => {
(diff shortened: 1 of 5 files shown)


yiw190:
4.6 is on proxy for prod

PR #722, opened 2026-08-12
fix: release automation session streams once they settle

yiw190:
## Bug

Every scheduled-task tick (and webhook trigger) creates a session and subscribes a WebSocket to the runtime container, but nothing ever closes it: idle finalization (`finalizeIdle`) flips session state and broadcasts `session_idle` without touching the subscription, and the stream keepalive ping prevents any transport-level idle cut. Leaked streams accumulate one per automation run.

On runtimes that cap concurrent connections per container, an agent with a frequent cron eventually exhausts the cap and every new `POST /sessions` / `GET /health` is rejected with HTTP 429 (`Failed to create session: Too Many Requests`) until some sockets happen to die. On Docker the same leak just accumulates fds.

## Repro

1. Create an agent with a recurring scheduled task (e.g. every 10 minutes) on a connection-capped runtime.
2. Let it run until the number of completed runs reaches the runtime container's connection cap (host logs show one `WebSocket connection established` per run, zero closes).
3. The next tick fails: `Failed to create session: Too Many Requests`, and the agent's health checks 429 as well.

## Change

- `message-persister.ts`:
  - `doSubscribeToSession` now resolves (non-blocking) whether the subscription belongs to an unpromoted cron/webhook session from session metadata, cached on the streaming state. Both automation paths register metadata before subscribing, so the read can't miss them.
  - `finalizeIdle` synchronously unsubscribes such sessions once they truly settle. Synchronous on purpose: an async gap would race the wake path's `isSubscribed` check. Gated on `stateEventsAuthority` — only the runtime's own idle event proves the container's message queue is drained, so legacy result-driven idles keep their stream (pre-fix behavior).
  - `promoteAutomatedSession` clears the flag: promoted sessions behave interactive and keep one live stream.
- Interactive sessions are untouched. All resume paths (message send route, wake delivery) already re-subscribe on demand, so a released session costs one reconnect if it is ever resumed.
- Tests: settle-releases for scheduled and webhook sessions, keep-alive for interactive/promoted/legacy sessions, and clean re-subscribe + re-release across a wake cycle. Two replay-test mock factories gained the `getSessionMetadata` export the subscribe path now uses.

## Verification

- `src/shared/lib/container/` + `src/shared/lib/scheduler/` suites: 1077 passed.
- Full suite failures (db/migration/policy suites) reproduce identically without this change — environmental, unrelated.
- `eslint` clean on touched files; `tsc` errors limited to pre-existing `tools/verify-*` scripts.

### DIFF
--- a/src/shared/lib/container/base-container-client.ts
+++ b/src/shared/lib/container/base-container-client.ts
@@ -1241,12 +1241,7 @@ export abstract class BaseContainerClient extends EventEmitter implements Contai
   async deleteSession(sessionId: string): Promise<boolean> {
     const port = await this.getPortOrThrow()
 
-    // Close WebSocket if exists
-    const ws = this.wsConnections.get(sessionId)
-    if (ws) {
-      ws.close()
-      this.wsConnections.delete(sessionId)
-    }
+    this.closeTrackedWebSocket(sessionId)
 
     const response = await fetch(
       `${this.getBaseUrl(port)}/sessions/${sessionId}`,
@@ -1363,6 +1358,16 @@ export abstract class BaseContainerClient extends EventEmitter implements Contai
     return response.ok
   }
 
+  // Drop listeners before close so a deliberate teardown cannot fire
+  // connection_closed into a newer socket's map slot (SUP-572).
+  private closeTrackedWebSocket(sessionId: string): void {
+    const ws = this.wsConnections.get(sessionId)
+    if (!ws) return
+    ws.removeAllListeners()
+    ws.close()
+    this.wsConnections.delete(sessionId)
+  }
+
   subscribeToStream(
     sessionId: string,
     callback: (message: StreamMessage) => void
@@ -1377,9 +1382,8 @@ export abstract class BaseContainerClient extends EventEmitter implements Contai
     const setupWebSocket = async () => {
       const port = await this.getPortOrThrow()
 
-      const existing = this.wsConnections.get(sessionId)
-      if (existing) {
-        existing.close()
+      if (this.wsConnections.has(sessionId)) {
+        this.closeTrackedWebSocket(sessionId)
       }
 
       const ws = new WebSocket(
@@ -1452,11 +1456,7 @@ export abstract class BaseContainerClient extends EventEmitter implements Contai
     })
 
     const unsubscribe = () => {
-      const ws = this.wsConnections.get(sessionId)
-      if (ws) {
-        ws.close()
-        this.wsConnections.delete(sessionId)
-      }
+      this.closeTrackedWebSocket(sessionId)
     }
 
     return { unsubscribe, ready }
--- a/src/shared/lib/container/base-container-client.ts
+++ b/src/shared/lib/container/base-container-client.ts
@@ -1364,6 +1364,8 @@ export abstract class BaseContainerClient extends EventEmitter implements Contai
     const ws = this.wsConnections.get(sessionId)
     if (!ws) return
     ws.removeAllListeners()
+    // close() on CONNECTING emits 'error'; zero listeners is an uncaughtException.
+    ws.on('error', () => {})
     ws.close()
     this.wsConnections.delete(sessionId)
   }
--- a/src/shared/lib/container/message-persister.ts
+++ b/src/shared/lib/container/message-persister.ts
@@ -119,6 +119,10 @@ interface StreamingState {
   isCompacting: boolean // True while compaction is in progress, cleared on compact completion
   agentSlug?: string // The agent slug for this session
   notAutomationSession?: boolean // Cached "not a cron/webhook session" verdict; skips the automation-status metadata write on later results
+  // Unpromoted cron/webhook session: release its container stream when the
+  // session settles. Resolved from session metadata at subscribe time so the
+  // settle-time teardown in finalizeIdle stays synchronous (race-free).
+  releaseStreamOnSettle?: boolean
   // True when the most recent result was a clean success (not error-shaped,
   // not an interrupt, not a resume-exit). Consumed by finalizeIdle: a success
   // result alone is NOT terminal — queued messages or background work can keep
@@ -470,7 +474,12 @@ class MessagePersister {
       lastResultSubtype: null,
       lastResultCleanSuccess: false,
       isRetrying: false,
+      // Carried over so a transport reattach mid-run doesn't lose the verdict
+      // before the refresh below lands.
+      releaseStreamOnSettle: prior?.releaseStreamOnSettle ?? false,
     })
+
+    this.resolveReleaseStreamOnSettle(sessionId, agentSlug)
 
     // Store container client for reconnection checks
     this.containerClients.set(sessionId, client)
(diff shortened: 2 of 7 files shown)


src/shared/lib/container/message-persister.ts
```diff
@@ -535,6 +571,25 @@ class MessagePersister {
...
+    // the queue is drained — a legacy result-driven idle can still have a
+    // queued message about to start a turn on this stream.
+    if (
+      state.releaseStreamOnSettle &&
+      state.stateEventsAuthority &&
+      this.subscriptions.has(sessionId)
+    ) {
+      console.log(`[MessagePersister] Releasing settled automation stream for session ${sessionId}`)
+      this.unsubscribeFromSession(sessionId)
+    }
```
iddogino, line 593: **Failed runs never release — the original 429 symptom survives for erroring crons.**

An error-shaped result settles the session *before* the authoritative idle arrives: the result handler sets `state.isActive = false` directly (line ~2008), so when the runtime's `session_state_changed: 'idle'` lands, the guard at line ~1891 (`state.isActive && lastResultSubtype !== null`) skips `finalizeIdle` entirely and this release never runs. Verified empirically: a scheduled session settled with `is_error: true` + authoritative idle keeps `isSubscribed === true`.

This matters because the agents most likely to hit the connection cap are exactly the ones whose cron fails every tick (expired credentials, rate limits): one leaked socket per failed run, same as before the fix. (Interrupted runs leak too via the `isInterrupted` frame swallow, but that path is rare and usually followed by a send.)

Suggested direction: evaluate the release on the authoritative idle even when the turn already settled via error — in the `'idle'` branch, run this same gated check when `!state.isActive`, still requiring `lastResultSubtype !== null` so a bare/stale idle can't release. That stays safe against the wake race because `markSessionActive` nulls `lastResultSubtype` on every new send. Worth a companion test mirroring the success-settle one with `is_error: true`.

src/shared/lib/container/message-persister.ts
```diff
@@ -493,6 +502,33 @@ class MessagePersister {
...
+    const stateRef = this.streamingStates.get(sessionId)
+    void getSessionMetadata(agentSlug, sessionId)
+      .then((meta) => {
+        // A resubscribe replaced the state and kicked off its own resolution.
+        const current = this.streamingStates.get(sessionId)
+        if (!current || current !== stateRef) return
+        current.releaseStreamOnSettle = Boolean(
+          (meta?.isScheduledExecution || meta?.isWebhookExecution) &&
+            !meta?.promotedToInteractive
+        )
```
iddogino, line 527: **Nit — promotion can be overwritten by an in-flight resolution.** This `.then` can land *after* `promoteAutomatedSession` set `releaseStreamOnSettle = false`, having read metadata *before* the `promotedToInteractive` write — flipping the flag back to `true`. The window is ms-wide (subscribe-time read vs. a promotion) and the cost is only one reconnect on the promoted session's next message, so acceptable as-is — but a synchronous "promoted" marker on the streaming state that this resolver respects would make it airtight.

iddogino (changes requested):
The mechanism is sound and the design claims hold up under scrutiny — metadata registration precedes subscribe on both automation paths, the check-to-`markSessionActive` windows in all four resume paths are microtask-only so the synchronous release can't race a new turn, and the teardown side effects (session-scoped request drops, capability grants) are safe. Tests and lint pass locally.

Requesting changes for one real coverage gap (inline comment on the release block) plus two smaller items:

**2. Deliberate close still routes `connection_closed` back into the persister** (`base-container-client.ts` — outside this diff, so noting it here): `unsubscribe()` just calls `ws.close()`, and the `close` handler then fires `callback({type: 'connection_closed'})` and unconditionally deletes the session's `wsConnections` entry — even when the map now holds a newer socket. Today the release path is benign only because `unsubscribeFromSession` deletes `streamingStates` first, so `handleMessage` bails on the missing state. But release-then-quick-resume (a wake landing within the close handshake window) recreates state before the stale `close` event dispatches; `handleConnectionClosed` then sees `isRunning` and opens a second stream for the same session, and the old socket's close handler having deleted the new socket's map entry desyncs the bookkeeping — one stream ends up orphaned and every frame is processed twice. The shape is mostly pre-existing (any transport reattach can do this), but this PR adds the first unsubscribe-on-a-live-container path, so it's newly reachable. One-line hardening: remove the `close` listener (or set a deliberate-close flag) before `ws.close()` in `unsubscribe`. Fine as a fast-follow if you'd rather keep this PR tight.

**3. Promotion vs. resolver race** (inline comment on `resolveReleaseStreamOnSettle`) — nit, ms-wide window, but cheap to make airtight.

src/shared/lib/container/message-persister.ts
yiw190: this is fixed in the new PR as well.

src/shared/lib/container/message-persister.ts
yiw190: good catch! fixed in new commit

yiw190:
All good finds — addressed all 3 of them here rather than opening another PR.

src/shared/lib/container/base-container-client.ts
```diff
@@ -1363,6 +1358,16 @@ export abstract class BaseContainerClient extends EventEmitter implements Contai
     return response.ok
   }
 
+  // Drop listeners before close so a deliberate teardown cannot fire
+  // connection_closed into a newer socket's map slot (SUP-572).
+  private closeTrackedWebSocket(sessionId: string): void {
+    const ws = this.wsConnections.get(sessionId)
+    if (!ws) return
+    ws.removeAllListeners()
```
iddogino, line 1366: **Confirmed crash vector — `removeAllListeners()` also strips the `'error'` listener, and `ws` is a Node EventEmitter: an `'error'` emit with zero listeners throws as an uncaughtException** (which the fatal handler turns into an app quit).

Two real windows:

1. **Socket still CONNECTING:** `ws@8.21.0`'s `close()` on a CONNECTING socket calls `abortHandshake` → `websocket.emit('error', …)` ("WebSocket was closed before the connection was established"). Reproduced against the repo's actual `ws`: the exact `removeAllListeners(); close()` sequence dies with an uncaughtException. Reachable because the socket is tracked in `wsConnections` synchronously *before* `'open'` fires, and `deleteSession` / `unsubscribeFromSession` / an overlapping resubscribe (the reconnect in `handleConnectionClosed` never awaits `ready`) can all hit a not-yet-open socket.
2. **OPEN socket:** an ECONNRESET during the close handshake after listeners are removed also emits `'error'` unhandled — plausible in `deleteSession`, where the container side is being torn down at the same moment.

Pre-PR, the always-attached `'error'` handler swallowed both — and this path now runs on every automation tick teardown.

```suggestion
    ws.removeAllListeners()
    // Deliberate teardown: an 'error' with no listeners throws (uncaughtException),
    // e.g. ws emits one when close() lands on a still-CONNECTING socket.
    ws.on('error', () => {})
```

iddogino (changes requested):
The two prior findings are properly fixed (verified: error-settle release + the stale-metadata-vs-promotion pin, both with solid tests), and the close-listener commit is load-bearing — without it the release would synthesize `connection_closed` and `handleConnectionClosed` would auto-resubscribe, defeating the whole fix. All 308 tests in the touched suites pass, eslint/tsc clean.

Requesting changes for one new issue that commit 36528afe introduces: `closeTrackedWebSocket` can crash the process via an unhandled `'error'` event (inline comment with repro + one-line fix). With that fixed, this is good to merge.

PR #746, opened 2026-08-13
fix(composer): refresh transcript before idle rescue restores a prompt (SUP-427)

JeremyBischoff:
Closes https://linear.app/datawizz/issue/SUP-427/sent-message-text-sometimes-stays-in-the-composer-after-a-successful

Production: 1 file, +55 net (`message-list.tsx`).
Tests: 2 files, +290 net.

Type: bug fix

A successful send can put the text back in the composer and turn Send on.
The agent may already be working, or have already replied.

```
before
  you send
    → the server accepts
    → the composer clears
    → a local copy stays in memory
  the session goes idle
    → wait 1.5s
    → if that copy is still there, put the text back
    → do not read a fresh transcript

after
  you send
    → the server accepts
    → the composer clears
    → a local copy stays in memory
  the session goes idle
    → wait 1.5s
    → fetch a fresh transcript
         ├─ fetch failed → leave the copy, leave the composer empty
         ├─ fetch came back with an older cache → same
         ├─ the prompt is in the transcript → drop the copy, leave the composer empty
         └─ the prompt is still missing → put the text back so you can resend
```

A send still in flight with no server id is left alone.
Failed send restore and Stop restore are unchanged.

Other people's leftover copies drop on their own 1.5s timer.
Clearing them in the same wait cancelled the fresh read and restarted the clock.

## Worth reviewing

The 1.5s wait is still a beat, not proof the transcript has the prompt.
If the source of truth itself is empty at that fetch, we still restore.

A cancelled fetch can look like success with yesterday's list.
We refuse that the same way we refuse a failed fetch.

Peer copies sit on a second timer so they cannot cancel the local read.
One shared timer would leave their copies up until the fetch returns.

## Why not a longer wait

A longer wait still treats a stale list as loss.
The E2E that holds the transcript read for 4.5s can also pass if recovery is deleted.
The unit case (cache empty, fresh read has the id) is what pins this shape.

## What this does not fix

Lost HTTP acknowledgement.
The server accepted.
This window never sees that.
Composer restore runs as a failed send.
That needs request correlation.
Separate design.

A hung transcript fetch that never finishes can leave rescue unresolved.
The messages query does not abort.

## How to verify

1. From this worktree, run `gamut-test electron`.
2. Open `test-agent`. Type a short prompt. Send.
3. Wait 3 seconds.
4. Composer stays empty. Send stays off. One transcript copy. Mock reply can arrive.

Smoke screenshot (drop into the PR): `/tmp/stuck-prompt-electron-smoke.png`

## Validation

- `npx vitest run src/renderer/components/messages/message-list.test.tsx src/renderer/components/messages/message-input.test.tsx` - 160 passed
- `E2E_MOCK=true E2E_PORT=3017 PORT=3017 npx playwright test e2e/specs/composer-failure-recovery.spec.ts --project=web-chromium` - 4 passed
- `npm run typecheck` - pass
- eslint on the touched files - pass
- Electron mock smoke: three accepted short prompts, composer empty at idle

### DIFF
--- a/e2e/specs/composer-failure-recovery.spec.ts
+++ b/e2e/specs/composer-failure-recovery.spec.ts
@@ -166,4 +166,41 @@ test.describe('Composer failure recovery', () => {
     // ~1.5s and was still present at this point)
     await expect(sessionPage.getMessageInput()).toHaveText('')
   })
+
+  test('a delayed transcript refetch does not restore an accepted short prompt', async ({ page }) => {
+    // Regression for the idle-rescue race: POST and SSE complete normally, but
+    // the messages GET is held past the 1.5s grace. The accepted prompt must
+    // not be written back into the composer just because the cached transcript
+    // has not caught up yet.
+    const text = 'short accepted prompt'
+    await page.route('**/sessions/*/messages', async (route) => {
+      if (route.request().method() !== 'GET') return route.continue()
+      await new Promise((r) => setTimeout(r, 4500))
+      return route.continue()
+    })
+
+    await sessionPage.typeMessage(text)
+    const sentAt = Date.now()
+    await sessionPage.getSendButton().click()
+    await expect(sessionPage.getMessageInput()).toHaveText('')
+
+    // The live reply can arrive while the transcript GET is still held.
+    await expect(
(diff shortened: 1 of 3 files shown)


iddogino:
@JeremyBischoff working on followup

PR #747, opened 2026-08-13
fix(dashboard): replace node_modules mtime skip with install preflight

yiw190:
## Summary
- Fixes [SUP-578](https://linear.app/datawizz/issue/SUP-578/dashboard-manager-replace-node-modules-mtime-heuristic-with): `runBunInstallIfNeeded` treated `node_modules` mtime as integrity, so a gutted tree (missing direct dep or dangling `.bin`) still skipped install.
- `bun run start` then exited 127 → `handleCrash` restarted with `forceInstall: false` → skip again → 3× in 5 min → permanent `crashed`.
- Before start, preflight every direct `dependencies`/`devDependencies` package and its declared bins. On failure, repair with frozen Bun install (plain install if there is no lockfile), re-check, and only then launch.

## Repro (before)
1. Start a dashboard whose `node_modules` directory is newer than `package.json` but missing a declared `.bin` (e.g. `open-slide`).
2. Boot/crash-restart uses `forceInstall: false` and skips install.
3. `bun run start` exits 127; crash loop until the dashboard is permanently `crashed`.

## What changed
- New `dashboard-install-preflight` helper (zod at the `package.json` boundary). Does not use `bun pm ls` or `--dry-run`.
- Boot/crash path: healthy preflight skips install even if `package.json` is newer than `node_modules`. Unhealthy + lockfile → `bun install --frozen-lockfile` (no unfrozen fallback). Unhealthy + no lockfile → plain install.
- Frozen install non-zero or post-repair preflight failure → status `crashed`, no `bun run start`, no restart scheduled.
- `forceInstall: true` still always installs, then preflights before start.
- Independent of SUP-577 (MicroVM overlay).

## Test plan
- [ ] `npx vitest run src/dashboard-install-preflight.test.ts src/dashboard-manager.test.ts` (69 passed locally)
- [ ] Boot a dashboard with a healthy tree: install skipped, start runs
- [ ] Delete a direct dep or `.bin` shim, restart with `forceInstall: false`: frozen repair runs, then start
- [ ] Frozen install failure: dashboard stays `crashed`, no restart loop
- [ ] Agent-initiated start (`forceInstall: true`) still always installs


yiw190:
superseeded

PR #752, opened 2026-08-13
fix: pipe GET /messages to avoid N full transcript copies

yiw190:
## Summary
- Host-app OOM on `GET /messages`: overlapping list refetches each `readFile` + parse + `c.json()` the whole Claude JSONL (screenshots included). A 1 GB Fargate task already at 400–500 MB dies (exit 137).
- Response shape is unchanged: `transformMessages`, author annotation, interrupted-subagent resolution, and awaiting-input recovery still run.
- Two cuts on that path:
  1. `readJsonlFile` streams line-by-line instead of one `readFile` string.
  2. The transformed array is piped through `createJsonArrayStringifyTransform`, so the host does not also hold the full `JSON.stringify` response string.
- Peak per request is still the parsed + transformed array. Concurrent refetches still hold N of those. This PR drops the extra whole-file string and the extra response string.

## RSS (re-measured on this branch)

Incident file `/tmp/iddo-9b2c4d80.jsonl` (35.7 MB), includes `transformMessages`, separate processes, `--expose-gc`:

- Old path (`readFile` + parse + `JSON.stringify`) 3 concurrent: **91 → 556 MB** RSS
- This PR 3 concurrent: **91 → 278 MB** RSS
- This PR 10 concurrent: **570 MB** RSS (N transformed arrays still retained)

The 924 → 183 MB numbers from the first commit were for a raw JSONL pipe with no transform. They do not apply to HEAD.

## What this does not do
- Image base64 is still parsed and sent (~17 MB transformed body).
- Client refetch stacking is unchanged.
- Follow-up: page / stream the transform so the host does not accumulate the full array per request.

## Test plan
- [x] `npx vitest run src/shared/lib/utils/file-storage.test.ts src/api/routes/agents.test.ts`
- [ ] Open a session with a large illustrated transcript and confirm `GET /messages` returns 200 with `content.text` / `toolCalls`
- [ ] Confirm host RSS during a tool-result burst does not jump by ~file-size × in-flight × 3

### DIFF
--- a/src/api/routes/agents.ts
+++ b/src/api/routes/agents.ts
@@ -180,7 +180,7 @@ import { logAuditEvent, logAuditEventOrThrow } from '@shared/lib/services/audit-
 import { loadSessionUsageTotals } from '@shared/lib/services/usage-service'
 import { captureException } from '@shared/lib/error-reporting'
 import * as fs from 'fs'
-import { Readable } from 'stream'
+import { Readable, pipeline } from 'stream'
 import pLimit from 'p-limit'
 import * as path from 'path'
 import type { ApiAgent } from '@shared/lib/types/api'
@@ -1828,12 +1828,15 @@ agents.get('/:id/sessions/:sessionId/messages', AgentRead(), async (c) => {
     const source = Readable.from(transformed)
     const stringify = createJsonArrayStringifyTransform()
     const reportStreamError = (err: unknown) => {
+      const code = (err as NodeJS.ErrnoException)?.code
+      if (code === 'ABORT_ERR' || code === 'ERR_STREAM_PREMATURE_CLOSE') return
       console.error('Failed to stream messages:', err)
       captureException(err, { tags: { component: 'agents', operation: 'stream-messages' } })
     }
-    source.on('error', reportStreamError)
-    stringify.on('error', reportStreamError)
-    return c.body(Readable.toWeb(source.pipe(stringify)) as ReadableStream, 200, {
+    pipeline(source, stringify, (err) => {
+      if (err) reportStreamError(err)
+    })
+    return c.body(Readable.toWeb(stringify) as ReadableStream, 200, {
       'Content-Type': 'application/json',
     })
   } catch (error) {
--- a/src/api/routes/agents.ts
+++ b/src/api/routes/agents.ts
@@ -54,6 +54,7 @@ import {
   readSessionMetadata,
   updateSessionName,
   registerSession,
+  getSessionMessagesWithCompact,
   getSession,
   getSessionMetadata,
   sessionExists,
@@ -66,7 +67,7 @@ import {
   removeMessage,
   removeToolCall,
 } from '@shared/lib/services/session-service'
-import { getSessionJsonlPath, readFileOrNull, getAgentSessionsDir, readJsonlFile, writeJsonFileAtomic, displaySlug, createJsonlToJsonArrayTransform } from '@shared/lib/utils/file-storage'
+import { getSessionJsonlPath, readFileOrNull, getAgentSessionsDir, readJsonlFile, writeJsonFileAtomic, displaySlug, createJsonArrayStringifyTransform } from '@shared/lib/utils/file-storage'
 import {
   MAX_UPLOAD_TOTAL_SIZE,
   UploadTooLargeError,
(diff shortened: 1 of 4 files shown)


src/api/routes/agents.ts
```diff
@@ -1825,7 +1825,17 @@ agents.get('/:id/sessions/:sessionId/messages', AgentRead(), async (c) => {
       }
     }
 
-    return c.json(transformed)
+    const source = Readable.from(transformed)
+    const stringify = createJsonArrayStringifyTransform()
+    const reportStreamError = (err: unknown) => {
```
iddogino, line 1830: **Client aborts will hit Sentry on every mid-body disconnect** (confirmed with a live repro against `@hono/node-server`): when the client disconnects, node-server calls `reader.cancel(error)` on the web stream, which destroys the Node pipe and fires this handler with `code: 'ABORT_ERR'`. Since this endpoint streams multi-MB bodies and is refetched aggressively (session switches, SSE reconnect invalidation, navigating away), aborts are routine, not exceptional — this will flood Sentry.

```suggestion
    const reportStreamError = (err: unknown) => {
      const code = (err as NodeJS.ErrnoException)?.code
      if (code === 'ABORT_ERR' || code === 'ERR_STREAM_PREMATURE_CLOSE') return
```

src/shared/lib/utils/file-storage.ts
```diff
@@ -422,6 +430,28 @@ function parseJsonlLine<T>(line: Buffer): T | undefined {
...
 }
 
+/** Object-mode items in → JSON array bytes out. Used as `Readable.from(items).pipe(this)`. */
+export function createJsonArrayStringifyTransform(): Transform {
+  let first = true
+  return new Transform({
+    writableObjectMode: true,
+    transform(obj, _enc, cb) {
+      try {
+        const json = JSON.stringify(obj)
```
iddogino: **Invalid JSON for `undefined` elements.** `JSON.stringify(undefined)` returns the *value* `undefined`, so this pushes the literal text `[undefined` — malformed output the client can't parse. The old `c.json(transformed)` path serialized `undefined` array elements as `null` (`JSON.stringify([undefined, 1])` → `"[null,1]"`). `transformMessages` shouldn't produce `undefined` entries today, but this is an exported shared utility, so future callers inherit the trap.

```suggestion
        const json = JSON.stringify(obj) ?? 'null'
```

src/api/routes/agents.ts
```diff
@@ -1825,7 +1825,17 @@ agents.get('/:id/sessions/:sessionId/messages', AgentRead(), async (c) => {
...
+      console.error('Failed to stream messages:', err)
+      captureException(err, { tags: { component: 'agents', operation: 'stream-messages' } })
+    }
+    source.on('error', reportStreamError)
+    stringify.on('error', reportStreamError)
+    return c.body(Readable.toWeb(source.pipe(stringify)) as ReadableStream, 200, {
```
iddogino: Nit (non-blocking): `.pipe()` doesn't forward errors — if `source` errored, it would unpipe without destroying `stringify`, leaving the response hanging open instead of aborting. Can't realistically happen with `Readable.from(array)`, but `Readable.from(transformed).compose(stringify)` (or `stream.pipeline`) would make the pattern safe for the next caller who copies it with a real error-capable source.

iddogino (changes requested):
Reviewed in depth: pulled the branch, ran the PR's test files (436 + 112 passing across the four `readJsonlFile` callers), typecheck + lint clean, and verified streamed bodies survive every serving path (`@hono/node-server` in Electron, CORS/auth-only middleware, cloud proxy pass-through). The approach is right and the response shape/semantics are preserved — ENOENT still returns `[]`, malformed/blank lines still skipped, Dates still ISO strings.

Requesting changes for two small pre-merge fixes (details inline):

1. **Client aborts will spam Sentry** — confirmed with a live repro against `@hono/node-server`: disconnecting mid-body fires `reportStreamError` with `ABORT_ERR`, and this endpoint's multi-MB, frequently-refetched responses make mid-body aborts routine. Filter abort/premature-close codes before `captureException`.
2. **`createJsonArrayStringifyTransform` emits invalid JSON for `undefined` elements** — old `c.json` path serialized them as `null`; the new per-object stringify pushes the literal text `undefined`. One-liner: `JSON.stringify(obj) ?? 'null'`.

Plus one non-blocking nit on `.pipe()` error forwarding.

Everything else looks good — the test-mock cleanup (dropping the fake `vi.mock('stream')` so GET /messages tests consume the real streaming path) is a genuine improvement, and the PR body is honest about what peak memory it does and doesn't fix.

src/api/routes/agents.ts
yiw190: stream.pipeline is something I read when implementing as well. Will use that to fix it

src/api/routes/agents.ts
yiw190: make sense -- will apply the suggested change

iddogino (approved):
Re-reviewed `40018cee` — all three findings addressed, and the fixes go beyond the minimum in the right ways. Approving.

Verified on the new head:
- **Abort filter**: re-ran the live repro against the exact new pattern (`pipeline` + filter) — a client disconnect mid-body now resolves as `ABORT_ERR`, gets filtered, zero `captureException` calls, no uncaught `'error'` events, and the `pipeline` callback settles cleanly.
- **`?? 'null'`**: fixed, plus a regression test asserting `[undefined, 1]` → `[null, 1]`, matching the old `c.json` semantics exactly.
- **`pipeline` instead of `.pipe()`**: the `pipeline(source, stringify, cb)` + `toWeb(stringify)` wiring is correct — pipeline wires and monitors the writable side without competing with `toWeb` for the readable side. Confirmed the happy path still delivers a complete, parseable 5000-element body through `@hono/node-server`, and cancel-side destroys surface only as the filtered premature-close/abort codes.
- Test suites for all four `readJsonlFile` callers (549 tests), `tsc --noEmit`, and eslint all pass.

Nice touch refactoring the transform tests into the `stringifyJsonArray` helper — it also exercises the same `pipeline` pairing the route uses.

PR #753, opened 2026-08-13
feat: cursor-page GET /messages instead of the full transcript

yiw190:
## Summary
- Independent of #752 (stream-pipe). Both PRs target `main` and can merge in either order.
- The renderer already windowed the last 300 messages and loaded 200 more on scroll-up, but only after `GET /messages` downloaded the whole JSONL.
- `?limit=` / `?cursor=` now reads a tail of the file (older screenshot rows are not `JSON.parse`d), returns `{ messages, nextCursor }`, and the chat fetches the next older page on scroll-up.
- No query params keeps the existing full-array response (`c.json(transformed)`) so #752 can still land on that return, and other callers stay unchanged.
- Optional host cap: `MESSAGES_PAGE_LIMIT` / `MESSAGES_PAGE_OLDER_LIMIT` (default 300 / 200). The renderer still requests 300/200; the API min()s against the env.

## Test plan
- [ ] Open a long session (300+ visible messages). First load is the trailing page; scroll to the top loads older messages.
- [ ] SSE / new messages while scrolled up do not drop the older page or jump the viewport.
- [ ] Small sessions (under 300) still render in full; `nextCursor` is null.
- [ ] `GET /messages` with no query params still returns a JSON array (compat).
- [ ] Merging after #752 (or before it) does not conflict.
- [ ] With `MESSAGES_PAGE_LIMIT=100`, first page is 100 even if the client asked for 300.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,6 +20,11 @@ ANTHROPIC_API_KEY=sk-ant-your-key-here
 # How long to cache Docker/Podman availability checks (in seconds)
 # RUNNER_AVAILABILITY_CACHE_TTL_SECONDS=60
 
+# Optional: cap GET /messages page size when the client sends ?limit=
+# (no-limit callers still get the full array). Defaults 300 / 200, max 500.
+# MESSAGES_PAGE_LIMIT=300
+# MESSAGES_PAGE_OLDER_LIMIT=200
+
 # Platform: Base URL for the platform web app (login / auth exchange)
 # PLATFORM_BASE_URL=https://platform.example.com
 
--- a/e2e/specs/api-error-display.spec.ts
+++ b/e2e/specs/api-error-display.spec.ts
@@ -30,6 +30,9 @@ test.describe('API Error Display', () => {
     await expect(errorCard.first()).toBeVisible({ timeout: 15000 })
     await expect(errorCard.first()).toContainText('LLM Provider Error')
     await expect(errorCard.first()).toContainText('Invalid API key')
+    await expect(errorCard.first()).toHaveAttribute('aria-expanded', 'false')
+    await errorCard.first().getByText('More details', { exact: true }).click()
+    await expect(errorCard.first()).toHaveAttribute('aria-expanded', 'true')
     await expect(errorCard.first()).toContainText('external LLM provider API')
   })
 
--- a/e2e/specs/message-queueing.spec.ts
+++ b/e2e/specs/message-queueing.spec.ts
@@ -114,10 +114,12 @@ test.describe('Message queueing while agent is working', () => {
 
     // Hold transcript GETs so the picked-up ghost cannot materialize while we
     // stage the lost race. Send (POST) and cancel (DELETE …/queued-messages/:uuid)
-    // use other method/URL shapes and pass through.
+    // use other method/URL shapes and pass through. The trailing * covers the
+    // ?limit= query string (Playwright globs match the full URL) without
+    // matching …/queued-messages/* (the literal /messages can't match /queued-).
     let holdTranscript = true
     let pickupRefetchSeen = false
-    await page.route('**/sessions/*/messages', async (route) => {
+    await page.route('**/sessions/*/messages*', async (route) => {
       if (route.request().method() !== 'GET') return route.continue()
       pickupRefetchSeen = true
       while (holdTranscript) await new Promise((r) => setTimeout(r, 100))
--- a/e2e/specs/subagent-browser-input-status.spec.ts
+++ b/e2e/specs/subagent-browser-input-status.spec.ts
@@ -28,6 +28,15 @@ test.describe('Subagent Browser Input Status', () => {
 
     agent = await createAgent(request, uniqueName(testInfo, 'Subagent Browser Agent'))
     await gotoAgentHome(page, agent)
+
+    // Establish the session route and its event stream before issuing a
+    // short-lived request. Otherwise the create-session navigation can miss
+    // request events while the chat is still mounting.
+    await sessionPage.sendMessage(`slow response establish subagent session ${uniqueSuffix(testInfo)}`)
+    await sessionPage.waitForResponse(15000)
+    await page.reload()
+    await expect(sessionPage.getMessageList()).toBeVisible({ timeout: 15000 })
+    await agentPage.waitForStatus('idle', 15000)
   })
 
   test('browser input requested by a subagent shows the card and flips status to awaiting_input', async ({ page }, testInfo) => {
--- a/electron.vite.config.ts
+++ b/electron.vite.config.ts
@@ -50,6 +50,8 @@ export default defineConfig({
       __WEB__: JSON.stringify(false), // all Electron (dev http + prod file://) → hash history
       __AMPLITUDE_API_KEY__: JSON.stringify(process.env.AMPLITUDE_API_KEY || analyticsConfig.defaultAmplitudeKey),
       __RENDER_TRACKING__: JSON.stringify(process.env.RENDER_TRACKING === 'true'),
+      'process.env.MESSAGES_PAGE_LIMIT': JSON.stringify(process.env.MESSAGES_PAGE_LIMIT ?? ''),
+      'process.env.MESSAGES_PAGE_OLDER_LIMIT': JSON.stringify(process.env.MESSAGES_PAGE_OLDER_LIMIT ?? ''),
       'globalThis.__PLATFORM_BASE_URL__': JSON.stringify(process.env.PLATFORM_BASE_URL || ''),
       'globalThis.__PLATFORM_PROXY_URL__': JSON.stringify(process.env.PLATFORM_PROXY_URL || ''),
       'globalThis.__PLATFORM_AUTH_ISSUER_URL__': JSON.stringify(process.env.PLATFORM_AUTH_ISSUER_URL || ''),
--- a/electron.vite.config.ts
+++ b/electron.vite.config.ts
@@ -50,8 +50,6 @@ export default defineConfig({
       __WEB__: JSON.stringify(false), // all Electron (dev http + prod file://) → hash history
       __AMPLITUDE_API_KEY__: JSON.stringify(process.env.AMPLITUDE_API_KEY || analyticsConfig.defaultAmplitudeKey),
       __RENDER_TRACKING__: JSON.stringify(process.env.RENDER_TRACKING === 'true'),
-      'process.env.MESSAGES_PAGE_LIMIT': JSON.stringify(process.env.MESSAGES_PAGE_LIMIT ?? ''),
-      'process.env.MESSAGES_PAGE_OLDER_LIMIT': JSON.stringify(process.env.MESSAGES_PAGE_OLDER_LIMIT ?? ''),
       'globalThis.__PLATFORM_BASE_URL__': JSON.stringify(process.env.PLATFORM_BASE_URL || ''),
       'globalThis.__PLATFORM_PROXY_URL__': JSON.stringify(process.env.PLATFORM_PROXY_URL || ''),
       'globalThis.__PLATFORM_AUTH_ISSUER_URL__': JSON.stringify(process.env.PLATFORM_AUTH_ISSUER_URL || ''),
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -287,6 +287,7 @@ vi.mock('@shared/lib/services/session-service', () => ({
   updateSessionName: vi.fn(),
   registerSession: vi.fn(),
   getSessionMessagesWithCompact: vi.fn(),
+  getSessionMessagesPage: vi.fn(),
   getSession: vi.fn(),
   getSessionMetadata: vi.fn(),
   sessionExists: vi.fn().mockResolvedValue(true),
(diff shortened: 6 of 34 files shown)


iddogino (changes requested):
Reviewed in depth with the branch checked out locally — typecheck, lint, and all 654 touched-suite tests pass, and the overall design (tail-read + transform + slice server-side, accumulate-older client-side, full-array default kept for #752) is sound. Three issues need fixing before merge; the first is confirmed with a reproduction.

## Blocking

### 1. Unstable cursor when the tail cut lands inside a multi-entry assistant message (confirmed repro)

`transformMessages` merges assistant entries sharing one `message.id` into a single item whose `id` is the **first** entry's `uuid`. When `readJsonlTailLines`'s cut lands mid-message and the growth loop exits with `transformed.length === limit` exactly (or `end === limit` in the cursor branch), `messages[0]` is a partially-merged head whose id is a mid-message entry uuid — an id the full transform never emits.

Repro (vitest, using the session-service test scaffolding): a transcript ending in `…, asst(X-0, msg-X), asst(X-1, msg-X), 4 user rows, 27 isMeta rows` with `limit: 5` (initial tail = 32 lines, cutting between X-0 and X-1):

```
first page ids: [ 'X-1', 'tail-u-1', 'tail-u-2', 'tail-u-3', 'tail-u-4' ]  nextCursor: X-1
older page ids: []  nextCursor: null      // 40 older turns exist in the file
```

Three consequences:
- The follow-up `?cursor=X-1` request can never find the id (the fully-merged item is keyed `X-0`), doubles the tail all the way to `MAX_TAIL_LINES` (re-reading and re-transforming up to 50k lines per doubling), then returns an empty page with `nextCursor: null` — **scroll-up pagination silently dies and all older history becomes unreachable** for the rest of the session view.
- The first page renders the assistant message with its leading content blocks missing.
- Same empty-page dead end for any cursor id that has vanished for other reasons (e.g. the message was deleted), after the same worst-case 50k-line growth.

Suggested fix, mirroring the `lines.shift()` partial-line drop one level down: when `!reachedStart`, treat `transformed[0]` as potentially partial and never let it be `messages[0]` — grow while `transformed.length <= limit` (strict) in the no-cursor branch, and require `end > limit` / clamp `start ≥ 1` in the cursor branch. The dropped item stays reachable through the next older page, so there's no gap. Cursor-not-found should also stop being an expensive dead end — cap the growth for the miss case and/or fall back to something non-terminal rather than `{messages: [], nextCursor: null}`.

### 2. Scroll-up permanently wedges when an older fetch fails or comes back empty (`message-list.tsx`)

`prevScrollHeightRef` is set before `fetchOlder()` and only ever cleared by the `[windowSize]` layout effect, which fires only when the arriving page grows `data` (via the anchor effect). If the fetch **rejects** (caught and logged in `use-messages.ts`) or resolves with an **empty page** (finding 1), nothing clears the ref — and since it also guards the `hiddenCount > 0` local-expansion branch, *both* scroll-up paths are dead until remount, while the "N earlier messages hidden — scroll up to load" banner keeps advertising the broken gesture. Suggest having `fetchOlder` report whether it prepended anything and clearing the guard in the caller when it didn't (clearing it unconditionally in a `finally` would run before the layout effect and break the scroll compensation on the success path).

### 3. `lastPersistedAssistantMatches` never matches the new cache shape (`use-message-stream.ts` ~224)

It reads the messages query cache via `getQueriesData` and guards with `Array.isArray(data)` — but this PR changes the cached value under `['messages', sessionId, agentSlug]` to `{ messages, nextCursor }`. The guard now always skips, so `reconcileMessagesAfterIdle` never early-exits and runs all 3 invalidate/refetch rounds after **every** turn end. The UI still reconciles (the refetches do the work), but the match-and-stop contract is silently dead and every idle costs 3 extra GETs. Needs to unwrap `data.messages` when the envelope shape is present.

## Non-blocking

- **Viewport jump race (the PR's own test-plan bullet):** if a new message streams in while an older fetch is in flight, the anchor effect grows `windowSize`, and the `[windowSize]` restore effect misapplies the held `prevScrollHeightRef` delta (jump down by the appended height) *and* clears the guard — so the older page then prepends with no compensation (second jump). Worth reworking the guard to be fetch-scoped rather than held across the network round-trip.
- **Seam flicker:** the slid-off accumulation is a plain `useEffect`, so the render where `latest` slides forward paints once *without* the slid-off messages before they land in `older`. `useLayoutEffect` (or deriving the union synchronously against a ref) removes the visible blink.
- **Deleted old messages never disappear:** `useDeleteMessage` invalidates the query, which refetches only the latest page; a message living in the `older` accumulation stays rendered until session switch.
- **Window-bounded consumers:** `use-pending-requests`, `agent-activity-indicator`, and `browser-activity-log` share `useMessages` and now see only the trailing ~300 messages (their hook instances never page). All three are tail-anchored so this looks fine, but it's a silent behavior change worth a comment on the hook.
- `MESSAGES_PAGE_LIMIT` / `MESSAGES_PAGE_OLDER_LIMIT` are undocumented outside the code.

Nice touches: the 80KB-row test proves rows larger than `TAIL_READ_CHUNK` are handled; keeping the no-param path byte-identical for #752 (now approved) was verified — no conflict in either merge order.

iddogino (changes requested):
Re-reviewed `aeb9e6c2` + `aec96b0c` with the branch checked out — typecheck, lint, and all 763 tests in the affected suites pass. *(Updated after triaging the red Tests workflow and a second look at the paging bounds — adds the E2E spec break as blocking and two scoped follow-ups.)*

## Fixed and verified

- **Unstable cursor (finding 1)** — properly fixed. `dropPartialHead` + the strict growth conditions implement exactly the right invariant, and my X-0/X-1 repro is now a committed regression test. I also ran a backward-walk sweep (5 transcript shapes: 80–400 turns, 2–7 assistant entries per message, 0–9 interleaved meta rows, limits 2–11, walking every page to the start): full coverage, zero duplicates, zero stalls, cursors always advance. The cursor-miss policy in `aec96b0c` is also the right call — your livelock analysis of the "point at a newer message" alternative is correct, and terminating with a null cursor on a genuinely vanished id is the safe behavior.
- **Reconcile envelope unwrap (finding 3)** — fixed, tests updated, early-exit works against both cache shapes.
- Non-blocking items too: slid-off accumulation is now `useLayoutEffect` (seam flicker gone), delete tombstones filter both accumulations, `fetchOlder` failures report to Sentry, cursor selection prefers server-issued cursors, env vars documented in `.env.example`.

## Blocking 1: the wedge fix broke scroll compensation on the normal path (confirmed empirically)

Adding `visibleMessages.length` to the restore effect's deps (`message-list.tsx:761`) makes the guard get consumed one commit too early. Sequence when an older page arrives on the standard windowed path (e.g. the very first scroll-up fetch: 300 loaded, windowSize 300):

1. Commit N (data grown to 500, windowSize still 300): the rendered slice `slice(-300)` is unchanged, so the DOM hasn't grown — but the restore effect fires anyway (`visibleMessages.length` changed), computes `scrollHeight` delta = **0**, and clears `prevScrollHeightRef`.
2. The anchor effect's `setWindowSize(n + 200)` then re-renders synchronously: commit N+1 mounts the 200 prepended rows, the restore effect fires again (`windowSize` changed) — but the guard is already null, so nothing compensates.

Repro: I added a jsdom test that derives `scrollHeight` from the rendered row count (10px/row) and drives the real `fetchOlder(onBeforePrepend)` flow. Expected `scrollTop` 50 → 2050 after the 200-row prepend; actual: stays **50** — the viewport shows the top of the prepended chunk instead of the content the user was reading. Reverting the deps to `[windowSize]` alone makes the same test pass (scrollTop = 2050).

`[windowSize]` alone is sufficient and correct here: the anchor effect always grows `windowSize` when a prepend lands while scrolled up, so the restore effect reliably fires at the commit where the rows are actually in the DOM. And since `aeb9e6c2` now captures the guard **at prepend time** (`onBeforePrepend`), the mid-flight-append race that might have motivated the extra dep is already handled — an append during the fetch finds the guard null and compensates nothing, correctly.

One residual edge worth a line while you're in there: if the user scrolls back to the bottom while the fetch is in flight, `onBeforePrepend` still sets the guard but the anchor effect skips (`isScrolledToBottomRef` true), so the guard lingers and blocks the next scroll-up gesture. Clearing `prevScrollHeightRef` in the scroll handler's at-bottom path (or skipping the capture when already at bottom) covers it.

Suggest also keeping a DOM-derived-scrollHeight compensation test like the repro above in the suite — the existing pagination tests all use fixed mocked geometry, which is why this regressed silently.

## Blocking 2: `message-queueing.spec.ts:103` is broken by this diff (verified in CI run 31757297522)

Playwright URL globs match the **full URL including the query string**, so the transcript-hold interceptor `page.route('**/sessions/*/messages', …)` no longer matches now that the renderer appends `?limit=300`. `pickupRefetchSeen` never flips and the `expect.poll` times out — failed all 3 attempts. Fix in this PR: add a trailing wildcard (e.g. `'**/sessions/*/messages*'`), taking care the pattern still excludes `…/queued-messages/*`; worth grepping `e2e/` for other `/messages`-suffix route patterns while at it.

For the record, the other red in that run is **not** yours: `api-error-display.spec.ts:23` fails identically (all 3 attempts) on main's tip run [31755396364](https://github.com/SkillfulAgents/SuperAgent/actions/runs/31755396364), and home-graph / search / session-long-sleep were flaky-but-passed, consistent with main.

## Should fix (or explicitly accept)

- **History deeper than `MAX_TAIL_LINES` (50k raw rows) is unreachable via paging.** Every cursor request re-scans backwards from EOF, so a full backward walk is O(depth²) and a cursor deeper than the cap terminates as an empty page even though older history exists — a silent regression vs the old full-array response (which, to be fair, was its own failure mode at that size). The clean fix removes both the cap and the quadratic rescans at once: make the cursor opaque and carry the byte offset (`uuid@offset`), so older-page reads seek to the known position instead of re-scanning from EOF. If you'd rather defer that, make the cap explicit (constant comment + a `log`/metric when a walk hits it) so it reads as a decision rather than an accident — and note the current long-transcript test (600 rows) doesn't exercise it.
- **`useDeleteToolCall` has the same staleness shape as the message-delete you already fixed:** a tool call deleted from a message living in the `older` accumulation stays rendered until session switch (invalidation only refreshes the latest page). The tombstone plumbing now exists, so extending it here is small; fine as a follow-up if you'd rather keep this PR tight.

yiw190:
Addressed the re-review in d95c05bc:

**Blocking 1 (scroll compensation)** — fixed as suggested: the restore effect's deps are back to `[windowSize]` only, and the capture in `onBeforePrepend` is skipped when the user has returned to the bottom mid-fetch (covers the residual lingering-guard edge you flagged; no compensation is needed there since the trailing slice doesn't move). Added the DOM-derived-geometry regression test you recommended — scrollHeight computed from the mounted row count at 10px/row driving the real `fetchOlder(onBeforePrepend)` flow. Verified it discriminates: with the old `visibleMessages.length` dep it fails with exactly your repro numbers (`expected 50 to be 2050`); with the fix it passes.

**Blocking 2 (e2e glob)** — the transcript-hold route in `message-queueing.spec.ts` is now `'**/sessions/*/messages*'`, with a comment noting why the trailing `*` is needed and why it can't match `…/queued-messages/*`. Grepped `e2e/` for other `/messages`-suffix patterns: `composer-failure-recovery.spec.ts` and `file-preview.spec.ts` only intercept POSTs (no query string), so they're unaffected. Not run locally — relying on CI for the Playwright confirmation.

**Depth cap** — explicitly accepted for this PR: `MAX_TAIL_LINES` now carries a comment stating the reachability bound and the O(depth²) walk, and both terminal paths (cursor miss at the cap, and a sequential walk truncating to an empty page) log a warning when deeper history exists. The opaque `uuid@offset` cursor is the right long-term fix; deferring it to a follow-up.

**`useDeleteToolCall` staleness** — deferring to a follow-up as offered, to keep this PR scoped.

All 321 tests in the touched suites pass, typecheck/lint clean.

iddogino (approved):
Approving at `d95c05bc`. All three actionable items from the last round are fixed, and I ran a full manual validation pass against the real app on top of the automated layers. No regressions found.

## Round-3 fixes verified

- **Scroll compensation restored** — deps back to `[windowSize]` with the mechanism documented in place, and the DOM-derived-geometry compensation test is exactly the guard this needed (fixed-geometry mocks can't catch a guard consumed one commit early).
- **Guard-linger edge** — capture skipped when the user returned to the bottom mid-fetch; correct pairing, since at bottom the anchor effect doesn't grow the window and the prepend never mounts.
- **Queueing spec glob** — `'**/sessions/*/messages*'`; passes locally (all 5 tests) and in CI.
- **Depth cap explicitly accepted** — documented on the constant with the offset-cursor lift path, warn-logged on both termination paths (verified the warn fires exactly once per capped walk).

## Validation performed (beyond unit suites: 764 tests, typecheck, lint all green)

**Live API battery** (real server, seeded 1,200-item transcript with multi-entry assistant messages, tool_use/tool_result pairs, thinking blocks, isMeta rows, 60KB fat rows):
- No-param GET still returns the bare full array (compat for #752 and external callers).
- `?limit=300` trailing page is byte-identical to the tail slice of the full response; tool results and thinking blocks intact.
- Full cursor walk reassembles **exactly** the full-array sequence (5 hops, 153ms, zero dupes/gaps).
- `limit=0`/`501`/`abc` and oversized cursors → 400; vanished cursor → clean `{[], null}` terminal in 30ms.
- 58.7k-line / 30MB transcript: paged GET **13ms vs 418ms** for the old full-array path; deep walk terminates cleanly at the documented cap (18,749 items) with the warn logged once, no stall.
- `MESSAGES_PAGE_LIMIT=100` / `MESSAGES_PAGE_OLDER_LIMIT=50` live: client asking 300/200 gets 100/50; no-param response unaffected.

**Real-browser pass** (headless Chromium via Playwright against the live app, real layout engine):
- First load renders the trailing ~300 rows and requests `?limit=300` with no cursor.
- Four consecutive scroll-to-top page loads (300→500→700→900→1100 rows): the content under the viewport moved **0.0px** every time, with scrollTop compensated by exactly the prepended height; `?cursor=&limit=200` requests observed on the wire.
- Live SSE while scrolled up with older pages loaded (real mock-container session, 342 items): new turn appended, reading position moved **0.0px**, oldest loaded row still present, new message visible after scrolling down.
- Small session: full render, `nextCursor: null`, no hidden-banner, scroll-to-top fires no cursor fetch.

**CI** (run 31758856366): 286 passed; the queueing spec is green; the only 2 hard failures are `api-error-display:23` and `subagent-browser-input-status:66`, both reproduced identically on main's tip run — pre-existing, not this PR.

Remaining follow-ups already agreed as non-blocking: `useDeleteToolCall` tombstone parity, and the offset-carrying cursor if 50k-line sessions ever need deeper scrollback.

PR #755, opened 2026-08-13
Show and send inherited model and effort on chat, cron, and webhook (SUP-579)

JeremyBischoff:
Closes https://linear.app/datawizz/issue/SUP-579/chat-cron-and-webhook-show-and-send-the-wrong-inherited-effort

11 production files, +549 / −110 net.
9 test files.

Type: bug fix

After connect, the chat card said Select model and Medium.
Cron and webhook said High.
The missing model name was a label bug.
Unset effort was not.
The host sent no effort.
The session treated that as High.
Settings said Medium.

One inherit rule now feeds the card and every new session.
Override, else the agent default, else the app default.
A click still saves.
Opening does not.
Reset clears.

## The card, before and after

Before:

```
open the card
 └─ no override stored
     ├─ chat → show "Select model · Medium"
     └─ cron / webhook → show High
```

After:

```
open the card
 ├─ Settings not loaded → show a dash
 └─ Settings loaded
     → override, else agent default, else app default
     → snap the label if this model cannot do that effort or speed
     → do not write
```

## A new session, before and after

Before:

```
start a session
 → use override, else agent default
 → if neither has effort, send nothing
     └─ session treats nothing as High
```

After:

```
start a session
 → override, else agent default, else app default
 → send that effort
 └─ if no rung has one, omit it and still start
```

Chat, cron, webhook, run-now, composer, and cross-agent invoke all use this path.
Composer still omits an unpicked value on the wire.
The server fills it.
Cross-agent invoke has no per-call pick, so it is agent default, else Settings.

## Opening must not save

Showing the inherit hands the picker a real value.
If that value is illegal for the model, the picker would write a snap on first paint.

```
open, inherited effort the model cannot do
 ├─ before this PR: picker writes Medium
 └─ after: label snaps, stored row stays unset

open, leftover Fast on a model with no speed choice
 ├─ before: picker writes Normal
 └─ after: label snaps, stored row stays unset
```

## Worth reviewing

The inherit helper is the whole contract.
Cards and the six start paths must stay on it.
A future card that invents its own fallback will drift again.

Opening still must not write.
The snap is display-only.
The shared picker is unchanged.

A click holds the whole cron card still until the save lands.
If the save fails, the other knobs stay frozen until you leave the page.
That window is narrow.

## Why not just fix the labels

We could show High on the card to match omit.
That would make the label honest and leave unset sessions on High.
The product default is the Settings effort, not the old High guard.
The rejected path keeps every existing unset session on High forever.

## What this does not change

The shared picker.
Connect does not persist a model.
Speed is still two-rung: agent default, else omit.
The container still treats a missing effort as High.

Existing unset sessions move from High to the app default on the next session.
That is the fix.
That now includes composer chats and cross-agent invokes, not only automations.

## Try it

1. Open a chat with no override.
   The card should show the real inherited model and effort, not Select model or a lone Medium.
2. Reload without picking.
   Nothing should save.
3. Open a scheduled task with no override.
   Same inherit.
   Reset should be hidden.
4. Optional: pick Low, then Reset.
   It should return to inherit.
5. Start a composer chat with no pick.
   The new session should run at Settings' effort, not High.

## Validation

CI green on this head, including e2e after one unrelated flake rerun.
Focused unit: inherit helper, both cards, composer sessions, and cross-agent invoke.
`npm run typecheck` clean.

### DIFF
--- a/src/shared/lib/container/runtime-options.ts
+++ b/src/shared/lib/container/runtime-options.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod'
+import type { AgentPreferences } from '@shared/lib/types/agent-preferences'
 import { EFFORT_LEVELS, SPEED_LEVELS, type EffortLevel, type SpeedLevel } from './types'
 
 /**
@@ -93,22 +94,21 @@ export type RuntimeInherit = {
 
 /**
  * Surface override → agent default → app default.
- * Junk surface/agent values are treated as unset (do not throw).
+ * Junk surface values are treated as unset (do not throw).
  * Effort is omitted when no rung has one. Speed stays two-rung.
  */
 export function resolveRuntimeInherit(
   surface: unknown,
-  agent: unknown,
+  agent: Partial<AgentPreferences> | null | undefined,
   models: unknown,
 ): RuntimeInherit {
   const s = asRecord(surface) ?? {}
-  const a = asRecord(agent) ?? {}
   const raw = asRecord(models) ?? {}
   const m = inheritModelsSchema.parse({ agentModel: raw.agentModel })
 
-  const model = presentString(s.model) ?? presentString(a.defaultModel) ?? m.agentModel
-  const effort = optionalEffort(s.effort) ?? optionalEffort(a.defaultEffort) ?? optionalEffort(raw.agentEffort)
-  const speed = optionalSpeed(s.speed) ?? optionalSpeed(a.defaultSpeed)
+  const model = presentString(s.model) ?? presentString(agent?.defaultModel) ?? m.agentModel
+  const effort = optionalEffort(s.effort) ?? optionalEffort(agent?.defaultEffort) ?? optionalEffort(raw.agentEffort)
+  const speed = optionalSpeed(s.speed) ?? optionalSpeed(agent?.defaultSpeed)
 
   return {
     model,
@@ -127,12 +127,17 @@ export function clampEffortForDisplay(
   return supported.includes('medium') ? 'medium' : supported[0]
 }
 
-/** Snap a resolved speed for display only. Mirrors useSpeedClamp, which always snaps to 'normal'. */
+/**
+ * Snap a resolved speed for display only — mirrors useSpeedClamp. A found
+ * catalog model with no supportedSpeeds allows only 'normal'; an unknown
+ * model leaves the speed alone (useSpeedClamp is inert without a model).
+ */
 export function clampSpeedForDisplay(
   speed: SpeedLevel | undefined,
-  supported: SpeedLevel[] | undefined,
+  catalogModel: { supportedSpeeds?: readonly SpeedLevel[] } | undefined,
 ): SpeedLevel | undefined {
   if (!speed) return undefined
-  if (!supported || supported.length === 0 || supported.includes(speed)) return speed
-  return 'normal'
+  if (!catalogModel) return speed
+  const available = catalogModel.supportedSpeeds ?? ['normal']
+  return available.includes(speed) ? speed : 'normal'
 }
--- a/src/shared/lib/container/runtime-options.ts
+++ b/src/shared/lib/container/runtime-options.ts
@@ -60,3 +60,79 @@ export function parseRuntimeOptions(raw: unknown): RuntimeOptions {
 
   return result
 }
+
+const inheritModelsSchema = z.object({
+  agentModel: z.string().min(1),
+})
+
+function presentString(value: unknown): string | undefined {
+  if (typeof value === 'string' && value.length > 0) return value
+  return undefined
+}
+
+function optionalEffort(value: unknown): EffortLevel | undefined {
+  const parsed = z.enum(EFFORT_LEVELS).safeParse(value)
+  return parsed.success ? parsed.data : undefined
+}
+
+function optionalSpeed(value: unknown): SpeedLevel | undefined {
+  const parsed = z.enum(SPEED_LEVELS).safeParse(value)
+  return parsed.success ? parsed.data : undefined
+}
+
+function asRecord(raw: unknown): Record<string, unknown> | null {
+  if (!raw || typeof raw !== 'object' || Array.isArray(raw)) return null
+  return raw as Record<string, unknown>
+}
+
+export type RuntimeInherit = {
+  model: string
+  effort?: EffortLevel
+  speed?: SpeedLevel
+}
+
+/**
+ * Surface override → agent default → app default.
+ * Junk surface/agent values are treated as unset (do not throw).
+ * Effort is omitted when no rung has one. Speed stays two-rung.
+ */
+export function resolveRuntimeInherit(
+  surface: unknown,
+  agent: unknown,
+  models: unknown,
+): RuntimeInherit {
+  const s = asRecord(surface) ?? {}
+  const a = asRecord(agent) ?? {}
+  const raw = asRecord(models) ?? {}
+  const m = inheritModelsSchema.parse({ agentModel: raw.agentModel })
+
+  const model = presentString(s.model) ?? presentString(a.defaultModel) ?? m.agentModel
+  const effort = optionalEffort(s.effort) ?? optionalEffort(a.defaultEffort) ?? optionalEffort(raw.agentEffort)
+  const speed = optionalSpeed(s.speed) ?? optionalSpeed(a.defaultSpeed)
+
+  return {
+    model,
+    ...(effort ? { effort } : {}),
+    ...(speed ? { speed } : {}),
+  }
+}
+
+/** Snap a resolved effort to what the catalog model allows, for display only. */
+export function clampEffortForDisplay(
+  effort: EffortLevel | undefined,
+  supported: EffortLevel[] | undefined,
+): EffortLevel | undefined {
+  if (!effort) return undefined
+  if (!supported || supported.length === 0 || supported.includes(effort)) return effort
+  return supported.includes('medium') ? 'medium' : supported[0]
+}
+
+/** Snap a resolved speed for display only. Mirrors useSpeedClamp, which always snaps to 'normal'. */
+export function clampSpeedForDisplay(
+  speed: SpeedLevel | undefined,
+  supported: SpeedLevel[] | undefined,
+): SpeedLevel | undefined {
+  if (!speed) return undefined
+  if (!supported || supported.length === 0 || supported.includes(speed)) return speed
+  return 'normal'
+}
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -5961,6 +5961,7 @@ describe('session model/effort resolution — POST /:id/sessions', () => {
       agentModel: 'global-agent-model',
       browserModel: 'browser-model',
       dashboardBuilderModel: 'dashboard-model',
+      agentEffort: 'medium',
     })
   })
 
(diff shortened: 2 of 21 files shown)


src/shared/lib/container/runtime-options.ts
```diff
@@ -60,3 +60,79 @@ export function parseRuntimeOptions(raw: unknown): RuntimeOptions {
...
+
+/** Snap a resolved speed for display only. Mirrors useSpeedClamp, which always snaps to 'normal'. */
+export function clampSpeedForDisplay(
+  speed: SpeedLevel | undefined,
+  supported: SpeedLevel[] | undefined,
+): SpeedLevel | undefined {
+  if (!speed) return undefined
+  if (!supported || supported.length === 0 || supported.includes(speed)) return speed
+  return 'normal'
+}
```
iddogino: **Confirmed: this doesn't mirror `useSpeedClamp` for the common case, and it breaks "opening does not save."**

`availableSpeeds()` treats a catalog model with **no `supportedSpeeds`** as `['normal']` ("Omitted means no speed choice" — `speed-section.tsx`), so `useSpeedClamp` fires `onSpeedChange('normal')` for any non-normal speed on such a model. This helper passes the speed through instead (`!supported → return speed`).

Result: an inherited `defaultSpeed: 'fast'` from agent prefs, resolved onto a model without `supportedSpeeds` — most of the catalog; e.g. a stale defaultSpeed after the agent's default model changed, or a surface model override that isn't speed-capable — reaches the picker as `speed='fast'`, and the picker's internal clamp **persists `{speed: 'normal'}` on first paint**, on both cards (failing repro test in the review body).

The distinction the helper needs is "model not in catalog" (clamp inert — pass through, matching `useSpeedClamp`'s `if (!model) return`) vs "model found, no speeds" (only `'normal'` allowed). Passing `catalogModel?.supportedSpeeds` conflates the two, so the fix has to take the model, not the list:

```ts
/**
 * Snap a resolved speed for display only — mirrors useSpeedClamp. A found
 * catalog model with no supportedSpeeds allows only 'normal'; an unknown
 * model disables the clamp (useSpeedClamp is inert without a model).
 */
export function clampSpeedForDisplay(
  speed: SpeedLevel | undefined,
  catalogModel: { supportedSpeeds?: readonly string[] } | undefined,
): SpeedLevel | undefined {
  if (!speed) return undefined
  if (!catalogModel) return speed
  const available = (catalogModel.supportedSpeeds as SpeedLevel[] | undefined) ?? ['normal']
  return available.includes(speed) ? speed : 'normal'
}
```

with the two call sites (`runtime-options-card.tsx:53`, `integration-settings-controls.tsx:163`) passing `catalogModel` — which also drops the `as SpeedLevel[]` casts there.

Two notes while here: the `supported.length === 0` branch is dead — the catalog schema requires `supportedSpeeds` to be non-empty and to contain `'normal'` when present. And the effort mirror is exact (verified against `useEffortClamp`; `supportedEfforts` is a required field), so no equivalent hole exists there.

iddogino (changes requested):
Reviewed on a local checkout of dbe2fe25 — the full diff plus the contracts around it: `useEffortClamp`/`useSpeedClamp`, `getEffectiveModels` and `/api/settings/models`, every `createSession` implementation, and the container's `warm-profile.ts`. `npm run typecheck` is clean; all 1,941 tests across the touched directories pass, including the untouched consumers (`webhook-trigger-view.test.tsx`, `chat-integration-side-panel.test.tsx`).

The shape is right: one inherit rule feeding both the cards and the four start paths is the correct contract, the effort clamp mirrors `useEffortClamp` exactly, and moving the cron/webhook card off admin-gated `useSettings` onto `useModelSettings` quietly fixes those cards for non-admin users in auth mode.

Requesting changes for one confirmed bug (inline comment); item 2 below is an asymmetry this PR introduces, so it should either land here or in an immediately-filed follow-up rather than being dropped.

### 1. Opening the card can still save — the speed clamp doesn't mirror `useSpeedClamp` (confirmed, inline)

An inherited `defaultSpeed: 'fast'` resolved onto a model with no `supportedSpeeds` (most of the catalog) reaches the picker unclamped, and the picker's internal `useSpeedClamp` persists `{speed: 'normal'}` on first paint — on both cards. That breaks the PR's core "opening does not save" invariant and also flips `hasCustom`, surfacing Reset. Details and a suggested fix are in the inline comment on `clampSpeedForDisplay`.

<details>
<summary>Failing repro test (drop next to <code>override-inherit.test.tsx</code> — both cases fail with exactly one <code>{speed: 'normal'}</code> write on mount)</summary>

```tsx
// @vitest-environment jsdom
import { describe, it, expect, vi, beforeEach } from 'vitest'
import { render, screen } from '@testing-library/react'
import { IntegrationModelEffort } from '@renderer/components/chat-integrations/integration-settings-controls'
import { makeChatIntegration } from '@renderer/components/chat-integrations/test-factories'
import { RuntimeOptionsCard } from './runtime-options-card'
import type { ModelDefinition } from '@shared/lib/llm-provider'
import type { EffortLevel } from '@shared/lib/container/types'

const STD: EffortLevel[] = ['low', 'medium', 'high']
// No supportedSpeeds on the model — like most catalog entries.
const CATALOG: ModelDefinition[] = [
  { id: 'claude-opus-4-8', label: 'Opus 4.8', family: 'opus', isLatest: true, icon: 'anthropic', supportedEfforts: STD },
]

const useSettingsMock = vi.fn()
const mutateChat = vi.hoisted(() => vi.fn())

vi.mock('@renderer/hooks/use-chat-integrations', () => ({
  useUpdateChatIntegration: () => ({ mutate: mutateChat, isPending: false }),
}))
vi.mock('@renderer/hooks/use-settings', () => ({
  useSettings: () => useSettingsMock(),
  useModelSettings: () => useSettingsMock(),
}))
vi.mock('@renderer/hooks/use-agent-preferences', () => ({
  // Agent default speed set (e.g. stale from a prior speed-capable default model).
  useAgentPreferences: () => ({ data: { defaultSpeed: 'fast' } }),
}))

beforeEach(() => {
  mutateChat.mockReset()
  useSettingsMock.mockReturnValue({
    data: {
      llmProvider: 'anthropic',
      llmProviderStatus: [{ id: 'anthropic', catalog: CATALOG, defaultModels: { agent: 'opus' } }],
      models: { agentModel: 'claude-opus-4-8', agentEffort: 'medium' },
    },
  })
})

describe('inherited speed on a non-speed model', () => {
  it('chat card: opening must not write', () => {
    render(<IntegrationModelEffort integration={makeChatIntegration()} />)
    expect(screen.getByTestId('settings-model-trigger')).toBeInTheDocument()
    expect(mutateChat).not.toHaveBeenCalled() // fails: called with { speed: 'normal' }
  })

  it('cron card: opening must not write', () => {
    const onUpdate = vi.fn()
    render(
      <RuntimeOptionsCard agentSlug="a" model={null} effort={null} speed={null} onUpdate={onUpdate} />,
    )
    expect(screen.getByTestId('settings-model-trigger')).toBeInTheDocument()
    expect(onUpdate).not.toHaveBeenCalled() // fails: called with { speed: 'normal' }
  })
})
```
</details>

### 2. The interactive path was left on the old ladder — same bug, plus a new prewarm cost

`src/api/routes/agents.ts` still sends `effort: runtimeOptions.effort ?? agentPrefs.defaultEffort` with no app-default rung, and `prewarmDefaults.effort` likewise. Two consequences:

- The composer displays the same inherit this PR standardizes (`agentDefaultEffort ?? settings.models.agentEffort` → Medium) but omits unseeded effort from the wire — so an unset **interactive** session still shows Medium and runs High. That's the exact SUP-579 display≠sent bug, alive on the main surface. It isn't in the "What this does not change" list, so flagging in case it's an oversight rather than a scoping call.
- Warm-process reuse is exact key equality including `effort` (`agent-container/src/warm-profile.ts` → `warmProfileKey`). Automated sessions now ask for `effort: 'medium'` while interactive sessions warm a profile with `effort: undefined` — before this PR both omitted effort and their profiles matched. On agents that mix interactive with cron/webhook sessions, every alternation now spawns cold. (Chat-integration sessions never matched anyway — they carry a systemPrompt.)

Moving `agents.ts` — the session shape *and* `prewarmDefaults` — onto `resolveRuntimeInherit` fixes both at once.

### Minor

- New lint warning: `catalog` in `runtime-options-card.tsx` should be wrapped in `useMemo` (react-hooks/exhaustive-deps at line 50; as-is it also churns `handleReset`'s identity every render).
- `resolveRuntimeInherit` types all three params `unknown` and reads `a.defaultModel` / `a.defaultEffort` / `a.defaultSpeed` stringly. Junk tolerance is right for the DB-sourced surface values, but the agent param is always a typed `AgentPreferences` — a future field rename there would silently fall through to the app default with no type error and no test failure. `agent: Partial<AgentPreferences>` (or parsing with `agentPreferencesSchema`) keeps the tolerance where it matters.

### Verified non-issues, for the record

- `effort: resolved.effort` putting an explicit `undefined` key on `createSession` is safe end-to-end: the real client `JSON.stringify`s the body (which drops it) and the mock reads the property — no consumer checks key presence, so the old "must be omitted entirely" test comment has no live counterpart.
- The cards' dash-gate can't stick: `/api/settings/models` always serves `agentModel` + `agentEffort` (`getEffectiveModels` defaults effort to `medium`), so the gate always clears once settings load, and `resolveRuntimeInherit`'s `.parse` can't throw server-side either.
- The acknowledged trade-offs (a failed save leaves `picked` set, so the card stops following inherit until props change; existing unset automations migrate High → Medium on their next run) behave as described.

JeremyBischoff:
Done in `879f6c31` - composer sessions now send the app default effort. Memo and Partial landed in `02288c6e`. This also turned up the same omit on cross-agent invoke - Done in `bfe9d73d`.

src/shared/lib/container/runtime-options.ts
JeremyBischoff: Done in `02288c6e` - clamp now takes the catalog model, so leftover Fast on a no-speed model no longer writes on open.

iddogino (approved):
Re-reviewed at bfe9d73d — all findings resolved, approving.

- **Write-on-open (finding 1)**: `clampSpeedForDisplay` now takes the catalog model and treats "found, no `supportedSpeeds`" as normal-only while leaving unknown models alone — an exact mirror of `useSpeedClamp` in both directions. I re-ran my original repro (inherited `defaultSpeed: 'fast'` on a speed-less model, both cards): **passes, zero writes on mount**. The leftover-Fast case is also pinned in `override-inherit.test.tsx` now, so it can't regress silently.
- **Interactive path (finding 2)**: `agents.ts` resolves both the session shape and `prewarmDefaults` through `resolveRuntimeInherit`, so unset interactive sessions send the app-default effort and the warm-profile key (`warm-profile.ts` is exact-match on `effort`) lines up across interactive ↔ cron/webhook again. The prewarm assertions in `agents.test.ts` pin exactly that.
- **Bonus**: `x-agent /invoke` moved onto the same helper — cross-agent spawns were the last path off the ladder, so every session start now shares one inherit rule. Worth a one-line update to the PR description, which still says agent-to-agent spawn omits effort.
- **Minors**: `catalog` memoized (lint clean again) and the agent param typed as `Partial<AgentPreferences>`, so a prefs field rename is now a compile error instead of a silent fallback.

Verified locally at bfe9d73d: repro test green, 2,343 tests across the touched areas pass, `npm run typecheck` clean, eslint clean on all changed files. CI is green across the board.

iddogino:
Pushed `ada798b9` on top: extracted the inherit derivation both cards repeated (settings + prefs + `resolveRuntimeInherit` + catalog lookup + display clamps + ready-gate) into a single `useInheritedRuntimeSelection(agentSlug, surface)` hook, per discussion with @iddogino. Pure refactor — no behavior change; the existing card tests (which mock the underlying settings/prefs hooks) pin it, and typecheck/lint are clean. The hook's doc comment states the contract so a future override surface can't quietly invent its own fallback chain — the drift risk the PR description called out. Feel free to drop or reshape the commit if you'd rather keep it out of this PR.

PR #764, opened 2026-08-14
Stop leftover iMessage events from starting a session after connect

JeremyBischoff:
Closes https://linear.app/datawizz/issue/SUP-587/stop-leftover-imessage-events-from-starting-a-session-after-connect

1 production file, +31 / −9.
New leftover-inbound unit file.

## The bug

After iMessage connects, leftover setup traffic is treated as if you typed it.

You finish setup. We send the contact card. That is intended.

The gateway then dumps leftovers back at us: your `/setup`, our own send, the card caption, the contact file.
Each leftover is treated as a new user text.
Each one tries to start a conversation.

```
connect + send contact card
  → leftovers come back as "user" texts
  → each leftover starts a conversation
       ├─ start fails → "Failed to start a new session" on the phone
       └─ start partly works → agent is already running
            → it asks for GitHub
            → you never asked for that
```

That is why the phone gets a burst of the same error, then later talks about a GitHub request you never made.

## What should happen

```
connect + send contact card
  → leftovers dropped
  → quiet
  → you text something real
  → one conversation starts
  → agent replies
```

## Before

```
finish iMessage setup / reconnect
  → we send the contact card
  → gateway flushes leftovers (/setup, our own send, card caption, vcf)
  → each leftover counts as a user text
  → each one starts a session
  → phone: "Failed to start a new session"
  → agent may already be running (GitHub was that side effect)
```

## After

```
finish iMessage setup / reconnect
  → we still send the contact card
  → leftovers are dropped
  → no session
  → you text Hello
  → one session starts from that text
  → agent replies
```

## What changed

The iMessage connector drops leftover inbound before emit.

- Own sends (`fromMe` / `is_from_me`)
- Echoes of what we just sent (`message.sent` ids, last contact-card caption)
- Handshake flush (nothing before `_connected`, then skip the next `queuedCount` receives)

A later real text still emits.

## Unchanged

Contact card after setup still sends.
Slack, Telegram, and session-save order stay as they are.

## Why not a manager-wide filter

Slack already drops bot-authored inbound.
Telegram never echoes the bot's own sends.
The missing filter is local to iMessage.
A manager caption-string skip would also drop a real user text that matches the card.

## Worth reviewing

- Handshake skip is by `queuedCount` after `connected`, including reconnect.
  The live gateway payload shape is still unproven.
- Caption match lasts until the next outbound.
  A user who types the exact card caption in that window is dropped.
- Own-send detection trusts `fromMe` / `is_from_me` on the gateway payload.

## Known leftovers

This does not add an `IncomingMessage` provenance field on every connector.
That is a larger interface change than this defect.

Reconnect with a non-zero queued flush was not hit in live smoke.
This boot and reconnect both reported `0 queued events`.
The unit file covers the non-zero flush.

## Validation

- [x] Red then green on leftover-inbound cases, plus existing connector tests
- [x] Typecheck
- [x] Live smoke: reconnect stayed quiet, `Hello` started one session, reply landed on the phone

### DIFF
--- a/src/shared/lib/chat-integrations/imessage-connector-leftover-inbound.test.ts
+++ b/src/shared/lib/chat-integrations/imessage-connector-leftover-inbound.test.ts
@@ -1,0 +1,216 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
+import { IMessageConnector } from './imessage-connector'
+
+vi.mock('ws', () => {
+  const MockWebSocket = vi.fn() as any
+  MockWebSocket.OPEN = 1
+  return { default: MockWebSocket }
+})
+
+vi.mock('@shared/lib/error-reporting', () => ({
+  captureException: vi.fn(),
+}))
+
+const CONTACT_CAPTION = "Save me as a contact so I'm not just a number. Text me anytime."
+const USER_PHONE = '+15551234567'
+const OTHER_PHONE = '+10005551234'
+
+class MockWs {
+  readyState = 1
+  sent: string[] = []
+  send(data: string) { this.sent.push(data) }
+  close() {}
+  on() {}
+}
+
+function createConnector(): IMessageConnector {
+  return new IMessageConnector({
+    gatewayUrl: 'ws://localhost:3456',
+    phoneNumber: USER_PHONE,
+    token: 'test-token',
+  })
+}
+
+function wireUp(connector: IMessageConnector): MockWs {
+  const ws = new MockWs()
+  ;(connector as any).ws = ws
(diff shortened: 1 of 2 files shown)


JeremyBischoff:
Closing. We could not reproduce the phone burst.

A clean setup logged `0 queued events` and stayed quiet until a real Hello.
The one incident is more likely two clients on the same iMessage line (dev, then the app) than leftover setup traffic.
The gateway kicks the older socket (`replaced`). That fight can look like leftover inbound.

Not shipping a filter we cannot demo.

PR #767, opened 2026-08-14
Teach Replicate list then schema then create

JeremyBischoff:
Depends on SkillfulAgents/platform#288. That PR is on `main`.
File-duration models are out of the platform table until after-run billing exists.
Refresh staging from `main` before you test this there. Staging may still list those rows.

Type: prompt change.
Production: 1 file.
Tests: 1 file.

Teach list → schema → create for platform Replicate.
Drop the old catalog scrape.
Confirm before video, music, 3D, talking-head, or voice cloning.
Those confirms include the cost from the list row.
No hardcoded slugs. The platform table stays the menu.

```
list GET /v1/replicate/models?kind=
schema GET /v1/replicate/models/{owner}/{name}
create POST /v1/replicate/models/{owner}/{name}/predictions
 └ unknown slug → 403
```

## What this does not do

- No SuperAgent allowlist of its own
- Does not add lip-sync or source-file talking-head. Those wait on platform after-run billing.

## How to check

1. Connect a platform account so the media section renders.
2. Ask the agent to generate an image. It should `GET /v1/replicate/models?kind=image` first.
3. Ask for video. It should confirm with a cost from the list row.
4. Lip-sync or talking-head from a source file should stay off the list.

## Validation

- `npm --prefix agent-container test -- src/system-prompt-vars.test.ts` - 40 passed

### DIFF
--- a/agent-container/src/system-prompt-vars.test.ts
+++ b/agent-container/src/system-prompt-vars.test.ts
@@ -42,6 +42,9 @@ describe('generateSystemPrompt rendering', () => {
     expect(out.includes('## Built-in media generation')).toBe(webhook)
     expect(out.includes('v1/replicate')).toBe(webhook)
     expect(out.includes('ANTHROPIC_AUTH_TOKEN')).toBe(webhook)
+    expect(out.includes('GET .../models')).toBe(webhook)
+    expect(out.includes('?kind=image|video|audio|3d|talking_head')).toBe(webhook)
+    expect(out.includes('Available models')).toBe(false)
   })
 
   // The pause/resume guidance must always render: without it agents reach for
--- a/agent-container/src/system-prompt-vars.test.ts
+++ b/agent-container/src/system-prompt-vars.test.ts
@@ -43,8 +43,10 @@ describe('generateSystemPrompt rendering', () => {
     expect(out.includes('v1/replicate')).toBe(webhook)
     expect(out.includes('ANTHROPIC_AUTH_TOKEN')).toBe(webhook)
     expect(out.includes('GET .../models')).toBe(webhook)
-    expect(out.includes('?kind=image|video|audio|3d|talking_head')).toBe(webhook)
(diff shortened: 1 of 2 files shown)


iddogino:
Superseded by #803.

#803 moves the procedural media-generation guidance out of the base system prompt and into a root-owned, read-only guide at `/opt/gamut/docs/media-generation.md`, so the prompt block this PR edits no longer exists on that branch — there is nothing left to merge cleanly into.

The content here is not lost. It has been folded into #803 (commit `a6045483`), split by layer:

**Into `docs/media-generation.md`** — the API procedure:
- list → schema → create, in that order
- `GET /v1/replicate/models?kind=…`, always filtered; the six kinds are `image`, `video`, `audio`, `talking_head`, `3d`, `document`
- the list is the allowlist; no separate SuperAgent/Gamut allowlist
- `GET /models/{owner}/{name}` for input fields before building a request
- `POST …/predictions` with `{"input": {...}}`; `Prefer: wait` for images and speech, omitted for video and anything long-running, poll `GET /predictions/{id}` instead
- `403` → go back to the list rather than retrying or guessing a variant
- media inputs take a reachable `http(s)` URL to the file itself, not a page that embeds it
- the old `models/_/_` scrape and its `Available models` error fallback are gone

**Into the system prompt** — the parts that are approval rules rather than procedure, so they stay always-on:
- the capability line now covers generating *or editing* images, video, speech, music, 3D, and talking-head clips
- the confirmation list is widened from "video or music" to video, music, 3D, talking-head, or voice cloning
- the confirmation must quote the cost from that model's list row
- "Never invent a model slug"

**Tests** — the five assertions here pointed at prompt strings that no longer exist, so they were re-pointed: the prompt-level rules assert against the rendered prompt under the `platformServices` gate, and a new guide test covers `?kind=`, all six kind values, both endpoint shapes, `403`, `Prefer: wait`, and the two negative assertions.

The platform dependency is unchanged — #803 now documents `?kind=` as the only discovery path, so it needs SkillfulAgents/platform#288 live, exactly as this PR did.

Closing as superseded.

PR #769, opened 2026-08-15
Open agent file links in the preview tray (SUP-591)

JeremyBischoff:
Agent markdown that points at `/workspace/...` opened a dead new tab. Those links now open the existing preview tray, the same one deliver-file already uses. Chat, thought cards, workflow results, and links inside a previewed file all go there.

Closes https://linear.app/datawizz/issue/SUP-591/open-agent-file-links-in-the-preview-tray

38 files, +1569 / -205

```
click a workspace file link
- `/workspace/...` or workspace `file:`, and we know the agent → open that file in the existing side panel
- relative link inside a previewed file → same, from that file's folder
- bare `report.md` in chat, or anything else → same as today

same file from chat and from a delivery → one tab
heading jump → try the full name first (a `#` in the filename wins), 404 → load without the hash
path that leaves /workspace/ → error, not a blank pane
collapsed workflow row → file link only; headings and lists flatten to one line
```

wired: chat (streaming too), thought cards, workflow result card, previewed `.md` (absolute, `file:`, relative link and image), workflow Result line, collapsed agent row

not: notifications, web-search / web-fetch, wiki `[[name]]`, HTML preview, pictures in thought cards, permission copy, explore templates

path collapse → `src/shared/lib/utils/workspace-path.ts`
chat / relative / delivery entry → `src/renderer/lib/workspace-file-url.ts`
when `file:` and relatives rewrite → `src/renderer/lib/markdown-url-transform.ts`
click opens the tray → `src/renderer/components/messages/markdown-file-link.tsx`
heading probe + tab merge → `src/renderer/components/file-preview/use-preview-file-source.ts`, `adoptFilePath` in `src/renderer/context/file-preview-context.tsx`


src/renderer/lib/workspace-file-url.ts
```diff
@@ -8,6 +40,10 @@ export function encodeWorkspaceFilePath(filePath: string): string {
     .join('/')
 }
 
-export function getAgentFileApiPath(agentSlug: string, filePath: string): string {
+export function getAgentFileApiPath(agentSlug: string, filePath: string): string | null {
+  // Callers such as delivered files and bookmarks do not pass through
+  // workspaceFilePathFromHref, so keep the URL-construction boundary safe too.
+  // Do not decode here: a literal "%2e%2e" filename is re-encoded safely.
+  if (hasDotPathSegment(filePath)) return null
```
iddogino: **Regression for delivered files.** The deliver-file tool stores the agent's path verbatim, and the tray, download pill, bookmarks, and folder browser pass it straight into `getAgentFileApiPath`. A delivery of `./output/report.md` or `output/../report.md` now opens a tab with an empty body and no download button, where main served it fine after dot-normalization. Suggest normalizing with `path.posix.normalize` and rejecting only if the result escapes `/workspace/`, which keeps the cloud-proxy protection this comment describes.

src/renderer/lib/workspace-file-url.ts
```diff
@@ -1,3 +1,35 @@
...
+  const pathOnly = href.split(/[?#]/)[0] ?? ''
+  if (!pathOnly.startsWith(WORKSPACE_PREFIX)) return null
+  if (pathOnly.length <= WORKSPACE_PREFIX.length) return null
+
+  let decoded: string
+  try {
+    decoded = decodeURIComponent(pathOnly)
+  } catch {
+    return null
+  }
```
iddogino, line 30: Micromark treats `%` followed by two alphanumerics as already encoded, so a filename like `90%done.md` reaches here undecoded and `decodeURIComponent` throws. Returning null renders a plain new-tab anchor to a path that 404s. Fall back to the raw path instead of null.

src/renderer/lib/workspace-file-url.ts
```diff
@@ -1,3 +1,35 @@
...
+}
+
+/**
+ * Chat markdown file links that point at a container workspace path.
+ * Only `/workspace/...` counts. Bare filenames, in-app routes, and schemes do not.
+ * Decodes once so a CommonMark `%20` becomes a real space before the file API encodes.
+ */
+export function workspaceFilePathFromHref(href: string | undefined | null): string | null {
+  if (!href) return null
+  const pathOnly = href.split(/[?#]/)[0] ?? ''
```
iddogino: Stripping at `?`/`#` before decoding truncates filenames that contain a literal `#` or `?`. `[x](/workspace/output/Issue #12 notes.md)` is valid CommonMark and yields `/workspace/output/Issue ` here, so the tray shows a missing-file error. Only the pre-encoded `%23` case is tested. Consider trying the full path first and falling back to the stripped one.

Also, `?? ''` is dead: `split()[0]` is typed `string` without `noUncheckedIndexedAccess`, and it can never be undefined at runtime.

src/renderer/lib/workspace-file-url.ts
```diff
@@ -1,3 +1,35 @@
...
+  if (!pathOnly.startsWith(WORKSPACE_PREFIX)) return null
+  if (pathOnly.length <= WORKSPACE_PREFIX.length) return null
+
+  let decoded: string
+  try {
+    decoded = decodeURIComponent(pathOnly)
+  } catch {
+    return null
+  }
+  if (!decoded.startsWith(WORKSPACE_PREFIX)) return null
```
iddogino: Lines 24 and 25 are unreachable. Line 15 already requires the literal `/workspace/` prefix, which decoding cannot alter, and every `%XX` decodes to at least one character, so the length check on line 16 already covers this.

src/renderer/components/messages/agent-transcript.tsx
```diff
@@ -35,10 +36,14 @@ export function flattenAssistantMessages(messages: ApiMessageOrBoundary[] | unde
...
 
 /** A markdown text block styled for a transcript (prose, xs). */
-export function TranscriptText({ children }: { children: string }) {
+export function TranscriptText({ children, agentSlug }: { children: string; agentSlug?: string }) {
   return (
     <div className="prose prose-sm max-w-none break-words dark:prose-invert text-xs">
-      <ReactMarkdown remarkPlugins={[remarkGfm]} urlTransform={markdownUrlTransform}>
+      <ReactMarkdown
+        remarkPlugins={[remarkGfm]}
+        components={markdownLinkComponents(agentSlug)}
```
iddogino: `markdownLinkComponents(agentSlug)` builds a new closure on every render, and react-markdown passes that function as the element type. A new identity means React unmounts and remounts every link, and this component re-renders on every streaming delta from the subagent, so a click that lands mid-stream can be lost. Wrap in `useMemo(() => markdownLinkComponents(agentSlug), [agentSlug])` like `message-item.tsx:230` does.

src/renderer/components/messages/workflow-result-card.tsx
```diff
@@ -44,7 +51,11 @@ export function WorkflowResultCard({ notification }: { notification: WorkflowRes
         <div className="flex items-center gap-2 px-2 py-1.5 bg-muted/40">{header}</div>
       )}
       <div className="px-3 py-2 prose prose-sm max-w-none min-w-0 break-words dark:prose-invert prose-strong:font-medium">
-        <ReactMarkdown remarkPlugins={[remarkGfm]} urlTransform={markdownUrlTransform}>
+        <ReactMarkdown
+          remarkPlugins={[remarkGfm]}
+          components={markdownLinkComponents(agentSlug)}
```
iddogino: Same remount issue as `TranscriptText`: this card re-renders on every subagent SSE event, and the fresh `components` object recreates the link elements each time. Memoize on `agentSlug`.

src/renderer/components/messages/markdown-file-link.tsx
```diff
@@ -0,0 +1,82 @@
...
+function MarkdownFileLink({
+  children,
+  href,
+  agentSlug,
+}: {
+  children?: ReactNode
+  href?: string
+  agentSlug: string
+}) {
+  const { openFile } = useFilePreview()
```
iddogino: `useFilePreview()` runs before the workspace check, so every anchor in every mounted message subscribes to the file-preview context, plain https links included. That context value is rebuilt on each tab switch, PDF page scroll, and folder-query keystroke, so all of them re-render. `openFile` is already a stable callback. Either compute `filePath` here and mount a hook-using child only for `/workspace/` links, or expose `openFile` through a separate never-changing context.

Side effect: any render of a markdown link with a truthy `agentSlug` now hard-requires a `FilePreviewProvider`. No current path lacks one, but nothing enforces it.

src/renderer/components/messages/message-item.tsx
```diff
@@ -327,10 +328,18 @@ function MessageItemComponent({ message, isStreaming, agentSlug, sessionId, isSe
...
+  // Don't render assistant messages that have no text, no tool calls, no
+  // thinking, and no workflow result card (and aren't streaming). These are
+  // transient empty entries from partially-persisted JSONL that will be filled
+  // in on the next refetch.
+  if (
+    isAssistant
+    && !hasText
+    && toolCalls.length === 0
+    && thinking.length === 0
+    && workflowResults.length === 0
```
iddogino, line 366: This is a genuine fix: on main, an assistant message whose entire text was a workflow notification hit `return null` and its result card was never shown. It is not mentioned in the PR description and no test exercises the notification-only path. Worth a unit test. Also note `WorkflowCompletePayloadSchema` has no `.min(1)` on `result`, so an empty result renders a "Workflow completed" card with no body.

src/renderer/components/messages/markdown-file-link.tsx
```diff
@@ -0,0 +1,82 @@
...
+      <MarkdownAnchor onActivate={() => openFile(filePath, agentSlug)}>
+        {children}
+      </MarkdownAnchor>
+    )
+  }
+  return <MarkdownAnchor href={href}>{children}</MarkdownAnchor>
+}
+
+/** Override only `a`. Callers that already customize other tags should spread this last. */
+export function markdownLinkComponents(agentSlug?: string): Pick<Components, 'a'> {
```
iddogino, line 83: The anchor extraction is incomplete. `linkify.tsx:62` still says the anchor lives in `message-item.tsx`, and `thinking-block-item.tsx`, `file-preview/renderers/markdown-renderer.tsx`, and `tool-renderers/web-search.tsx` each hand-roll the same `<a>`. A `/workspace/` link therefore opens the tray in a chat bubble but a new tab everywhere else. Thinking blocks are called out as out of scope, but the stale comment and the other duplicates are not.

src/shared/lib/container/mock-container-client.ts
```diff
@@ -1927,6 +1927,31 @@ export class MockContainerClient extends EventEmitter implements ContainerClient
       'I\'ve scheduled the daily issue summary task.'
     )],
     // Wide, many-column markdown table to exercise table breakout (SUP-319).
+    // Longer 'link *' triggers first: substring match is first-hit.
```
iddogino: The SUP-319 wide-table comment on the line above now describes the `link *` entries instead of `['wide table', ...]`. Also, none of the seven `link ...` triggers contain `link file` as a contiguous substring, so the ordering note is guarding nothing.

src/renderer/components/messages/message-item-markdown-file-links.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+import { describe, it, expect, vi, afterEach } from 'vitest'
+import { screen, cleanup, fireEvent } from '@testing-library/react'
+import { renderWithProviders } from '@renderer/test/test-utils'
+import { createAssistantMessage } from '@renderer/test/factories'
+import { useFilePreview } from '@renderer/context/file-preview-context'
+import { MessageItem } from './message-item'
+import { TranscriptText } from './agent-transcript'
+import { WorkflowResultCard } from './workflow-result-card'
+
+vi.mock('./subagent-block', () => ({
```
iddogino: These five `vi.mock` blocks are a trimmed copy of the ones in `message-item.test.tsx`. Consider hoisting them into a shared test helper so the two files cannot drift.

src/renderer/components/messages/message-item-markdown-file-links.test.tsx
```diff
@@ -0,0 +1,160 @@
...
+  TooltipTrigger: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  TooltipContent: ({ children }: { children: React.ReactNode }) => <span>{children}</span>,
+}))
+
+vi.mock('./insufficient-balance-card', () => ({
+  usePlatformBillingUrl: () => null,
+  InsufficientBalanceCard: () => <div data-testid="insufficient-balance-card" />,
+}))
+
+function PreviewProbe() {
```
iddogino: `PreviewProbe` duplicates `FilePreviewProbe` in `message-list.test.tsx`, but reads `openTabs[0]` where the original reads `openTabs[activeTabIndex]`. That differs once more than one tab is open. Reuse the existing probe.

iddogino (changes requested):
Nice feature. A few correctness issues in the href parsing and one regression for delivered files, plus some small cleanups. Inline comments below.

src/renderer/lib/workspace-file-url.ts
JeremyBischoff: Done in `e1346478f` - dotted deliveries now normalize and only refuse if the result leaves `/workspace/`.

src/renderer/lib/workspace-file-url.ts
JeremyBischoff: Done in `e1346478f` - a malformed percent-escape keeps the raw name instead of becoming a new-tab link.

src/renderer/lib/workspace-file-url.ts
JeremyBischoff: Done in `e1346478f` - try the full destination first, then fall back to the stripped path. Dropped the dead `?? ''`.

src/renderer/lib/workspace-file-url.ts
JeremyBischoff: Done in `e1346478f` - dropped the two unreachable post-decode checks.

src/renderer/components/messages/agent-transcript.tsx
JeremyBischoff: Done in `e1346478f` - memoized the link components on `agentSlug`.

src/renderer/components/messages/workflow-result-card.tsx
JeremyBischoff: Done in `e1346478f` - same memo on the workflow card.

src/renderer/components/messages/markdown-file-link.tsx
JeremyBischoff: Done in `e1346478f` - only workspace file links subscribe to the preview hook.

src/renderer/components/messages/message-item.tsx
JeremyBischoff: Done in `e1346478f` - added a unit test for a notification-only assistant message.

src/renderer/components/messages/markdown-file-link.tsx
JeremyBischoff: Done in `e1346478f` - pointed the leftover comment at the file-link renderer and left the other anchors for a follow-up.

src/shared/lib/container/mock-container-client.ts
JeremyBischoff: Done in `e1346478f` - rewrote the two fixture comments so they describe the link cases.

src/renderer/components/messages/message-item-markdown-file-links.test.tsx
JeremyBischoff: Done in `e1346478f` - shared the five mocks so the two test files cannot drift.

src/renderer/components/messages/message-item-markdown-file-links.test.tsx
JeremyBischoff: Done in `e1346478f` - both tests now use the same active-tab probe.

iddogino (commented):
Thanks, all 12 earlier comments are addressed and the rebase merges cleanly. Typecheck, lint, and the unit suite pass locally. A second pass turned up a few more items, none blocking on their own. Three land on files outside the diff, so they go here:

- **`file-preview/renderers/markdown-renderer.tsx:75`** still hard-codes a plain new-tab anchor, so a `/workspace/` link inside a previewed `.md` file, the surface this feature opens into, goes nowhere. `agentSlug` is available from `file-renderer.tsx` but not forwarded. Thinking blocks have the same shape. I know the other anchors were deferred, but the in-tray case feels like part of this feature.
- **`context/file-preview-context.tsx:150`** dedups tabs on raw string equality. A delivery of `output/report.md` and a chat link to `/workspace/output/report.md` now open two tabs for one file, with comments, rename, and remove keyed on different strings. Normalizing in `openFile` with the new helper would fix it.
- **`file-preview/file-preview-tray-content.tsx:95`** renders an empty pane, no error, and no Download button when the API path is null. After normalization that only happens for paths that escape `/workspace/`, but an explicit message would beat a blank body.

src/renderer/components/messages/message-item.tsx
```diff
@@ -219,8 +215,9 @@ export const MarkdownBlock = memo(function MarkdownBlock({ text, embeddedImageAl
     () => createMarkdownUrlTransform({ aliases: embeddedImageAliases, agentSlug }),
     [embeddedImageAliases, agentSlug]
   )
+  const components = agentSlug ? markdownComponentsForAgent(agentSlug) : MARKDOWN_COMPONENTS
```
iddogino: This one was missed in the memo fix. `StreamingMarkdownBlock` below wraps the same call in `useMemo`, but the settled block rebuilds the `a` component type on every render. `memo` holds for most props, but when a new embedded image alias appears in any tool result every settled block re-renders and every link in the transcript is torn down and recreated. Same `useMemo(..., [agentSlug])` as line 253.

src/renderer/lib/workspace-file-url.ts
```diff
@@ -1,3 +1,64 @@
...
+ * Tries the full destination first so a literal `#` or `?` in the name is kept,
+ * then falls back to the query/hash-stripped form.
+ * Decodes once so a CommonMark `%20` becomes a real space before the file API encodes.
+ * A malformed percent-escape keeps the raw name.
+ */
+export function workspaceFilePathFromHref(href: string | undefined | null): string | null {
+  if (!href) return null
+  const stripped = href.split(/[?#]/)[0]
+  const fromFull = decodeHrefCandidate(href)
+  if (fromFull) return fromFull
```
iddogino, line 48: Consequence of my earlier ask, flagging so it is deliberate: because the full destination is tried first, `[see results](/workspace/output/report.md#results)` opens a tab for `report.md#results` and the tray shows a load error. The stripped fallback on line 58 only fires when the query or hash contains a dot segment. Heading anchors into a markdown file are probably more common than literal `#` in a filename. A tray-level fallback to the stripped path on 404 would cover both.

src/renderer/lib/workspace-file-url.ts
```diff
@@ -1,3 +1,64 @@
...
+  let decoded: string
+  try {
+    decoded = decodeURIComponent(pathOnly)
+  } catch {
+    decoded = pathOnly
+  }
+  // URL parsers resolve dot segments before the request reaches the file API.
+  // In cloud mode that could remove the keyed proxy prefix and retarget the
+  // preview at the loopback API, so reject navigation segments after decoding.
+  if (hasDotPathSegment(decoded)) return null
```
iddogino: Now that `getAgentFileApiPath` normalizes and checks for escape, every tab consumer is already protected, so this reject adds no safety. It just makes `/workspace/output/../report.md` a dead link in chat while the identical string delivered by the tool opens fine. The comment above describes the old policy. Consider one shared `parseWorkspacePath` for href, image (`workspaceImageUrl` in `markdown-url-transform.ts` has a third copy), and API-path builders, with the security note in exactly one place.

src/renderer/lib/workspace-file-url.ts
```diff
@@ -8,6 +69,34 @@ export function encodeWorkspaceFilePath(filePath: string): string {
...
 }
 
-export function getAgentFileApiPath(agentSlug: string, filePath: string): string {
-  return `/api/agents/${encodeURIComponent(agentSlug)}/files/${encodeWorkspaceFilePath(filePath)}`
+/**
+ * Collapse `.` / `..` and refuse only if the result leaves `/workspace/`.
+ * Relative deliveries (`./output/report.md`) are resolved under `/workspace/`.
+ * An absolute path that is not already under `/workspace/` is an escape.
+ */
+function normalizeWorkspaceFilePath(filePath: string): string | null {
```
iddogino: This mirrors `normalizeWorkspaceContainerPath` in `src/api/routes/agents.ts` with slightly different semantics, so what counts as inside `/workspace` can drift between client and server. A pure-string module under `src/shared/lib/utils` could serve both.

src/renderer/lib/workspace-file-url.ts
```diff
@@ -1,3 +1,64 @@
+const WORKSPACE_PREFIX = '/workspace/'
+
+function hasDotPathSegment(filePath: string): boolean {
+  return filePath.split('/').some(segment => segment === '.' || segment === '..')
+}
+
+/** POSIX-style `.` / `..` collapse. Matches `path.posix.normalize` for these paths. */
```
iddogino: Nit: this does not quite match `path.posix.normalize`, which keeps a trailing slash. Also, every caller passes an absolute string, so the `!absolute` branch and the `body || '.'` fallback are unreachable, and the trailing-slash strip on line 89 is a no-op.

src/renderer/components/messages/markdown-file-link.tsx
```diff
@@ -0,0 +1,98 @@
...
+function MarkdownFileLink({
+  children,
+  href,
+  agentSlug,
+}: {
+  children?: ReactNode
+  href?: string
+  agentSlug: string
+}) {
+  const filePath = workspaceFilePathFromHref(href)
```
iddogino, line 71: `[report](file:///workspace/output/report.md)` never reaches the tray: the URL transform blanks the `file:` scheme before this override sees the href, while the same PR's image transform explicitly accepts that form for `src`. The test pins this as null, so it is a design choice, but link and image surfaces now disagree on an obvious form.

src/renderer/components/messages/message-item-test-mocks.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+export const tooltip = {
+  TooltipProvider: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  Tooltip: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  TooltipTrigger: ({ children }: { children: React.ReactNode }) => <>{children}</>,
+  TooltipContent: ({ children }: { children: React.ReactNode }) => <span data-testid="tooltip-content">{children}</span>,
+}
+
+export const insufficientBalanceCard = {
```
iddogino: `./insufficient-balance-card` no longer exists after main renamed it to the provider error card in #888. Vitest tolerates mocking a missing module, so this export and the two `vi.mock` lines that use it are dead.

src/renderer/test/file-preview-probe.tsx
```diff
@@ -0,0 +1,13 @@
+import { useFilePreview } from '@renderer/context/file-preview-context'
+
+export function FilePreviewProbe() {
```
iddogino, line 3: Nit: this and `message-item-test-mocks.tsx` are test-only but live in the production components folder. `src/renderer/test/` already holds the other shared test helpers.

src/renderer/components/messages/message-item.tsx
JeremyBischoff: Done in `cccb868d2` - same memo on the settled block.

src/renderer/lib/workspace-file-url.ts
JeremyBischoff: Done in `cccb868d2` - a heading jump that 404s now retries the path without the hash.

src/renderer/lib/workspace-file-url.ts
JeremyBischoff: Done in `cccb868d2` - chat hrefs now use the same collapse-and-refuse helper as deliveries.

src/renderer/lib/workspace-file-url.ts
JeremyBischoff: Done in `cccb868d2` - client and server now share `src/shared/lib/utils/workspace-path`.

src/renderer/lib/workspace-file-url.ts
JeremyBischoff: Done in `cccb868d2` - dropped the unreachable relative and trailing-slash branches.

src/renderer/components/messages/markdown-file-link.tsx
JeremyBischoff: Done in `cccb868d2` - reversing the earlier reject: workspace `file:` links now open the tray.

src/renderer/components/messages/message-item-test-mocks.tsx
JeremyBischoff: Done in `cccb868d2` - dropped the dead `insufficient-balance-card` mock.

src/renderer/test/file-preview-probe.tsx
JeremyBischoff: Done in `cccb868d2` - moved both test helpers into `src/renderer/test/`.

JeremyBischoff (commented):
Done in `cccb868d2` - previously said leftover anchors were deferred; decided to wire them for previewed markdown (including relative links and images), thought cards, the workflow Result line, and the collapsed agent row. A delivery and a chat link to the same file now share one tab, and an escaped path shows an error instead of a blank pane.

PR #770, opened 2026-08-15
fix: stop messages OOM — strip inline tool images, propagate aborts, single-flight pages

yiw190:
## What was broken

A production host-app (1,024 MB ECS task serving 9 agents) was OOM-killed twice (exit 137) while one image-heavy session was open. Root cause was three defects multiplying:

1. **Inline base64 images end-to-end.** Browser-tool screenshots are persisted as base64 inside `tool_result` blocks — 91.2% of the 49 MB transcript, with each image stored twice per line (`message.content[].source.data` + `toolUseResult.file.base64`). `GET /messages?limit=300` returned 19–20 MB (94.8% base64) and cost ~350 MB RSS per request.
2. **Broken cancellation.** The renderer refetches messages on a 15 s poll plus ~10 SSE invalidation paths. TanStack's `cancelRefetch` aborted client-side only: the `queryFn` discarded its `AbortSignal`, and the server never checked for disconnects — superseded requests ran to completion as zombies.
3. **No concurrency guard.** Identical concurrent requests each paid the full transcript parse. Measured: 3 concurrent ≈ +900 MB RSS → over the 1 GB task limit.

## Repro

Copy an image-heavy session transcript (~49 MB, 2,068 lines) into a local agent, then:
- `curl 'GET /api/agents/:slug/sessions/:id/messages?limit=300'` → ~20 MB response, ~350 MB RSS spike
- 3 concurrent curls → ~+800 MB RSS
- 3 curls killed after 50 ms (`--max-time 0.05`) → ~+645 MB RSS (server keeps computing for dead clients)

## What changed

- **`message-transform.ts`**: `stripInlineImages` replaces inline image blocks in tool results with `image_ref { toolUseId, index, mimeType }`; `stripEntryInlineImages` applies it at parse time on the paged path (and drops the unused duplicate `toolUseResult.file.base64`) so megabyte base64 strings never accumulate across the tail.
- **New route `GET /:id/sessions/:sessionId/tool-images/:toolUseId/:imageIndex`**: serves one image on demand via a byte-scan that JSON-parses only lines containing the toolUseId; `Cache-Control: immutable` (tool results never change).
- **Renderer**: `parseToolResult` understands `image_ref`; `ToolCallItem` (expanded view only) loads refs lazily from the new route; `sessionId` threaded to the two call sites.
- **Abort propagation**: `useMessages`' `queryFn` now passes its `AbortSignal` through `apiFetch`; `getSessionMessagesPage` checks the signal between tail-read / parse / transform phases and the route returns 499 for disconnected clients.
- **Single-flight**: `getSessionMessagesPageShared` dedupes identical concurrent page computations; the shared computation aborts only when every waiter's signal has fired.

## Ablation measurements (same 49 MB transcript, dev Electron host)

| Metric | Before | +strip response | +abort | +single-flight | Full stack |
|---|---|---|---|---|---|
| Response size (limit=300) | 20.10 MB | **1.05 MB** | 1.05 MB | 1.05 MB | 1.05 MB |
| RSS, 1 request | +348 MB | +361 MB | — | — | **+300 MB** |
| RSS, 3 concurrent | +797 MB | +680 MB | +444 MB | **+41 MB** | +227 MB |
| RSS, 3 aborted clients | +645 MB | +637 MB | **+178 MB** | **+0 MB** | +57 MB |

Worst case drops from unbounded N × ~350 MB to a structural ~1 × ~300 MB (one computation per page at a time, aborted when all clients disconnect).

## Test plan

- [x] 912 tests pass across `agents.test.ts` (349), transform/service/renderer suites
- [x] `src/` typecheck clean
- [x] Concurrent responses byte-identical under single-flight
- [x] `tool-images` endpoint returns valid JPEG (13 ms, correct content-type)
- [x] Sanity: normal request unchanged (HTTP 200, same message shape, 79 `image_ref` blocks for 79 images)
- [ ] UI walkthrough on the image-heavy session (images load on expand)
- [ ] Subagent transcript images (stripped but no ref context in drawer yet — renders nothing; follow-up if needed)

## Not in scope

- Single-request parse cost (~300 MB) — needs streaming parse / byte-budget; follow-up.
- Transcript storage format (base64 duplication at write time) — SDK-owned resume history; needs SDK contract support for external image refs.

### DIFF
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -304,6 +304,9 @@ vi.mock('@shared/lib/services/session-service', () => ({
   registerSession: vi.fn(),
   getSessionMessagesWithCompact: vi.fn(),
   getSessionMessagesPage: vi.fn(),
+  getSessionMessagesPageShared: vi.fn(),
+  getToolResultImage: vi.fn(),
+  RequestAbortedError: class RequestAbortedError extends Error {},
   getSession: vi.fn(),
   getSessionMetadata: vi.fn(),
   sessionExists: vi.fn().mockResolvedValue(true),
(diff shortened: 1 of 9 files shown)


yiw190:
superseded by #771

PR #771, opened 2026-08-15
Throttle SSE-driven messages refetches and abort superseded fetches (SUP-593)

iddogino:
Client half of the session-OOM fix (SUP-592), plus the server-side abort handling added during review — the /messages read path now honors client cancellation end to end. Fixes [SUP-593](https://linear.app/datawizz/issue/SUP-593).

## Problem

On a long session, every SSE event that touches the transcript (`tool_call`, `tool_result`, `messages_updated`, `session_idle`, …) fired its own `invalidateQueries(['messages', sessionId])`, and the queryFns never passed React Query's AbortSignal — so superseded refetches ran to completion server-side. Each refetch of the incident session is a 17.8 MB response built from two full 37 MB transcript reads; under NFS-grade latency, in-flight concurrency grows as event rate × latency. This is the proven mechanism behind the 1 GB cloud app-container OOM on session resume (repro: OOMKilled in 25 s at 4 req/s fast disk, ~45 s at 1 req/s over real NFS).

## Changes

**Throttle (`use-message-stream.ts`)** — all ~17 SSE-driven `['messages', sessionId]` invalidation sites now funnel through `invalidateMessagesThrottled`, a per-session leading-edge throttle (750 ms): the first event refetches immediately, the rest of the burst collapses into at most one trailing refetch. Collapsed callers share a promise that resolves when the trailing refetch settles, so the post-idle reconcile loop keeps its bounded-retry semantics. Not throttled: `['sessions']` / `['pending-user-requests']` / `['subagent-messages']` invalidations, and all user-initiated ones (delete message/tool call, interrupt).

**Abort (`use-messages.ts`)** — `useMessages` and the sibling queryFns (`useSubagentMessages`, `useWorkflowTree`, `useWorkflowAgentMessages`) now take React Query's per-fetch `{ signal }` and pass it through `apiFetch`, so TanStack v5's default `cancelRefetch` genuinely aborts a superseded HTTP request instead of orphaning it. Nuance (documented in tests): RQ only cancels once the cache has data — initial fetches are reused, never aborted — which is exactly the incident's steady state.

Why not a native TanStack mechanism: `staleTime` is ignored by `invalidateQueries` (it force-marks stale); `cancelRefetch: false` collapses only while a fetch is in flight (window scales with latency, not event rate) and can strand the UI on data fetched *before* the last event, with nothing refetching after; `notifyManager.batch` batches renders, not fetches. The trailing edge guarantees one refetch strictly after the last event of a burst.

## Server-side changes (added during review)

**Server honors aborts (`c3a3590f`)** — the client abort now propagates: route → `getSessionMessagesPage` → `readJsonlTailLines`, checked per 64 KB chunk of the backward tail walk. An aborted request answers 499 (no error log) and stops paying for transcript reads — probed on a real node server with a 20 ms/read fs shim: abort at 150 ms cost 7 reads / 0.46 MB vs 38 reads / 2.49 MB for the control (−82 %), zero reads after.

**Arm the signal before async work (`b27580de`)** — `@hono/node-server` creates the request `AbortController` lazily on first `.signal` access, and its close handler only aborts a controller that already exists — a hangup landing before first access (during auth/session-existence awaits) was permanently missed. An `armAbortSignal` middleware now runs first in `src/api/index.ts`; a real socket-disconnect test pins both the fix and the adapter behavior that requires it. Also: the tail reader observes aborts landing during the final chunk read, and the idle reconcile loop is generation-scoped to its EventSource so a stale wake can't undo the unmount cleanup.

**Review round 4 (`c7314d35`)** — a newer `session_idle` supersedes a sleeping reconcile loop (ownership token) instead of being dropped, so a second short turn or rapid remount doesn't wait on the 15 s poll for its final text; the tail reader checks the signal both before and after each chunk read, so an abort landing during `open()`/`stat()` no longer starts one unnecessary read.

Known residual (deliberate): the paginated response is still eagerly serialized via `c.json`, so an abort arriving mid-stringify is only observed afterward — that's the two-pass-streaming scope of SUP-595.

## Validation

- Unit suite green: 9 526 tests. ~12 existing sync-invalidation tests adapted to fake-timer trailing-edge assertions.
- New tests pin the contract: burst-coalescing bound (6-event burst → exactly leading + 1 trailing; separate windows refetch immediately; per-session independence; `connected` still refetches immediately as the late-join recovery path), real abort semantics in the new `use-messages.test.ts` (superseding invalidation flips `signal.aborted`, superseding refetch still delivers), an `apiFetch` signal pass-through pin in `api.test.ts`, route 499 + signal-forwarding tests, tail-reader abort tests (pre-open / during stat / mid-walk / final read), a real node-server socket-disconnect pair for the arming middleware, and reconcile lifecycle tests (unmount cleanup, stale-loop generation bail, newer-idle supersession). Each regression test was verified to fail against the unfixed code.
- E2E: 20/20 passed across chat-flow, message-queueing, cross-session-messages, concurrent-tabs-request-sync, persistence.
- Manual validation in a real Chromium renderer against a mock-mode server: **3 GET `/messages` per 22 s tool-using turn, down from the measured baseline of 7** (leading +0.07 s, trailing +0.82 s, 15 s safety poll), **zero same-moment duplicates**, a real `net::ERR_ABORTED` captured on a superseded `?limit=300` fetch. Streaming UX intact: live delta rendered 224 ms after turn start; final persisted text ~70 ms after idle.

## Acceptance (from SUP-593)

- [x] One tool-using turn produces ≤ 2–3 messages refetches, zero same-moment duplicates
- [x] Superseded requests actually cancelled client-side
- [x] Streaming UX unchanged; final state reconciles promptly after idle
- [x] `connected` still triggers an immediate refetch (late-join/drift recovery)
- [x] No regression in interrupt flow, queued messages, pending-input recovery (E2E green)

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -51,11 +51,16 @@ import { authEnforcementMiddleware } from './middleware/auth-enforcement'
 import { getAuthSettings } from '@shared/lib/auth/auth-settings'
 import { getPublicAuthProviders } from '@shared/lib/auth/provider-config'
 import { LocalModeAuth, isContainerFacingPath } from './middleware/local-mode-auth'
+import { armAbortSignal } from './middleware/arm-abort-signal'
 
 const app = new Hono()
 
 // Background services start AFTER HTTP bind (web/server.ts, main/index.ts,
 // vite.config.ts) so cold-wake health is not gated on settings/DB/auth.
+
+// Must run before any middleware that awaits: a client hangup during those
+// awaits is otherwise invisible to every later signal check (see the module).
+app.use('*', armAbortSignal)
 
 // Enable CORS for all routes
 const trustedOrigins = process.env.TRUSTED_ORIGINS?.split(',').map(o => o.trim()).filter(Boolean)
--- a/src/api/middleware/arm-abort-signal.test.ts
+++ b/src/api/middleware/arm-abort-signal.test.ts
@@ -1,0 +1,81 @@
+import { describe, it, expect, afterEach } from 'vitest'
+import net from 'net'
+import { serve } from '@hono/node-server'
+import type { ServerType } from '@hono/node-server'
+import { Hono } from 'hono'
+import { armAbortSignal } from './arm-abort-signal'
+
+// Real @hono/node-server + real socket disconnects: the adapter creates the
+// request AbortController lazily on first `.signal` access and its close
+// handler only aborts a controller that already exists, so these tests can't
+// run against Hono's in-process `app.request()` helper — the laziness under
+// test lives in the node adapter.
+
+interface Probe {
+  abortedSeenByHandler?: boolean
+}
+
+function buildApp(withArming: boolean): { app: Hono; probe: Probe } {
+  const app = new Hono()
+  if (withArming) app.use('*', armAbortSignal)
+  const probe: Probe = {}
+  app.get('/probe', async (c) => {
+    // Async pre-work before the route touches the signal — the window in which
+    // the real /messages route awaits auth and session-existence checks.
+    await new Promise((r) => setTimeout(r, 200))
+    const signal = c.req.raw.signal
+    // Give a late-created controller every chance to fire before we look.
+    await new Promise((r) => setTimeout(r, 100))
+    probe.abortedSeenByHandler = signal.aborted
+    return c.text('done')
+  })
+  return { app, probe }
+}
+
+let server: ServerType | undefined
+afterEach(() => {
+  server?.close()
+  server = undefined
+})
+
+function serveApp(app: Hono): Promise<number> {
+  return new Promise((resolve) => {
+    server = serve({ fetch: app.fetch, port: 0, hostname: '127.0.0.1' }, (info) =>
+      resolve(info.port)
+    )
+  })
+}
+
+// Connect, send the request, then destroy the socket while the handler is
+// still inside its pre-signal await.
+async function hitAndHangUp(port: number): Promise<void> {
+  await new Promise<void>((resolve) => {
+    const sock = net.connect(port, '127.0.0.1', () => {
+      sock.write('GET /probe HTTP/1.1\r\nHost: t\r\n\r\n')
+      setTimeout(() => sock.destroy(), 50)
+    })
+    sock.on('error', () => {})
+    sock.on('close', () => resolve())
+  })
+  // Let the handler run to completion and record what it saw.
+  await new Promise((r) => setTimeout(r, 400))
+}
+
+describe('armAbortSignal', () => {
+  it('a hangup during pre-signal awaits is observed once the signal was armed up front', async () => {
+    const { app, probe } = buildApp(true)
+    const port = await serveApp(app)
+    await hitAndHangUp(port)
+    expect(probe.abortedSeenByHandler).toBe(true)
+  })
+
+  // Control: pins the adapter behavior that makes the middleware necessary.
+  // If an upgraded @hono/node-server starts observing pre-access hangups on
+  // its own, this fails — delete the middleware and this file together.
+  it('without arming, the same hangup is permanently missed', async () => {
+    const { app, probe } = buildApp(false)
+    const port = await serveApp(app)
+    await hitAndHangUp(port)
+    expect(probe.abortedSeenByHandler).toBe(false)
+  })
+})
--- a/src/api/middleware/arm-abort-signal.ts
+++ b/src/api/middleware/arm-abort-signal.ts
@@ -1,0 +1,18 @@
+import type { MiddlewareHandler } from 'hono'
+
+/**
+ * Touch the request's AbortSignal before any asynchronous work runs.
+ *
+ * @hono/node-server materializes the request's AbortController lazily on the
+ * first `.signal` access, and its socket-close handler only aborts a controller
+ * that already exists at close time — a client hangup that lands before the
+ * first access is dropped, and the controller a later access creates never
+ * fires. Routes that check `c.req.raw.signal` mid-pipeline (the /messages
+ * transcript reads) would then run their full read/parse/serialize path for a
+ * client that is already gone. Arming the signal here, synchronously and ahead
+ * of every other middleware, guarantees downstream checks observe every hangup.
+ */
+export const armAbortSignal: MiddlewareHandler = (c, next) => {
+  void c.req.raw.signal
+  return next()
+}
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -3773,7 +3773,7 @@ describe('GET /:id/sessions/:sessionId/messages pagination', () => {
     expect(body.messages).toHaveLength(1)
     expect(body.messages[0].id).toBe('m1')
     expect(body.nextCursor).toBe('m1')
-    expect(getSessionMessagesPage).toHaveBeenCalledWith('test-agent', 'sess-1', { limit: 2, cursor: undefined })
+    expect(getSessionMessagesPage).toHaveBeenCalledWith('test-agent', 'sess-1', { limit: 2, cursor: undefined, signal: expect.any(AbortSignal) })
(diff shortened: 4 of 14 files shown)


iddogino:
**Re: [P1] server ignores the abort — confirmed and fixed in c3a3590f.**

The finding was correct: the signal died at the HTTP boundary. `c.req.raw.signal` flipped aborted (as the probe showed) but nothing in the route → `getSessionMessagesPage` → `readJsonlTailLines` pipeline ever looked at it, so every aborted request still paid the full read/parse/serialize cost. At NFS latency (~4.5s/request) even the throttled 1-per-750ms rate stacks ~6 concurrent orphaned jobs — the measured OOM regime.

**Fix:** the request signal now threads through the paginated branch down to the tail reader, which checks it before each 64KB chunk read; the service re-checks before parse/transform, and the route checks before annotation and before serialization. A client abort answers `499` (same idiom as `cloud-proxy.ts`) without logging a false failure. The legacy unpaginated branch gets a post-read check only — its full-file reader is being reworked wholesale by the streaming follow-up (SUP-595).

**Re-run of the probe, same shape, on the deployment's `@hono/node-server` stack** (5.4 MB transcript, 20 ms-per-read fs shim emulating NFS):

| | control (no abort) | aborted at 150 ms |
|---|---|---|
| handler outcome | `handler_completed` | `handler_aborted` → 499 |
| transcript chunk reads | 38 (2.49 MB) | **7 (0.46 MB, −82%)** |
| reads after abort (4 s watch) | — | **0** |

The abort lands at the next 64 KB chunk boundary (7 reads × 20 ms ≈ 140 ms), so worst-case orphaned work per aborted request is one chunk read plus nothing downstream.

New tests pin all three layers: `readJsonlTailLines` (pre-aborted signal never opens the file; mid-walk abort stops after the in-flight read), `getSessionMessagesPage` (rejects with `AbortError`; a never-aborted signal changes nothing), and the route (aborted request → 499; abort landing after the read still skips annotation + serialization).

iddogino:
**Re: [P2] reconcile can slip to ~1.5s / [P3] throttle state survives unmount — both addressed in cb415e05.**

**P2** — the reconcile loop no longer shares the throttle's trailing edge. It refetches immediately via `invalidateMessagesNow`, which *folds into* the throttle rather than stacking on it: a pending trailing timer is cancelled (its waiters settle with the fresh refetch) and the window restarts, so surrounding events coalesce onto the reconcile's fetch instead of adding another. The loop is single-flight per session with its own bounded backoff, so the exemption can't reopen the burst. Finalization under persistence lag recovers on the original **250 / 750 / 1500 ms** schedule; the reconcile-lag test now pins the exact count (three refetches, the folded trailing timer must not fire a fourth). Worth noting for the record: the pre-throttle code had the same class of miss (readable-at-760ms also waited until 1500ms), and on NFS the ~4.5s refetch RTT dominates either schedule — but the fix is strictly better and free.

**P3** — `releaseEventSource` now does symmetric cleanup when the last subscriber unmounts: the pending trailing timer is cancelled, its joined promise is settled (nothing awaiting it can ever hang — this also forecloses the deadlock a naive clear would have created for the reconcile guard), and the map entry is deleted. The map stays bounded by open sessions and a remount starts on a fresh leading edge instead of joining the previous mount's pending work. Two of the finding's consequences were theoretical in production — the QueryClient is an app-lifetime singleton, and a post-unmount invalidate is a no-op mark-stale — but the cleanup is cheap and pins the behavior either way.

New test: burst → unmount before the trailing fires → no refetch ever lands; remount gets an immediate leading-edge refetch. Re-validated in the browser: still 3 GET /messages per tool turn (leading +0.07s / trailing +0.83s / 15s poll), zero duplicates, streaming delta at ~250ms, final text ~160ms after idle. Full suite green (9,520).

iddogino:
All three findings confirmed and fixed in `b27580de`.

**P1 — request cancellation armed too late.** Verified at the source level in the locked `@hono/node-server` 1.19.14: `request.js` creates the `AbortController` lazily on first `.signal` access, and `listener.js`'s close handler guards on `if (abortController)` — a hangup before first access is dropped, and the controller a later access creates never fires. Reproduced with a raw-socket probe against the real adapter: `late-arm → aborted:false`, `early-arm → aborted:true`. Fix: an `armAbortSignal` middleware registered first in `src/api/index.ts` touches `c.req.raw.signal` synchronously before any other middleware runs, so every downstream `throwIfAborted()` observes every hangup — including ones landing during the auth/ACL/session-existence awaits (which are disk stats, i.e. real milliseconds on NFS). Covered by a real node-server socket-disconnect test with two halves: the armed app observes the hangup, and a control app without the middleware pins the adapter behavior that makes it necessary (if a future adapter upgrade fixes the laziness, the control fails and tells us to delete both middleware and test).

**P2 — abort during the final chunk read.** `readJsonlTailLines` only checked at the top of each loop iteration; an abort landing while the final read was in flight had no next iteration to observe it and still paid `Buffer.concat` + a full byte-scan of the tail window (the service-level check after return did stop parse/transform). The check now runs immediately after each awaited read, before the chunk is counted or the tail materialized. New test: single-chunk file, abort fired while the only read is in flight → rejects with AbortError instead of returning the scanned tail.

**P3 — idle reconcile survives final unmount.** The loop's sleeps could straddle the last subscriber's unmount; a stale wake called `invalidateMessagesNow`, whose get-or-create recreated the throttle entry the unmount cleanup had just deleted and restamped the window — delaying a rapid remount's leading-edge refetch by up to one throttle window. The loop now captures its EventSource generation at start and bails on any wake where `eventSources.get(sessionId)` no longer matches — covering both release (entry deleted) and remount (new EventSource). New test walks idle → unmount mid-loop (zero invalidations land) → rapid remount right after the stale loop's would-be retry (leading edge fires immediately) → 5s drain (no late invalidations trickle in).

On the residual eager-`c.json()` serialization risk: agreed, and it's the scope of SUP-595 (two-pass streaming of the paginated response), tracked there rather than patched here.

Validation: full unit suite 9,524 passed / 0 failed (three new tests), typecheck clean, changed-file lint clean.

https://claude.ai/code/session_01NAEmgqH41CiCcaAUvDBs8Q

iddogino:
Both findings fixed in `c7314d35`, and the PR description is updated — it no longer claims "frontend-only, no API change" and now documents the server-side abort work added during review.

**Finding 1 — stale reconcile suppressing a newer turn.** Confirmed, and the remount branch was partly a regression of `b27580de`: the newer idle was dropped by the one-loop-at-a-time Set while the older loop slept, and the older loop then exited on its generation check without invalidating — leaving the new turn's final text to the 15 s poll. The guard is now an ownership token (`idleReconcileOwners: Map<sessionId, symbol>`): a newer `session_idle` overwrites the token, only the current owner invalidates, a superseded loop exits at its next wake, and the finally-cleanup deletes the entry only if still owner. At most one loop ever issues invalidates, so the retry cadence is unchanged. New test covers exactly the scenario you specified — remount plus a second completed turn, all before the stale loop's first 250 ms wake: the new loop's first retry lands at ~250 ms (folding the idle's pending trailing refetch) and the full drain shows exactly its three retries, zero from the superseded loop. Verified to fail against the unfixed code (`expected +0 to be 1` — the dropped reconcile).

**Finding 2 — pre-read abort check.** Restored; `readJsonlTailLines` now checks the signal immediately before *and* after each awaited chunk read, so an abort landing during `open()`/`stat()` (themselves RPC-priced on network volumes) no longer starts one unnecessary read. New test aborts mid-`stat` through a handle proxy whose `read` throws if ever invoked: rejects with AbortError and zero reads. Also verified to fail against the unfixed code (the read guard fired).

On the residual eager-serialization risk: agreed it's follow-up territory, not a regression here — it's the two-pass-streaming scope of SUP-595 and now noted as a known residual in the PR description.

Validation: full unit suite 9,526 passed / 0 failed (two net-new tests), typecheck clean, changed-file lint clean.

https://claude.ai/code/session_01NAEmgqH41CiCcaAUvDBs8Q

yiw190 (approved):

PR #772, opened 2026-08-16
Forward-delta cursor on GET /messages (?after=) with upsert-by-id merge (SUP-594)

iddogino:
Implements SUP-594 (child of SUP-592 session-OOM): a forward-delta cursor on `GET /messages` so live-session refetches stop re-reading and re-transferring the whole trailing page. Stacked on #771 (SUP-593) — review only the last commit; the base branch carries the throttle/abort work.

## Why upserts, not appends

New JSONL lines can **mutate** display items the client already holds: a `tool_result` attaches to a `tool_use` inside an earlier assistant item, and a streaming assistant message keeps merging content blocks (text, thinking, usage) into the same display item across lines. So the protocol returns an upserted **window**: everything at-or-after an anchor, as full current versions, spliced over the client's cached suffix by id.

## Design

**Shared definition of "settled"** (`src/shared/lib/messages-delta.ts`, used by both server and renderer):
- An unresolved tool call can only receive its result while its turn is alive; a non-queued user message ends the turn, so open calls before the last one (interrupted turns) are dead forever and never drag the anchor back.
- The trailing assistant message is always still-mutable (it may merge more streamed blocks), even behind trailing system items.
- `pickDeltaAnchor` = the item before the first still-mutable one; `mergeDeltaMessages` splices the window over the cache, returning null (→ full fetch) when the splice point is unknown.

**Server** (`getSessionMessagesDelta`): growing tail read (128 → 10,000-line cap). Anchor resolution has a free safety property: a tail window that cuts an assistant message mid-merge gives the partial item a *different* uuid, so anchors can only ever resolve against complete items — a miss forces growth. The response window widens backward defensively past still-mutable items **and** past parents of any `tool_result` recorded after the anchor entry (covers a client that anchored past a queued mid-turn message while a call was open and the result has since closed it). Anchor miss / exhausted growth ⇒ `{ messages: [], anchor: null, resync: true }` — which doubles as drift protection for file rewrites (message deletion, retention cleanup).

**Route**: `?after=` on the existing endpoint, 400 when mixed with `cursor`, reuses `annotateAndRecoverMessages` (settled input requests, subagent resolution, auth-mode senders), and answers 499 on client abort exactly like the page branch.

**Client** (`useMessages` queryFn): refetches as a delta when the cache has a pickable anchor and the last full fetch is under 60s old (`MESSAGES_FULL_REFETCH_INTERVAL_MS`; the stamp lives inside the cache entry so it's GC'd with it). `resync`, merge misses, and pre-delta servers all fall back to a full page **in the same call** — `limit` rides along with `after` so an old server answers a detectable page envelope (missing `anchor` field) instead of the legacy unpaginated array. The periodic full refetch repairs cross-client drift and re-bounds the delta-grown page; overflow slides into the existing older-history buffer. Old clients never send `after` → byte-identical behavior (all pre-existing route/page tests untouched and green).

## Validation

**Unit**: full suite 9,553 passed / 0 failed. New coverage: 13 anchor/merge tests, 9 service delta tests (late-tool-result widening, block-split merge, bounded-tail resync on a 10k+-line file, abort), 5 route tests, 6 client hook tests (delta URL, upsert merge, resync fallback, 60s cadence, old-server downgrade, splice-miss fallback). One existing test fixed: `poll-cadence.test.ts` mocks `useQuery` while `useMessages` now also calls `useQueryClient` — needed a provider wrapper.

**E2E**: 63 message-flow tests green (chat-flow, tool-rendering, message-queueing, persistence, cross-session, session-delete, thinking-display, typing-indicator, user-input-requests, awaiting-input-status).

**Browser (real Chromium against the E2E mock server)**, per SUP-594 acceptance:
- Turn refetches are all deltas: leading edge +0.10s, throttle trailing +0.85s, and the 15s poll rides the delta too with a correctly advanced anchor. Zero resyncs.
- On a 5.4MB grown transcript: the one full page load transfers 603KB (300 items) and reads ~15MB server-side (the known pagination degeneracy — SUP-595's scope); per-event deltas transfer **4.1KB / 4.6KB / 0.5KB (~150× less)** and read a **262–327KB bounded tail** regardless of file size — zero full-file reads (verified with fs-level telemetry).
- Streaming UX unchanged: streamed text visible at 340ms, final persisted line reconciled 148ms after idle; transcript, tool card, and final text render correctly.

Typecheck and lint clean. ⚠ Stacked-PR note: CI on a non-main base runs no tests — the results above are from local runs at this head.

### DIFF
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -304,6 +304,7 @@ vi.mock('@shared/lib/services/session-service', () => ({
   registerSession: vi.fn(),
   getSessionMessagesWithCompact: vi.fn(),
   getSessionMessagesPage: vi.fn(),
+  getSessionMessagesDelta: vi.fn(),
   getSession: vi.fn(),
   getSessionMetadata: vi.fn(),
   sessionExists: vi.fn().mockResolvedValue(true),
@@ -559,7 +560,7 @@ import {
   importSkillFromZip,
 } from '@shared/lib/services/skillset-service'
 import { getAgent, getAgentWithStatus, listAgentsWithStatus } from '@shared/lib/services/agent-service'
-import { listSessions, listSessionsByIds, getSessionMessagesWithCompact, getSessionMessagesPage, getSessionSummary, sessionExists, sessionBelongsToAgent, reserveSessionOwnership, sessionIsKnown, isSessionRegistered, deleteSession, getSession, updateSessionName, readSessionMetadata, updateSessionMetadata } from '@shared/lib/services/session-service'
+import { listSessions, listSessionsByIds, getSessionMessagesWithCompact, getSessionMessagesPage, getSessionMessagesDelta, getSessionSummary, sessionExists, sessionBelongsToAgent, reserveSessionOwnership, sessionIsKnown, isSessionRegistered, deleteSession, getSession, updateSessionName, readSessionMetadata, updateSessionMetadata } from '@shared/lib/services/session-service'
 import { listPendingScheduledTasks } from '@shared/lib/services/scheduled-task-service'
 import { listArtifactsFromFilesystem } from '@shared/lib/services/artifact-service'
 import { deleteNotificationsBySessionIds, getSessionIdsWithUnreadNotifications, getUnreadNotificationsByAgents } from '@shared/lib/services/notification-service'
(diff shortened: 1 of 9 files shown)


iddogino:
All six findings addressed at `e4bb2ded` (fix commit `e3eb2e78` + merge of #771's `b27580de`). Triage first, then what changed:

**1. Stale base — confirmed, fixed.** Merged `b27580de` in, so the merge candidate now carries early signal arming, the post-read abort recheck, and the generation-scoped reconcile.

**2. Queued message shields a streaming assistant — confirmed (reproduced against the real module).** One nuance: the turn-boundary loop already skipped queued users; it was the *trailing-assistant* scan that stopped at any message item, so a queued message shielded the still-merging assistant while a trailing informational didn't. Fixed: the trailing scan now skips queued users the same way it skips system items — `[u1, a1(streaming), q1(queued)]` anchors on `u1`, and the server (same shared function) re-serves `a1`'s merged blocks. Pinned on both sides: anchor unit tests + a service test with a block entry landing *after* the queued message (`streaming` → `streaming continued`).

**3. Resync leaves stale `older` items — real behavior, but pre-existing:** the slide-off buffer has stranded vanished items on *any* full refetch after a rewrite since the cursor-paging PR; deletions were always masked by the `deleted` overlay, which is why it never surfaced. That said, `resync` is a first-class rewrite signal we didn't have before, so it now clears the older buffer and the slide-off baseline (ordered before the slide-off effect so vanished items are never captured). The test now asserts `[u9]`, not `[u1,a1,u9]`.

**4. Late tool-result widening without growth — confirmed, fixed.** The server now grows the tail until every tool_result recorded after the anchor has its parent `tool_use` in view, and resyncs at the cap. A result whose parent is genuinely gone (rewritten away) in a fully-read file stays benign — no resync storm. Tested: parent 200+ lines past the initial window gets pulled into the response with its result attached; orphan case returns normally.

**5a. Deletion staleness — confirmed as a genuine regression** (pre-delta, the invalidation always full-fetched). `useDeleteToolCall` and `useDeleteMessage` now expire the cached `fetchedFullAt` before invalidating, so the refetch is authoritative. Browser-validated with the real context menu: tool card gone **489ms** after "Remove from history", post-delete request is a full page.

**5b. `olderCursor` null latch — mostly pre-existing** (#753 machinery; a deletion between two full fetches hits it identically). Deltas only lengthen the stale-cursor window. Left as a follow-up rather than folded in here.

**6. Resume-replay duplicates — confirmed, fixed.** `mergeDeltaMessages` dedupes the kept prefix against the window's ids; the replayed copy wins at its replayed position, matching what a full tail fetch shows. Unit-tested for id uniqueness and content.

**Validation on the merged head** (stacked PR still runs no CI tests — all local): full suite **9,564 passed / 0 failed**, typecheck + changed-file lint clean. Browser reruns: turn cadence unchanged (leading delta +0.08s, trailing +0.83s with advanced anchor, 15s poll delta, 536B–1KB payloads, zero resyncs), streaming text at 224ms, final reconcile 71ms after idle, and the deletion repro from the review now converges in under half a second.

iddogino:
All five confirmed and fixed at `028f7275`. Thanks for the real-agent repro — it caught exactly the case the mock can't produce (the E2E mock's tool_use message also carries a text block, so `removeToolCall` never deletes the whole entry; real agents emit tool-only assistant entries, which is why my earlier browser check passed while yours resurrected the card).

**1. Deletion resurrection (P1)** — a rewrite-mutation refetch (`fetchedFullAt === 0` marker) is now treated exactly like a resync full fetch: the response is authoritative, so the hook clears the older-history buffer and resets the slide-off baseline before the turnover effect can capture the removed entry. The test you flagged now resolves the full response omitting the item and asserts the rendered list is exactly what the server returned — no `older` resurrection.

**2. Replay reordering (P1)** — agreed, the client knows more than the server window here. `mergeDeltaMessages` now upserts replayed ids in place at their cached chronological positions and appends only genuinely new ids: `[u1,a1,u2] + [u2,u1,a1,a2]` → `[u1,a1,u2,a2]`. The test that codified the reordering now pins the corrected order (and that the window's copy wins on content).

**3. fetchOlder across resync (P2)** — added a generation fence: session switches and resyncs bump a generation counter, and an older-page response from a previous generation is discarded before any state commit (no stale history, no `olderCursor = null` latch). New test: page in flight → resync lands → stale page resolves → list stays at the post-resync transcript and `hasOlder` reflects the new cursor state.

**4. Resync loop on beyond-cap orphans (P2)** — the unmatched-parent check now only forces growth while growth is possible. At the cap the delta proceeds without the parent's upsert: a parent deeper than the 10k-line tail is also deeper than any anchor a client could still hold open, so skipping the upsert is a bounded staleness, while resyncing there made every poll a full fetch — the exact cascade this protocol exists to prevent. New test: 10,200-line transcript + trailing parentless result → normal delta, no resync.

**5. Author query bound (P2 perf)** — `annotateAndRecoverMessages` now filters by `sessionId` AND `inArray(messageAuthor.id, userMessageIds)`, so a delta loads author rows only for the handful of user messages it returns. (The legacy unpaginated branch keeps its inline copy — that whole path is reworked in SUP-595.)

**Validation at `028f7275`** (stacked PR, CI runs no tests — all local): full suite **9,566 passed / 0 failed**, typecheck + changed-file lint clean. Browser (mock server): delta cadence unchanged (leading/trailing/poll all deltas, 536B–1KB, zero resyncs, final text 405ms), tool-call deletion converges in ~1.3s via a full refetch. The tool-only entry case rides on the strengthened hook test; a real-agent re-check of the deletion flow on your harness would be a welcome confirmation.

iddogino:
Two of the four addressed at `cefed8bb`; the other two I'm pushing back on with evidence.

## Fixed

**Cross-client deletions (P2)** — took this one even though the mechanism predates the PR (the slide-off buffer has retained rewrite-vanished items on every full refetch since the cursor-paging PR; local deletes were masked by the `deleted` overlay). The slide-off effect now distinguishes positionally: an item is genuine turnover only if it sat *before* the new page's first item in the previous page; anything missing from within the still-covered range was rewritten away and is dropped. This also closes a latent variant nobody had hit yet — a delta window dropping a rewritten suffix item used to slide it into `older` too. When the page advances wholesale between fetches the ambiguity resolves toward turnover, as before. Tests: remote-deletion drop, and legit turnover still sliding.

**Hanging fetchOlder (P2)** — agreed, the fence stopped the commit but not the request. Generation resets (session switch, resync) now abort the in-flight older-page request outright; the existing `finally` clears `isFetchingOlder` immediately and the abort rejection is recognized rather than reported. The resync test now asserts the abort fires, the busy flag clears, and fresh paging starts immediately against the post-resync cursor.

## Pushing back

**Replay reordering across page stores (P1)** — the core scenario doesn't reproduce. I built a transcript with the *entire* 40-turn history replayed verbatim plus fresh turns and ran the real page reader: it returns canonical chronological order (`u-37,a-37,…,u-41,a-41`), no replayed positions, no duplicates. The reason is an interaction the finding missed: uuid-dedup keeps the *first* occurrence in the window, and a duplicate-heavy window keeps `transformed.length ≤ limit`, which forces the page reader's growth loop to pull the originals into view before it will serve a page. So the periodic full refresh reconstructs canonical order for any transcript within the reader's 50k-raw-line cap — the reorder requires originals beyond 50k lines, a corner every paging path in the app shares. The `older`-buffer sub-claim is true but is the buffer's pre-existing frozen-snapshot property: no refetch, delta or full, has ever updated items after they slide off; that's independent of this PR.

**Delta cap conflation (P2)** — reachability requires all of: an input-request call stamped settled by annotation (so the client anchors past it), sitting >10,000 *raw* lines back, while still inside the client's 300-item display page. Even then the failure is one tool call showing its settled placeholder instead of the final result text, and the next 60-second full fetch (50k window) repairs it — bounded, cosmetic staleness in an extreme geometry. Resyncing at the cap instead would convert a *persistent* orphan into a full fetch on every poll, which is the cascade this protocol exists to prevent — and persistent orphans are the more plausible input, even though today's rewrite paths (`removeToolCall` deletes result and parent together) make both rare. The cap-rescan cost for a transient orphan is real but bounded (~20k lines once per poll while it persists) and I'd rather keep the simple rule than special-case it; happy to revisit if a real transcript ever hits it.

## Validation at `cefed8bb`

Full suite **9,568 passed / 0 failed**, typecheck + changed-file lint clean. Browser (mock): turn cadence unchanged (all deltas, 536B–1KB, final text 410ms), tool-call deletion converges in 659ms via a full refetch.

yiw190 (approved):

PR #774, opened 2026-08-16
fix(x-agent): cap sync waits at 120s and promote to async on timeout

iddogino:
## Problem

A sync `invoke_agent` call against a slow target failed after exactly ~5 minutes with `Network error calling x-agent invoke: fetch failed`, even though the invoked session was running fine. The calling agent then re-invoked, spawning a duplicate run — a retry storm in the making.

Root cause: the sync paths in `/api/x-agent/invoke` and `/api/x-agent/get-transcript` hold the HTTP response open with **zero bytes written** until `waitForIdle` resolves (default 10 min), while the container's bare `fetch` (Node ≥22 / undici) aborts any request whose response headers don't arrive within **300 s**. Any target turn longer than 5 minutes was guaranteed to surface as a retry-inducing network error, and since the host never sees the disconnect, the original run kept going — so every retry duplicated the whole job.

## Fix

Sync is meant for fast turns, so a slow turn now promotes to the async contract instead of erroring:

- `waitForIdle` rejects its timeout branch with a typed `WaitForIdleTimeoutError` so callers can tell "target still working" apart from real failures.
- Both `/invoke` sync branches cap the wait at **120 s** (`SYNC_WAIT_TIMEOUT_MS`, env-overridable) — safely under the 300 s undici cliff. On timeout they return `{sessionId, status: 'running'}` with explicit guidance in the existing `error` field: *do NOT re-invoke, poll `get_agent_session_transcript`*. Reusing the `error` field keeps already-deployed containers compatible (they render it as `note:`).
- `/get-transcript` sync becomes a bounded long-poll: same 120 s cap; on timeout it returns the transcript-so-far with `status: 'running'` (200) so the caller can call again, instead of a 504. Non-timeout failures still 504.
- The route matches the timeout by `error.name` rather than `instanceof`: many test suites mock the persister module wholesale, and `instanceof` against a missing mock export throws. The persister's own test pins the class name so the contract can't drift silently.
- Container tool descriptions (`invoke_agent`, `get_agent_session_transcript`) now document the ~2-minute cap, the promotion behavior, and the poll-don't-re-invoke rule; also fixed a stale `get_session_transcript` tool-name reference.

The host-side cap alone already eliminates the "fetch failed" mode for containers built before this change, since no sync response can outlive 300 s anymore.

## Testing

- `x-agent.test.ts` 65/65: new tests for both invoke promotion paths (guidance text + `timeoutMs: 120_000` cap asserted), get-transcript timeout long-poll (200 + transcript-so-far), and the non-timeout 504 path. Timeout errors in tests are built by name, not imported, so they exercise the wire contract independent of the fix.
- `message-persister.test.ts` 272/272: timeout rejection now asserted to carry the `WaitForIdleTimeoutError` class and name.
- `npx tsc --noEmit` and eslint clean on touched files.

https://claude.ai/code/session_015mcU8vDZ7D6BkixhPnDNpr

### DIFF
--- a/agent-container/src/tools/agents/get-session-transcript.ts
+++ b/agent-container/src/tools/agents/get-session-transcript.ts
@@ -17,7 +17,7 @@ export const getSessionTranscriptTool = tool(
   'get_agent_session_transcript',
   `Read the message transcript of a session belonging to another agent. Returns a status line ('running' | 'idle' | 'awaiting_input') followed by the messages.
 
-If sync=true and the session is currently running, the tool waits until the target agent's turn is complete before returning. Otherwise it returns the current transcript immediately.
+If sync=true and the session is currently running, the tool waits up to ~2 minutes for the target agent's turn to complete before returning. If the turn is still in progress after that, it returns the transcript so far with status 'running' — call again with sync=true to keep waiting. Otherwise it returns the current transcript immediately.
 
 Tool calls in the transcript are summarized — the raw tool input/output is omitted to keep the result compact.`,
   {
--- a/agent-container/src/tools/agents/invoke-agent.ts
+++ b/agent-container/src/tools/agents/invoke-agent.ts
@@ -14,9 +14,9 @@ export function makeInvokeAgentTool(getCallerSessionId: () => string) {
     'invoke_agent',
     `Send a message to another agent in this workspace.
 
-If session_id is omitted, a new session is started on the target agent. If session_id is provided, the message is appended to that existing session — the session must exist and not currently be running (use get_session_transcript with sync to wait).
+If session_id is omitted, a new session is started on the target agent. If session_id is provided, the message is appended to that existing session — the session must exist and not currently be running (use get_agent_session_transcript with sync to wait).
 
-If sync=true, the tool waits for the target agent's turn to finish and returns its last message. If sync=false (default), the tool returns immediately with status 'running' and you can later read the transcript with get_session_transcript.
+If sync=true, the tool waits up to ~2 minutes for the target agent's turn to finish and returns its last message. If the target is still working after that, the call returns status 'running' with the session_id — the invocation was delivered and is in progress; do NOT re-invoke (that would start a duplicate run), poll get_agent_session_transcript instead. If sync=false (default), the tool returns immediately with status 'running' and you can later read the transcript with get_agent_session_transcript.
 
 Use list_agents first to discover available slugs.
 
--- a/src/api/routes/x-agent.test.ts
+++ b/src/api/routes/x-agent.test.ts
@@ -126,6 +126,21 @@ function waitForIdleTimeoutError(): Error {
     name: 'WaitForIdleTimeoutError',
   })
 }
+
+// Sync waits pass the REMAINING end-to-end budget (deadline stamped at handler
+// entry minus pre-wait work), so the exact timeoutMs varies — assert the shape
+// and that it stays within (0, 120s]. requireActiveFirst is off because sync
+// callers mark/observe the session active before waiting, so "inactive" means
+// the turn already finished.
+function expectBoundedSyncWait(sessionId: string): void {
+  expect(mockWaitForIdle).toHaveBeenCalledWith(sessionId, {
+    timeoutMs: expect.any(Number),
+    requireActiveFirst: false,
+  })
+  const opts = mockWaitForIdle.mock.calls.at(-1)?.[1] as { timeoutMs: number }
+  expect(opts.timeoutMs).toBeGreaterThan(0)
+  expect(opts.timeoutMs).toBeLessThanOrEqual(120_000)
+}
 const mockIsSessionActive = vi.fn((_sessionId?: string): boolean => false)
 const mockIsSessionAwaitingInput = vi.fn((_sessionId?: string): boolean => false)
 const mockWaitForIdle = vi.fn(async (..._args: unknown[]) => {})
@@ -184,7 +199,7 @@ vi.mock('@shared/lib/error-reporting', () => ({
 // Imports (after mocks)
 // ----------------------------------------------------------------------------
 
-import xAgentRoute from './x-agent'
+import xAgentRoute, { resolveSyncWaitTimeoutMs } from './x-agent'
 
 // ----------------------------------------------------------------------------
 // Test app + helpers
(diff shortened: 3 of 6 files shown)


iddogino:
Pushed `9e71372b` addressing all three findings:

1. **[P1] End-to-end deadline** — `syncDeadline` is stamped at handler entry in both `/invoke` and `/get-transcript`, before policy review / container startup / session creation. A new `waitForTurnWithinBudget(sessionId, deadline)` helper passes only the *remaining* budget to `waitForIdle`, and an already-exhausted budget promotes to async immediately without waiting (the prompt was delivered; only the wait is skipped). New test simulates 150s of pre-delivery work via a `Date.now` spy advanced inside `sendMessage` and asserts promotion with `waitForIdle` never called.

2. **[P2] Override validation** — `resolveSyncWaitTimeoutMs()` rejects non-finite / non-positive values (falls back to 120s) and clamps everything to 240s, keeping any override under the 300s transport cliff. Exported and unit-tested (`undefined`/`''`/garbage/`0`/negative/`Infinity`/`NaN` → 120s; `300000`/`600000` → 240s).

3. **[P2] Fast completion** — sync waits now pass `requireActiveFirst: false`. Both invoke paths mark the session active before delivering the prompt (and get-transcript checks `isSessionActive` immediately before waiting), so an inactive session at wait time means the turn already finished — it now resolves as `completed` instead of rejecting "never became active" and mislabeling the result `running`. This also fixes the latent get-transcript race where a session idling between the active-check and the wait produced a 504.

Local verification: `x-agent.test.ts` 69/69 (4 new), `message-persister.test.ts` 272/272, `tsc --noEmit` and eslint clean.

iddogino:
Pushed `9ecc0b7a` addressing the four remaining findings:

1. **[P1] Delivery cutoff** — a second clock, `deliveryCutoff` (handler entry + 240s), is checked immediately before the two delivery points: `sendMessage` on the existing-session path and `createSession` (whose `initialMessage` *is* the prompt) on the new-session path. Past the cutoff the handler returns 504 with "the prompt was NOT delivered. It is safe to retry; prefer sync=false" — so a caller whose fetch died at 300s never gets a ghost run to duplicate. It applies to **async invokes too**, since they also respond only after delivery. Scope note: the container delivery calls themselves aren't cancellable, so enforcement is a boundary check rather than mid-call cancellation; the 60s margin to the 300s cliff covers the delivery call's own duration. Two new tests advance a `Date.now` spy inside `ensureRunning` and assert 504 with `sendMessage`/`createSession` never called.

2. **[P2] Exhausted budget checks activity** — `waitForTurnWithinBudget` with no remaining budget now returns `isSessionActive ? 'timeout' : 'completed'`. The flagged exhaustion test now marks the session active statefully (mirroring the real persister), and a new companion test has the turn finish during slow delivery and asserts `status: 'completed'` with the correct reply.

3. **[P2] Turn-correlated replies** — the existing-session invoke path captures the last assistant entry's uuid immediately before delivering the prompt; `readLastAssistantMessage` keeps polling while that boundary entry is still last, instead of returning the previous turn's answer. Sync get-transcript does the same: boundary captured before waiting, and on completion it reconciles (bounded, ~5s) until a newer assistant entry is flushed, so `idle` responses can't ship a transcript missing the reply they waited for. Two new tests stage the flush race (boundary read → stale read → flushed read) and assert the fresh answer wins on both paths.

4. **[P3] Docs vs cap** — resolved by capping rather than documenting: the env override can now only *shorten* the 120s default (`Math.min(parsed, 120_000)`), so the tools' "up to ~2 minutes" promise stays accurate. Resolver unit tests updated.

Local verification: `x-agent.test.ts` 74/74 (+5 new, 2 updated), `message-persister.test.ts` 272/272, `tsc --noEmit` and eslint clean.

iddogino:
Pushed `7ddd4502` for round 3. Two findings fixed in full, one fixed in half with a deliberate decline of the other half:

**1. [P1] Existing-session cutoff — fixed.** The check now sits AFTER every unbounded pre-send await: boundary read → bounded stream attach → **cutoff check** → `markSessionActive` → attribution insert (~ms local SQLite) → `sendMessage`. The root unboundedness was the persister's WebSocket `ready` promise (no timeout of its own), so the attach is now wrapped in a 30s bound (`X_AGENT_SUBSCRIBE_TIMEOUT_MS` for tests); pre-delivery a stall fails fast with nothing sent.

**2. [P1] New-session end-to-end — fixed via revocation rather than deadline plumbing.** `createSession` *is* the delivery and cannot be cancelled mid-flight, so it's raced against the cutoff: if it resolves late — when the caller's fetch is certainly dead — the session is deleted the moment it materializes (same cleanup as the failed-registration path), so a retry cannot duplicate a ghost run. Post-delivery, the only unbounded await was again the stream attach; on timeout there the session is likewise revoked (unsubscribed = nothing persists its transcript = invisible ghost). The remaining post-delivery awaits (ownership, registration, metadata) are local DB/file writes — I deliberately did not thread a deadline through them; bounding the two genuinely unbounded operations covers the failure mode without plumbing a deadline through fast local work. Tests: late-create revocation (create resolves 120ms past a ~50ms-remaining cutoff → 504 + `deleteSession('late-sess')` awaited via `vi.waitFor`), attach-stall on both paths, attach-eats-the-cutoff on the existing path.

**3. [P2] Idle-before-flush — the timeout-gap case is fixed; the cleared-before-check case is declined with rationale.** Reconciliation now also runs when the wait returned 'timeout' but the session is inactive by the time we'd read status — same flush race as a completion, and we observed the activity in-request. The other variant (activity clears just before the `isSessionActive` check) is not soundly fixable with the suggested barrier: the transcript JSONL is written by the CLI process inside the container — the host only reads the file and has **no write queue to barrier on**. The only host-side mechanism is boundary polling, and running it whenever a response is "about to report idle" charges up to the full ~5s poll budget to every sync read of a genuinely idle session (the boundary never changes, so the poll always runs to exhaustion). A caller in that sliver gets plain read-what's-flushed semantics and picks up the entry on its next poll; the harmful variant — "I waited for this turn and was told idle without its reply" — is covered by the completed path and this fix. If this sliver matters in practice, the sound fix is a `wentIdleAt` stamp in the persister's streaming state to gate reconciliation on recent idleness; happy to do that as a follow-up if you want the extra persister surface.

**Follow-up noted, not in this PR:** client-supplied invocation idempotency keys would make retries safe by construction for this entire class, but require container-tool, wire, and storage changes.

Local verification: `x-agent.test.ts` 79/79 (+5 new), `message-persister.test.ts` 272/272, `tsc --noEmit` and eslint clean.

PR #783, opened 2026-08-17
Two-pass streaming messages page with byte budget (SUP-595)

iddogino:
Implements SUP-595 (parent SUP-592): makes `GET /messages` pagination structurally un-OOM-able server-side. Same JSON shape, chunked transfer — invisible to clients. Follows #771 (SUP-593) and #772 (SUP-594).

## Why the old reader blew up

`getSessionMessagesPage` sized its tail window in lines with a message-count goal, but the raw-line→display-item ratio (~0.18 on real transcripts) made the `maxLines *= 2` loop escalate until it re-read the **whole file from EOF, repeatedly**, materializing every chunk, parsed entry, and transformed item — then serialized the response as one multi-MB `c.json` string. Measured: ~50–150MB transient per in-flight request; OOM-killed at 14–15 concurrent requests in a 1GB cgroup.

## Design

**Pass 1 — backward index scan** (`scanMessagesPageWindow` + new `iterateJsonlLinesBackward`). Walks backward from EOF one line at a time, parsing each transiently to count display-item starts, retaining only offsets and id sets. The classifier (`countsAsDisplayStart`) mirrors every rule `transformMessages` uses to decide an entry starts a new item: uuid dedup, assistant merge by `message.id`, meta / tool-result-only / task-notification / compact-summary skips, queued-command attachments, the informational banner's synthetic user copy. The scan stops at `limit+1` items (one sacrificial head item, same `dropPartialHead` contract as before), the **byte budget**, `MAX_TAIL_LINES`, or file start. The doubling loop is gone — one scan, extended incrementally, never re-read.

Cursor pages locate the cursor by **raw byte search** for the quoted id (no JSON parse above the cursor; candidates are verified by parsing, which also catches queued-command cursors whose id lives in `source_uuid`). A cursor 17k raw lines deep resolves in 63ms.

**Pass 2 — bounded materialization.** One deliberate deviation from the ticket's sketch: instead of rewriting the transform as a streaming operator, the byte budget bounds the window and the **unchanged `transformMessages`** runs over just that window. Page semantics are preserved by construction — all nine pre-existing pagination tests passed untouched on first run. Slicing logic is byte-for-byte the old code.

**Byte budget** (`MESSAGES_PAGE_BYTE_BUDGET`, 4MB): a hard cap on raw bytes a window may materialize, with a ≥2-item floor so a single item larger than the budget still serves rather than producing an empty page. Truncated pages are valid short pages: `nextCursor` points at their first item and the client's existing scroll-up paging walks the rest.

**Streamed envelope**: the route emits `{"messages":[…],"nextCursor":…}` item-by-item through `Readable.from` with real backpressure instead of one monolithic JSON string. (Chunks must be Buffers — `Readable.toWeb`'s byte consumers reject strings; a route test caught this immediately.)

**Untouched**: the `?after=` delta path (already bounded, SUP-594) and the legacy unpaginated branch (per ticket).

## Documented deviations from the old reader

- A cursor page's window extends 512KB (`CURSOR_WINDOW_GRACE_BYTES`) past the cursor line so tool_use blocks at the page's newest edge still meet their results (results land within the same turn, typically adjacent). The old code read from the cursor to EOF; a result further out than the grace region is not attached — bounded cosmetic staleness on a historical page, never on the trailing page (whose window always ends at EOF). Pinned by test: a result recorded past the cursor line (queued message interleaved between use and result) still attaches.
- Backward-order counting can place an item boundary on a different line than the forward transform for replayed/duplicated ids; the shift lands on the sacrificial head item that page slicing already drops. The mixed-transcript parity test walks tiny budget-truncated pages across every classifier rule and reconstructs the full transform's output exactly.

## Acceptance

**Memory** — 37MB transcript, 16 concurrent `limit=300` requests:
- In-process probe: peak **+10MB heap / +10MB external for all 16 combined**, 50ms wall (baseline: 50–150MB *per request*, OOM at 14–15 concurrent).
- Live dev server burst: peak RSS 274MB from a 111MB baseline ≈ **~10MB transient per in-flight request** (target was ≤~15MB).

**Content parity** — full unit suite **9,599 passed / 0 failed**, including all pre-existing pagination tests unmodified. 31 new tests: backward iterator (offsets, chunk-spanning lines, abort), ranged `streamFileLines`, budget truncation + full cursor-walk reconstruction, giant-item floor, mixed-transcript parity walk, grace-region result attachment, depth-cap/vanished-cursor termination.

**Byte-budget paging** — `byteBudget: 1024` against a 60-item thread: every truncated page returns a valid `nextCursor`, and the walk reconstructs the exact full item sequence.

**Browser** (mock server, real Chromium): initial 300-item page renders from the streamed endpoint (603KB, 32ms); scroll-up prepends cursor pages (200-item cap via `capMessagesPageLimit`, contiguous, no gaps); a live turn still runs entirely on SUP-594 deltas (~4KB each) with the final text rendered. 13 chat/messages E2E specs green.

⚠ Base is `main`, so CI runs the full suite on this PR.

### DIFF
--- a/src/api/routes/agents.ts
+++ b/src/api/routes/agents.ts
@@ -1824,6 +1824,19 @@ async function annotateAndRecoverMessages(
   }
 }
 
+// Buffers, not strings: Readable.toWeb hands chunks straight to the web
+// ReadableStream, whose byte consumers reject non-Uint8Array chunks.
+function* messagesPageJsonChunks(page: {
+  messages: TransformedItem[]
+  nextCursor: string | null
+}): Generator<Buffer> {
+  yield Buffer.from('{"messages":[')
+  for (let i = 0; i < page.messages.length; i++) {
+    yield Buffer.from((i === 0 ? '' : ',') + JSON.stringify(page.messages[i]))
+  }
+  yield Buffer.from(`],"nextCursor":${JSON.stringify(page.nextCursor)}}`)
+}
+
 // Unpaginated path stays inlined so the stream-pipe PR can still land on `return c.json(transformed)`.
 // GET /api/agents/:id/sessions/:sessionId/messages - Get messages for a session
 agents.get('/:id/sessions/:sessionId/messages', AgentRead(), async (c) => {
@@ -1884,7 +1897,14 @@ agents.get('/:id/sessions/:sessionId/messages', AgentRead(), async (c) => {
       c.req.raw.signal.throwIfAborted()
       await annotateAndRecoverMessages(page.messages, agentSlug, sessionId)
       c.req.raw.signal.throwIfAborted()
-      return c.json({ messages: page.messages, nextCursor: page.nextCursor })
+      // Serialize item-by-item instead of one JSON.stringify of the whole
+      // envelope: pages hold multi-MB tool results, and a monolithic response
+      // string was one of the transients that made concurrent page fetches
+      // OOM the process. Readable.from pulls lazily, so writes see real
+      // backpressure from the socket.
+      return c.body(Readable.toWeb(Readable.from(messagesPageJsonChunks(page))) as ReadableStream, 200, {
+        'Content-Type': 'application/json',
+      })
     }
 
     const messages = await getSessionMessagesWithCompact(agentSlug, sessionId)
--- a/src/shared/lib/services/session-service.test.ts
+++ b/src/shared/lib/services/session-service.test.ts
@@ -1148,9 +1148,9 @@ describe('session-service', () => {
     it('hard cap bounds a huge non-display gap and still serves the trailing item', async () => {
       // Tool-result-only rows are not display items, so the two-item budget
       // floor alone would scan through an arbitrarily large run of them. The
-      // hard cap stops the scan unconditionally; the trailing item (complete,
-      // single-entry) must still be served rather than sacrificed, and the
-      // walk must terminate rather than error on the unreachable gap.
+      // hard cap stops the scan once its one-servable-item floor is met; the
+      // trailing item (complete, single-entry) must be served rather than
+      // sacrificed, and cursor paging must cross the gap to older history.
       const ts = (s: number) => new Date(Date.UTC(2026, 0, 8, 0, 0, s)).toISOString()
       const rows: object[] = [
         ...makeThread(2),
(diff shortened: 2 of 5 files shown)


iddogino:
All seven findings verified with repros against the previous head and addressed at `bf95038a`. Two turned out worse than reported.

## P1s

**Replayed cursors cycle forever** — confirmed exactly as described: the 6×2 replay with limit=3 cycled `a1 → u0 → a1` indefinitely (27 duplicate items in 9 hops before I stopped it). Cursor resolution is now canonical: the scan keeps byte-searching below the first match and re-anchors on every deeper verified occurrence, matching the transform's oldest-occurrence uuid dedup. The repro now terminates in two pages with each item returned exactly once — identical to the parent implementation. Cost: cursor scans now always run to file start or the 50k-line cap; on the 37MB probe transcript that put a shallow cursor page at ~51ms (was ~8ms) and left the deep page unchanged at 62ms. Correctness over speed; the memchr-only search keeps it flat.

**Byte budget not a hard bound** — confirmed by inspection; the two-item floor had no ceiling behind it. There is now an unconditional hard cap at 2× the budget (8MB default). To keep the cap from producing empty pages in the geometry that motivated the floor (a small trailing item behind a huge tool-result run), the sacrificial-head rule became precise instead of blanket: the head item is only dropped when it can actually be partial — an assistant merge group whose older entries may lie below the window start. Single-entry items (user, system, id-less assistant) are complete wherever the window cuts, so a complete trailing item behind a >cap gap now serves instead of being sacrificed into an empty page. History behind a gap larger than the hard cap is unreachable through paging — the accepted trade-off for a structural bound; the old code's answer to that geometry was materializing it.

**Quadratic carry** — confirmed: an 8MB newline-free row cost 80ms of copying. Fragments now accumulate in an array with one concat at the line boundary (the forward reader's existing pattern); same row now 9ms, linear.

**Grace truncates results** — confirmed: a 600KB result row after a queued-command cursor came back unattached. The scan now records line-start offsets while searching, and the window end is the first line start at or past the grace distance — any row *beginning* inside the 512KB grace region is included whole, regardless of its size. The repro attaches the full 614,400-byte result.

## P2s

**Reordered system items** — confirmed and worse than reported: the `info → recall → boundary → info2` walk didn't just return the wrong item, it lost **three of five** items (`u0`, `mr`, `info1`) and terminated early through an empty page. Two compounding bugs fixed: counting stops on a system entry now extend through the contiguous system run (so the transform's recalls→boundaries→informationals reordering happens entirely inside one window), and `reachedStart` is now derived from the window actually touching offset 0 rather than from which stop condition fired. The walk now returns all five items in display order and terminates.

**Short reads** — confirmed by inspection, and worth noting it predates this PR: `readJsonlTailLines` has carried the identical single-read pattern since the cursor-paging PR, feeding the delta reader on the same NFS deployments. Both now fill each chunk range in a loop; a fill that comes up short (concurrent truncation) stops the walk cleanly instead of splicing lines across the gap. Pinned with a 3-byte-capped read mock over both readers.

**Abort within a large row** — confirmed. `streamFileLines` now takes the signal and checks it per chunk, so a multi-MB row no longer defers cancellation until it fully materializes; `readEntriesRange` threads the request signal through.

## Validation at `bf95038a`

Full suite **9,605 passed / 0 failed** — all pre-existing pagination expectations unchanged, 6 new regression tests (replay walk termination, system-run walk completeness, grace-region oversized result, hard-cap gap behavior, short-read masking, per-chunk abort). Memory probe on the 37MB transcript: 16 concurrent limit=300 pages in 48ms at ~4.4MB combined transient per request. HTTP smoke on the live server: trailing page, cursor page, and `?after=` delta all correct against the 37MB session.

On the process note: agreed that SUP-597's cost/bytes-read safety net remains open — these regression tests narrow the pagination-behavior gap but assert no read-cost invariants.

iddogino:
All six verified with repros against `bf95038a`; five fixed at `117d8807`, one accepted with the bound documented. Findings 2–4 reproduced exactly as described.

## Fixed

**Empty capped tail (P1)** — reproduced: two visible messages behind a 10KB trailing result run returned `[]`/`null`. The hard cap now has a floor mirroring the budget's: it cannot settle until the window holds at least one complete servable item (two when the deepest item is a merge group the head-drop would sacrifice). A turn's freshly written tool-result run at EOF — the transient geometry a live degenerate session produces constantly — can no longer blank the session. Repro now returns `[u0, A-use]` with the results attached, matching the parent.

**Vanishing cursor from a split merge group (P1)** — reproduced exactly: `[A1, N]` served with `A1` a partial duplicate of `A0`'s item, then cursor `A1` → terminal `[]`, with `old`/`A0`/`Q` permanently unreachable. Your diagnosis was right that classifying only the deepest item is insufficient. The fix is stronger than patching the flag: **no stop condition may fire while a merge group is open.** The scan tracks the open group's `message.id` through interleaved queued/system/filtered rows and only allows target, budget, or cap stops once a turn boundary proves the group closed (a different assistant response or a non-queued user row) — bounding the overshoot by one turn's span. The repro now pages `[Q, N]` → `[old, A0]` with `A0` fully merged (`part one part two`), canonical order, clean termination.

**Filtered rows splitting system runs (P2)** — reproduced: `info1` permanently lost behind a meta row. Extension now judges adjacency the way the transform does: rows removed before ordering (meta, non-display entry types, malformed lines, compact summaries, informational synthetic copies) continue the run; rows that act as transform anchors — including tool-result-only user rows, which genuinely split the attachment grouping — still settle it. The repro walk now returns all five items in display order.

**Cancellation holes (P2)** — the refill loop observes the signal before every physical read (pinned: abort during the first short read stops after exactly one read), and the forward reader rechecks the signal at EOF before concatenating an unterminated trailing row.

**Double retention on final yield (P2)** — carry fragments are released before the yield, same pattern as `streamFileLines`' pending buffer.

## Accepted: oversized rows exceed the cap (P1)

Confirmed — a 64KB row inflates the selected range past the declared 2KB cap. Accepting this one deliberately: classifying a row requires assembling it, so peak memory is ≥ one row under any windowing scheme, and any row inside a served item must be materialized by every implementation that ever renders it — the pre-cursor-paging code, the parent, and all alternatives included. The honest per-request bound is now documented at the constant: **O(hardCap + one turn's span + largest single row)**, where each exception term is content that must be held to serve anything at all. Amplification is ~3× the largest single row, versus ~3× the whole file before this PR. The real fix for multi-MB embedded payloads is moving them behind references — the SUP-596 follow-up, which this bound is the argument for.

## Validation at `117d8807`

Full suite **9,609 passed / 0 failed**, twice consecutively (one interleaved run under heavy load bailed with 88 skips — the two complete runs are authoritative). 6 new regression tests: capped-tail visible history, split-merge-group walk with full-merge assertion, meta-split system-run walk, refill-loop abort, plus the round-1 hard-cap test updated for its improved semantics — the walk now *crosses* the capped gap (serving the items behind it) instead of terminating at it. All pre-existing pagination expectations unchanged.

iddogino:
Findings 2–4 fixed at `3b817cd6`; finding 1 I'm holding as the documented bound, with a concrete follow-up proposal.

## Fixed

**Capped gap + reordered system run terminates paging (P1)** — reproduced exactly: cursor `mr` returned `[]`/`null` with the three older items permanently hidden. The mechanism was that raw rows below a *system-item* cursor can transform to display-**after** it (recalls sort before boundaries before informationals), so the count was satisfied by items that contribute nothing to the page. Counting below a system cursor is now suppressed until the first anchor row past the cursor's run — every item counted after that provably displays before the cursor (the run attaches above it), so the page makes progress whenever older history exists, and empty/`null` only remains for a cursor that genuinely has nothing below. The repro now walks `[mr, cb, info, new-u, new-a]` → `[old-a, A-use]` → `[old-u]`, complete and terminating.

**Cap cuts a heterogeneous system run (P2)** — I could not reproduce the permanent loss (two parameterizations of your fixture self-healed through subsequent pages in my hands), but the mechanism is plainly present: the extension's cap guard could bisect a run, which round 1 already established as loss-capable. Fixed at the mechanism: run/filtered extension is now cap-exempt (bounded by the run's span — system rows are small, same rationale as the merge-group exemption), and hard-cap stops route through extension instead of settling directly, so no cap ever places a boundary inside a reorderable group. Pinned with a property test walking the sub-cap-rows/super-cap-run fixture at three (limit, budget) combinations against the full transform.

**Repeated compact boundaries over-counted (P2)** — reproduced exactly: `[c4, N]` then terminal, `U`/`A` lost. Two halves: the system-cursor suppression above already repairs the walk (cursor `c4` is a system cursor), and counting now mirrors the transform's newest-wins collapse — a compact boundary only counts when no other boundary has been counted since the last anchor row, with recalls/informationals correctly NOT breaking the span (they aren't message entries, so they don't reset the anchor key). The repro serves `[c4, N]` → `[U, A]` → done.

## Holding: multi-row gaps exceed the aggregate cap (P1)

The 122KB selection for 100 hidden rows is the one-servable-item floor introduced for your round-2 finding, and it is term "(a)" of the bound documented at the constant: **O(hardCap + one turn's span + largest single row)**. The reason I'm not converting it to a mid-gap continuation: those hidden rows are the tool results of the display item below the gap, and they attach inline to that item on the wire — any request that *serves* the floor item must materialize its turn. A continuation that stops mid-gap can only ever emit that item with silently missing results; it bounds the request by degrading the content, not by avoiding the cost.

That said, the direction is cheaper than it looks and worth ticketing: `nextCursor` is already fully opaque to the client (it round-trips it without interpretation), so an offset-carrying continuation token needs zero client changes — it's the "offset cursor" follow-up already noted on the cursor-paging PR, and it pairs with SUP-596's media refs, which remove the giant-gap geometry at the source rather than paging through it. Happy to file both as one follow-up if you want the aggregate cap made airtight.

## Validation at `3b817cd6`

Full suite **9,612 passed / 0 failed**; typecheck + lint clean. 3 new regression tests (system-cursor gap walk asserting full-transform equality, cap-exempt run property walk, boundary-collapse paging); all 138 pre-existing pagination/session tests unchanged.

iddogino:
Finding 1 fixed at `60e5ae5e`; finding 2 accepted as a known replay-geometry corner, logged for the ticketed follow-up.

## Fixed: duplicate replay anchors reset the boundary span (P2)

Reproduced exactly — `[c4, X]` then cursor `c4` → `[]`/`null`, `U`/`A` permanently hidden. Your diagnosis and suggested fix were both right: the transform keys system-item anchors **by uuid**, so a byte-identical replayed row is the same anchor as its original, not a new one. The duplicate check now runs before the classifier registers the row's uuid, and a duplicate neither resets the boundary-collapse span nor lifts system-cursor suppression. The repro walks `[c4, X]` → `[A]` → `[U]`, complete and terminating.

One honest residual, inherent to backward scanning: the *newest* replay copy can't be recognized as a duplicate (its original lies deeper, unseen), so one boundary between it and the true anchor can still over-count. I traced this through the fixture — the only effect is a shorter page (the walk above pages 1-at-a-time instead of 2), never a lost item or a terminal empty page; a larger limit returns the exact slice. Short pages are already valid protocol.

## Accepted: window-local replay dedup repositions system items (P2)

Confirmed as a mechanism, with a data point worth recording: my parameterization of your exact `[U,A,C]×2` fixture produces the *correct* result on the current head — pages `[U, A]` then `[C]`, concatenating to precisely the full transform's `[C, U, A]` with nothing lost. The failure is window-geometry-sensitive; in your layout the boundary marker renders at the wrong position, in mine it doesn't. Either way the invariants hold: no item is lost, no walk terminates early — it's positional cosmetics for system markers in verbatim-replayed transcripts.

I'm accepting it because the real fix is architectural, not a patch: canonical anchor resolution *across* windows means either carrying anchor context between requests or extending windows to cover duplicate chains that can legally sit 50k lines apart. That's the same design space as the opaque offset-continuation proposed for the aggregate-cap finding, and it's where this belongs — designed once, deliberately, in the follow-up ticket alongside SUP-596's media refs. It's also the same replay-position-approximation class knowingly accepted on the delta side in #772 (frozen older-buffer snapshots, replay ordering at the periodic full).

## Validation at `60e5ae5e`

Full suite **9,613 passed / 0 failed**; typecheck + lint clean; the dup-anchor walk promoted to a regression test asserting full-transform equality and termination. All prior pagination regressions unchanged.

With this, every terminal-loss and memory-bound finding across four review rounds is closed; the two open items (aggregate-cap continuation, cross-window replay canonicalization) are accepted-and-documented cosmetic/bounded corners slated for one follow-up ticket.

PR #785, opened 2026-08-17
Cache CLAUDE.md and session metadata by file mtime on GET /api/agents

yiw190:
## Why

`GET /api/agents` re-reads each agent's `CLAUDE.md` and `session-metadata.json` on every tray/UI poll. Local SSD hides that. Cloud `/data` is S3 Files, so those reads are network RTTs stacked on one Node process.

The correct cache for a single file is the one `getSessionSummary` already uses: **stat, then serve the parsed result only when mtime (and size) still match.** No TTL. A write — host or container — changes mtime, so the next list re-reads.

## What changed

- `CLAUDE.md` and `session-metadata.json`: in-process cache keyed by resolved path + `mtimeMs` + size. Overlapping lists share one load. Missing files are not stored, so a create is visible on the next stat.
- `updateAgent` goes through the same `stat` path, so a host rename cannot merge into a stale parse after a container edit.
- `readSessionMetadata` (read-only) is cached. `readSessionMetadataStrict` (the write path) still hits disk.
- `artifacts/`: still **single-flight only**. Directory mtime is not a correct generation — editing `package.json`, adding `screenshot.png`, or creating `node_modules` often does not bump the parent dir. The walk result is dropped when the walk finishes.

No event invalidation. No 20s TTL.

## Staging check (S3 Files `/data`)

One-off Fargate task on `yitian-superagent` (org `e72242c1-…`), same access point as the host-app, write + `stat` on `/data/.mtime-probe`:

| check | result |
|---|---|
| `mtime` after create | non-zero, current time |
| restat with no write | same `mtimeMs` |
| append after 1.1s | `mtime` changed (`deltaMs` 1114 / 1126) |
| `birthtime` | epoch 0 (already known) |
| existing `CLAUDE.md` | real `mtime` (`2026-07-07`), not epoch |

Same-client write → `stat` is usable. Cross-client NFS attribute cache (MicroVM write, host-app `stat`) can lag a few seconds — the same caveat `getSessionSummary` already lives with. Not a TTL.

## Test plan

- [x] Warm `getAgent` / `readSessionMetadata` does not `readFile` when mtime is unchanged
- [x] Write then read re-reads (including `updateAgent` after a container-style `CLAUDE.md` write)
- [x] Overlapping reads share one load
- [x] A miss is not cached; a later create is visible
- [x] `mtimeMs === 0` is not used as a cache key
- [x] Two overlapping artifact lists share one walk; a later list walks disk again
- [ ] After deploy: warm `/api/agents` polls stop re-reading unchanged `CLAUDE.md` / `session-metadata.json`

### DIFF
--- a/src/shared/lib/services/agent-service.test.ts
+++ b/src/shared/lib/services/agent-service.test.ts
@@ -48,6 +48,7 @@ import {
   agentExists,
   getAgentClaudeMdContent,
   setAgentClaudeMdContent,
+  _resetAgentClaudeMdMtimeCacheForTest,
 } from './agent-service'
 
 describe('agent-service', () => {
@@ -66,9 +67,11 @@ describe('agent-service', () => {
 
     // Reset mocks
     vi.clearAllMocks()
+    _resetAgentClaudeMdMtimeCacheForTest()
   })
 
   afterEach(async () => {
+    _resetAgentClaudeMdMtimeCacheForTest()
     // Restore env
     if (originalEnv) {
       process.env.SUPERAGENT_DATA_DIR = originalEnv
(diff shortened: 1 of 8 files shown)


yiw190:
hold off

yiw190:
not the main reason

PR #793, opened 2026-08-18
Guard live-edge follow through clamp-producing transitions

iddogino:
Follow-up to #784. Fixes three regressions from the `use-stick-to-bottom` migration, all instances of one bug family: **clamp-echo escapes**. When the scroll range shrinks while pinned to the live edge, the browser clamps `scrollTop`; the clamp's scroll event is byte-identical to a user scrolling up, so the library classified it as the reader escaping — silently killing follow and/or showing a stale "Scroll to bottom" pill.

## The three transitions

1. **Vertical viewport shrink** — the library's ResizeObserver watches only the content element, so window resizes never re-pinned; browsers anchor the top edge, sliding the newest content behind the fold. Growing the window back also produced a grow-clamp misread as an escape.
2. **Send-time turn collapse** — the finished turn collapsing (same commit as the optimistic message) shrank content → clamp → spurious escape → pill shown over the blank new-turn reserve, and the reserve→follow handoff started disengaged.
3. **Idle-time turn collapse** — the "Worked for Nm" summary swallowing the turn's tool cards clamped `scrollTop` and stranded the reader above the final answer.

## The fix

A shared `protectFollowTransition()` guard with two layers, called from the send effect, the viewport-resize pin, and a `completedTurns`-growth layout effect:

- **Shield**: extend the library's own `resizeDifference` window across the transition so the clamp echo is discarded by its scroll classification (commit-hooked where possible, so it's race-free).
- **Verification backstop**: 40ms later, if follow reads escaped-at-bottom and no user gesture (wheel/touch/keys/held pointer) happened since the transition, reverse the escape via `scrollToBottom`. A real gesture suppresses the reversal, so user intent always wins. This covers the ~1-in-8 case where ResizeObserver delivery slips a frame behind the clamp event.

Plus the viewport RO now re-pins on viewport height changes (`state.scrollTop` write, marked as the library's own) and a catch-all guards any content shrink at the bottom — thinking-card collapses, streamed→persisted swaps, and future shrink sources.

## Verification

- Every guard is negative-tested: disabling it makes a specific test fail deterministically (the resize shield: 2/2 E2E failures without it; the idle-collapse guard: unit test red with the guard short-circuited).
- New E2E: `keeps the live edge pinned through vertical window resizes` — shrink 250px → still pinned, grow back → still pinned, stream continues → still following, no pill. 10/10 serial passes.
- New unit tests for all three transitions (viewport shrink pin + escaped-stays-put, send-collapse clamp → no pill, idle-collapse clamp → growth still followed to bottom + no pill), via a controllable FakeResizeObserver harness.
- Full suite: transcript-follow E2E 9/9 across 3 serial repeats; messages-dir unit tests 796/796 (51 files); typecheck + lint clean (including `e2e/`).

https://claude.ai/code/session_01GGDfc6bd4NeNio2DwmBT39

### DIFF
--- a/e2e/specs/transcript-follow.spec.ts
+++ b/e2e/specs/transcript-follow.spec.ts
@@ -107,4 +107,41 @@ test.describe('Transcript live-edge follow', () => {
       .toBeLessThan(90)
     await expect(pill).toBeHidden()
   })
+
+  test('keeps the live edge pinned through vertical window resizes', async ({ page }) => {
+    await sessionPage.sendMessage(SAGA_PROMPT)
+    await sessionPage.waitForResponse(15000)
+    // Get past the new-turn reserve, whose spacer handles resizes on its own.
+    await expect(page.getByTestId('turn-anchor-spacer')).toBeHidden({ timeout: 20000 })
+    await expect
+      .poll(async () => (await scrollMetrics(page)).distanceFromBottom, { timeout: 15000 })
+      .toBeLessThan(90)
+
+    // Browsers anchor the top edge on a viewport shrink, which would slide
+    // the newest content behind the fold — the transcript must re-pin so
+    // content leaves from the top, not the bottom.
+    const viewport = page.viewportSize()!
+    await page.setViewportSize({ width: viewport.width, height: viewport.height - 250 })
+    await expect
+      .poll(async () => (await scrollMetrics(page)).distanceFromBottom, { timeout: 5000 })
+      .toBeLessThan(90)
+
+    // Growing back stays pinned too — the browser's own clamp scroll must
+    // not be misread as the reader escaping.
+    await page.setViewportSize({ width: viewport.width, height: viewport.height })
+    await expect
+      .poll(async () => (await scrollMetrics(page)).distanceFromBottom, { timeout: 5000 })
+      .toBeLessThan(90)
+
+    // Following survived both resizes: as the stream continues, the viewport
+    // keeps up and the escape affordance never appears.
+    const mark = (await scrollMetrics(page)).contentHeight
+    await expect
+      .poll(async () => (await scrollMetrics(page)).contentHeight, { timeout: 10000 })
+      .toBeGreaterThan(mark + 150)
+    await expect
+      .poll(async () => (await scrollMetrics(page)).distanceFromBottom, { timeout: 15000 })
+      .toBeLessThan(90)
+    await expect(page.getByRole('button', { name: 'Scroll to bottom' })).toBeHidden()
+  })
 })
--- a/e2e/specs/transcript-follow.spec.ts
+++ b/e2e/specs/transcript-follow.spec.ts
@@ -108,6 +108,33 @@ test.describe('Transcript live-edge follow', () => {
     await expect(pill).toBeHidden()
   })
 
+  test('holds the reading line steady while the agent works', async ({ page }) => {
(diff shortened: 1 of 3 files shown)


iddogino:
Pushed a second commit (`31cb7625`) fixing a fourth member of the clamp family this PR unmasked: **the reserve-hold sag**.

While the new-turn reserve holds, the viewport sits at max scroll by design — so a transient content shrink under it (working indicator swapping forms, streamed→persisted swap) clamps `scrollTop` before the spacer re-inflates the range. The escape guards in the first commit keep following *alive* through that clamp, which turned the old failure (follow dies + stale pill) into a visible bounce: send → glide up → sag ~40–48px → recover on next growth.

Fix: `syncTurnReserve` now restores the anchored reading line in the same pass that re-sizes the spacer (the spacer write is imperative, so the restored range is already in place). Skipped while the reader is gesturing, while a scroll animation is in flight, and on the send-time call (the glide travels to the reserve by design).

Verified with a frame-by-frame scroll trace against the mock `work slowly` scenario: sag depth 47px → 0, send glide intact (68 animation steps), no pill. New unit test models the sticky clamp and is red-green proven (guard disabled → deterministic fail); new E2E smoke test locks the hold-through-swap contract (reserve holds, viewport re-converges, no pill — the mock's turn-end growth lands too soon after the swap for E2E to time-discriminate the restore itself, so the unit test carries the deterministic guard). Full follow spec 12/12 across 3 serial repeats; messages-dir units 797/797; typecheck + lint clean.

iddogino:
Both re-review findings addressed in `ed4d7f7f` — each was verified real against the library source before fixing.

**Swallowed real escapes.** Confirmed: the library's deferred scroll classification bails whenever `resizeDifference` is set, and keyboard/touch/scrollbar escapes ride on that classification (only wheel-up bypasses it via `handleWheel`). Fix: the scroll handler now stamps gesture-attributed *upward* scrolls (a stronger signal than the plain gesture stamp, which a downward wheel also sets), and the 40ms verification gained a complementary branch — if an upward gesture landed inside the window yet following still reads engaged and the reader is beyond the 70px re-stick range, the escape was swallowed: mark it via the library's `stopScroll()`. The distance condition keeps reserve eating unaffected (eating re-bases the reading line onto the reader, so they stay within the re-stick range) and avoids misfiring on clamp echoes that carry a stale gesture stamp (those land at or get restored to the bottom). Regression test: PageUp inside the collapse shield → pill appears, later growth leaves the reader at their position. Fails deterministically with the branch disabled.

**Pre-animation restore preempting the send glide.** Confirmed: `state.animation` is assigned inside `scrollToBottom`'s first rAF callback, so the `!state.animation` guard was blind in the gap where the POST's uuid-assignment commit lands. Fix: a `sendScrollInFlightRef` brackets the whole send scroll — set before the reserve is sized, cleared when the scroll promise settles — and replaces the per-call `skipRestore` option (which only covered the first sync). Regression test: non-reduced-motion send followed by a uuid-assignment commit before the first frame → viewport stays at the pre-glide position. Fails deterministically with the ref guard disabled (even with `!state.animation` still present, confirming the original guard alone was ineffective).

Verification: 799/799 messages-dir unit tests (115 in message-list.test.tsx, both new cases red-green proven), transcript-follow E2E 12/12 across 3 serial repeats, typecheck + lint clean.

PR #11, opened 2026-03-04
feat: Voice input with real-time streaming STT (Deepgram & OpenAI)

yiw190:
## Summary

- **BYOK voice input** — users bring their own Deepgram or OpenAI API key, configure in Settings > Voice, and use the mic button to dictate messages with real-time streaming transcription
- **STT adapter layer** with unified interface abstracting Deepgram Nova 3 (16kHz) and OpenAI Realtime Transcribe (24kHz, gpt-4o-mini-transcribe) WebSocket protocols
- **Voice settings tab** with provider selection, API key validation, and live mic test with waveform visualization and streaming transcript preview
- **STT usage tracking** — recording duration and cost are written to a `stt_usage` table and merged into the existing Usage dashboard alongside LLM models (totalCost, byModel, byAgent)

### New files (5)
- `src/renderer/lib/stt.ts` — `SttAdapter` interface + `DeepgramAdapter` / `OpenaiAdapter` implementations (direct browser-to-provider WebSocket)
- `src/renderer/hooks/use-voice-input.ts` — React hook managing mic capture, PCM conversion, adapter lifecycle, interim/final transcript accumulation, and usage reporting
- `src/renderer/components/settings/voice-tab.tsx` — Settings UI: provider picker, API key input with validate/save, live test with canvas waveform + animated transcript
- `src/api/routes/stt.ts` — `GET /token` relays stored API key to frontend; `POST /usage` records STT session duration with cost calculation
- `src/shared/lib/db/migrations/0010_amused_sumo.sql` — Creates `stt_usage` table with indexes on created_at, user_id, agent_slug

### Modified files (10)
- `src/shared/lib/db/schema.ts` — Added `sttUsage` table definition (provider, model, durationMs, cost in micro-dollars, agentSlug, userId)
- `src/shared/lib/config/settings.ts` — Added `SttProvider`, `VoiceSettings`, `deepgramApiKey`, `openaiApiKey` to settings schema
- `src/api/routes/settings.ts` — Added Deepgram/OpenAI key save/remove/validate endpoints, voice settings in GET/PUT responses
- `src/api/routes/usage.ts` — Queries `stt_usage` table, merges STT costs into daily aggregation (totalCost + byModel + byAgent)
- `src/api/index.ts` — Mounted `/api/stt` router
- `src/renderer/hooks/use-settings.ts` — Extended `UpdateSettingsParams` with voice + new key fields
- `src/renderer/components/settings/global-settings-dialog.tsx` — Added Voice tab
[...]

PR #10, opened 2026-03-04
Browser + Chrome Profile Wizard

apai25:
- Add browser selection + chrome profile selection to onboarding wizard
- Add separate module for profile selection (used in settings + onboarding wizard)
- Modularize onboarding wizard, cleanup
- Chrome profile selection renders Chrome profile pictures
Screenshot 2026-03-04 at 1 16 37 PM
Screenshot 2026-03-04 at 1 17 10 PM

- TODO (later PR): Chrome profile isn't automatically logged in, even when selected (bug not related to this PR)

PR #804, opened 2026-08-19
Media references instead of inline base64 in message payloads (SUP-596)

iddogino:
Images in transcripts ride as base64 inside the JSONL row, and every message page re-shipped all of it to every viewer. `GET /messages?media=ref` replaces them with an address the client resolves against a new media endpoint, on demand.

Opt-in: without the param, responses are byte-identical to today.

## Addressing

A ref is the byte span of the **base64 payload itself**, not of the row holding it. The rows are multi-MB precisely because of the images, so "seek to the row, then find the image" would materialize the very thing this avoids. Serving a ref is a ranged read of exactly the payload piped through a streaming base64 decoder: O(chunk) memory, no line assembly, no JSON parse.

That works because base64's alphabet needs no JSON escaping, so the payload's bytes on disk are byte-identical to the decoded string value. Anything not provably locatable is left inline — correctness never depends on a ref being minted.

Refs are minted **while rows are parsed**, not after the transform. A page window holds every row at once, so stripping there keeps the base64 out of peak memory as well as off the wire. `transformMessages` is untouched; the ref object flows through `toolCall.result` by reference.

## Validity

Deletion and retention rewrite transcripts in place, shifting every offset after the edit. A ref therefore carries its row's uuid **and the byte offset that uuid sat at**, so a read confirms the row is still there with a 36-byte read rather than a scan. With exact quote delimiters around the payload and an image magic-number sniff of the head, a moved or vanished payload fails closed (410) instead of serving unrelated bytes.

Those same checks are what make client-supplied refs safe to honor without signing: a forged span has to land on a quote-delimited base64 string whose head decodes to a known image type — which in a transcript means an actual image, one the caller could already read through the messages endpoint they are authorized for. The response's Content-Type comes from the sniff, **never from the ref**, so a ref cannot dictate how bytes are interpreted. SVG is deliberately excluded: not sniffable, and it executes script when served inline.

⚠️ The uuid check needed a test that isolates it. Deleting the comparison left the whole suite green — the delimiter and magic checks were catching my rewrite fixtures instead. The regression test now swaps in a **same-length uuid**: offsets undisturbed, payload intact, only identity changed. That is the only shape that fails without this check.

## Two duplicate copies, one of them invisible

Measured on real transcripts: the base64 is stored **twice** — once in the `tool_result` block (the only copy that reaches the wire) and again under `toolUseResult.file.base64`. 4.91MB each in a 14.59MB session. The second copy is never serialized but is half of an image-heavy window's retained bytes, so it is dropped outright — it needs no ref, only release.

## Client

`tool-call-item.tsx` already rendered result images only inside the expanded branch, so collapsed calls never mount an ``. Pointing the src at the endpoint means the browser fetches **on expand**, with no observer and no fetch hook; `Cache-Control: private, immutable` makes re-expands and later polls free. A ref that no longer resolves shows a placeholder instead of a broken image.

The ref carries a `url` API path so no consumer needs the session's identity threaded down to it. That puts `lib/parse-tool-result.ts` on the `DIRECT_BASE_URL_CONSUMERS` inventory in the api-origin characterization suite, with its reason.

Both the page and the delta reader honor the param — the delta is what carries live screenshots during an active session.

## Validation

Live HTTP against a real 15MB / 21-image transcript, trailing page `limit=300`:

| | inline (today) | `media=ref` |
|---|---|---|
| response | 1,222,010 B | 288,930 B |
| time | 1.88 s | 0.13 s |

Every image byte is gone; the remaining 289KB is genuine text, which is what bounds the ratio on *this* transcript. Media endpoint: 3–8 ms per image, correct type and length. Forged ref → 410, garbage → 400.

- **Byte-identical parity** with `origin/main` in default mode: 7 sha256 hashes across limits 10/50/300, their cursor pages, and a delta, captured on both trees.
- **Byte-for-byte image recovery**: every ref on a real page re-read through the endpoint and compared to the original decoded payload.
- **Browser (Playwright)**: 4 images render 919×1998, `loading="lazy"`, 0 broken placeholders, and **0 media requests on initial page load**.
- Full suite **9698 passed / 0 failed**; typecheck + lint clean.

## Deliberately out of scope

- The subagent transcript route (`/subagent/:agentId/messages`) still reads whole files with `readJsonlFile` and ships inline base64.
- Oversized **text** tool results stay inline. Unlike an ``, a text ref cannot fetch itself — it needs a real fetch-on-expand hook and loading state. Images were ~95% of the bytes; happy to follow up if a real transcript shows otherwise.

## Unrelated, but worth knowing

Any worktree branched before #784 is missing `use-stick-to-bottom`, which makes `message-list.test.tsx` fail 115/115 with a misleading "Invalid hook call". `npm install` fixes it.

### DIFF
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -1,7 +1,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
 import { Hono } from 'hono'
 import { runInNewContext } from 'node:vm'
-import { Writable } from 'node:stream'
+import { Readable, Writable } from 'node:stream'
 import { createHash } from 'node:crypto'
 
 // ============================================================================
@@ -297,6 +297,11 @@ vi.mock('@shared/lib/services/agent-service', () => ({
   agentExists: (...args: unknown[]) => mockAgentExists(...args),
 }))
 
+vi.mock('@shared/lib/services/session-media', () => ({
+  decodeMediaRef: vi.fn(),
+  openMediaBlob: vi.fn(),
+}))
+
 vi.mock('@shared/lib/services/session-service', () => ({
   listSessions: vi.fn(),
   listSessionsByIds: vi.fn(),
@@ -553,6 +558,7 @@ vi.mock('hono/streaming', () => ({ streamSSE: (...args: unknown[]) => mockStream
 
 // Import the agents router after all mocks are set up
 import agents from './agents'
+import { decodeMediaRef, openMediaBlob } from '@shared/lib/services/session-media'
 import { UploadTooLargeError } from '@shared/lib/utils/chunked-upload'
 import {
   importAgentFromTemplate,
@@ -3867,6 +3873,36 @@ describe('GET /:id/sessions/:sessionId/messages pagination', () => {
     expect(getSessionMessagesPage).not.toHaveBeenCalled()
   })
 
+  it('forwards the media mode to the page reader', async () => {
+    vi.mocked(getSessionMessagesPage).mockResolvedValue({ messages: [], nextCursor: null })
+
+    const res = await getReq(app, `${URL}?limit=2&media=ref`)
+    expect(res.status).toBe(200)
+    expect(getSessionMessagesPage).toHaveBeenCalledWith(
+      'test-agent',
+      'sess-1',
+      expect.objectContaining({ media: 'ref' })
+    )
+  })
+
+  it('forwards the media mode to the delta reader', async () => {
+    vi.mocked(getSessionMessagesDelta).mockResolvedValue({ messages: [], anchor: null })
+
+    const res = await getReq(app, `${URL}?after=m1&media=ref`)
+    expect(res.status).toBe(200)
+    expect(getSessionMessagesDelta).toHaveBeenCalledWith(
+      'test-agent',
+      'sess-1',
+      expect.objectContaining({ media: 'ref' })
+    )
+  })
+
+  it('rejects an unknown media mode rather than silently serving inline', async () => {
+    const res = await getReq(app, `${URL}?limit=2&media=inline`)
+    expect(res.status).toBe(400)
+    expect(getSessionMessagesPage).not.toHaveBeenCalled()
+  })
+
   it('caps first-page limit to MESSAGES_PAGE_LIMIT', async () => {
     process.env.MESSAGES_PAGE_LIMIT = '100'
     vi.mocked(getSessionMessagesPage).mockResolvedValue({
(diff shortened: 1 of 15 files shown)


iddogino:
All nine confirmed defects fixed at `80fe0ad3`; findings 7 and 10 deferred with a documented reason. Every finding reproduced before I touched anything, and each fix ships with the failing case as a regression test.

Thank you for the reproductions — the offsets and byte counts made these fast to confirm, and two of them landed squarely on design decisions I'd defended in the PR body.

## P1 — stream teardown

Confirmed, and worse than "leaks a descriptor". I traced the second half: `main/index.ts:1820` answers `uncaughtException` with `gracefulShutdown()` → `app.quit()`, so an `EIO` on the source would **quit the desktop app**. My probe showed `handleStillOpen = true` after destroying the returned decoder, and `0` error listeners on the source.

Now `pipeline()` in both directions, the request signal wired for the stream's lifetime, and a benign-error filter so a client hanging up mid-image isn't logged as a fault. Probes now read `handleStillOpen = false` and `3` source error listeners.

## P1 — self-declared media URLs

Confirmed for external, protocol-relative, and same-origin paths. I also checked whether CSP blunted it: there is no app-wide policy, only `img-src` on two asset routes, so nothing did.

Worth recording *why* this is a regression rather than a pre-existing condition: before this PR, untrusted result content could only ever become a `data:` URL, so it could not cause a request at all.

The fix goes further than validating the path — **`url` is gone from the wire entirely**. A field the client deliberately ignores is a trap for the next consumer. The renderer now builds the single address it will request from `agentSlug` + `sessionId` it already holds, and the block contributes only an `id` that must match the base64url shape we mint. No context, no request. A hostile block with a plausible id resolves to *our* endpoint, where it simply doesn't validate.

That also retires the reasoning I gave for putting the URL in the payload ("no consumer needs the session identity threaded down"). The convenience was what moved the trust decision to the client; `sessionId` is now threaded to `ToolCallItem`, which is two call sites.

## Ref identity — I'd rate this P1, not P2

Your scenario reproduced exactly, and the probe output is the argument for raising it: the old ref did not fail, it **served image B** through image A's address, advertised `immutable` for a year. Wrong content shown confidently is worse than a placeholder. (Not a security issue — both images are in a session the caller already reads.)

Refs now carry a sampled content fingerprint (head/middle/tail + length, three small reads), so the address names an image rather than a location. Two payloads agreeing on all of it are the same image for serving purposes, which is the property that makes `immutable` honest.

## Unsupported formats — also higher than P2

An SVG renders inline today and became a permanent placeholder, which is a functional regression, not a missed optimization. Self-inflicted, too: my own comment documented excluding SVG from sniffing and I never connected that to minting. Minting is now gated on the same sniff that serves, so unsupported formats stay inline by construction.

## The rest

- **Textual base64** — the payload search now requires the JSON string delimiters and keeps scanning past matches that lack them. The repro serves the real image instead of minting a ref that fails its own check.
- **`media` ignored without pagination** — confirmed; `media` now selects the validated path. Live: `?media=ref` alone went from a 3MB inline response to 288,350 bytes, and `?media=bogus` alone from 200 to 400.
- **Fractional Content-Length** — `Math.floor((length * 3) / 4) - pad`, correct for padded and unpadded. Probe: declared 20480, served 20480. Padding placement is now checked while decoding too (the charset regex had allowed `=` anywhere). Writing that check surfaced a bug in my first attempt — the carry retains padding it already counted — which the regression test caught.
- **Transient failures as 410** — only ENOENT and failed validation return "gone"; `EIO`/`EMFILE`/`EACCES` propagate and the route answers 500, which is retryable. Good catch that this contradicted `readFileRangeFully`, introduced three files away in #783 for exactly the short-read reason; validation reads now loop the same way.
- **Error copy** — an `` error carries no reason, so it no longer claims one: "Couldn't load image" with a Retry that re-requests rather than re-showing a failed cache entry.
- **Duplicate keys** — keyed by index and src.

## Deferred, with reasons in the code

**JSON-serialized block arrays (7).** Real in code, but I scanned 12 real transcripts: 282 image tool results, **all arrays, zero serialized**. A robustness gap with no known occurrences.

**Stale older-history refs (10).** Real. The honest fix is a transcript generation carried through responses so a client can tell its addresses are stale — the same mechanism the paged reader wants for cursor continuity, and it should be designed once, with that, rather than patched here.

Both are documented under KNOWN GAPS at the top of `session-media.ts` so they're visible to the next reader rather than living only in this thread.

## Scope

Left as-is, and I agree it's the sharpest remaining point. I confirmed the number that makes it sharp: workflow subagent transcripts poll at `refetchInterval: 1500` while running, against a route that reads whole files and ships inline base64 — a heavier path than the 15s message poll this PR fixes. It needs its own ticket rather than a quiet expansion of this one; happy to file it.

## Verification at `80fe0ad3`

Full suite **9710 passed / 0 failed**; typecheck and lint clean. Live re-validation on the 15MB/21-image transcript: page 1,222,010 → 288,350 bytes, all four refs serve as JPEG, and in the browser all four images render with **0 media requests on initial load** and no placeholders.

iddogino:
All four remaining blockers fixed at `c4c0d64b`, plus the accessibility gap. Every one reproduced first.

One correction on status: **blocker 1 was already fixed before this review landed.** You reviewed `80fe0ad3`; `c231d2e1` removed the pass-through catch about an hour earlier, and CI's `unit-tests` job has been green on it since (8m24s). Your diagnosis was exactly right, including that it fails at the lint step of a job named "unit-tests" — worth noting for anyone else triaging, since the visible symptom is a missing `coverage-summary.json` forty lines further down.

That one was mine twice over: I'd "verified" lint with `npx eslint … | grep -cE " error "`, which cannot match because ESLint wraps severities in ANSI colour codes. It reported zero while a real error sat there. I now run the repo's own script and read the exit code.

## Sampled fingerprint (#3) — you're right, and I'd overreached

Reproduced with two 30,054-byte BMPs differing only around 28% in: `accepted = true`, and the old address served the second image.

I chose sampling to keep verification to three small reads. That bought a probabilistic guarantee while the response advertised `immutable` for a year — a deterministic promise. Those don't reconcile, and the sampling was the part that had to give.

The digest now covers the **whole payload**, verified by one pass over the span before the serving stream opens. Two things made that easier to accept than I expected when I first avoided it:

- It costs nothing measurable. End-to-end per image on the 15MB transcript: **3.6–6.5ms**, against 3–8ms before the pass existed. It's a sequential read of a span already in page cache, against a decode and a socket write.
- Memory is unchanged — 64KB chunks, nothing retained.

So the address now either names those exact bytes or resolves to nothing, which is what `immutable` was claiming all along.

## Malformed padding (#8)

Confirmed: `encodedLen % 4 = 1`, advertised 30,053, streamed 30,054. I'd checked padding *placement* and missed *length congruence* — a payload that isn't a whole base64 quantum decodes to a count no formula predicts, and that count is the Content-Length. Payloads with `length % 4 === 1`, or padding on a non-multiple-of-four length, are now left inline, and the same check guards serving.

## Existence preflight (#9)

Confirmed by reading `fileExists`: its catch-all turns any stat failure into `false`. You were also right that my route test masked it by mocking `sessionExists(true)`.

The preflight is gone rather than narrowed. Ownership is still checked; existence is the read's business, and `openMediaBlob` already separates ENOENT (410, genuinely gone) from operational failure (throws → 500, retryable). The new test asserts the route does not call `sessionExists` at all, so a future reinstatement fails loudly instead of silently restoring 404-on-EIO.

## Retry corrupting data URLs (#11)

Confirmed — my bug, and a clean example of a fix creating a defect one layer over. Cache-busting now applies only when `image.isRef`; inline images remount unchanged via `key`. The failure notice is also `role="status"` / `aria-live="polite"`, since it appears asynchronously — thank you for the accessibility catch.

## Verification at `c4c0d64b`

Run as CI runs them, by exit code:

- `npm run typecheck` → 0
- `npm run lint` → 0
- `npx vitest run --coverage` → 0, **9714 passed / 0 failed**

Live on the 15MB/21-image transcript: page 288,382 bytes, all four refs serve as JPEG at 3.6–6.5ms.

Four new regression tests, each failing against `80fe0ad3`: the equal-length BMP collision, the malformed-padding payload, the absent existence preflight, and the inline-data-URL retry.

## Still deferred, per direction

#7 (serialized block arrays) and #10 (stale older-history refs), both documented under KNOWN GAPS in `session-media.ts`. The subagent and workflow transcript surfaces also remain out of scope; I'd still like to file that as its own ticket rather than widen this PR — the workflow drawer polls those at 1.5s intervals against whole-file reads, which is a heavier path than the 15s message poll this PR fixes.

PR #805, opened 2026-08-19
fix: stream agent export zips instead of buffering them in memory

yiw190:
## Why

A large **Export full** took the host down. The zip was built entirely in RAM, so a big workspace OOM-killed a 1 GB host and returned 503. Export is also slow enough that someone can close Settings and click again, which used to start a second zip.

## What

1. **Stream the export on the host.** The zip is written out as it is built instead of held in host RAM. The host starts sending bytes as files are added. In the desktop app the renderer still buffers via `downloadBlob` (`await res.blob()`) before the save dialog — the OOM we hit was on the cloud host, where the browser is on someone else's machine. Streaming also drops `Content-Length`, so browsers and `curl -O` lose download progress.
2. **One export at a time, host-wide.** Template and full export share one lock. A second request gets 409 until the first zip finishes. Skill export is unchanged. This is deliberate: a 1 GB host cannot run two large zips. In multi-user auth mode that means one person's export blocks everyone else's.
3. **Status check so the spinner survives closing Settings.** Both export buttons stay disabled and spinning while the host is still zipping, so it is obvious an export is still happening.

(2) and (3) exist so a slow large-agent export cannot be stacked by a retry.

zlib: full export uses level 1 (level 9 pegs 0.5 vCPU hosts on large workspaces). Template export stays at 9 — the shareable `.agent` is small and size matters more.

Lock lifetime: the lock is released when the zip stream closes. If anything throws between starting the export and returning the `Response`, the stream is destroyed so the lock cannot leak. Cancel uses `archive.abort()` then `destroy()` so pending stat/append queues stop.

## Staging

Tested on staging with a workspace in the same size range as the agent that OOM'd (~500 MB / ~700 files). Host stayed healthy; RAM rose about 100–130 MB over idle during the zip, then returned to baseline. No OOM.

## Test plan

- [x] `npx vitest run src/shared/lib/services/agent-template-service.test.ts src/api/routes/agents.test.ts src/renderer/hooks/use-agent-templates.export-status.test.ts`
- [x] Staging export of a ~500 MB / ~700-file agent — host stayed healthy, memory did not climb with zip size
- [ ] Export template still downloads a valid `.agent` zip
- [x] Cancel the download mid-flight — host walk stops (covered by the real-socket hangup test)
- [ ] Start a second export while the first is still downloading — 409 "An export is already in progress"
- [ ] Close and reopen Settings during an export — both export buttons stay disabled and spinning

### DIFF
--- a/src/api/routes/agents.ts
+++ b/src/api/routes/agents.ts
@@ -4766,7 +4766,10 @@ function exportRouteError(c: Context, error: unknown, fallback: string) {
 agents.post('/:id/export-template', AgentAdmin(), async (c) => {
   try {
     const slug = getAgentId(c)
-    const [agent, zipStream] = await Promise.all([getAgent(slug), exportAgentTemplate(slug, c.req.raw.signal)])
+    const agent = await getAgent(slug)
+    // Lock is taken inside the export call and released only when the stream
+    // closes, so it must be the last throwing step before the Response.
+    const zipStream = await exportAgentTemplate(slug, c.req.raw.signal)
 
     logAuditEvent({ userId: getCurrentUserId(c), object: 'agent', objectId: slug, action: 'exported', details: { type: 'template' } })
     return packageDownloadResponse(zipStream, `${agent?.frontmatter.name || slug}-template${AGENT_PACKAGE_EXTENSION}`)
@@ -4779,7 +4782,10 @@ agents.post('/:id/export-template', AgentAdmin(), async (c) => {
 agents.post('/:id/export-full', AgentAdmin(), async (c) => {
   try {
     const slug = getAgentId(c)
-    const [agent, zipStream] = await Promise.all([getAgent(slug), exportAgentFull(slug, c.req.raw.signal)])
+    const agent = await getAgent(slug)
+    // Lock is taken inside the export call and released only when the stream
+    // closes, so it must be the last throwing step before the Response.
+    const zipStream = await exportAgentFull(slug, c.req.raw.signal)
 
     logAuditEvent({ userId: getCurrentUserId(c), object: 'agent', objectId: slug, action: 'exported', details: { type: 'full' } })
     return packageDownloadResponse(zipStream, `${agent?.frontmatter.name || slug}-full${AGENT_PACKAGE_EXTENSION}`)
--- a/src/api/routes/agents.ts
+++ b/src/api/routes/agents.ts
@@ -4759,6 +4759,16 @@ function packageDownloadResponse(body: Readable | Buffer, filename: string): Res
   return new Response(Readable.toWeb(nodeStream) as ReadableStream, { status: 200, headers })
 }
 
+// Lock lives on the stream ('close' releases it). Destroy if Response construction throws.
+function sendLockedExportStream(zipStream: Readable, build: () => Response): Response {
+  try {
+    return build()
+  } catch (err) {
+    zipStream.destroy()
+    throw err
+  }
+}
+
 function exportRouteError(c: Context, error: unknown, fallback: string) {
   if (error instanceof Error && error.name === 'ExportInProgressError') {
     return c.json({ error: error.message }, 409)
@@ -4773,12 +4783,11 @@ agents.post('/:id/export-template', AgentAdmin(), async (c) => {
   try {
     const slug = getAgentId(c)
     const agent = await getAgent(slug)
-    // Lock is taken inside the export call and released only when the stream
-    // closes, so it must be the last throwing step before the Response.
     const zipStream = await exportAgentTemplate(slug, c.req.raw.signal)
-
-    logAuditEvent({ userId: getCurrentUserId(c), object: 'agent', objectId: slug, action: 'exported', details: { type: 'template' } })
-    return packageDownloadResponse(zipStream, `${agent?.frontmatter.name || slug}-template${AGENT_PACKAGE_EXTENSION}`)
+    return sendLockedExportStream(zipStream, () => {
+      logAuditEvent({ userId: getCurrentUserId(c), object: 'agent', objectId: slug, action: 'exported', details: { type: 'template' } })
+      return packageDownloadResponse(zipStream, `${agent?.frontmatter.name || slug}-template${AGENT_PACKAGE_EXTENSION}`)
+    })
   } catch (error) {
     return exportRouteError(c, error, 'Failed to export template')
   }
@@ -4789,12 +4798,11 @@ agents.post('/:id/export-full', AgentAdmin(), async (c) => {
   try {
     const slug = getAgentId(c)
     const agent = await getAgent(slug)
-    // Lock is taken inside the export call and released only when the stream
-    // closes, so it must be the last throwing step before the Response.
     const zipStream = await exportAgentFull(slug, c.req.raw.signal)
-
-    logAuditEvent({ userId: getCurrentUserId(c), object: 'agent', objectId: slug, action: 'exported', details: { type: 'full' } })
-    return packageDownloadResponse(zipStream, `${agent?.frontmatter.name || slug}-full${AGENT_PACKAGE_EXTENSION}`)
+    return sendLockedExportStream(zipStream, () => {
+      logAuditEvent({ userId: getCurrentUserId(c), object: 'agent', objectId: slug, action: 'exported', details: { type: 'full' } })
+      return packageDownloadResponse(zipStream, `${agent?.frontmatter.name || slug}-full${AGENT_PACKAGE_EXTENSION}`)
+    })
   } catch (error) {
     return exportRouteError(c, error, 'Failed to export agent')
   }
--- a/src/api/routes/agents.ts
+++ b/src/api/routes/agents.ts
@@ -4753,6 +4753,15 @@ function packageDownloadResponse(body: Readable | Buffer, filename: string): Res
   return new Response(Readable.toWeb(nodeStream) as ReadableStream, { status: 200, headers })
 }
 
+function exportRouteError(c: Context, error: unknown, fallback: string) {
+  if (error instanceof Error && error.name === 'ExportInProgressError') {
+    return c.json({ error: error.message }, 409)
+  }
+  const message = error instanceof Error ? error.message : fallback
+  console.error(fallback, error)
+  return c.json({ error: message }, 500)
+}
+
 // POST /api/agents/:id/export-template - Export agent as ZIP download
 agents.post('/:id/export-template', AgentAdmin(), async (c) => {
   try {
@@ -4762,9 +4771,7 @@ agents.post('/:id/export-template', AgentAdmin(), async (c) => {
     logAuditEvent({ userId: getCurrentUserId(c), object: 'agent', objectId: slug, action: 'exported', details: { type: 'template' } })
     return packageDownloadResponse(zipStream, `${agent?.frontmatter.name || slug}-template${AGENT_PACKAGE_EXTENSION}`)
   } catch (error) {
-    const message = error instanceof Error ? error.message : 'Failed to export template'
-    console.error('Failed to export template:', error)
-    return c.json({ error: message }, 500)
+    return exportRouteError(c, error, 'Failed to export template')
   }
 })
 
@@ -4777,9 +4784,7 @@ agents.post('/:id/export-full', AgentAdmin(), async (c) => {
     logAuditEvent({ userId: getCurrentUserId(c), object: 'agent', objectId: slug, action: 'exported', details: { type: 'full' } })
     return packageDownloadResponse(zipStream, `${agent?.frontmatter.name || slug}-full${AGENT_PACKAGE_EXTENSION}`)
   } catch (error) {
-    const message = error instanceof Error ? error.message : 'Failed to export agent'
-    console.error('Failed to export full agent:', error)
-    return c.json({ error: message }, 500)
+    return exportRouteError(c, error, 'Failed to export agent')
   }
 })
 
--- a/src/api/routes/agents.ts
+++ b/src/api/routes/agents.ts
@@ -4740,26 +4740,27 @@ agents.post('/:id/skills/:dir/publish', AgentAdmin(), async (c) => {
  * display name (slugs are opaque minted ids), encoded per the same quoted +
  * RFC 5987 `filename*` convention as workspace-file downloads.
  */
-function packageDownloadResponse(zipBuffer: Buffer, filename: string): Response {
+function packageDownloadResponse(body: Readable | Buffer, filename: string): Response {
   const encoded = encodeURIComponent(filename)
-  return new Response(new Uint8Array(zipBuffer), {
-    status: 200,
-    headers: {
-      'Content-Type': 'application/octet-stream',
-      'Content-Disposition': `attachment; filename="${encoded}"; filename*=UTF-8''${encoded}`,
-      'Content-Length': zipBuffer.byteLength.toString(),
-    },
-  })
+  const headers: Record<string, string> = {
+    'Content-Type': 'application/octet-stream',
+    'Content-Disposition': `attachment; filename="${encoded}"; filename*=UTF-8''${encoded}`,
+  }
+  if (Buffer.isBuffer(body)) {
+    headers['Content-Length'] = body.byteLength.toString()
+  }
+  const nodeStream = Buffer.isBuffer(body) ? Readable.from(body) : body
+  return new Response(Readable.toWeb(nodeStream) as ReadableStream, { status: 200, headers })
 }
 
 // POST /api/agents/:id/export-template - Export agent as ZIP download
 agents.post('/:id/export-template', AgentAdmin(), async (c) => {
   try {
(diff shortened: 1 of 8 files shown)


src/api/routes/agents.ts
```diff
@@ -4740,45 +4746,57 @@ agents.post('/:id/skills/:dir/publish', AgentAdmin(), async (c) => {
...
 }
 
 // POST /api/agents/:id/export-template - Export agent as ZIP download
 agents.post('/:id/export-template', AgentAdmin(), async (c) => {
   try {
     const slug = getAgentId(c)
-    const [agent, zipBuffer] = await Promise.all([getAgent(slug), exportAgentTemplate(slug)])
+    const agent = await getAgent(slug)
+    // Lock is taken inside the export call and released only when the stream
+    // closes, so it must be the last throwing step before the Response.
```
iddogino: **Blocking.** This comment asserts an invariant the code doesn't hold. Two statements follow the export call, and both can throw: `getCurrentUserId(c)` (throws by design when the user is missing from context) and `packageDownloadResponse` -> `encodeURIComponent`.

When either throws, the lock is taken but the stream is never consumed and no abort ever arrives, so `hostExportBusy` stays `true` for the process lifetime. I reproduced this against a real server:

```
firstStatus:              500
hostExportBusyAfter3s:    true
secondStatus:             409  ("An export is already in progress")
directCallError:          ExportInProgressError
```

Every export on the host, for every user, 409s forever. `/export-status` reports `inProgress: true` permanently, so both UI buttons stay disabled and spinning until restart.

Neither trigger is reachable today (`AgentAdmin` guarantees the user is set; the hand-rolled frontmatter parser in `file-storage.ts:168` can't produce a lone surrogate). So this is latent - but the blast radius is a permanently wedged host, and the fix is structural rather than order-dependent:

```ts
const zipStream = await exportAgentFull(slug, c.req.raw.signal)
try {
  logAuditEvent({ ... })
  return packageDownloadResponse(zipStream, ...)
} catch (err) {
  zipStream.destroy() // emits 'close' -> releases the lock
  throw err
}
```

I confirmed `destroy()` does emit `close` on the archiver, so that path releases correctly. Same change needed on the `export-template` route below.

src/shared/lib/services/agent-template-service.ts
```diff
@@ -254,81 +256,136 @@ async function walkFullExportFiles(workspaceDir: string): Promise<string[]> {
...
+  const archive = archiver('zip', { zlib: { level: 1 } })
+  archive.once('close', release)
+  archive.once('error', release)
 
-    archive.on('data', (chunk: Buffer) => chunks.push(chunk))
-    archive.on('end', () => resolve(Buffer.concat(chunks)))
-    archive.on('error', reject)
+  const onAbort = () => {
+    release()
+    if (!archive.destroyed) archive.destroy()
```
iddogino: **Blocking (small).** archiver exposes `abort()` for exactly this - it kills `_queue` and `_statQueue` (`archiver/lib/core.js:77`). `destroy()` is a plain stream method archiver doesn't override, so the pending stat/append queues keep running after cancellation.

Measured on a 200-file archive, cutting at the same point:

```
destroy():  entries at cut 11 -> 14 after 3s  (3 more files read + zipped)
abort():    entries at cut 14 -> 14 after 3s  (0)
```

Backpressure stalls it quickly, so the absolute cost is small - this is correctness rather than an emergency. Suggest `archive.abort()` then `archive.destroy()`, since `abort()` alone won't tear down the stream.

Related, non-blocking: `finalize()`'s promise never settles after either `destroy()` or `abort()`, so this IIFE never reaches its `finally` and the `removeEventListener` never runs. I checked whether that accumulates - 40 consecutive cancelled exports showed no heap growth, so it isn't worth chasing.

src/shared/lib/services/agent-template-service.test.ts
```diff
@@ -1931,6 +1968,49 @@ describe('exportAgentFull', () => {
     )
   })
 
+  it('rejects a second export while the first stream is still open', async () => {
```
iddogino, line 1987: **Blocking.** This covers the happy-path lock, but the cancel path has no coverage - and that's the highest-risk behavior in the PR, since a lock that fails to release is a permanent host-wide wedge. It's also the item still unticked on the PR checklist ("Cancel the download mid-flight - host walk stops").

It's cheap to test. This repo already has the real-socket pattern in `src/api/middleware/arm-abort-signal.test.ts` and `src/api/routes/notifications.stream.test.ts`: stand up `@hono/node-server`, start the export, read a few hundred KB, `req.destroy()`, then assert `isHostExportBusy()` goes false. Mine ran in ~5s.

Worth adding alongside it: a test for the leak flagged on `agents.ts` - stub a throw between the export call and the `Response` and assert the next export still succeeds. That one keeps working as a regression test even if the route is later restructured.

src/shared/lib/services/agent-template-service.ts
```diff
@@ -254,81 +256,136 @@ async function walkFullExportFiles(workspaceDir: string): Promise<string[]> {
...
+    return await fn()
+  } catch (err) {
+    endHostExport()
+    throw err
   }
+}
 
-  const templateFiles = await walkTemplateFiles(workspaceDir)
+// Queue files into an archiver and return it immediately so the HTTP response
+// can start flushing. Level 1: level 9 pegged 0.5 vCPU hosts on large exports.
```
iddogino: Non-blocking, but worth reconsidering the scope: this applies to template export too, not just full export.

Measured on this repo's own `src/` tree (2049 files, 19.9 MB - about as close to a real mixed agent workspace as I could get locally):

```
level 1:  5.81 MB   764 ms
level 9:  5.20 MB   754 ms     -> level 1 is 12% bigger and not faster
```

And on mixed-entropy synthetic data, level 9 was 13% slower for a 0% size gain. My machine is IO-bound here, so this doesn't refute the 0.5 vCPU claim - but that claim isn't demonstrated anywhere in the PR either, and `.agent` templates are the small shareable artifact where size matters more than host CPU.

Suggest keeping level 1 (or zlib's default 6) for full export and leaving template export at 9 - or, if you'd rather keep it uniform, dropping the measured host-CPU numbers into the PR body so the tradeoff is on the record.

src/renderer/hooks/use-agent-templates.ts
```diff
@@ -96,6 +131,12 @@ export function useExportAgentFull() {
 
       await downloadBlob(res, `${agentName || agentSlug}-full${AGENT_PACKAGE_EXTENSION}`)
```
iddogino, line 132: Non-blocking, but the PR description says "Download can start before the archive is finished" and that isn't true for this path - `downloadBlob` does `await res.blob()`, so the renderer buffers the entire zip before the save dialog appears.

The host-side win is real and I measured it (600 MiB workspace, 223 MB zip, 12 MB RSS growth). Just note that in desktop mode the memory moves to the renderer rather than disappearing; the reported OOM was on a cloud host where the browser is on someone else's machine, so the fix does address the actual incident. Worth correcting the claim in the description.

Also gone with the buffer: `Content-Length`, so browsers and `curl -O` lose download progress. Inherent to streaming - just flagging it.

src/shared/lib/services/agent-template-service.ts
```diff
@@ -254,81 +256,136 @@ async function walkFullExportFiles(workspaceDir: string): Promise<string[]> {
...
+    super('An export is already in progress')
+    this.name = 'ExportInProgressError'
   }
+}
 
-  const claudeMdPath = getAgentClaudeMdPath(agentSlug)
-  const claudeMdContent = await readFileOrNull(claudeMdPath)
-  if (!claudeMdContent) {
-    throw new Error('CLAUDE.md not found in agent workspace')
+let hostExportBusy = false
```
iddogino, line 279: Non-blocking, wanted to make sure it's a deliberate call: this is host-wide, so in multi-user auth mode one person's export blocks everyone else's, and the 409 text ("An export is already in progress") doesn't tell the second user it's someone else's. `/api/agents/export-status` is `Authenticated()` but not admin-gated, so any signed-in user sees the flag and, through the UI, has both buttons disabled by an export they didn't start.

Given the 1 GB host that motivated this, serializing is probably right. A per-agent lock plus a host-wide concurrency cap of 1 would give the same protection with a message that can distinguish the two cases.

src/renderer/hooks/use-agent-templates.ts
```diff
@@ -61,7 +68,28 @@ export function useDiscoverableAgents() {
...
 
+export function useHostExportStatus() {
+  return useQuery<HostExportStatus>({
+    queryKey: HOST_EXPORT_STATUS_QUERY_KEY,
+    queryFn: async () => {
+      const res = await apiFetch('/api/agents/export-status')
+      if (!res.ok) throw new Error('Failed to fetch export status')
+      return hostExportStatusSchema.parse(await res.json())
+    },
+    refetchInterval: (query) => (query.state.data?.inProgress ? HOST_EXPORT_STATUS_POLL_MS : false),
```
iddogino, line 79: Minor: polling stops once `inProgress` is false, so a tab that already has Settings open when someone else starts an export won't notice until the query refetches (staleTime is 5s, `refetchOnWindowFocus` is off in `query-client.ts:79`). It self-heals on close/reopen, which is the case the PR is actually targeting, so this is just noting the window rather than asking for a change.

iddogino (changes requested):
The core fix is real and I measured it working. Three things I'd like fixed first, one of which the code's own comment claims to have handled but doesn't.

### What I verified

Built a harness running the real service behind a real `@hono/node-server` with real socket hangups (following the pattern in `src/api/middleware/arm-abort-signal.test.ts`):

| Check | Result |
|---|---|
| 600 MiB workspace -> 223 MB zip | **12 MB RSS growth** (was O(zip) plus `Buffer.concat` doubling) |
| 40 consecutive cancelled downloads | lock always released < 5s, **no heap growth** |
| `/export-status` registered before `/:id` | correct - the route test proves `getAgent` is never called |
| typecheck / eslint / unit | exit 0 / exit 0 (13 pre-existing warnings elsewhere) / 521 pass |

The OOM fix is genuine, and lock release on client disconnect holds up under repeated real hangups - that was my main worry going in.

### Blocking

1. The host export lock leaks permanently if anything throws between the export call and the `Response` (see inline on `agents.ts`). I reproduced the consequence end to end: one 500 wedges *every* export on the host, for every user, until process restart.
2. `archive.destroy()` should be `archive.abort()` - archiver has an explicit abort that kills the pending queues; `destroy()` does not (measured inline).
3. No test covers the cancel path, which is the highest-risk behavior in the PR and is still unticked on your own checklist.

### Non-blocking, worth a decision

Inline comments on the zlib level change (applies to template export too), the `downloadBlob` buffering vs. the "download can start before the archive is finished" claim, the host-wide lock in multi-user auth mode, and the status-polling window.

### CI

`e2e-tests` failed on `session-long-sleep.spec.ts:98` through both retries. No shared code with this diff. `main`'s recent Tests runs are green and the only other recent failure is a different job on a different branch, so I don't have enough history to call it a known flake - I'd just re-run rather than assert it.

src/shared/lib/services/agent-template-service.ts
yiw190: Yeah, deliberate. Left a one-liner on the lock.

src/shared/lib/services/agent-template-service.ts
yiw190: Good call. Split: full stays at 1, template back to 9.

src/shared/lib/services/agent-template-service.ts
yiw190: Done — abort() first, then destroy().

src/api/routes/agents.ts
yiw190: Both routes now destroy the stream if anything after export* throws, so the lock can’t stick. Added a test that forces the throw and checks the next export still works.

src/renderer/hooks/use-agent-templates.ts
yiw190: Fixed the PR text.

src/shared/lib/services/agent-template-service.test.ts
yiw190: Added the test for cancel and it passed.

src/shared/lib/services/agent-template-service.ts
```diff
@@ -254,81 +256,142 @@ async function walkFullExportFiles(workspaceDir: string): Promise<string[]> {
...
+  const release = () => {
+    if (released) return
+    released = true
+    endHostExport()
   }
 
-  const templateFiles = await walkTemplateFiles(workspaceDir)
+  const archive = archiver('zip', { zlib: { level: zlibLevel } })
+  archive.once('close', release)
+  archive.once('error', release)
```
iddogino: `once` should be `on`. After the first error event there's no `error` listener left on the archive, and a second one becomes an `uncaughtException` -> `src/main/index.ts:1820` -> `app.quit()`.

This is reachable specifically through the new `sendLockedExportStream` destroy path, because that's the one path where `packageDownloadResponse` never ran, so `Readable.toWeb` never attached its own error handling and this `once` is the *only* listener. `zipStream.destroy()` also doesn't go through `stopArchive`, so the file queue keeps opening files.

Scenario: an export is streaming, `build()` throws, and the agent gets deleted (`deleteAgent` rm -rf's the agent dir, `agent-service.ts:365`, with nothing coordinating against the export lock). The queue then opens files from a directory that's gone and archiver emits a burst of ENOENT errors - I measured 4 to 193 per run. `once` eats the first; the second kills the process.

Measured on `d9d331a`, no instrumentation on the archive (an extra listener is itself a permanent handler and masks this - it fooled me on my first pass):

```
normal streamed export + agent deleted mid-zip   -> 0/6 runs crashed   (toWeb's handler covers it)
sendLockedExportStream destroy path, same delete -> 6/6 runs crashed
```

Deterministic, not a race. With `archive.on('error', release)` and nothing else changed, the same harness goes to **0/6**. `release` is already idempotent via the `released` flag, so `on` is safe, and it's what the pre-PR code did (`archive.on('error', reject)`).

Optional extra, not needed for the crash: `sendLockedExportStream` could route through `stopArchive` so the external destroy also aborts the queue. I measured ~180 ms of CPU still burned after the route's bare `destroy()`. Small, and `on('error')` alone makes it safe - your call.

iddogino (changes requested):
Re-reviewed `d9d331a`. **All three blockers are fixed**, and I verified each independently rather than reading the diff - re-running the same harnesses I used the first time:

| Fix | Independent check |
|---|---|
| Lock leak on throw after export | forced throw -> 500, **lock released**, next export 200 with a valid 7.87 MB zip |
| `abort()` before `destroy()` | entries processed after the cut: **4 -> 0** |
| Cancel-path tests | 3 new service tests incl. a real `@hono/node-server` socket hangup through the real `armAbortSignal` - exactly the gap |
| zlib split (non-blocking ask) | identical content: template 43.9 KB (L9) vs full 86 KB (L1) - the per-call level is live |

526 tests pass, typecheck exit 0, eslint exit 0. Thanks for the PR description rewrite too - the renderer-still-buffers note, the dropped `Content-Length`, and the multi-user lock consequence are all stated plainly now.

### One new thing, and then I think this is done

`sendLockedExportStream` is a good shape, but on its throw path it turns a leaked lock into a **process quit**. Details inline on `agent-template-service.ts:324` - it's a one-word change (`once` -> `on`) and I verified the fix flips it from 6/6 crashing to 6/6 clean.

I went back and forth on whether this is worth another round given the trigger is the same currently-unreachable throw that `sendLockedExportStream` exists to guard. What decided it for me: `process.on('uncaughtException')` in `src/main/index.ts:1820` calls `app.quit()`, so the failure mode of the new safety net is worse than the bug it catches. One word, already proven, so the round-trip should be cheap.

Nothing else outstanding from my side.

iddogino (approved):
Pushed the one-word fix myself as `d2f9ba7c` (you asked, so I didn't send it back for another round).

`archive.once('error', release)` -> `archive.on('error', release)`, plus a comment on the line so it doesn't get tidied back. Nothing else touched - the diff is 5 lines.

Verified before pushing, same harness as the review, no instrumentation on the archive:

```
sendLockedExportStream destroy path + agent deleted mid-zip:  6/6 crashed -> 0/6
normal streamed export, same delete:                          0/3 (unchanged)
```

526 tests pass, typecheck exit 0, eslint exit 0.

That clears the last item from my side - approving. Two things still outstanding that aren't mine to call:

- **`e2e-tests` needs to go green.** It was red on `session-long-sleep.spec.ts:98` before my push. Unrelated to this diff, and `main` is green, but I don't have the run history to call it a known flake - worth a look at the new run rather than merging past it.
- **The branch is behind `main`** (`5a04482b`, the 0.5.9-rc.1 bump). Worth a rebase so what merges is what CI actually tested.

Nice work on this one - the streaming fix measures well (600 MiB workspace -> 223 MB zip, 12 MB RSS growth) and the cancel-path tests you added are the real thing.

PR #822, opened 2026-08-21
fix: restore the cloud proxy prefix on dashboard urls (SUP-646)

JeremyBischoff:
**Footprint:** 3 files, +263/-4. One line of existing code changes.

## The bug

In the desktop app driving a cloud workspace, any dashboard that fetches a same-origin subresource renders blank. Confirmed live: `open-slide-studio` returned a 200 document and 404 on `main.tsx` and `@vite/client`, leaving `#root` empty.

The desktop reaches the deployment through `/cloud/<key>` on the loopback origin. The deployment is never told that prefix exists, so it writes the dashboard's mount unprefixed into the markup and into every module specifier. Root-absolute urls ignore `<base>`, so those requests land on the local API, which has no such agent.

The predicate is whether the dashboard fetches anything same-origin, not which toolchain built it. Our own tooling emits single self-contained files, which is why this went unnoticed. Relative urls do not avoid it either: the injected `<base>` names the mount without the prefix, and Chromium's preload scanner resolves against that before the inline runtime can correct it.

## The change

`cloud-proxy.ts` substitutes the prefix back onto occurrences of the request's own mount, in html, javascript and css bodies. The replaced string is the mount the request was served from, so an occurrence of it in that body is by construction a url pointing back at the same dashboard. Nothing is parsed and nothing is inferred.

Applied at the handler's exit rather than threaded through `forwardRequest`, so `forwardRequest` and `toClientResponse` are byte-identical to `main` and the streaming path is untouched.

## Verification

- `dashboard-proxy-prefix.test.ts` drives the real proxy and the real runtime injection with no network, modelling the browser as the preload scanner behaves. Three cases: document asset urls, a relative asset url, and the entry module's import specifiers. All three were red before the fix and pass after, unedited.
- The real proxy on a real socket against a live deployment, driven by headless Chromium at the real prefix: dashboard mounts, zero 4xx.
- By hand in a dev desktop build driving a cloud workspace, dashboard rendering in the real iframe.
- Full suite: 599 files, 9838 tests. Typecheck and lint clean.

## Review notes

- **Bodies are buffered, not streamed**, for dashboard html/js/css. That is the only way to substitute across chunk boundaries. Everything else, including SSE, still streams untouched.
- **JSON is deliberately excluded.** A mount inside a data payload is not necessarily a url. A dashboard that fetches self-referential urls as JSON would still break.
- **Sourcemap offsets shift** after substitution. Dev-server dashboards only.
- **`etag` is dropped** on rewritten responses, because the bytes no longer match what upstream validated. That makes revalidation the common case, so a bodyless `304` bypasses the rewrite rather than being rebuilt — reconstructing a null-body status throws, and without the guard the renderer got a 500 instead of a 304.
- `docs/cloud-workspace.md` claimed documents come back unchanged. That claim is now false, so it is corrected in place; the surrounding text about injected shims deriving their own prefix is untouched.

## What this does not fix

The prefix is an authentication mechanism occupying the application's url namespace. This substitution works and is contained, but it inherits whatever a future dashboard invents. SUP-646 stays open with the architecture options and a spike result on the most promising one.

### DIFF
--- a/docs/cloud-workspace.md
+++ b/docs/cloud-workspace.md
@@ -168,9 +168,12 @@ Other properties worth not regressing:
   `cloud-proxy.integration.test.ts` is for.
 - **Only `/api` paths are forwarded**; anything else 404s, as does a bad key
   (a prober learns nothing either way).
-- **Documents served through the proxy don't inherit its prefix.** A dashboard
-  iframe is loaded from the keyed URL, but the HTML comes back unchanged, and a
-  root-relative `/api/...` inside it resolves against the *loopback* origin. The
+- **A dashboard's own urls get the prefix back; nothing else does.** A dashboard
+  iframe is loaded from the keyed URL, and `cloud-proxy.ts` substitutes the
+  prefix onto occurrences of that dashboard's own mount in html, javascript and
+  css bodies — without it a root-absolute entry module resolves against the
+  *loopback* origin, 404s on the local API, and the dashboard renders blank. Any
+  other `/api/...` inside the document is left alone. The
   LLM and speech shims injected into dashboards therefore derive their prefix
   from `location` at runtime (`api/polyfill-api-prefix.ts`) instead of
   hardcoding it — otherwise a cloud dashboard's LLM calls would silently run on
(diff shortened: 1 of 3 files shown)


JeremyBischoff:
Superseded by https://github.com/SkillfulAgents/SuperAgent/pull/829. We iframe the cloud URL instead of rewriting through the door.

PR #834, opened 2026-08-21
perf(dashboards): fast-path the standalone popout wrapper

iddogino:
## Summary

Stacked on #833 (→ #832 → #831). Item 9 from the dashboard cold-start audit: the popout wrapper ran three serial round trips (artifacts → agent status → start) and then polled at a fixed 1 s × 120 before creating the iframe, making popped-out dashboards systematically slower than in-app ones.

- **Parallel seed**: the start POST fires immediately (idempotent, fast no-op for a running agent) with the artifacts fetch in parallel; the agent-status round trip is gone. The warm path paints from the single artifacts response without awaiting the start call.
- **Optimistic mount**: the iframe mounts hidden behind the spinner as soon as the start succeeds — #833's held-request proxy resolves the document the moment the dashboard server binds, so the fetch overlaps startup. The spinner drops on the first load event observed after `running`; a document that loaded earlier (a hold that timed out into an error body) is refetched exactly once — same recovery contract as the in-app view.
- **Poll cadence**: 300 ms (matching the in-app view) for the first 30 s, then 1 s backoff; overall budget slightly larger than the old 120 s.

## Test plan

- Rewrote the wrapper-script tests for the new flow (they execute the served script in a sandbox against a mocked DOM/fetch): warm path paints from one artifacts fetch with spinner-until-load, first-run install messaging still surfaces, and a new case covers the early-load → refetch-once-on-running recovery.
- Full `agents.test.ts` suite: 409 passed. `tsc --noEmit` and eslint clean.

### DIFF
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -648,18 +648,26 @@ describe('GET /:id/artifacts/:artifactSlug/view', () => {
     mockAgentExists.mockResolvedValue(true)
   })
 
-  it('reuses the initial running status instead of entering the poll loop', async () => {
-    const res = await getReq(createApp(), '/api/agents/test-agent/artifacts/sales/view')
-    const html = await res.text()
-    const fetchMock = vi.fn()
-      .mockResolvedValueOnce(new Response(JSON.stringify([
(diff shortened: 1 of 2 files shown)


yiw190 (approved):
LGTM

PR #837, opened 2026-08-21
Mount Team Brain for every agent (SUP-650)

JeremyBischoff:
The Team Brain is one shared folder on the host, beside the agent workspaces rather than inside them. Every agent box gets that same folder attached at start so members read it as files and only the curator writes. A member who wants a change asks the curator.

Closes https://linear.app/datawizz/issue/SUP-650/team-brain-v1-shared-pages-curator-persist-request-via-invoke

```
flag off
  no folder on start
  curator setting is hidden
  assigning a curator fails

flag on
  start the curator → folder is writable
  start anyone else → folder is readable only
    a write request reads the curator name from the folder and invokes that agent
  change the curator
    always stop the old writer first
      stop fails → nothing else changes
    then write the new name
    then restart the new curator if it is already running
  flip the flag
    restart the running curator only
    other running agents keep the folder they already have until they next start
```

The old host page and read tool are gone. A running member keeps its current attach across a flag change until it next starts. Cloud MicroVM needs the companion image in gamut-infra #196. Local and Kubernetes do not wait on that image, and either order boots. This does not add multiple brains, a registry, edit history, or a nav page.


agent-container/src/system-prompt.md
```diff
@@ -225,6 +228,27 @@ A memory that names a specific function, file, or flag is a claim that it existe
 
 A memory that summarizes repo state (activity logs, architecture snapshots) is frozen in time. If the user asks about *recent* or *current* state, prefer `git log` or reading the code over recalling the snapshot.
 
+<%#brainMounted%>
+## Team Brain
+
+This workspace has a shared Team Brain. It is not this agent's MEMORY.md.
+<%#brainCurator%>
+You are the Team Brain curator. The brain is the folder `/brains/global`, mounted read/write.
```
iddogino: The curator specific instructions feel like they belong in the curator's special sys prompt, not here...

JeremyBischoff:
Closing: Team Brain v1 pivoted to a Cloud Volumes design (opt-in shared read/write folders between agents, no curator). A new PR will replace this.

PR #849, opened 2026-08-24
Upload files on drop and fail stalls instead of hanging Send (SUP-653)

JeremyBischoff:
Attachments now upload the moment they land in the composer, one at a time, with each chip showing waiting, progress, done, or a retryable error. Send waits for the queue and retries failures. A 30-second stall fails that chip instead of hanging Send forever, which is why uploads go through XHR.

Closes https://linear.app/datawizz/issue/SUP-653

13 production files. Tests are extra.

```
drop / paste / pick
  → chip shows waiting
  → one upload at a time
  → bar (files) or spinner (folder copy / zip)
  → check, or error with the reason on hover
  └─ retry on the chip, or Send re-uploads failed chips then sends
```

Waiting is text, not a spinner. Nothing is happening to that chip yet.
The error reason is the same hover the chip already uses for the filename.
Mounts still attach at Send.
The agent still gets `[Attached files:]` plus workspace paths.
Create Agent no longer offers a paperclip. There is no workspace yet.

This does not change the 50MB chunk threshold.
It does not delete orphans when a chip is removed.
It does not touch the file-request card.

### DIFF
--- a/e2e/specs/composer-failure-recovery.spec.ts
+++ b/e2e/specs/composer-failure-recovery.spec.ts
@@ -18,9 +18,9 @@ function attachmentPreview(page: import('@playwright/test').Page, fileName: stri
  * - a failed message POST restores the typed text into the composer (the
  *   optimistic ghost is dropped, nothing lands in the transcript), and the
  *   restored text can be resent as-is once the server recovers;
- * - a failed attachment upload short-circuits BEFORE the composer is cleared,
- *   surfacing a dismissible error while both the text and the attachment
- *   chips stay intact for a retry.
+ * - a failed attachment upload (now on drop, not on Send) flags the chip
+ *   with a retry control, keeps the typed text, and lets the user retry
+ *   from the chip or from Send.
  *
  * Failures are injected per-page with route interception (POST-only — the
  * transcript GETs on the same URL shape must keep flowing), so the tests are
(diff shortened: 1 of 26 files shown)


iddogino (changes requested):
Really solid queue design overall — the generation counter + per-id removal set, chip-carried jobs (so enqueue can't race the render), the StrictMode-proof re-enqueue/carryover strip, and the `handleUnauthorizedResponse` extraction are all careful work, and the transport tests are thorough. One confirmed blocker and one integrity ask before merge.

### Blocker: a Send-driven retry drops the retried file's path from the message

`handleSubmit` reads the done paths from `attachmentsRef.current` immediately after `await queue.retryAndWait()`. The `done`/path update from `uploadOne` is an ordinary setState from a promise continuation — React commits it on a macrotask, while the `await` resumes on a microtask — so the ref is guaranteed stale at that read. The retry uploads successfully, the message sends, but `[Attached files:]` omits the retried file; with no text and a single errored chip the composer submits an empty string. The file lands in the workspace and the agent is never told about it.

The design already acknowledges this exact lag — `failedRef` exists as a synchronous mirror so the ok/fail decision doesn't read React state — the path read just has no equivalent mirror.

Why the existing tests don't catch it:

- the unit test "submit continues after a successful retry" splices the mock attachments array **in place** ("Same array object" comment) — the one thing the real `updateAttachment` never does, since it produces a new array via setState;
- the e2e spec retries from the chip and waits for `done` before clicking Send, so Send never performs the re-upload itself;
- any manual repro with a mount chip attached is masked, because the mount mutation yields real macrotasks that let React commit first.

I pushed `use-message-composer-send-retry.test.tsx` (red, commit 4573daaf) that pins the intended behavior: real `useAttachments` + `useUploadQueue` + `useMessageComposer` wiring, `uploadFile` fails once then succeeds settling from a macrotask like a real XHR `onload`, and the assertion is that the submitted content contains the workspace path. It should go green with the fix.

Suggested fix shape: keep a synchronous id→path record in the queue alongside `failedRef` (or have `retryAndWait` return the done paths), and let `handleSubmit` fall back to it when `a.upload?.path` is absent.

### Cross-agent path integrity is enforced at the edges, unchecked at the sink

`UploadState.agentSlug` is written on every transition and read nowhere — its doc comment ("a different agent must re-upload") describes a check that doesn't exist. Today the invariant holds via `requeueAll` on Quick Dispatch slug change and the carryover strip in `useAttachments` (both tested), but any future composer host that swaps agents without one of those fails silently: the message to agent B ships `[Attached files:]` paths that only exist in agent A's workspace (dangling reference plus a filename leak across agents — not content, since each container mounts only its own workspace).

Ask: enforce it where paths are harvested in `handleSubmit` — only append `a.upload.path` when `a.upload.agentSlug === agentSlug`. That turns the comment into code, makes the field earn its keep, and composes with the blocker fix (the sync mirror can carry the slug too).

### Minor

- **30s stall vs final-chunk assembly.** The stall timer re-arms only on `upload.onprogress`, so once the last byte is sent it becomes a 30s cap on server response time. The final chunk of a chunked upload triggers O(file-size) work: `storeUploadChunk` assembles up to 2GB (sequential read+write), then `moveUploadedFile` may do a full cross-device copy. Fast on a local SSD; on slow-storage deployments (NFS-backed) it can exceed 30s → false `UploadStalledError`, the server completes anyway (orphan file), and the retry re-uploads everything and stalls again. The code comment assumes only template import does post-last-byte work — chunked composer files >50MB do too. Consider disarming or substantially extending the stall bound once the body is fully sent, at least on the final chunk.
- **A wedged Electron folder copy still blocks Send.** `uploadFolder` (server-side `fs.cp`) has no signal and no stall bound, and the chain is serial — chips queued behind a hung copy stay `queued` even after the user removes the folder chip itself, keeping Send disabled until every queued chip is removed.
- **Quick Dispatch can enqueue with an empty slug** if a file is dropped before agents load (`selectedSlug ?? ''` → POST `/api/agents//upload-file`), flashing an error chip. It self-heals via `requeueAll` when the slug arrives, but it's cheap to guard.

JeremyBischoff:
Done in `848fc767` - Send after a retry now takes the path from a live list on the queue, only attaches a path that belongs to this agent, and Quick Dispatch does not upload until an agent is picked.

Leaving the 30s-after-last-byte clock: no repro, and that quiet stretch is already the response deadline. We will take it if a large file on slow storage actually stalls.

Leaving the hung folder copy: a running copy still cannot be cancelled, and waiting chips can be removed so Send is not stuck. We will take it if a wedged copy shows up.

iddogino (approved):

PR #850, opened 2026-08-24
fix: home page flashes before the getting-started wizard opens

yiw190:
## What was broken

On a fresh boot (setup not completed), the home page rendered for about a second and was then replaced by the getting-started wizard. The wizard auto-open decision has to wait on two settings fetches (`/api/user-settings`, `/api/settings`) plus the wizard chunk preload, and during that whole window `RootLayout` rendered the `<Outlet/>` — i.e. the home page — before swapping it out.

## Repro

1. Reset `setupCompleted` to `false` (or boot a fresh install).
2. Launch the app.
3. Home page appears briefly, then the wizard replaces it.

## What changed

`RootLayout` now latches a `bootSettled` flag once the auto-open decision has played out — wizard actually open, setup already completed, or the auth-mode non-admin case where the wizard never auto-opens. Until then the outlet is held behind a bare `data-boot-surface` div, so the home page never commits behind onboarding.

Safety rule: if either settings query errors (post-retry), the gate releases and the app boots without the wizard — the pre-fix behavior — instead of sticking on a blank surface.

Trade-off: every boot now waits for both settings fetches before home appears. Local boots resolve in well under a second (the gap reads as the plain window background); a slow cloud workspace shows the boot surface slightly longer than before.

## Testing

- `src/renderer/components/layout/`, `wizard/`, and `auth/` vitest suites pass (156 tests).
- Lint clean; typecheck failures are pre-existing in `tools/` PoC scripts (missing optional `e2b`/`@daytonaio/sdk` deps), untouched here.
- Not verified: a manual first-boot run watching the flash disappear.


iddogino:
@yiw190 I feel like that's a pretty bad tradeoff for **every single app launch**:
>Trade-off: every boot now waits for both settings fetches before home appears. Local boots resolve in well under a second (the gap reads as the plain window background); a slow cloud workspace shows the boot surface slightly longer than before.

Is there a way to build it statically on page serve? If not, the juice doesn't seem to be worth the squeeze

yiw190:
> @yiw190 I feel like that's a pretty bad tradeoff for **every single app launch**:
> 
> > Trade-off: every boot now waits for both settings fetches before home appears. Local boots resolve in well under a second (the gap reads as the plain window background); a slow cloud workspace shows the boot surface slightly longer than before.
> 
> Is there a way to build it statically on page serve? If not, the juice doesn't seem to be worth the squeeze

will look into it

yiw190:
Static injection at serve time isn't really doable here (Electron loads a static file, web entry is static too). A compromise: persist a "wizard settled" flag in localStorage and use it to decide who takes the serial wait — only the true first boot does. WDYT? @iddogino

iddogino:
@yiw190 how long is the flash ? trying to decide how much engineering is worth here...

yiw190:
> @yiw190 how long is the flash ? trying to decide how much engineering is worth here...

was about less than 1s

PR #855, opened 2026-08-25
Rename sessions, dashboards, folders, and agents in place (SUP-685)

JeremyBischoff:
Sessions, dashboards, folders, and agents now rename in place on the row instead of through a dialog. The field replaces the link so a double-click cannot open the dashboard and a text drag cannot pick up the agent. Enter or click-away saves, Escape discards, and a failed write stays in the field with a toast.

Closes https://linear.app/datawizz/issue/SUP-685/one-rename-dialog-for-sessions-dashboards-and-folders

- Right-click Rename, or a new folder mounts already editing
- The field replaces the row and selects the current name
- Enter or click away: empty or unchanged closes with no write, otherwise the write runs with the field locked
- Success closes the field and replays the click that left
- Failure keeps the field open, toasts, and drops that click
- Escape closes if nothing is in flight
- A dialog opened from the row does not start an agent drag

Out of scope: file preview, the home-title click-to-edit, scheduled-task titles, the Settings name field.

### DIFF
--- a/e2e/specs/session-rename.spec.ts
+++ b/e2e/specs/session-rename.spec.ts
@@ -49,11 +49,11 @@ test.describe('Session Rename (names must persist)', () => {
     await sessionItem.click({ button: 'right' })
     await page.locator('[data-testid="rename-session-item"]').click()
 
-    const dialog = page.getByRole('dialog', { name: 'Rename Session' })
-    await expect(dialog).toBeVisible()
-    await dialog.getByPlaceholder('Session name').fill(newName)
-    await dialog.getByRole('button', { name: /^Rename$/ }).click()
-    await expect(dialog).not.toBeVisible()
+    const input = page.getByLabel('Session name')
+    await expect(input).toBeVisible()
+    await input.fill(newName)
+    await input.press('Enter')
+    await expect(input).toHaveCount(0)
     await expect(sessionItem).toContainText(newName, { timeout: 10000 })
   }
 
--- a/src/renderer/components/agents/agent-context-menu.test.tsx
+++ b/src/renderer/components/agents/agent-context-menu.test.tsx
@@ -325,3 +325,22 @@ describe('moving an agent into a left-nav folder', () => {
     expect(screen.getByTestId('move-agent-to-new-folder-item')).toBeInTheDocument()
   })
 })
+
+describe('rename', () => {
+  it('stays off the menu until a row asks to start the field', () => {
+    render(<AgentContextMenu agent={AGENT}><span>row</span></AgentContextMenu>)
+    expect(screen.queryByTestId('rename-agent-item')).not.toBeInTheDocument()
+  })
+
+  it('calls onRenameRequest from the owner rename item', async () => {
+    const onRenameRequest = vi.fn()
+    render(
+      <AgentContextMenu agent={AGENT} onRenameRequest={onRenameRequest}>
+        <span>row</span>
+      </AgentContextMenu>,
+    )
+
+    await userEvent.click(screen.getByTestId('rename-agent-item'))
+    expect(onRenameRequest).toHaveBeenCalledTimes(1)
+  })
+})
--- a/src/renderer/components/agents/agent-context-menu.test.tsx
+++ b/src/renderer/components/agents/agent-context-menu.test.tsx
@@ -17,8 +17,9 @@ vi.mock('@renderer/hooks/use-agents', () => ({
 vi.mock('./agent-settings-dialog', () => ({ AgentSettingsDialog: () => null }))
 vi.mock('@tanstack/react-router', () => ({ useNavigate: () => vi.fn() }))
 vi.mock('@tanstack/react-query', () => ({ useQueryClient: () => ({ invalidateQueries: vi.fn() }) }))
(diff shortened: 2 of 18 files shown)


iddogino (changes requested):
Reviewed on the branch. `tsc --noEmit` clean, eslint clean on all changed files, 101 unit tests pass across the five affected specs, CI green.

The core move is right, and I confirmed the mechanism rather than taking the description's word for it (notes at the bottom). Two things I'd like fixed first — both reproduced with throwaway tests.

## Should fix

### 1. A live name update wipes what the user typed — `src/renderer/components/ui/rename-dialog.tsx:73`

```js
useLayoutEffect(() => { if (open) setName(currentName) }, [open, currentName])
```

This re-runs on every `currentName` change, not just on the open edge. Sessions are auto-named by the summarizer (`src/api/routes/agents.ts:1184-1222`) and `['sessions']` is invalidated on stream events, so a session renamed shortly after its first message can have its generated title land mid-edit and silently replace the typed text. Reproduced:

```
render <RenameDialog open currentName="New Session" …/>
type "Quarterly numbers"
push currentName="Reviewing the Q3 spreadsheet"   // what the summarizer does

Expected: "Quarterly numbers"
Received: "Reviewing the Q3 spreadsheet"
```

The code this replaces was immune — it snapshotted the name at menu-click time and never touched it again. Gating on the open transition with a `wasOpen` ref fixes it, or fold the reset into the generation effect just below, which already tracks that edge.

### 2. Folder creation lost its select-on-open — `src/renderer/components/ui/rename-dialog.tsx:115`

Both removed inline inputs called `inputRef.current?.select()`. The new field only `autoFocus`es, and the caret lands at the end:

```
selection: 10 10 "New Folder"  →  type "Clients"  →  "New FolderClients"
```

That lands on the create-a-folder gesture specifically: `handleCreateFolder` seeds `uniqueFolderName()` and its comment says creating a folder and naming it is meant to be one gesture. The e2e can't catch it because `input.fill()` clears first.

One line — `onFocus={(e) => e.currentTarget.select()}` on the `Input` — and it improves the session and dashboard paths too.

## Nits (not blocking)

- `submit`'s reentrancy guard reads `isPending` from the render closure (`rename-dialog.tsx:88`); the implementations it replaces used `submittedRef` / `settledRef`. Low risk — the button is disabled and separate keydowns land in separate tasks — but a ref is free.
- Folder and dashboard rename no longer commit on blur / outside click; they discard. The description states this so I read it as deliberate — flagging only because it's the one user-visible behavior change that isn't a bug fix.
- The folder input lost `aria-label="Folder name"`; its accessible name now comes from the placeholder alone. axe accepts a non-empty placeholder so the a11y job stays green, but an explicit label is stronger.
- Failures are console-only. Keeping the dialog open beats the old silent close, but there is still no message. Worth noting that folder rename cannot fail at all — `handleRenameFolder` is a fire-and-forget `updateSettings.mutate` — so the pending / failure machinery is inert on that path.
- `scrollIntoViewIfNeeded` and the assertion message in `agent-folders.spec.ts:561` are an unrelated flake fix riding along.
- Speculative, not observed: `createFolder` now returns as soon as the renamed row is visible, and the row stays mounted throughout, so the modal overlay may still be fading out (~150ms) when the next raw `page.mouse` drag fires. `dragRowOnto`'s 3-attempt retry should absorb it and CI is green — naming the vector in case that spec gets flaky later.

## What I checked on the mechanism

- The drag listeners really do land on the agent `<li>`: `SortableAgentMenuItem` spreads `{...listeners}` into `AgentMenuItem`, which puts them on `SidebarMenuItem` via `...rest` (`app-sidebar.tsx:431`). Session and dashboard sub-items — and so their dialogs — are React descendants of that `<li>`, and with `POINTER_SENSOR_OPTIONS = { distance: 5 }` (`app-sidebar.tsx:145`) a 5px text drag-select inside the field is exactly enough to pick up the agent. Diagnosis holds.
- `RowPointerSensor` drops dnd-kit's third `activationContext` argument, which is safe: `PointerSensor`'s handler destructures only `nativeEvent` and `onActivation` (`@dnd-kit/core/dist/core.cjs.development.js:1637`).
- The guard is the same one dnd-kit's own `KeyboardSensor` already applies (`event.target !== active.activatorNode.current`, line 1348) — which is why keyboard never had this bug.
- Bonus fix the description doesn't claim: the agent *and* folder Radix context-menu items are also portaled inside the listener subtree, so presses on those no longer arm a drag either.
- Escape still dismisses despite `DialogContent`'s `onKeyDown` `stopPropagation`, because Radix listens on the document with `capture: true` (`@radix-ui/react-use-escape-keydown/dist/index.js:48`). That same `stopPropagation` is what keeps Enter/Space in the field from reaching the row's KeyboardSensor activator. Both halves check out.
- `invalidateAgentArtifacts` is strictly better than the old literal `['artifacts', agentSlug]` key — it covers `displaySlug` aliases.
- Only one `DndContext`/`PointerSensor` exists in the renderer, so the sensor swap is complete; and `e2e/specs/session-rename.spec.ts` keys (dialog name, `Session name` placeholder, `Rename` button) are untouched, as claimed.

JeremyBischoff:
Done in `3f3f3d077` - the field now resets only when the dialog opens, and it selects the seeded name on focus.

PR #861, opened 2026-08-25
feat(sessions): add "Mark as Unread" to the session context menu

iddogino:
## What

Adds a **Mark as Unread** option to the session right-click menu, bringing the blue dot back on a session until you reopen it.

Linear: [SUP-686](https://linear.app/datawizz/issue/SUP-686/feat-mark-session-as-unread)

## Why this shape

The sidebar dot isn't a session field — it's a projection of unread rows in the `notifications` table, joined onto the session list as `hasUnreadNotifications`. The obvious implementation (flip those rows back to `isRead: false`) was rejected:

- read state on notification rows is **shared across all users** (deliberately — see the comment in `schema.ts`), and it drives the notification inbox and its badge, so un-reading would resurface an old "Session complete" entry for everyone;
- a session that never produced a `session_complete` / `session_waiting` notification has **no row to flip**, so the menu item would silently do nothing.

Instead this adds a `session_unread_marks` table — a second source of truth for the dot, OR'd into the existing unread projection. The notification inbox, the unread count, and the dock badge are untouched: this drives the dot only.

Marks sit in the database rather than on session metadata because the dot is projected on polled endpoints. Session metadata is a per-agent JSON map, so answering "is anything marked?" from it costs a file read plus a Zod parse of every session's entry — and the api-perf suite pins the notable fast path at **zero file reads**. See Performance.

They are also **per-user**, unlike notification read state. That is only affordable because of the storage choice: in a JSON map read on every poll, per-user meant a second dimension in the map; in a table it is a column and a composite key.

## Changes

- `session_unread_marks` table (migration `0037`, PK `(session_id, user_id)`) + `session-unread-service`, with a batched `getSessionIdsMarkedUnreadByAgents` mirroring `getUnreadNotificationsByAgents` for the agents list. Every read and write is scoped to `getCurrentUserId(c)` — the `'local'` sentinel outside auth mode, so no FK to `user` (same reason as `push_subscriptions`).
- `POST` / `DELETE /api/agents/:id/sessions/:sessionId/unread`, guarded by `sessionIsKnown` so an unknown id can't be marked.
- Marks are dropped on session delete and auto-delete next to the notification rows; the table joins `FACTORY_RESET_TABLES` and the per-agent cleanup.
- Menu item + `useSetSessionMarkedUnread` hook.

### The flag is OR'd into every consumer of the unread projection

A mark is a second source of truth for one dot, so the thing worth reviewing is whether it reaches everywhere the notification set does. It's five places, all in `agents.ts`:

| Consumer | Surface |
|---|---|
| agent rollup in `enrichAgentsWithSummary` | sidebar agent dot, home cards |
| full session list (`listSessionsFromSummary`) | sidebar session rows, agent home |
| `?notable=true` fast path | badge/toolbar consumers |
| `getLatestVisibleSessionTail` | `latestVisibleSession` (iOS) |
| `getAttentionOutsideLatest` | `attentionOutsideLatest` (iOS) |

The last two arrived on `main` while this branch was open and are wired in the merge commit. The notable path needs the ids in the **fetch set** passed to `listSessionsByIds`, not just in the projection, or the session is never statted at all. Hidden automated sessions are excluded from the marked set for the same reason unread notifications are: they appear in no session list, so a dot raised by one could never be cleared.

Nothing outside `agents.ts` computes `hasUnreadNotifications` — every renderer consumer reads the API field, so there are no renderer-side projections to keep in sync.

## Decisions worth a look

**Clearing happens on the SessionView mount effect, not the `visibilitychange` handler next to it.** Both currently mark notifications read. Only mount clears `markedUnread` — otherwise marking the session you're *currently looking at* would be undone by the next window refocus. As written, the dot appears immediately and survives until you actually navigate away and come back, which is what the ticket asks for.

**Both verbs are `AgentRead()`,** unusually for writes under this path. A mark raises a dot on the acting user's sidebar and nowhere else, and only they can clear it — there is no shared state for a permission to protect, and gating it higher would leave a read-only viewer unable to dismiss their own dot. The menu item is correspondingly ungated.

**Marks are per-user.** A teammate opening the session does not clear your reminder, and your reminder does not appear on their sidebar. This is the one place the behaviour deliberately diverges from notification read state, which stays shared because it records a team-visible acknowledgement rather than one person's intent.

Single-user installs are just the one-user case: `getCurrentUserId()` returns the `'local'` sentinel outside auth mode, never nothing, so marks behave identically there. The acting user is a **required** field on the summary projection rather than an optional one — a caller that omitted it would silently drop marks from the agent rollup, so that is a compile error instead.

**Known gaps, accepted:** "Mark all as read" in the inbox doesn't clear marks — those dots survive until the session is reopened, which is the ticket's contract. And no SSE event is emitted for a mark, so another window picks it up on its next refetch rather than instantly.

## Performance

This is the part that shaped the design. The api-perf suite (SUP-658) pins each route's fs op counts exactly; the notable fast path's budget is `{ totalOps: 1, ops: { stat: 1 } }` — no file reads at all, because with nothing notable the id set is empty and `listSessionsByIds` returns before it stats anything.

A first cut kept the flag on session metadata, which broke that: finding marked sessions meant an unconditional `readSessionMetadata`, and the map is read + `JSON.parse`d + Zod-validated whole. Measured on an M-series Mac, warm page cache:

```
   100 sessions    0.04 MB   0.64 ms/read
 1,000 sessions    0.36 MB   2.67 ms/read
 5,000 sessions    1.83 MB  13.57 ms/read
20,000 sessions    7.31 MB  57.22 ms/read
```

A table lookup costs no filesystem work, so the marks moved there. **Every perf scenario now lands exactly at its existing budget — no budget was loosened:**

```
big:   sessions notable      ops=1  (budget 1)
big:   sessions page         ops=3  (budget 3)
big:   GET /api/agents warm  ops=15 (budget 15)
big:   home expansion warm   ops=22 (budget 22)
```

The one genuinely new operation is the clear that fires on every session open: one `stat` (the `sessionIsKnown` guard) plus one `DELETE` that usually matches no rows. The route reports whether anything was written as `changed`, and the mutation skips its cache invalidation when nothing was — otherwise every session open would refetch the session list and re-enrich every agent for a no-op.

## Testing

- `npm run test:run` — 10421 passed, 0 failures
- `npm run typecheck` — clean
- `npm run lint` — 0 errors

New: 13 service tests against a migrated in-memory DB (raise/clear round trip, agent scoping, `changed` return value, re-marking keeps the original timestamp, batched grouping, delete cleanup, and four on per-user isolation — one user's mark is invisible to another, one user opening the session leaves the other's mark standing, two users can hold a mark on the same session), 16 route tests (both verbs, 404 guard, no route collision, `changed` in the response, all five projections, the acting user reaching each of them, session-delete cleanup, and that the notable path never reads session metadata), 3 hook tests (no invalidation on a no-op, both invalidations when changed, verb selection), 4 context-menu tests (raises the mark, visible to members, visible to read-only viewers, hidden for live sessions), plus the perf suite covering the op-count contract.

Not covered: no E2E spec — the unread dot has no testid and there's no existing e2e coverage of it — and no visual verification in a running app, flagging that rather than claiming the dot was seen on screen. The `AgentUser` gate on `POST` also isn't directly asserted: `agents.test.ts` blanket-mocks the auth middleware as pass-through, so it's covered indirectly through the menu's `canUseAgent` test.


iddogino:
Icon swapped per review feedback — the **Mark as Unread** item now uses lucide's [`message-square-dot`](https://lucide.dev/icons/message-square-dot) instead of the blue filled `Circle`.

Pushed as `a867a38`. It also drops the hard-coded `fill-blue-500 text-blue-500`, so the icon inherits the menu's foreground colour like every other item (and stays correct in dark mode).

```diff
-import { Trash2, ClipboardCopy, Pencil, Circle } from 'lucide-react'
+import { Trash2, ClipboardCopy, Pencil, MessageSquareDot } from 'lucide-react'
...
-  <Circle className="h-4 w-4 mr-2 fill-blue-500 text-blue-500" />
+  <MessageSquareDot className="h-4 w-4 mr-2" />
```

Screenshot of the rendered menu is attached on the Linear ticket (GitHub's API can't upload images to comments): https://linear.app/datawizz/issue/SUP-686/feat-mark-session-as-unread#comment-94c60076

Typecheck clean, lint 0 errors, `session-context-menu` suite 6/6 green. The existing tests key off `data-testid`, not the icon, so no test changes were needed.

PR #862, opened 2026-08-25
Share one origin strip, and use registration time when a session is opened (SUP-247)

JeremyBischoff:
The three origin strips above a chat (scheduled job, webhook, x-agent) now share one banner, so the next origin can reuse the same chrome instead of another pasted strip. Opening a session also reports the registration time the list already uses, not the first transcript line, and a stored value that is not a real date is ignored so the payload stays a date. The agent title header is unchanged.

Related to https://linear.app/datawizz/issue/SUP-247/session-forking

3 production files, net −41.

```
open a session
 ├─ scheduled / webhook / x-agent → same strip and back control as before
 ├─ ordinary chat → no strip
 └─ createdAt
     ├─ real stored time → that time
     └─ missing or unparsable → first transcript time
```

Fork Session rebases onto this branch.


iddogino (changes requested):
Reviewed locally on a clean worktree. The refactor itself is solid — two asks before merge, neither large.

## Verified

**The renderer half is a proven no-op.** I rendered `main`'s `session-view.tsx` and this branch's side by side under identical mocks across 8 permutations — scheduled/webhook/x-agent each with and without a name, all three stacked at once, and none. Markup was **byte-identical** every time, `data-testid` placement included. The duplication is genuinely gone: exactly one `border-b bg-background px-4 py-2` left in the renderer.

**The service half is correctly scoped.** `parseSessionInfo` feeds only `getSession`, reached by `GET`/`PATCH /api/agents/:id/sessions/:sessionId` (plus the test-only `findSessionAcrossAgents`). The list path already preferred `metadata.createdAt` via `resolveSessionCreatedAt`, so this makes the detail endpoint agree rather than inventing a new rule. The rename-on-an-unregistered-session case is real — `updateSessionMetadata` merges and never stamps `createdAt` — and the new test pins it.

`tsc --noEmit` exit 0, `eslint` on the four files exit 0, `session-service.test.ts` + `session-view.test.tsx` 148 passed.

## Requested changes

### 1. The "unparsable date is ignored" guard covers one of four read sites

`src/shared/lib/services/session-service.ts:526`

Probed rather than inferred. With `createdAt: "not-a-date"` in metadata:

| path | result |
|---|---|
| `getSession`, transcript exists | `2026-01-24T01:30:58.661Z` ✅ (the new guard) |
| `getSession`, registered, **no transcript** → `emptySessionFromMetadata` (:568) | `Invalid Date` → payload `createdAt: null` |
| `listSessions` → `resolveSessionCreatedAt` (:586) | `Invalid Date` → `createdAt: null` |
| `listSessionsByIds` → same | `Invalid Date` → `createdAt: null` |

So "a stored value that is not a real date is ignored so the payload stays a date" holds for one branch of `getSession` and not the other, and list vs. detail still disagree in exactly the corrupt case. None of it is a regression — all three were already unguarded — but `sessionMetadataSchema` types `createdAt` as a bare `z.string()` on purpose (lenient by design, per its header comment), so the value really can be anything on disk. One shared helper across all three sites closes it:

```ts
function parseStoredDate(value: string | undefined): Date | null {
  if (!value) return null
  const d = new Date(value)
  return Number.isNaN(d.getTime()) ? null : d
}
```

`Number.isNaN` also reads more directly than `Number.isFinite` here, and matches `message-list.tsx:510`.

### 2. `backLabel` is required but only ever read when `onBack` is set

`src/renderer/components/layout/session-provenance-banner.tsx:7`

The JSDoc one line below documents omitting `onBack` "to render the sentence with no back button" — but a caller doing that is still forced by the type to pass a meaningless `backLabel`. Make it optional, or pair them:

```ts
back?: { label: string; onClick: () => void }
```

## Non-blocking

- **Let the banner own the icon sizing.** `session-view.tsx:91,102,113` all pass the identical `className="h-3 w-3 shrink-0"`. If the point is that the chrome is shared, take `icon: LucideIcon` and render `<Icon className="h-3 w-3 shrink-0" />` — three duplicated class strings gone, and one less way for a future caller to drift.
- **The `onBack`-less branch has no caller and no test.** Fine as prep for the fork PR that rebases on this; flagging so it doesn't stay dark.

## CI

`e2e-tests` red on `home-graph.spec.ts:133 › the fade survives a selection change under the cursor` (1 failed, 4 flaky, 311 passed) — **not this PR**. `main`'s own runs fail the same specs: run `32904263307` (ba075af8a) hard-failed `home-graph.spec.ts:223`; `32902328737` flaked `home-graph.spec.ts:54` and `:168`. This branch touches nothing on that path. Unit, auth-e2e, a11y, container and both builds are green.

JeremyBischoff:
Done in `bfbfce1a` - shared helper on all three date reads, and back is `{ label, onClick }`. Left the no-back test for the fork PR.

PR #879, opened 2026-08-26
fix(notifications): reopen dead streams so new children keep a composer (SUP-697)

JeremyBischoff:
When another agent creates a child, a dead notification stream means the create announce never arrives, so the child lists but stays read-only until reload. The browser closes EventSource for good on a non-200 or non-event-stream reply, which is the Electron cloud-proxy 502 that started this. This PR reopens that wire after a fatal close.

Closes https://linear.app/datawizz/issue/SUP-697/reopen-dead-notification-streams-so-agent-created-children-keep-a

2 production files. Tests are extra.

```
stream dies (fatal close)
  → reopen
  → onopen refetches agents + rights
  → later agent_created still lands
```

The live red vs green on 2026-08-26 was that bug: kill the wire, create via `POST /api/x-agent/create`, open the child without reload. Red stayed view-only. Green had a composer.

Human `POST /api/agents` still does not announce. That path is out of scope.

### DIFF
--- a/src/renderer/components/notifications/global-notification-handler.tsx
+++ b/src/renderer/components/notifications/global-notification-handler.tsx
@@ -10,7 +10,7 @@
  * - Scheduled task updates - updates task list
  */
 
-import { useCallback, useEffect, useRef } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
 import { useQueryClient } from '@tanstack/react-query'
 import { getApiBaseUrl, isElectron } from '@renderer/lib/env'
 import { apiFetch } from '@renderer/lib/api'
@@ -46,6 +46,7 @@ import {
   USER_ACTIONABLE_NOTIFICATION_TYPES,
 } from '@shared/lib/notifications/notification-preferences'
 import { useRenderTracker } from '@renderer/lib/perf'
+import { STREAM_RECONNECT_MS, watchStreamLiveness } from '@renderer/lib/stream-liveness'
 
 // Optimistic status echo per session lifecycle event. Deliberately exhaustive
 // with NO fallback: an event type added to the lifecycle case group without a
@@ -92,6 +93,9 @@ export function GlobalNotificationHandler() {
   userSettingsRef.current = userSettings
   const canAccessAgentRef = useRef(canAccessAgent)
   canAccessAgentRef.current = canAccessAgent
+  const [reconnectKey, setReconnectKey] = useState(0)
+  const reconnectTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
+  const streamRef = useRef<EventSource | null>(null)
 
   // Sync dock badge count with unread notifications (macOS Electron only)
   useEffect(() => {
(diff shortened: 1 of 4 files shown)


src/renderer/lib/stream-liveness.ts
```diff
@@ -0,0 +1,24 @@
+export const STREAM_RECONNECT_MS = 1_000
```
iddogino, line 1: `STREAM_RECONNECT_MS` is a flat 1s retry with no backoff. In the exact scenario that motivated this PR — the cloud-proxy serving 502s — every reopen fatally closes almost immediately and retries again, so a down backend gets ~1 request/sec per open window until it recovers (before this PR the stream just stayed dead and sent nothing). Suggest a capped exponential backoff (1s → 2s → … → ~30s), reset to 1s on a successful open, so recovery stays fast without the hammer.

src/renderer/components/notifications/global-notification-handler.tsx
```diff
@@ -499,13 +504,27 @@ export function GlobalNotificationHandler() {
     }
 
     es.onerror = () => {
-      // EventSource will auto-reconnect; onopen above catches up.
+      // Network errors reconnect in the browser. Fatal close is the watcher.
     }
 
+    const disposeLiveness = watchStreamLiveness(es, () => {
```
iddogino, line 596: Catch-up after a fatal reconnect is partial: `onopen` invalidates only `['agents']` and `['my-agent-roles']`, but this stream feeds ~30 other invalidation sites in this file (notifications, sessions, pending-user-requests, scheduled-task queries, …). Anything in those families that happened during the dead window stays stale after reopen until its next live event — and the dead window can be arbitrarily long.

Suggestion: on reopen **after a fatal close specifically** (`reconnectKey > 0`, or a `wasDeadRef` set in the `onDead` callback — not every `onopen`, which also fires on mount and after transient browser auto-reconnects like sleep/wake), do a blanket `queryClient.invalidateQueries()` with a small exclusion `predicate` for known-expensive queries (e.g. the agent-template publish-info query, whose server side runs an LLM completion per fetch). `invalidateQueries()` only refetches actively-mounted queries, so the cost is one burst of GETs for the open screen after a real outage. This inverts the maintenance burden from keeping a ~30-key allowlist in sync with the message handler to a 1–2 key denylist.

src/renderer/lib/stream-liveness.ts
JeremyBischoff: Done in `5dadad59c` - the reopen wait now doubles per consecutive fatal close, 1s to a 30s cap, and resets to 1s on a successful open.

src/renderer/components/notifications/global-notification-handler.tsx
JeremyBischoff: Done in `5dadad59c` - the first open after a fatal close now invalidates every mounted query except the four LLM-backed ones (template and skill publish-info and pr-info). Mount and browser reconnects keep the narrow agents + roles refetch.

iddogino (approved):

PR #881, opened 2026-08-27
Show the cloud deployment version in the sidebar footer (SUP-701)

JeremyBischoff:
The Cloud tab still runs the desktop bundle, so the footer can show both versions. Local stays one number. When the builds differ, or when they match and either is behind the update feed, the pair shows a laptop and a cloud, with a blue or orange dot on the side that is behind.

Closes https://linear.app/datawizz/issue/SUP-701

8 files, +567 / −17.

- Same versions and nobody behind: one number
- Differ, or same and behind the feed: laptop left, cloud right
- Blue dot = patch behind the latest. Orange = major or minor behind
- Latest is the update feed when one exists, else the desktop build
- Left opens Settings. Right opens the platform Cloud tab
- Missing org or platform URL: right click does nothing
- Missing or unreadable cloud version: desktop number only
- Version comes from `GET /api/runtime-status`, not platform discovery
- A desktop GitHub release can read as latest before the cloud image is rollable
- Blue vs orange is color only

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -85,6 +85,7 @@
         "@types/react": "^18.2.0",
         "@types/react-dom": "^18.2.0",
         "@types/react-file-icon": "^1.0.4",
+        "@types/semver": "^7.8.0",
         "@types/web-push": "^3.6.4",
         "@types/ws": "^8.5.12",
         "@types/yauzl": "^2.10.3",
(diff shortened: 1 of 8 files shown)


iddogino:
@JeremyBischoff what would be the argument to show the cloud version and not the local version? the rendering code depends on local version, API depends on cloud version... What's the instinct on showing one versus the other ?

PR #883, opened 2026-08-27
Add a copy-contents action to the file preview header

iddogino:
Adds a one-click "copy file contents" action to the file preview header, next to Download.

Linear: [SUP-709](https://linear.app/datawizz/issue/SUP-709/when-viewing-a-markdown-text-file-in-the-app-offer-a-copy-file-option)

## What changed

- **`copy-file-button.tsx`** — new header action. Shows `ClipboardCopy`, swaps to a green `Check` for 2s on success, and toasts via sonner (`Copied contents of "notes.md"`). Errors surface as `toast.error` with the reason.
- **`lib/clipboard.ts`** — shared `copyTextToClipboard` / `copyLazyTextToClipboard`.
- **`file-types.ts`** — `isCopyableTextFile` inverted to a binary-extension denylist; new `looksBinary` content sniff.
- **`folder-file-context-menu.tsx`** — its existing "Copy contents" item now shares the same helper and binary guard instead of calling `navigator.clipboard.writeText` directly.

## On the three points in the ticket

**1. Generic text detection.** There is no extension list that covers every text file an agent might write, so the extension check now only rules formats *out* (images, video, audio, pdf, archives, office docs, executables, fonts, sqlite). Everything else — including extensionless files like `LICENSE` and `Procfile`, and unfamiliar extensions like `.zig` or `.hcl` — is treated as copyable. The real guard is content-based: `looksBinary` applies git's heuristic (a NUL byte in the first 8 KB) plus a check for content that decoded mostly to U+FFFD replacement characters. A wrong guess therefore costs an error toast, not garbage on the clipboard.

**2. Toast + icon toggle.** Both implemented; the icon reverts after 2s and the timeout is cleared on unmount.

**3. Browser clipboard rules.** This was the fiddly part. A browser only honours a clipboard write while the triggering user gesture is still live, and awaiting a fetch first spends that gesture — Safari then rejects with `NotAllowedError`. Two mitigations:

- The text renderers already load through the `['file-content', url]` react-query key, so for the file on screen the content is normally *already in the cache*. The button reads it from there and writes without awaiting anything — the shape browsers reliably accept.
- When the cache is cold, `copyLazyTextToClipboard` hands the pending text to `ClipboardItem` as a promise, which keeps the gesture alive across the fetch. Engines that reject promise-valued `ClipboardItem` fall back to await-then-`writeText`, and `navigator.clipboard` being absent entirely (non-secure context, e.g. a self-hosted workspace on plain HTTP) falls back to a hidden-textarea `execCommand` copy.

## Tests

- `copy-file-button.test.tsx` — cached-copy without refetch, icon toggle and revert, cold-cache fetch, binary refusal, failed load, truncated-file wording.
- `clipboard.test.ts` — `ClipboardItem` promise path, fallback on rejection (loader called once), `execCommand` path, and that a loader error is surfaced rather than the clipboard's generic one.
- `file-types.test.ts` — extensionless/unknown extensions now copyable, binary ones not; `looksBinary` cases including the 8 KB sample window.
- `file-preview.spec.ts` — new Playwright test asserting the real clipboard actually holds the file text after a click (this is what would catch a gesture regression), plus copy hidden on an image tab.

`npm run test:run` (10,553), `typecheck`, and `lint` pass. `file-preview.spec.ts` passes end-to-end in Chromium.

## Note

I could not open the screenshot attached to the ticket (Linear's uploads host isn't reachable from the agent's proxy), so I placed the button next to the existing Download action in the file-preview header — the app's only file-viewer toolbar. Happy to move it if the screenshot meant somewhere else.

### DIFF
--- a/e2e/specs/file-preview.spec.ts
+++ b/e2e/specs/file-preview.spec.ts
@@ -503,8 +503,39 @@ test.describe('File Preview', () => {
     await expect(markdown(page).getByRole('heading', { name: 'Report Content' })).toBeVisible({ timeout: 5000 })
     await expect(page.locator('img[alt="chart.png"]')).not.toBeVisible()
 
+    await expect(page.getByTestId('file-preview-copy')).toBeVisible()
+
     // Switch forward to the image tab again → image renderer returns.
     await fileTab(page, 'chart.png').click()
     await expect(page.locator('img[alt="chart.png"]')).toBeVisible({ timeout: 5000 })
+
+    // A PNG has nothing copyable, so the header drops the action.
+    await expect(page.getByTestId('file-preview-copy')).toHaveCount(0)
+  })
+
+  test('copies the open text file from the preview header', async ({ page }) => {
+    await page.context().grantPermissions(['clipboard-read', 'clipboard-write'])
+    await agentPage.createAgent(`CopyFile ${Date.now()}`)
+    const agentSlug = await getLatestAgentSlug(page)
+    seedWorkspaceFile(agentSlug, 'output/report.md', '# Test Report\n\nThis is a test with **bold** text.')
+
+    await sessionPage.sendMessage('deliver file')
+    await sessionPage.waitForResponse(15000)
+    const filePill = getFilePill(page, 'report.md').first()
+    await expect(filePill).toBeVisible({ timeout: 10000 })
+    await filePill.click()
+    await expect(markdown(page).getByRole('heading', { name: 'Test Report' })).toBeVisible({ timeout: 10000 })
+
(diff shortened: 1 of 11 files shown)


iddogino:
Follow-up note found while recording the demo: a `LICENSE` file (no extension) now correctly shows the copy button, but the preview pane underneath still renders "Preview is not available for this file type" — `file-renderer.tsx` still dispatches on the old `TEXT_EXTS` allowlist. That's pre-existing behaviour and I left it alone to keep this PR scoped to the copy action, but it's the obvious next change if we want extensionless files to preview as well as copy.

PR #894, opened 2026-08-28
Fork a session into a copy that keeps the conversation (SUP-247)

JeremyBischoff:
Right-click a session and Fork Session opens a copy that already has the conversation, the source model and effort, and whatever was sitting unsent in the composer. The original stays where it is. The copy is idle until you talk. A Forked from strip links back using the saved listing name, or just names the gap if that chat is gone.

Closes https://linear.app/datawizz/issue/SUP-247/session-forking

25 production files.

- Menu click snapshots the unsent draft
- Host refuses an unknown, mid-turn, or gone source
- The agent copies the transcript and we register the new session as "{name} (fork)"
- A copy of a scheduled or webhook chat is a normal user session
- Subagent files and sender badges copy when they can
- The app seeds the cache and draft, then opens the copy


iddogino (changes requested):
## Blocking

1. `src/api/routes/agents.ts:2821` — `forkedFromSessionName` comes from the parent's *metadata name* only, so a live transcript-only or nameless parent renders "Forked from a deleted session" after the first refetch — and `session-view.tsx:129` drops the Back link too, because it gates on the name instead of `forkedFromSessionId`.

2. `src/shared/lib/container/base-container-client.ts:1424` — every container 404 is read as "container predates the fork endpoint": a container that genuinely lost the session (empty/corrupt/schema-dropped `sessions.json` — all three paths exist in `session-persistence.ts`) gets perpetual "restart the agent" advice that cannot fix it. The two 404s are body-distinguishable (route miss = plain text, session miss = JSON).

3. `src/api/routes/agents.ts:2945` — the rollback's container `deleteSession` is a guaranteed no-op for a cold fork (`SessionManager.deleteSession` skips persistence for non-live sessions), orphaning a `sessions.json` record that points at a deleted transcript, with no GC path — the exact unowned state the container-side error path avoids.

4. `src/shared/lib/utils/file-storage.ts:851` — symlink safety is opt-in (`regularFilesOnly`), leaving all five legacy call sites (`agent-template-service.ts`, `skillset-service.ts`) dereferencing symlinks out of cloned third-party repos, and the refusal is a silent return. Invert the polarity (safe default, explicit opt-out at the five sites — behavior-preserving) and log the refusal.

## Minor

- `src/api/routes/agents.ts:2943` — if the rollback's host delete throws, it shadows the original `registerSession` error out of the logs; wrap it in its own try/catch.
- `src/renderer/hooks/use-sessions.ts:167` — the status-blind `Error('Failed to fork session')` flattens the designed-in 409 "Session is currently running" (and the 500's guidance) out of the toast; throw the response body's text.
- `src/renderer/components/layout/app-sidebar.tsx:172` — non-selected rows hardwire `isStreaming` to false, so Fork enables during the `isActive` invalidation window; UX-only (the server 409 backstops it).
- `src/api/routes/agents.ts:2821` — the handler now parses the full session-metadata map a third time per request; one `readSessionMetadata` read can serve both ids.
- `src/api/routes/agents.ts:3000` — `copyForkAttribution` materializes the entire transcript just for `{uuid, forkedFrom}` pairs and inserts rows one at a time; stream via `streamJsonlFile` and batch the inserts.
- `src/renderer/components/layout/agent-header.tsx:65` — a second full `useMessageStream` subscription to read one boolean; take `isStreaming` as a prop from `AgentShell`, which already holds the identical subscription.
- `src/renderer/components/sessions/session-context-menu.tsx:110` — the detail-cache seed + draft carry belong inside `useForkSession` (`onMutate`/`onSuccess`), so a second caller doesn't silently lose the composer settings and draft.
- `src/renderer/components/layout/agent-header.test.tsx:173` (and `app-sidebar.test.tsx:672`) — neither suite asserts `data-is-active="false"`; a hardcoded `isActive={true}` passes every test.
- `src/renderer/components/layout/app-sidebar.test.tsx:4` — the guarded `localStorage` stub can never execute under the file's pinned jsdom environment; delete it.
- `src/shared/lib/container/mock-container-client.ts:2747` — the hand-rolled JSONL loop is `readJsonlFile`'s semantics verbatim; use the helper.

JeremyBischoff:
@iddogino

1. Done in `07c806473` - parent title now comes from the live session lookup, and Back hides only when that lookup returns nothing.
2. Done in `07c806473` - a JSON 404 is a gone session; a plain 404 is still the old-image restart path.
3. Done in `07c806473` - forget now drops a never-opened copy so rollback does not leave a listing pointed at a deleted transcript.
4. Done in `07c806473` - copy-a-folder skips shortcuts by default, the five install paths opt back in, and a skip is logged.

- Done in `07c806473` - host rollback delete is in its own try/catch so it cannot shadow the register error.
- Done in `07c806473` - the toast now throws the response body's error text.
- Left as UX-only - the server 409 backstops it.
- Absorbed - the live lookup replaced the extra listing read.
- Done in `07c806473` - sender badges stream via `streamJsonlFile` and batch-insert.
- Done in `07c806473` - header takes `isStreaming` from AgentShell.
- Done in `07c806473` - draft snapshot and cache seed live in `useForkSession`.
- Done in `07c806473` - idle-menu tests assert `data-is-active="false"`.
- Left the stub - deleting it fails the suite because jsdom was not given `--localstorage-file`.
- Done in `07c806473` - the mock now uses `readJsonlFile`.

src/api/routes/agents.ts
```diff
@@ -2879,6 +2882,10 @@ agents.get('/:id/sessions/:sessionId', AgentRead(), async (c) => {
       invokedByAgentName: metadata?.invokedByAgentSlug
         ? invokingAgent?.frontmatter.name ?? metadata.invokedByAgentSlug
         : undefined,
+      forkedFromSessionId: metadata?.forkedFromSessionId,
+      forkedFromSessionName: metadata?.forkedFromSessionId
```
iddogino, line 2953: `getSession(parent)` streams the parent's entire transcript on every GET of a fork, and a parent read error turns the fork's own GET into a 500. The parent's registered name is already in session metadata — resolve it from there.

agent-container/src/session-manager.ts
```diff
@@ -731,6 +747,55 @@ export class SessionManager extends EventEmitter {
...
+      throw new SessionBusyError(sourceId);
+    }
+
+    // CLAUDE_CONFIG_DIR=/workspace/.claude + dir '/workspace' resolves to the
+    // same projects dir the source lives in; the fork lands beside it.
+    const { sessionId: newId } = await sdkForkSession(source.claudeSessionId, { dir: source.workingDirectory });
+
+    const now = new Date().toISOString();
+    try {
+      this.persistence.saveSessionChecked({
```
iddogino, line 800: `...source` copies `metadata.isAutomated`, so a fork of a scheduled/webhook/x-agent session is classed automated in-container (0ms idle eviction, wrong runtime class). Clear the automation flags in the spread like the capability grants — and add a test covering this.

agent-container/src/server.ts
```diff
@@ -176,6 +176,24 @@ app.post('/sessions/:id/interrupt', async (c) => {
...
+    const id = await sessionManager.forkSession(sessionId);
+    if (!id) {
+      return c.json({ error: 'Session not found' }, 404);
+    }
+    return c.json({ id }, 201);
+  } catch (error: any) {
+    if (error instanceof SessionBusyError) {
+      return c.json({ error: error.message, code: 'session_busy' }, 409);
+    }
+    console.error('Error forking session:', error);
```
iddogino, line 203: When the transcript JSONL was reaped but the record survived, `sdkForkSession` throws not-found and this returns an opaque 500. Match that error and return the JSON 404 so it routes through the existing session-gone path.

src/api/routes/agents.ts
```diff
@@ -2980,6 +2987,148 @@ agents.delete('/:id/sessions/:sessionId/unread', AgentRead(), async (c) => {
...
+  const slug = getAgentId(c)
+  const sourceId = c.req.param('sessionId')
+  try {
+    if (!(await sessionIsKnown(slug, sourceId))) {
+      return c.json({ error: 'Session not found' }, 404)
+    }
+    // Check-then-refuse, like the x-agent 409: no lock. The container checks its
+    // own turn tracker too; the residual window is one dropped partial line.
+    if (messagePersister.isSessionActive(sourceId)) {
+      return c.json({ error: 'Session is currently running' }, 409)
```
iddogino: This handler reads and Zod-parses the same session-metadata file three times (`sessionIsKnown`, `getSession`, `getSessionMetadata`), all serialized behind `ensureRunning`. Read the metadata once up front and run `ensureRunning` concurrently with the host-side reads.

src/api/routes/agents.ts
```diff
@@ -2980,6 +2987,148 @@ agents.delete('/:id/sessions/:sessionId/unread', AgentRead(), async (c) => {
...
+    }
+
+    // Sequenced before the response so the first messages request already sees
+    // them; best-effort so neither can fail a fork that exists.
+    const sessionsDir = getAgentSessionsDir(slug)
+    const sourceDir = path.join(sessionsDir, sourceId)
+    // lstat, not stat: a replaced session folder that is a symlink would
+    // otherwise be followed on the host and copy files from outside the workspace.
+    let sourceIsRealDir = false
+    try {
```
iddogino: This inline lstat guard duplicates the symlink defense `copyDirectoryFiltered` now has built in (file-storage.ts). Drop it and let the helper's default no-followSymlinks mode refuse a symlinked root.

src/renderer/components/sessions/session-context-menu.tsx
```diff
@@ -56,6 +56,8 @@ interface SessionContextMenuProps {
    * silent no-op.
    */
   sessionIsLive?: boolean
+  /** Mid-turn: Fork Session is disabled (the host refuses with 409 as the backstop). */
+  isActive?: boolean
```
iddogino: Two overlapping liveness props (`sessionIsLive` vs the new optional `isActive`) with different derivations, hand-blended at every call site — and the new one defaults to Fork-enabled when unthreaded. Derive the disabled state inside the menu from a single prop set, or at least rename it `forkDisabled`.

src/api/routes/agents.ts
```diff
@@ -2980,6 +2987,148 @@ agents.delete('/:id/sessions/:sessionId/unread', AgentRead(), async (c) => {
...
+      }
+      throw error
+    }
+    if (!forked) {
+      return c.json({ error: 'Container does not support fork; restart the agent to pull the latest image' }, 500)
+    }
+    const newId = forked.id
+    const name = `${source.name} (fork)`
+
+    const initialMetadata: Parameters<typeof updateSessionMetadata>[2] = {
```
iddogino: This model/effort/speed conditional-spread trio is repeated verbatim in the 201 body below. Build one `runtimeChoices` object and spread it into both.

src/api/routes/agents.ts
```diff
@@ -2980,6 +2987,148 @@ agents.delete('/:id/sessions/:sessionId/unread', AgentRead(), async (c) => {
   return setUnreadFlag(c, c.req.param('sessionId'), false)
 })
 
+// POST /api/agents/:id/sessions/:sessionId/fork - Fork Session: copy the
+// conversation into a new session carrying the full prior context.
+agents.post('/:id/sessions/:sessionId/fork', AgentUser(), async (c) => {
```
iddogino, line 3059: The whole fork choreography (container call, registerSession, rollback, sidecar copy, attribution) is inline in the route while its lifecycle siblings live in session-service. Extract a `forkSession(slug, sourceId, opts)` service so future entry points don't re-duplicate the rollback logic.

src/api/routes/agents.ts
```diff
@@ -2980,6 +2987,148 @@ agents.delete('/:id/sessions/:sessionId/unread', AgentRead(), async (c) => {
...
+    } catch {
+      sourceIsRealDir = false
+    }
+    if (sourceIsRealDir) {
+      await copyDirectoryFiltered(sourceDir, path.join(sessionsDir, newId))
+        .catch((error) => console.error(`fork: subagent/workflow copy for ${newId} failed (non-fatal)`, error))
+    }
+    if (isAuthMode()) {
+      await copyForkAttribution(slug, sourceId, newId)
+        .catch((error) => console.error(`fork: attribution copy for ${newId} failed (non-fatal)`, error))
```
iddogino: The sidecar copy and attribution rebuild touch disjoint state and each already swallows its own errors — `Promise.all` them instead of awaiting sequentially, keeping both ahead of the response at max(a,b) latency.

iddogino (changes requested):
Requesting changes for three correctness issues on the fork path (parent-transcript streaming on fork GETs, `isAutomated` inheritance in the container fork, reaped-transcript forks surfacing as opaque 500s) plus structure/efficiency cleanups. Details inline.

src/shared/lib/services/session-fork-service.test.ts
```diff
@@ -0,0 +1,241 @@
...
+    const attrGate = new Promise<void>((resolve) => {
+      releaseAttr = resolve
+    })
+    let copyCalled = false
+    let attrCalled = false
+    copyDirectoryFiltered.mockImplementation(async () => {
+      copyCalled = true
+      await copyGate
+    })
+    streamJsonlFile.mockImplementation(async function* () {
```
iddogino, line 173: @JeremyBischoff I think this is breaking our unit tests

iddogino (changes requested):

src/shared/lib/container/mock-container-client.ts
```diff
@@ -2758,6 +2766,51 @@ export class MockContainerClient extends EventEmitter implements ContainerClient
...
   }
 
+  /**
+   * Fork Session in mock mode: copy the source JSONL to a new id the way the
+   * SDK does (fresh uuids, parent chain remapped, `forkedFrom` backlink on
+   * every line) and register a cold session so the fork is sendable.
+   */
+  async forkSession(sessionId: string): Promise<{ id: string } | null> {
+    const source = this.sessions.get(sessionId)
+    if (!source) return null
```
iddogino: Returning `null` here hits the wrong host path: the client contract reserves `null` for "endpoint missing (old image)", so an unknown source surfaces as "Container does not support fork; restart the agent to pull the latest image" instead of the real container's JSON 404 → "Session not found". Throw `ContainerNotFoundError` like the real client would, so mock/E2E runs exercise the same error path. (Found by forking a pre-seeded session in a mock-mode dev server — anything not in the in-memory map reproduces it.)

src/shared/lib/services/session-fork-service.test.ts
JeremyBischoff: Done in `f1a41d047` - typecheck was failing on that mock.

src/shared/lib/container/mock-container-client.ts
JeremyBischoff: Done in `f1a41d047` - a missing session now throws the same not-found error as a real agent.

PR #900, opened 2026-08-29
fix(import): restore sessions from full agent imports

yiw190:
## Bug

Full `.agent` imports restore session transcripts and `session-metadata.json`, but imported sessions do not appear after the host-owned session ownership index has already been initialized. The import path never registers the restored session IDs, so session listing rejects them as unowned.

## Reproduction

1. Initialize `session-ownership.json` on the destination host.
2. Import a full `.agent` archive containing session metadata and JSONL transcripts.
3. Open the imported agent.
4. Observe that no restored sessions are listed.

## What changed

- Register all restored full-import session IDs in the host-owned ownership index after extraction.
- Validate all ownership collisions before writing, so a conflict cannot partially claim the remaining imported sessions.
- Roll back the new workspace when an import fails after agent creation.
- Release session ownership when an agent is deleted, including cleanup retries after its directory is already gone.
- Add regression coverage for visibility, collision isolation, import rollback, deletion cleanup, and stale-claim cleanup.

## Test plan

- [x] `npm run test:run -- src/shared/lib/services/agent-template-service.test.ts src/shared/lib/services/session-service.ownership.test.ts src/shared/lib/services/agent-service.test.ts` (227 tests)
- [x] ESLint on all changed files (0 errors)
- [x] `git diff --check`
- [ ] `npm run typecheck` — blocked by existing missing local declarations/dependencies in unrelated `semver`, `e2b`, and Daytona files

### DIFF
--- a/src/shared/lib/services/agent-service.test.ts
+++ b/src/shared/lib/services/agent-service.test.ts
@@ -49,6 +49,7 @@ import {
   getAgentClaudeMdContent,
   setAgentClaudeMdContent,
 } from './agent-service'
+import { registerSession, sessionBelongsToAgent } from './session-service'
 
 describe('agent-service', () => {
   let testDir: string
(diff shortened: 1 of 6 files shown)


iddogino:
superseded by 907

PR #913, opened 2026-08-31
Add Shared Volumes for cloud agents (SUP-739)

JeremyBischoff:
Host folders and shared volumes now travel the same mount pipe. One record, one list, one start contract, one Volumes card, one prompt env var. The runner decides what that record may bind. Cloud binds a workspace folder at `/volumes/<name>`. Desktop binds a folder from this computer at `/mounts/<name>`.

Related to https://linear.app/datawizz/issue/SUP-739/shared-volumes-opt-in-shared-readwrite-folders-for-cloud-agents

| | Files | Lines added |
| --- | ---: | ---: |
| Production | 32 | ~1,300 |
| Tests | 22 | ~1,400 |
| Drizzle snapshot | 2 | ~3,090 |

Migration `0040_shared_volumes` adds `shared_volumes` (id, display name, unique mount name, created at) and `agent_shared_volumes` (which agent is attached, unique per pair, attachments die when the volume dies).

What changed:

- One Volumes card. Desktop offers add-folder. Cloud offers create and attach.
- Create and attach refuse on a desktop runner. Add-folder refuses on a cloud runner.
- A shared-volume folder that used to drop silently now shows mount health on the card.
- Desktop agents see `/mounts/*` in the prompt. Cloud agents see `/volumes/*`.
- `SUPERAGENT_MOUNTS` is a JSON array of the paths the runtime accepted, so a comma in a folder name stays one path.
- Kubernetes with no prefix setting uses `agents/<slug>/workspace`, same as MicroVM and the host app. The three prefix settings are gone.

Start.

```
folder missing → skip that volume, warn, still boot
rest → mount read/write
prompt lists only what mounted
prompt: keep own work in /workspace
```

Delete.

```
other agents still attached → 409
only this agent's attach → detach then delete
no attachments → delete
```

Intended, not obvious:

- The agent cannot create or attach. A person does it from the card.
- A missing folder at start is skipped. The agent still boots.
- Cap is 19 per agent. That is the run-hook byte budget with real mTLS size.
- INDEX.md is prompt advice. Create does not plant one.
- Chat uploads still go to `/workspace/uploads`.
- The `restart` flag on the mount routes is dead and left as-is.
- The global volumes list stays guarded the same way.

Manual smoke (Gate 2) passed: desktop folder add, desktop refuses shared create, cloud create/attach/detach, delete hidden while two agents share the volume.


JeremyBischoff:
Closing this in favour of a chain of smaller PRs under SUP-739. The mechanism is unchanged: one mount record, one merged list, one runtime contract, and the registry as a second source that rides the same pipe.

Landed: #978, the prompt lists the folders the container has.
Next: runtimes own the mount list and accept `/volumes/` records on cloud (SUP-790), then the registry and merged list, the Volumes card, and the cloud e2e lane, one PR each.

The branch stays until the last slice merges, so the e2e spec and the card code can be carved from it.

PR #917, opened 2026-09-01
fix(microvm): do not treat PENDING as a dead generation

yiw190:
## Summary
- First chat after a cold MicroVM start can return **Failed to create session** in 5–14s (`MicroVM agent … failed to become healthy`) even though `RunMicrovm` succeeded. Retry works.
- Cause: `getInfoFromRuntime` treated every non-`RUNNING` / non-`SUSPENDED` state as terminal and called `cleanupLocalIf`. `waitForHealthy` fail-fasts on the first `/health` miss when `getInfo` is not `running`, so a `PENDING` flap aborted the 120s health wait and tore down the new proxy.
- Change: only `TERMINATED` / `TERMINATING` (plus the existing `SUSPENDED` / NotFound paths) drop local state. `PENDING` and other in-progress states keep the proxy so health polling can finish.

This is not the #856 dead-loopback-port bug (already in `0.5.14`). Same error string, different path. Seen on staging MI signups and on prod Fargate (same host-app image).

## Test plan
- [x] `npx vitest run src/shared/lib/container/lambda-microvm-runtime.test.ts` (82 passed)
- [ ] Cold cloud workspace: first message after deploy should create a session without a retry
- [ ] A truly dead VM (`TERMINATED`) still fail-fasts and a later start runs a new MicroVM

### DIFF
--- a/src/shared/lib/container/container-manager-concurrency.test.ts
+++ b/src/shared/lib/container/container-manager-concurrency.test.ts
@@ -456,6 +456,47 @@ describe('containerManager — stopContainer during in-flight start', () => {
     mockStop.mockResolvedValue({ forceStopUsed: false })
   })
 
+  it('ensureRunning does not start while stopContainer is in flight', async () => {
+    let releaseStop: (value: { forceStopUsed: boolean }) => void = () => {}
+    mockStop.mockImplementation(() => new Promise<{ forceStopUsed: boolean }>((resolve) => {
+      releaseStop = resolve
+    }))
+
+    const stopPromise = containerManager.stopContainer('test-agent')
+    await vi.waitFor(() => expect(mockStop).toHaveBeenCalled())
+
+    await expect(containerManager.ensureRunning('test-agent')).rejects.toThrow(
+      'Cannot start agent test-agent while it is stopping',
+    )
+    expect(mockStart).not.toHaveBeenCalled()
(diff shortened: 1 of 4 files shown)


iddogino:
@yiw190 a couple of risks here:

1. Seems like a generation can get stuck in pending with no self heal... Need a limit for how long it can be pending, or a per-generation reachedRunning latch so we only treat pending as alive after we see running....
2. Can race with stop during boot ->  stopContainer deletes startingAgents before awaiting the pre-stop hook and client.stop(). A message arriving in that window triggers a second start(), which now short-circuits on the PENDING generation and caches/broadcasts running with a port that is about to be terminated. The follow-up restartAgent then throws "Cannot restart agent while it is stopping".
3. Fallthrough returns running with current undefined. lambda-microvm-runtime.ts:1068. If local state is torn down during the GetMicrovm await, the new non-terminal branch returns running with a proxy port whose listener is already stopped. The deleted code returned stopped here.
4. Default flipped from fail-closed to fail-open. lambda-microvm-runtime.ts:1062. state is z.string().optional(), so an omitted or unrecognized state now reports running forever with no log or Sentry capture. Prefer an explicit PENDING allowlist and a logged default branch.

iddogino (changes requested):
see comment

PR #914, opened 2026-09-01
feat(errors): 402 paywall card with dashboard hand-off (SUP-725)

yiw190:
## Why

A platform 402 currently leaves users with a generic provider error and no clear next action. Reproduce by exhausting an unsubscribed or subscribed staging workspace, then sending a message.

## What changed

- Render a neutral paywall card with role/billing-aware Subscribe, Add usage, Add credit card, Fix payment, Ask admin, and fallback CTAs.
- Put the unresolved paywall in the composer slot across navigation/reload; suppress duplicate and historical cards.
- Open the organization dashboard for billing. Add usage includes `intent=topup` and the local-app return link.
- Refresh billing on dashboard return/focus. When the proxy gate allows access, remove the card and restore the composer for manual retry.
- Keep non-platform 402s on the generic provider-error path.

This PR does **not** automatically continue the interrupted request. That behavior is isolated in the stacked follow-up PR.

Depends on [platform #363](https://github.com/SkillfulAgents/platform/pull/363) for the billing snapshot and dashboard hand-off.

## Test plan

- [x] Focused unit/component suite: 1,039 tests
- [x] Manual staging: exhausted free workspace → Subscribe card
- [x] Manual staging: exhausted subscribed workspace → Add usage card
- [x] Manual staging: dashboard return refreshes access and restores composer
- [x] Navigation/reload keeps one card in the composer position
- [ ] Member staging account shows Ask admin without billing details

PR #931, opened 2026-09-02
feat(errors): resume sessions after billing recovery (SUP-725)

yiw190:
## Why

The base paywall PR restores the composer after billing recovers, but the user must manually retry the interrupted request. This stacked PR continues that exact session automatically.

## What changed

- Arm one initiating session when the user opens a write billing CTA.
- Poll the proxy-provided `access.allowed` gate after deep-link/focus return; stale or missing gate data never authorizes continuation.
- Send one fixed hidden continuation through an authenticated, agent-scoped API route.
- Scope and bound deduplication records and atomically claim session activity so a concurrent real user send wins safely.
- Clear terminal targets and coalesce deep-link/focus/visibility refresh signals.

Stacked on [the Paywall UI PR](https://github.com/SkillfulAgents/SuperAgent/pull/914). Review only the diff against `feat/sup-725-in-app-paywall`.

## Test plan

- [x] Focused paywall/recovery suite: 1,055 tests
- [x] Manual staging: subscribe restores access and continues the interrupted session
- [x] Manual staging: top-up restores access and continues the interrupted session
- [ ] Cancel checkout leaves the gate denied and does not continue
- [ ] Two sessions: only the session that opened billing continues
- [ ] Repeated hand-back signals produce one continuation

### DIFF
--- a/src/api/billing-resume-attempts.test.ts
+++ b/src/api/billing-resume-attempts.test.ts
@@ -1,0 +1,41 @@
+import { beforeEach, describe, expect, it } from 'vitest'
+
+import {
+  beginBillingResumeAttempt,
+  finishBillingResumeAttempt,
+  resetBillingResumeAttemptsForTests,
+} from './billing-resume-attempts'
+
+describe('billing resume attempts', () => {
+  beforeEach(() => resetBillingResumeAttemptsForTests())
+
+  it('scopes duplicate attempt ids to their session', () => {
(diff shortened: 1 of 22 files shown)


yiw190:
Superseded by the #961 → #962 stack. #962 reworks the paywall as a registered provider-error component; auto-resume after billing recovery is not carried over there and is noted as a follow-up.

PR #933, opened 2026-09-02
fix(renderer): mark read-only skillsets in publish pickers

yiw190:
## Summary
- The default connected skillset is `SkillfulAgents/public-skillset` (shown as **Gamut Public Skillset**). It is already read-only (`publishMode: none`), but Publish still listed it as a destination you could pick.
- Skill and agent publish pickers still list every connected skillset. Read-only ones are visible and labeled **Read only**, and cannot be selected.
- Publish copy uses **skillset**, not library.

## Test plan
- [ ] With only the default Gamut skillset connected, open skill Publish and agent Share → Publish: it appears as **Read only** and cannot be selected
- [ ] Add a GitHub skillset and confirm it is selectable in both pickers
- [ ] Settings → Skillsets still lists the default Gamut skillset (install/browse unchanged)

### DIFF
--- a/src/renderer/components/agents/agent-share-popover.tsx
+++ b/src/renderer/components/agents/agent-share-popover.tsx
@@ -37,7 +37,7 @@ import {
   useHostExportStatus,
 } from '@renderer/hooks/use-agent-templates'
 import { AgentTemplatePublishPanel } from '@renderer/components/agents/agent-template-publish-panel'
-import { useSkillsets } from '@renderer/hooks/use-skillsets'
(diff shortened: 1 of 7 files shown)


iddogino:
@yiw190 

**UX**: we should show non-publishable skillsets, but mark them as read only (otherwise it's confusing why you see the skillset in one place but not another) .

**Language**: current language is skillset, not libraries. Let's not mix and match

yiw190:
Updated for the review:

- Non-publishable skillsets stay in the pickers and are labeled **Read only**
- Publish copy now says skillset, not library

PR #935, opened 2026-09-02
Unify the session menu across the sidebar and the agent home list

westaskew:
## Summary

Follow-up to #934, applying the same unification and styling to the session menu.

The agent home's session list rows had their own popover (rename, copy raw log, delete): a second copy of what the sidebar's `SessionContextMenu` already offered, and it had drifted. There is now **one** session menu. The list row is the menu's trigger, and its three-dot replays a click as a context-menu event on the row, the same mechanism the agent home's three-dot uses for the agent menu. The row's duplicate handlers and dialogs are deleted. As a side effect the list rows gain Mark as Unread, Fork, and the usage totals, which the old popover never offered.

**Same treatment as the agent menu:** fixed `w-52`, 12px radius, 8px padding, dividers inset 12px, Delete in the default colour.

**Items:** Edit, Fork, Mark as Unread (eye icon), Copy Raw Log · Delete · Cost and Tokens in a muted panel with quiet values.

**One implementation change:** the menu read the route via TanStack's `useParams`, which needs a live router. It now uses the router-free `useRouteLocation` the list row already used, so it renders anywhere (this is what let it drop into the agent-home test without a router).

## Tests

- Agent-home test mock gains the session hooks the menu reads; the menu's own test stubs `useRouteLocation`.
- `session-delete` e2e drives the shared menu's delete item instead of the old row popover's button.
- Passing locally: typecheck, lint, the sessions / agent-home / layout / triggers / completed-tasks unit suites (264 tests), and the session-delete, session-rename, and session-fork e2e specs in mock mode.

### DIFF
--- a/e2e/specs/session-delete.spec.ts
+++ b/e2e/specs/session-delete.spec.ts
@@ -116,16 +116,12 @@ test.describe('Session deletion', () => {
 
     await gotoAgentHome(page, agent)
 
-    // exact:true — role-name matching is substring by default, and the row
-    // button's accessible name embeds the kebab's label, so a substring match
-    // resolves to both the row and its kebab
     const kebabFor = (session: TestSession) =>
-      page.getByRole('button', { name: `Actions for ${session.name}`, exact: true })
+      page.locator(`[data-testid="session-row-menu-${session.id}"]`)
     const deleteDialog = page.getByRole('alertdialog')
 
     // Cancel path: open the kebab (the shared session menu), choose Delete,
-    // then back out — the
-    // session must survive
+    // then back out — the session must survive
     await kebabFor(sessionB).click()
     await page.locator('[data-testid="delete-session-item"]').click()
     await expect(deleteDialog).toBeVisible()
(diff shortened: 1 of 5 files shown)


iddogino:
Pushed 4298f996 with three review fixes:

- **Rename** stays "Rename" rather than "Edit": the item opens a rename dialog and nothing else.
- **Delete** keeps its destructive red colour (menu item), matching the confirm dialog.
- **session-delete e2e** locates the row kebab by `data-testid="session-row-menu-<id>"` instead of a role+exact-name query, matching how the rest of the spec and the rename/fork specs find elements.

Two notes on the description: `useRouteLocation` is not router-free (it calls `useRouterState`); it works in unit tests because `src/renderer/test/setup.ts` mocks it globally. And the "same mechanism as the agent home three-dot" and the `w-52 rounded-xl` styling refer to #934, which is still open, so the menus will only match once both land.

Verified locally: lint clean, sessions + agent-home unit suites (91 tests), session-delete e2e in mock mode.

https://claude.ai/code/session_01Ngk3cS1JD2uW6FVVLQckc6

PR #940, opened 2026-09-02
fix(triggers): tear down upstream Composio subscription as the trigger creator (SUP-765)

yiw190:
## Bug

When an agent is deleted by a user who didn't create its Composio triggers, the upstream subscription teardown silently fails and the trigger is orphaned: Composio keeps sending webhook events forever, and since the local trigger row is cancelled, the events can never be claimed — they pile up as `pending` and hammer the host with realtime pings on every arrival.

Root cause: `cancelWebhookTriggerWithCleanup` runs `deleteComposioTrigger` under the **deleting user's** ambient attribution. The platform proxy's `DELETE /v1/composio/triggers/{id}` verifies ownership against the subscription's Composio `user_id` — the **creator's** member id — so any cross-member cleanup gets 404 "Trigger not found", which the catch block swallows (Sentry only; the UI reports success).

## Repro / evidence (datawizz incident, Sep 2)

- 4 calendar triggers created by one member, agent deleted by another member on Aug 18 → Sentry [ELECTRON-AT](https://datawizz.sentry.io/issues/ELECTRON-AT) `ComposioTriggerError: Trigger not found`, culprit `DELETE /api/agents/jp76x5tltm`, timestamp matches all 4 `cancelled_at` rows.
- ~3,400 unclaimable `pending` webhook_events accumulated since; a recurring-meeting deletion fanned out into ~1,500 events at once and amplified a restart into ~10 min of CPU saturation.
- Linear: [SUP-765](https://linear.app/datawizz/issue/SUP-765/composio-trigger-teardown-fails-cross-member-upstream-subscription)

## Change

- Run `deleteComposioTrigger` inside `runWithAttribution(...)` using the trigger **creator's** attribution, falling back to the connected-account owner (same candidate order as SUP-226), mirroring what the custom-endpoint branch already does via `resolveCleanupMemberId`.
- No behavior change when neither resolves (keeps ambient attribution, as today).

## Test plan

- [x] New `webhook-trigger-service.sup765.test.ts`: teardown under creator attribution; fallback to connected-account owner; ambient fallback when neither resolves.
- [x] Existing `custom-teardown` + `sup221` trigger tests still green (15/15).
- [x] `typecheck` clean for touched files (pre-existing errors only in unrelated `tools/verify-*` scripts), `eslint` clean.

## Out of scope (follow-ups tracked in SUP-765)

- Platform-side org-scoped teardown (needed when the creator's user no longer exists).
- One-off cleanup of the 4 orphaned datawizz triggers + pending event purge.
- The separate bare-org-token `acting member` error flood from AccountSync / proxy status checks.


src/shared/lib/services/webhook-trigger-service.ts
```diff
@@ -458,16 +459,19 @@ export async function cancelWebhookTriggerWithCleanup(
...
-          await disablePlatformWebhookEndpoint(
-            resolveCleanupMemberId(trigger),
-            trigger.composioTriggerId,
-          )
+          await disablePlatformWebhookEndpoint(resolveCleanupMemberId(trigger), upstreamId)
         } else {
-          await deleteComposioTrigger(trigger.composioTriggerId)
+          // The proxy scopes trigger DELETE to the subscription's Composio user
+          // (the creator); the ambient deleter 404s cross-member (SUP-765).
+          await runWithAttribution(
```
iddogino: **Blocking: the incident path doesn't go through here.** The Sentry culprit for SUP-765 is `DELETE /api/agents/:id`, which runs `cleanupAgentData` -> `cleanupWebhookTriggers` (`src/shared/lib/services/agent-cleanup-service.ts:89`). That call site invokes `deleteComposioTrigger(trigger.composioTriggerId)` directly with no `runWithAttribution`, so member B deleting an agent whose trigger was created by member A still sends the DELETE under B's ambient attribution, the proxy 404s, `captureException` swallows it, and the upstream subscription stays live.

Suggest either having `cleanupWebhookTriggers` call `cancelWebhookTriggerWithCleanup(trigger.id)` (it already duplicates the last-one-out / custom-vs-composio logic), or extracting a shared `deleteComposioTriggerForTrigger(trigger)` helper that applies the creator attribution and using it from all three call sites.

src/shared/lib/services/webhook-trigger-service.ts
```diff
@@ -501,6 +505,26 @@ function resolveCleanupMemberId(trigger: WebhookTrigger): string {
   return resolved?.memberId ?? getStoredPlatformMemberId() ?? 'local'
 }
 
+// Creator first, then connected-account owner (SUP-226 candidate order);
+// null keeps the ambient request attribution.
+function resolveCleanupAttribution(trigger: WebhookTrigger): Attribution | null {
```
iddogino: **Blocking: `createdByUserId` is not the minting principal.** The subscription is minted by `enableComposioTrigger` in `message-persister.ts` (~L4292), and that runs under whichever request opened the session's stream subscription (ALS flows into the ws `message` listener from `subscribeToSession`). `createdByUserId` is copied from session metadata instead. They diverge in a realistic case: session created by A, in-memory subscription dropped (detach / host restart), member B posts a message and re-subscribes under B's `runWithRequestUser` scope, agent calls `setup_trigger`. The subscription is now B's but the row says A, so this PR deletes as A and 404s. Trigger-manager-spawned sessions also register without `createdByUserId`, so the owner fallback there is a guess.

Robust fix: capture the minting principal at mint time (`getRequestUserId()` / `attribution.current()?.getKey()`) onto the trigger row and delete under that, rather than re-deriving it at teardown.

src/shared/lib/services/webhook-trigger-service.ts
```diff
@@ -458,16 +459,19 @@ export async function cancelWebhookTriggerWithCleanup(
...
+    const upstreamId = trigger.composioTriggerId
+    const remaining = await countActiveTriggersForComposioId(upstreamId)
     if (remaining === 0) {
       try {
         if (trigger.kind === 'custom') {
-          await disablePlatformWebhookEndpoint(
-            resolveCleanupMemberId(trigger),
-            trigger.composioTriggerId,
-          )
+          await disablePlatformWebhookEndpoint(resolveCleanupMemberId(trigger), upstreamId)
```
iddogino: The PR description says the composio branch now mirrors the custom branch, but the custom branch doesn't actually attribute to the creator either. `disablePlatformWebhookEndpoint` sets `Authorization: Bearer <token>::<creatorMember>` explicitly, but the platform fetch interceptor unconditionally overwrites `Authorization` from `attribution.current()` (`install-fetch-interceptor.ts:27-28`; `install-fetch-interceptor.test.ts:89-103` asserts exactly this override). In an HTTP route that resolves to the deleting user, so the `memberId` argument is dead. Pre-existing, but since you're here: wrap this call in the same `runWithAttribution(...)` so both branches use one mechanism, and the ALS override wins consistently.

src/shared/lib/services/webhook-trigger-service.ts
```diff
@@ -501,6 +505,26 @@ function resolveCleanupMemberId(trigger: WebhookTrigger): string {
...
+// Creator first, then connected-account owner (SUP-226 candidate order);
+// null keeps the ambient request attribution.
+function resolveCleanupAttribution(trigger: WebhookTrigger): Attribution | null {
+  return (
+    attribution.fromResourceCreator(trigger.createdByUserId) ??
+    attribution.fromResourceCreator(getConnectedAccountOwnerUserId(trigger.connectedAccountId))
+  )
+}
+
+function getConnectedAccountOwnerUserId(connectedAccountId: string | null): string | null {
```
iddogino: This is a byte-for-byte copy of the private `resolveConnectedAccountOwner` in `src/shared/lib/scheduler/trigger-manager.ts:45-54`, and the `[createdByUserId, ownerOf(connectedAccountId)]` candidate order is now built independently there (~L330) and in `resolveCleanupAttribution`. We also now have `resolveCleanupMemberId` (creator -> stored member -> `'local'`, returns a string) and `resolveCleanupAttribution` (creator -> owner -> null, returns an `Attribution`) side by side with different fallback chains, plus a third inline copy in `agent-cleanup-service.ts:84`. These will drift.

Suggest exporting one `resolveTriggerPrincipal(trigger)` built on the existing `resolvePlatformMemberForCandidates`, deleting the trigger-manager copy, and having both the string and attribution consumers derive from it.

src/shared/lib/services/webhook-trigger-service.sup765.test.ts
```diff
@@ -0,0 +1,140 @@
...
+vi.mock('@shared/lib/services/platform-auth-service', () => ({
+  getPlatformAccessToken: () => mockGetPlatformAccessToken(),
+  getStoredPlatformMemberId: () => mockGetStoredPlatformMemberId(),
+}))
+
+const mockFromResourceCreator = vi.fn()
+const mockRunWithAttribution = vi.fn(
+  (auth: unknown, fn: () => Promise<unknown>) => fn(),
+)
+vi.mock('@shared/lib/platform-attribution', () => ({
```
iddogino: Mocking `@shared/lib/platform-attribution` wholesale (`fromResourceCreator` returns sentinels, `runWithAttribution` is `(auth, fn) => fn()`) means these tests only pin call arguments and order. They'd stay green if production discarded the resolved attribution entirely, so the "teardown under creator attribution" claim isn't what's proven. The sibling `sup226.test.ts` seeds real `user` / `authAccount` rows and asserts on the real resolver. Suggest doing the same here: leave `platform-attribution` unmocked and assert `attribution.current()?.getKey() === 'member:<creatorMemberId>'` inside the `deleteComposioTrigger` mock.

src/shared/lib/services/webhook-trigger-service.sup765.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    const ownerAuth = { key: 'owner' }
+    mockFromResourceCreator.mockImplementation((userId: string | null) =>
+      userId === 'user-owner' ? ownerAuth : null,
+    )
+    await testDb.insert(schema.connectedAccounts).values({
+      id: 'ca_1',
+      providerConnectionId: 'pc_1',
+      toolkitSlug: 'googlecalendar',
+      displayName: 'Calendar',
+      userId: 'user-owner',
```
iddogino: Minor: this inserts a `connectedAccounts` row with `userId: 'user-owner'` and no matching `user` row. `connectedAccounts.userId` is an FK to `user.id` with cascade (`schema.ts:171`) and the real DB runs `PRAGMA foreign_keys = ON` (`db/index.ts:45`), so this only passes because the harness omits the pragma. Seed the owner via `insertUser` + `insertConnectedAccount` like `sup226.test.ts` does; that also models the realistic case (creator deleted, owner present with a platform authAccount row).

Also trivial, in the service: the `captureException` extra a few lines below the new alias still spells `upstreamId: trigger.composioTriggerId`; use the shorthand `upstreamId,` so there's one name in that block. And consider gating `resolveCleanupAttribution` on `attribution.requiresActingMember()` like `account-sync-service.ts:94` / `trigger-manager.ts:176` do, since in opaque-key mode the lookups can't change the outcome.

iddogino (changes requested):
Thanks for tracking this down. The creator-attribution idea is right, but as written it doesn't reach the path that produced the SUP-765 incident, and the principal it resolves to isn't guaranteed to be the one that minted the subscription. Details inline; the first two are blocking, the rest are follow-ups I'd like addressed in this PR since they're on touched lines.

src/shared/lib/services/webhook-trigger-service.ts
yiw190: Good catch — this was the actual incident path. Went with your first option: `cleanupWebhookTriggers` now delegates to `cancelWebhookTriggerWithCleanup` per trigger.

src/shared/lib/services/webhook-trigger-service.ts
yiw190: Agreed, re-deriving at teardown was a guess. Now capturing the acting member at mint time into a new `minted_by_member_id` column and deleting under it; the creator/owner chain stays only as fallback for pre-column rows. (Diff looks bigger than it is — drizzle regenerated the full meta snapshot json for the migration.)

src/shared/lib/services/webhook-trigger-service.ts
yiw190: You're right that the explicit bearer was dead under the interceptor. Both branches now run inside one `runWithAttribution`, and the `memberId` argument is derived from the same resolved attribution so the two can't disagree.

src/shared/lib/services/webhook-trigger-service.sup765.test.ts
yiw190: Fair — those tests only pinned call wiring. Rewrote them against the real resolver: org-shaped JWT, seeded `user`/`authAccount` rows, and each test asserts the `attribution.current()?.getKey()` actually observed inside the upstream delete.

src/shared/lib/services/webhook-trigger-service.sup765.test.ts
yiw190: All three done — the test harness now runs `PRAGMA foreign_keys = ON` too, so the seeds are forced honest going forward. Resolution is gated on `requiresActingMember()` since the lookups are meaningless in opaque-key mode.

src/shared/lib/services/composio-trigger-reconciler.ts
```diff
@@ -0,0 +1,141 @@
...
+  getSubscribedComposioTriggerIds,
+  resolvePlatformMemberForCandidates,
+} from '@shared/lib/services/webhook-trigger-service'
+
+/**
+ * Delete orphaned upstream subscriptions, acting as each subscription's
+ * creator member (the proxy scopes list/delete to the acting member).
+ * Returns the number of upstream subscriptions deleted.
+ */
+export async function reconcileComposioTriggers(): Promise<number> {
```
iddogino: Please remove this module (plus its test and the `startup.ts` wiring) in favour of the lazy reconcile described in the review summary. For the record, the specific problems if it were kept: (1) cancelled rows never reach a terminal state, so a host that once cancelled upstream id X re-deletes X every hour while a sibling host in the same org still holds it live; (2) `subscribed` is snapshotted before any network I/O, so a trigger re-enabled mid-pass gets the same upstream id back and is deleted under the new active row; (3) the try/catch wraps the whole per-member loop, so one failing delete aborts the remaining ids for that member in the same order every hour; (4) every pass re-scans and re-resolves all cancelled rows in history and makes one sequential proxy call per member that ever had one, including ex-members whose tokens 403 and raise a Sentry event hourly. The list-and-delete core is still useful as the one-time cleanup script.

src/shared/lib/scheduler/trigger-manager.ts
```diff
@@ -327,12 +314,10 @@ class TriggerManager {
     // creator has no platform member (SUP-226). If neither resolves to a
     // platform member (e.g. opaque-key / single-user mode), keep the prior
     // best-effort attribution (creator, else owner).
-    const candidates = [
-      trigger.createdByUserId,
-      resolveConnectedAccountOwner(trigger.connectedAccountId),
-    ]
-    const resolved = resolvePlatformMemberForCandidates(candidates)
-    const ownerUserId = resolved?.userId ?? candidates.find((c) => c) ?? null
+    const ownerUserId =
```
iddogino, line 356: (Anchored here because the target line isn't in the diff.) The reconcile point is the `activeTriggers.length === 0` branch in `processEventGroup`, ~L271-276. Suggested shape:

```ts
if (activeTriggers.length === 0) {
  const allTerminal = triggers.length > 0 && triggers.every((t) => t.status === 'cancelled' || t.status === 'failed')
  if (allTerminal) {
    // We once owned this subscription and no longer want it; the member we
    // claimed these events under is the one the proxy scopes DELETE to.
    await runWithAttribution(attribution.fromMemberId(memberId), () =>
      deleteComposioTrigger(composioTriggerId),
    ).catch((error) => captureException(error, { extra: { composioTriggerId, memberId } }))
  }
  await acknowledgeEvents(events.map((e) => e.id), memberId)
  return
}
```

Paused rows fall through to ack-only. Ids with no local rows stay ack-only too, since they may belong to another host sharing this org/member.

iddogino (changes requested):
Thanks for the fast turnaround. The six earlier comments are substantively addressed: agent-delete routes through `cancelWebhookTriggerWithCleanup`, both teardown branches run in one `runWithAttribution`, the sup765 tests use the real resolver, and the FK pragma / `upstreamId` / `requiresActingMember()` bits are in. Nice.

One direction change, and two small leftovers.

### Drop the hourly reconciler

I'd rather not ship a periodic job for this. The reconciler as written also has real problems on multi-host orgs (see inline), but even fixed, an hourly scan of every cancelled row in history plus one proxy call per member forever is a lot of machinery for a problem the root fix already prevents. Simpler shape:

1. **Keep the `minted_by_member_id` column** for explicit teardown. That part is right.
2. **Reconcile lazily, at the moment it matters.** `processEventGroup` in `trigger-manager.ts` already has the "no active local triggers for this Composio id, acking events" branch. That is the reconcile point: an event arrived for a subscription we no longer want. Delete upstream right there. Attribution is free in that branch: the events were claimed under `memberId`, and the proxy only gives a member the events for subscriptions it minted, so `attribution.fromMemberId(memberId)` is by construction the principal that can delete. No creator/owner/minted-by resolution needed. Two guards: only delete when local rows exist and every one is cancelled or failed (paused rows must stay ack-only), and leave ids with no local rows ack-only, since a cloud-workspace host sharing our org/member may own them.
3. **One-time cleanup as a script, not product code.** Take the reconciler's list-and-delete loop, add the per-id try/catch and an in-use re-check immediately before each delete, and run it once against the affected deployments to clear the legacy orphans. That does not need to live in `startup.ts`.

The one gap in the lazy path: the poll member set is built from active triggers only, so an orphan whose minting member has no other active trigger on this host is never polled. After the root fix that only recurs when a teardown fails transiently, which is already a Sentry event, and the one-time script covers the legacy rows. I'm fine with that trade.

So: delete `composio-trigger-reconciler.ts`, its test, and the `startup.ts` wiring; add the delete to the existing ack branch.

### Leftovers

- **Poll set should include `mintedByMemberId`.** `getDistinctPlatformMemberIdsForActiveTriggers` still derives members from creator / owner / stored member. A subscription minted under a non-creator member (the exact case the column was added for) is now torn down correctly but never polled, so it silently never fires. Add the column as the first candidate there too.
- **`'local'` placeholder persisted as the minting member.** In opaque-key mode the custom-endpoint mint site stores `memberId` from `resolvePlatformMemberForSession`, which can be the literal `'local'`. After a later switch to an org JWT that truthy value short-circuits every fallback and produces a `jwt::local` bearer at teardown. Persist only `actingMemberId()` (null otherwise), matching what the Composio mint site already does.

yiw190:
Dropped the hourly reconciler in c40713b in favor of the lazy shape: `processEventGroup`'s no-active-triggers branch now deletes the upstream subscription when every local row for that id is cancelled/failed, under `attribution.fromMemberId(memberId)` — the member the events were claimed as. Paused rows and ids with no local rows stay ack-only.

- Reconciler module, test, and `startup.ts` wiring removed; its list-and-delete core lives on as the one-time `scripts/cleanup-orphaned-composio-triggers.ts` (per-id try/catch, in-use re-check immediately before each delete, `--dry-run`).
- Poll set: `minted_by_member_id` is now the first candidate in `getDistinctPlatformMemberIdsForActiveTriggers`.
- Custom mint site persists only `actingMemberId()` — the `'local'` placeholder is never written to the column.

While making the sup226 suite hermetic I found it was already failing at HEAD on machines with a stored platform member (the stored-member fallback added earlier in this PR reads real settings); it now mocks `platform-auth-service`. Trigger suites 545/545 green, typecheck and eslint clean.

src/shared/lib/scheduler/trigger-manager.ts
```diff
@@ -278,10 +266,35 @@ class TriggerManager {
...
+      // Lazy reconcile (SUP-765): every local row is terminal, yet the upstream
+      // still delivers — a teardown crashed or 404ed cross-member. The member
+      // these events were claimed under is the one the proxy scopes DELETE to.
+      // Paused rows fall through to ack-only (subscription must stay alive),
+      // and ids with no local rows stay ack-only too — another host sharing
+      // this org/member may own them.
+      const allTerminal =
+        triggers.length > 0 &&
+        triggers.every((t) => t.status === 'cancelled' || t.status === 'failed')
+      if (allTerminal) {
```
iddogino: **Unreachable for the case it targets.** `pollAndClaimEvents` posts `trigger_ids: getSubscribedComposioTriggerIds()`, which is active + paused only, and the proxy filters claims on that list. Events for an id whose rows are all cancelled/failed are never claimed, so this branch never runs for an orphan. My suggestion was wrong on that point.

What does reach here: a cancel that lands between claim and processing. In that case `cancelWebhookTriggerWithCleanup` already deleted upstream, so this delete 404s and raises a Sentry event for a non-problem.

Revised shape, in the poll loop rather than the event path (see review summary): per pass, per member, select terminal rows with a null `upstream_deleted_at`, no active/paused sibling on the same upstream id, delete under minted-by, set the marker on success or 404. Two things to carry over from the script into whatever lands here: an in-use re-check (`countActiveTriggersForComposioId === 0`) immediately before the delete, since a same-slug re-enable gets the same upstream id back and would lose its subscription; and treating a 404 as already-gone rather than an error.

scripts/cleanup-orphaned-composio-triggers.ts
```diff
@@ -0,0 +1,117 @@
...
+    if (!ids) {
+      ids = new Set()
+      byMember.set(memberId, ids)
+    }
+    ids.add(upstreamId)
+  }
+
+  let deleted = 0
+  for (const [memberId, candidateIds] of byMember) {
+    await runWithAttribution(attribution.fromMemberId(memberId), async () => {
```
iddogino: **`runWithAttribution` has no effect in this process.** The platform fetch interceptor is installed only by `startup.ts:167`; `triggerFetch` (`composio/triggers.ts:29-36`) writes `Authorization: Bearer <token>` itself with no `::memberId` suffix. So in an org-JWT deployment, the only mode that produced orphans, every list and delete goes out as the bare org JWT: the proxy rejects the member-less token or 404s the delete, and the script prints `Done. Deleted 0` while every orphan survives. Call `installPlatformFetchInterceptor()` at the top of `main()`.

scripts/cleanup-orphaned-composio-triggers.ts
```diff
@@ -0,0 +1,117 @@
...
+  listActiveComposioTriggers,
+  deleteComposioTrigger,
+} from '../src/shared/lib/composio/triggers'
+import { attribution, runWithAttribution } from '../src/shared/lib/platform-attribution'
+import {
+  getSubscribedComposioTriggerIds,
+  resolvePlatformMemberForCandidates,
+} from '../src/shared/lib/services/webhook-trigger-service'
+
+const dryRun = process.argv.includes('--dry-run')
```
iddogino: **`--dry-run` is not write-free.** The first `db.` access runs `initDb()` which applies pending drizzle migrations and sets `journal_mode=WAL` (`db/index.ts:28-64`) against whatever DB the env names. An operator running this from a newer checkout against a live DB "to be safe" would migrate the production schema under the still-running older app. Open the DB read-only for dry-run (or skip `migrate()` entirely in the script and fail if the schema is behind), and say so in the header.

scripts/cleanup-orphaned-composio-triggers.ts
```diff
@@ -0,0 +1,117 @@
...
+ * One-time cleanup of orphaned upstream Composio subscriptions (SUP-765).
+ *
+ * A pre-SUP-765 cross-member teardown 404ed silently, leaving the upstream
+ * subscription live while the local trigger row is cancelled — Composio keeps
+ * delivering events that can never be claimed. This deletes exactly those
+ * subscriptions: ones a LOCAL cancelled composio-kind row claims and no local
+ * active/paused row still subscribes. Run once per affected deployment.
+ *
+ * Usage (against a deployment's data dir):
+ *   SUPERAGENT_DB_PATH=/path/to/superagent.db npx tsx scripts/cleanup-orphaned-composio-triggers.ts
```
iddogino: The documented invocation sets only `SUPERAGENT_DB_PATH`, but the token, member, and Composio mode come from `getSettings()` under `getDataDir()` (defaults to `~/.superagent` unless `SUPERAGENT_DATA_DIR`), and `getPlatformProxyBaseUrl()` needs `PLATFORM_PROXY_URL` because the build-time global is undefined under tsx. Following the usage line as written against another deployment's DB either exits early with a false all-clear ("Platform Composio inactive"), uses the operator's own org token for the deletes, or throws `Invalid URL` on every member. Document `SUPERAGENT_DATA_DIR` + `PLATFORM_PROXY_URL` and refuse to run if either is unset.

scripts/cleanup-orphaned-composio-triggers.ts
```diff
@@ -0,0 +1,117 @@
...
+      mintedByMemberId: webhookTriggers.mintedByMemberId,
+      createdByUserId: webhookTriggers.createdByUserId,
+      ownerUserId: connectedAccounts.userId,
+    })
+    .from(webhookTriggers)
+    .leftJoin(connectedAccounts, eq(connectedAccounts.id, webhookTriggers.connectedAccountId))
+    .where(
+      and(
+        eq(webhookTriggers.kind, 'composio'),
+        eq(webhookTriggers.status, 'cancelled'),
```
iddogino: **`'failed'` rows are orphans nobody cleans.** `markTriggerFailed` (`webhook-trigger-service.ts:448`) flips a row to `failed` with no upstream teardown, e.g. when a trigger-spawned session can't start because the agent was deleted. That row leaves the poll set, the runtime branch treats `failed` as terminal, and this script selects `cancelled` only. Use `inArray(status, ['cancelled', 'failed'])` here, and the poll-loop reconcile should cover both too.

Also, lines 60-75 re-implement the minted-by / creator / owner chain that `resolveCleanupMemberId` / `resolveCleanupAttribution` already centralise, and already diverge (no stored-member fallback, so a row the cancel path would tear down is SKIPped here). Export a row-shaped helper from the service and call `deleteOrphanedUpstreamSubscription` instead of duplicating the policy.

src/shared/lib/container/message-persister.ts
```diff
@@ -4401,6 +4408,10 @@ ${continuation}`
         const filterExp = input.filter_exp ?? undefined
 
         const memberId = await this.resolvePlatformMemberForSession(agentSlug, sessionId)
+        // Persist only a real acting member (SUP-765): `memberId` can be the
+        // 'local' placeholder in opaque-key mode, and a persisted 'local' would
+        // short-circuit every fallback after a later switch to an org JWT.
+        const mintedByMemberId = attribution.current()?.actingMemberId() ?? undefined
```
iddogino: This now drops a real member. `handleCreateWebhookEndpointTool` is dispatched from the container stream with no attribution wrapper, so `attribution.current()` is often null here, yet the endpoint two lines down is minted explicitly as `token::memberId` where `memberId` is a real `sub_` id (session creator or stored member). Storing null means a later teardown falls through to whatever the stored member is *then*, and a re-auth as a different member produces the same cross-member 404 on a public URL. Keep the explicit member when it isn't the placeholder:

```ts
const mintedByMemberId =
  attribution.current()?.actingMemberId() ?? (memberId === 'local' ? undefined : memberId)
```

src/shared/lib/services/webhook-trigger-service.ts
```diff
@@ -237,6 +249,17 @@ export async function getWebhookTriggersByComposioId(composioTriggerId: string):
     )
 }
 
+/** All local rows for an upstream id regardless of status; the lazy reconcile
+ * needs the terminal ones too (SUP-765). */
+export async function listAllWebhookTriggersByComposioId(
```
iddogino, line 251: With trigger-manager switched to `listAllWebhookTriggersByComposioId`, `getWebhookTriggerByComposioId` and `getWebhookTriggersByComposioId` above no longer have a production caller and survive only via their unit tests. Three near-identical selectors where the active-only ones silently miss paused/terminal rows is the exact blind spot this PR had to work around. Delete both (and their test block), or make the active-only one a filter over `listAll`.

iddogino (changes requested):
Reconciler removal, minted-by in the poll set, and the `'local'` fix all landed cleanly, and the lazy branch is implemented exactly as I asked. Unfortunately what I asked for doesn't work, and that's on me: `pollAndClaimEvents` scopes the claim to `getSubscribedComposioTriggerIds()` (active + paused only, `webhook-events-client.ts:83`), so the proxy never returns events for an all-cancelled id and the `allTerminal` branch is unreachable for the orphan it targets. The only way in is a cancel racing an in-flight poll, where teardown already ran and the branch just 404s into Sentry.

### Revised direction: reconcile on the poll loop, no events needed

Each poll pass already runs per member. Add one step: select rows with `status in ('cancelled','failed')`, a non-null upstream id, no active/paused sibling on that id, and a null terminal marker (new column, e.g. `upstream_deleted_at`, can go into migration 0039 since it hasn't shipped). For each, delete upstream under `mintedByMemberId` (creator/owner chain for pre-column rows), and set the marker on success **or 404**. That is bounded by the marker, piggybacks a loop that already exists, needs no listing call, and retries transient failures for free. It also fixes the `'failed'` gap below, since `markTriggerFailed` never tears down upstream today.

The cleanup script then only needs to handle legacy rows, and it needs the fixes inline before it can be run against a real deployment.

### Summary of inline comments

- `trigger-manager.ts`: replace the event-driven branch with the poll-loop step above; add an in-use re-check before the delete; treat 404 as already-gone.
- `scripts/cleanup-orphaned-composio-triggers.ts`: install the fetch interceptor (attribution has no effect without it); make `--dry-run` truly read-only; document `SUPERAGENT_DATA_DIR` and `PLATFORM_PROXY_URL`; include `'failed'` rows; reuse the service's member chain instead of re-implementing it.
- `message-persister.ts:4414`: keep the real member when it isn't `'local'`.
- `webhook-trigger-service.ts`: the two active-only by-composio-id selectors are now dead in production.

Checks on this revision: 37/37 tests in the touched files, typecheck and lint clean.

yiw190:
Poll-loop reconcile in 01248fa0; the `allTerminal` event branch is removed.

- Reconcile runs at the top of each poll pass over terminal rows with an upstream id, no active/paused sibling, and null `upstream_deleted_at` (new column in 0039). Deletes under `mintedByMemberId`, re-checks in-use right before the delete, 404 = gone, marker set on success or 404. 25 per pass, per-row failures retried next pass.
- Script: interceptor installed, `SUPERAGENT_DATA_DIR` + `PLATFORM_PROXY_URL` required, DB opened read-only (exits if schema is behind, never migrates), `failed` rows included, reuses the service helpers instead of its own member chain.
- `message-persister.ts` keeps the real member when it isn't `'local'`; the two active-only by-id selectors and their tests are deleted.

Verified against the real staging proxy: cross-member `DELETE` → 404 (subscription untouched), owner → 200, already-deleted → 404; bare token → 400, non-member → 403. Datawizz backlog is 67 rows / 61 ids → 3 passes; the 4 ids still emitting events already 404 (Composio-side).

src/shared/lib/services/webhook-trigger-service.ts
```diff
@@ -491,14 +487,186 @@ export async function cancelWebhookTriggerWithCleanup(
...
+  try {
+    await runWithAttribution(cleanupAuth, () =>
+      deleteUpstream(
+        trigger.kind,
+        cleanupAuth?.actingMemberId() ?? resolveCleanupMemberId(trigger),
+        upstreamId,
+      ),
+    )
+  } catch (error) {
+    if (!isUpstreamNotFound(error)) throw error
```
iddogino: **Blocking.** Your staging test showed the proxy 404s a cross-member DELETE. That means a 404 here is ambiguous: "gone" or "wrong member". For rows with `mintedByMemberId` set (or in opaque-key mode, where the proxy ignores the member) the member is known and 404 = gone is safe. For pre-column rows the member is guessed via `resolveCleanupAttribution`, and the datawizz-style orphan (minted under ambient member A, `createdByUserId` resolves to B) gets a 404 as B, is stamped `upstream_deleted_at`, and is excluded from `listOrphanedUpstreamTriggers` forever, silently. That is the population this PR exists for, and the deleted script cross-checked existence before deleting.

Suggest: accept 404 as terminal only when the acting member is known. For a guessed member, try each candidate in order (minted → creator → owner → stored) and mark only after all 404; otherwise leave the marker null and `captureException` so the row stays diagnosable.

src/shared/lib/scheduler/trigger-manager.ts
```diff
@@ -167,6 +155,16 @@ class TriggerManager {
     this.isProcessing = true
 
     try {
+      // Owed upstream teardowns are found from local rows, not events: the
+      // claim is scoped to active/paused ids, so an orphan never delivers
+      // an event here (SUP-765). Runs before the member early-return so a
+      // host with only terminal rows still converges.
+      if (getPlatformAccessToken()) {
```
iddogino: **Blocking.** Three things compound here: (a) this runs before `pollAndClaimEvents`, so up to 25 sequential upstream DELETEs now sit in front of every event claim; (b) it's inside the `isProcessing` guard, so a realtime INSERT nudge that lands mid-reconcile is dropped at `if (this.isProcessing) return`, and `triggerFetch` / `endpointsFetch` have no timeout, so a hung proxy stalls every triggered session on the host; (c) `pollAndProcess` has no periodic timer (the only `setInterval` is the 50-minute JWT refresh), so "converges over subsequent passes" assumes events keep arriving. On a quiet org, the first pass does 25 and the remainder waits for the next restart. The removed test "still acks events when the lazy teardown fails" encoded the invariant that teardown trouble never blocks delivery.

Move the reconcile after the claim and outside the guard (fire-and-forget with its own re-entrancy flag), or give it a cheap dedicated timer. Either way, add a timeout to the delete fetches.

src/shared/lib/db/migrations/0039_webhook_trigger_minted_by_member.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `webhook_triggers` ADD `minted_by_member_id` text;--> statement-breakpoint
+ALTER TABLE `webhook_triggers` ADD `upstream_deleted_at` integer;
```
iddogino: **Blocking.** No backfill, so every historical cancelled/failed row with an upstream id becomes an owed teardown on upgrade, including the overwhelming majority whose teardown succeeded. Two consequences: with the 404 rule above, most get stamped by a wrong-member 404; and where a member is shared across hosts (desktop + cloud workspace, same org/member), host A's stale terminal row for `ti_X` will DELETE `ti_X` while host B holds it live, since `countActiveTriggersForComposioId` only checks local rows. Backfill `upstream_deleted_at = updated_at` for rows terminal before this migration (their teardown either succeeded or is what the one-time script is for), so the runtime reconcile only ever sees rows that went terminal under the new code.

Also minor: 0039 was edited in place rather than adding 0040, and `_journal.json`'s `when` is unchanged, so a DB that already applied the earlier 0039 (anyone who ran commit b21fcc76) never gets `upstream_deleted_at` and every trigger SELECT throws. Low blast radius since it never shipped, but worth a 0040 to be safe.

src/shared/lib/services/webhook-trigger-service.ts
```diff
@@ -491,14 +487,186 @@ export async function cancelWebhookTriggerWithCleanup(
...
+// the remainder converge over subsequent passes.
+const RECONCILE_BATCH_SIZE = 25
+
+/**
+ * Poll-loop reconcile (SUP-765): retry owed teardowns — a crash, a pre-column
+ * cross-member 404, or `markTriggerFailed` (which never tears down upstream).
+ * Bounded by the `upstreamDeletedAt` marker; failures are captured per row.
+ */
+export async function reconcileOrphanedUpstreamSubscriptions(): Promise<number> {
+  const candidates = listOrphanedUpstreamTriggers().slice(0, RECONCILE_BATCH_SIZE)
```
iddogino: Head-of-line starvation: the batch is sliced to 25 **before** the per-row skips run, and two classes of candidate never get a marker: rows where `deleteOrphanedUpstreamSubscription` returns `false` (composio kind while platform Composio is inactive) and rows that throw non-404 every pass (401/403 for an ex-member). The SELECT has no ORDER BY, so those occupy the same 25 slots on every pass. A bring-your-own-Composio-key deployment with 25+ historical composio rows and one live custom-endpoint orphan at position 26 never reaches it, and reports 0 with no error.

Filter unreachable kinds in `listOrphanedUpstreamTriggers` (custom only when platform token, composio only when platform Composio active), and give persistent failures an attempt counter or ORDER BY so the batch rotates.

src/shared/lib/scheduler/trigger-manager.ts
```diff
@@ -327,12 +327,10 @@ class TriggerManager {
     // creator has no platform member (SUP-226). If neither resolves to a
```
iddogino, line 353: (Anchored on the nearest touched hunk.) `markTriggerFailed` at ~L346 is now interpreted by the reconcile as "owed teardown", but its only producer is `agentExists()` at ~L342, a filesystem probe that can be transiently false (data volume not yet mounted at boot, rename in progress). Before this PR a `failed` row left upstream intact and could be hand-fixed; now the next pass deletes the subscription and there is no route from `failed` back to `active`. Real agent deletion already goes through `agent-cleanup-service` → `cancelWebhookTriggerWithCleanup`. Have this branch call that explicitly (it has the last-one-out and reachability guards), and keep `failed` out of the orphan query.

src/shared/lib/container/message-persister.ts
```diff
@@ -4288,6 +4289,11 @@ ${continuation}`
           return
         }
 
+        // The proxy scopes the subscription to the acting member of THIS mint
+        // request (ambient ALS), which is not always the session creator.
+        // Record it so teardown can delete as the same member (SUP-765).
+        const mintedByMemberId = attribution.current()?.actingMemberId() ?? undefined
```
iddogino, line 4295: The custom-endpoint site now keeps the explicit member when it isn't `'local'`, but this Composio site still records ALS-only. Container-dispatched mints (the common path) have no ambient attribution, so `mintedByMemberId` is null and every such row is treated as pre-column at teardown, which is what feeds the 404 problem above. Apply the same fallback as the custom site: `actingMemberId() ?? (memberId === 'local' ? undefined : memberId)`, using whatever member `enableComposioTrigger` is actually sent as.

src/shared/lib/services/webhook-trigger-service.ts
```diff
@@ -491,14 +487,186 @@ export async function cancelWebhookTriggerWithCleanup(
...
+async function deleteUpstream(
+  kind: WebhookTrigger['kind'],
+  memberId: string,
+  upstreamId: string,
+): Promise<void> {
+  if (kind === 'custom') await disablePlatformWebhookEndpoint(memberId, upstreamId)
+  else await deleteComposioTrigger(upstreamId)
+}
+
+function isUpstreamNotFound(error: unknown): boolean {
```
iddogino, line 513: Reuse nits, non-blocking: (1) `WebhookEndpointsApiError` is a third structurally identical `{message, statusCode}` class next to `ComposioTriggerError` and `ComposioApiError`, so this has to enumerate by `instanceof`; a duck-typed `statusCode === 404` on any Error (as `nango-account-provider.ts` and `platform-k8s-runtime.ts` already do) means a new throw site can't silently regress to perpetual retries. (2) `['cancelled','failed']` is inlined in `markUpstreamDeleted` and `listOrphanedUpstreamTriggers`, and `['active','paused']` in six places in this file; hoist `TERMINAL_STATUSES` / `SUBSCRIBED_STATUSES`. (3) The sup765 test re-declares both error classes inside `vi.hoisted`; use `vi.importActual` so a constructor change fails a test.

scripts/cleanup-orphaned-upstream-subscriptions.ts
```diff
@@ -0,0 +1,134 @@
...
+
+function requireEnv(name: string): string {
+  const value = process.env[name]?.trim()
+  if (!value) fail(`${name} is not set; refusing to run against an ambiguous deployment.`)
+  return value
+}
+
+// Read-only pre-flight: the shared db module runs pending migrations on first
+// touch, so make sure there are none before importing anything that uses it.
+function assertSchemaCurrent(dbPath: string): void {
```
iddogino, line 86: Non-blocking: `assertSchemaCurrent` hand-parses `_journal.json` and re-implements drizzle's `folderMillis` vs `__drizzle_migrations.created_at` comparison. `readMigrationFiles({ migrationsFolder })` from `drizzle-orm/migrator` returns exactly what `migrate()` compares; use that and export `getMigrationsFolder()` from `db/index.ts` so there is one path resolution. Similarly the interceptor bootstrap at ~L101 duplicates `startup.ts:163-167` minus the `decodeOrgIdFromToken` guard; extract a shared helper. Both matter because this script's whole job is to not diverge from what the app does, and it isn't run in CI.

iddogino (changes requested):
All five asks from last round landed as requested, and the staging verification is exactly the data I needed. Thanks. Checks on this revision: 65/65 tests in the touched files, typecheck and lint clean.

The reconcile mechanics are sound, but your own staging result (cross-member DELETE returns 404) turns the "404 = gone" rule into a correctness regression on the incident population, and two other issues make the first upgrade on a large deployment risky. Holding on the first three inline; the rest are follow-ups I'd like in this PR.

**Blocking**

1. **404 under a guessed member hides the orphan instead of fixing it** (`tearDownUpstream`). For pre-column rows the member comes from the creator chain, so the exact row SUP-765 is about gets a wrong-member 404, is stamped `upstream_deleted_at`, and leaves the candidate set forever with no Sentry event.
2. **Reconcile sits in front of the claim, inside `isProcessing`, with no timer behind it.** Up to 25 sequential proxy DELETEs with no timeout now precede every event claim, nudges arriving mid-reconcile are dropped, and since `pollAndProcess` only runs on start / nudge, "converges over passes" assumes events keep arriving.
3. **No backfill.** Every historical terminal row becomes a candidate on upgrade. With (1) most get stamped by a wrong-member 404. With a member shared across hosts (desktop + cloud workspace), host A's stale terminal row deletes host B's live subscription, because the in-use re-check is local only.

**Follow-ups**

- Head-of-line starvation in the 25-row batch.
- `failed` now implies teardown, but its only producer is a filesystem probe that can be transiently false.
- The Composio mint site still records only ALS attribution, so container-dispatched mints stay pre-column and feed (1).
- 0039 edited in place; reuse nits (three identical error classes, inlined status lists, script re-implements journal parse + interceptor bootstrap).

yiw190:
**Resolved Blocking**

**1** A 404 only counts as gone when we actually know the member (mintedByMemberId, or opaque-key mode). Pre-column rows walk creator → owner → stored in order; if all 404, UpstreamOwnerUnresolvedError, marker stays null, Sentry.
**2** Reconcile runs after the claim, outside isProcessing, fire-and-forget with its own flag, 60s cooldown, drains only while a full batch succeeds, attempts capped at 10. 15s timeout on the two DELETEs only. Went fire-and-forget over a timer — boot drains everything that can succeed, the rest is transient and already in Sentry. Can add a timer if you'd rather.
**3** 0039 back to what was first pushed; 0040 adds marker + attempts and backfills pre-existing terminal rows (COALESCE(cancelled_at, created_at) — no updated_at on this table).


**Follow-ups** — unreachable kinds filtered; failed dropped and markTriggerFailed removed (agent-gone calls cancelWebhookTriggerWithCleanup, only ENOENT counts as gone). Duck-typed 404, SUBSCRIBED_STATUSES hoisted, importOriginal in the test, script on readMigrationFiles + shared interceptor helper.

**Didn't take**: the Composio mint site. Unlike the custom site there's no memberId in scope — the member comes purely from ALS. Anything I'd fall back to is a guess, and once it's in mintedByMemberId teardown treats it as known and trusts a 404 — the same hole as (1). Null keeps the row on the try-every-candidate path. And a mint with no ALS can't produce a row anyway: it goes out as a bare org JWT and the proxy rejects it (staging: 400).

**Self review** (fba3b7ea): the timeout was on the shared fetch clients so mints were bounded too — moved to DELETEs only. unknown presence logged "skipping" but the batch is acked and dropped (no claim lease on the proxy) — log fixed + Sentry, real fix is a follow-up. New bookkeeping columns leaked to every role — now owner-only.

PR #943, opened 2026-09-03
feat(models): add Muse Spark 1.3 and 1.3c to the platform catalog

yiw190:
## Summary
- Meta released Muse Spark 1.3 on the same platform-proxy route and prices as 1.2. The picker still treated 1.2 as the Meta flagship, so a new agent would not see 1.3.
- Add `muse-spark-1.3` and `muse-spark-1.3-contributor` to the platform catalog, mark 1.3 as latest/default, and keep 1.1 / 1.2 / 1.2c selectable.

Depends on the proxy allowlist in SkillfulAgents/platform#381. Without that deploy, picking 1.3 is rejected as an unsupported Meta model.

## Test plan
- [x] `npx vitest run src/shared/lib/llm-provider/model-catalog.test.ts src/renderer/components/messages/model-family-list.test.tsx`
- [ ] On a platform-provider workspace, Meta tab shows Muse Spark 1.3 as latest
- [ ] 1.3c still shows the product-improvement warning; standard 1.3 does not
- [ ] After platform#381 is deployed, a send on 1.3 completes instead of `unsupported_model`

### DIFF
--- a/src/renderer/components/messages/model-family-list.test.tsx
+++ b/src/renderer/components/messages/model-family-list.test.tsx
@@ -38,6 +38,8 @@ describe('familyDisplayName', () => {
     expect(familyDisplayName('sonnet')).toBe('Sonnet')
     expect(familyDisplayName('gpt')).toBe('GPT')
     expect(familyDisplayName('glm')).toBe('GLM')
+    expect(familyDisplayName('muse')).toBe('Muse Spark')
+    expect(familyDisplayName('muse-contributor')).toBe('Muse Spark Contributor')
   })
 })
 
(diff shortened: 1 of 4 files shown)


iddogino:
Picker after eabcaeae, on the Meta tab of the composer (mock web build with `llmProvider: platform`, so this is the real `PLATFORM_CATALOG`, not a fixture).

Two lineage rows, one per tier, with version chips. 1.3 is the default and latest for both; the standard row keeps 1.1 as a pinnable chip so existing pins don't break.

| Standard row hovered | Contributor row hovered |
| --- | --- |
| Meta tab: Muse Spark row with 1.3 / 1.2 / 1.1 chips, 1.3 selected | Meta tab: Muse Spark Contributor row with 1.3 / 1.2 chips |

Pinning a contributor version surfaces the data-use warning, same as before the family split:

Muse Spark Contributor 1.2 pinned, with the product-improvement data-use warning

PR #945, opened 2026-09-03
feat(renderer): deliver files as rows

westaskew:
## Summary

Stacked on https://github.com/SkillfulAgents/SuperAgent/pull/950 (which is stacked on https://github.com/SkillfulAgents/SuperAgent/pull/949) — merge those first.

- **Delivered files render as full-width rows** under the turn instead of pills: the icon tile, the filename, the agent's description plus byte size, and an outlined Download button. Clicking the row opens the drawer; the button downloads directly. Folders open the folder browser.
- **Size comes from the `deliver_file` tool result**, the only record of the file at delivery time. The parser accepts both the plain-string and content-block-array result shapes.
- **The composer's byte formatter** moves to a shared util so both use one.

## Test plan

- [x] messages / ui / lib / shared suites (1958 passing); new tests for the row (name, description, size, click, keyboard, folder, download-link isolation), the size parser, and the byte formatter
- [x] E2E `file-preview.spec.ts` updated to the `file-delivery-row` test id (not run locally)
- [ ] Visual check in the Electron app: delivered rows in a session, hover state, dark mode


iddogino:
Reviewed as part of the stack (main ← #949 ← #950 ← #945 ← #946 ← #948). Screenshots are from the E2E mock build at this PR's head, with the base (#950) alongside for comparison.

**Delivered files**, pills → rows (description and size come from the `deliver_file` result; the mock reports its own sizes):

Delivered files as pills vs rows

### Blockers

- **Enter on the Download link opens the preview instead of downloading.** The row's `onKeyDown` (`src/renderer/components/ui/file-delivery-row.tsx:58`) also receives the keydown bubbling up from the nested `<a>`, calls `preventDefault()` (which cancels the anchor's activation) and then `open()`. A keyboard user who tabs onto "Download report.pdf" and presses Enter gets the tray, never the file. The link's `stopPropagation` only covers click. Guarding with `e.target === e.currentTarget` fixes it, but the nested interactive (an `<a>` inside `role="button"`) also fails axe's `nested-interactive` rule, so moving the link out of the button element is the cleaner shape.

### Should fix

- **The size is scraped from prose written for the model.** `getDeliveredFileSize` regexes the container's "File "x" (N bytes) has been delivered…" sentence (`src/shared/lib/tool-definitions/deliver-file.ts:29`), and the container and app ship separately, so any rewording silently drops the size everywhere. The `(?:size: )?` alternative exists only for the E2E mock's invented format; making the mock emit the real sentence and dropping the alternative keeps test-only grammar out of production. Longer term, a structured `sizeBytes` on the persisted tool call would avoid parsing altogether.
- **First-match regex.** The result puts the quoted path before the size, so `report (2 bytes).txt` parses as 2 B. Anchoring on `" (\d+ bytes) has been delivered` avoids it.
- **`resultText` re-implements `parseToolResult(...).text`** (`src/renderer/lib/parse-tool-result.ts:159`), which also handles the JSON-string-of-blocks shape this copy misses. Passing the already-flattened text into `getDeliveredFileSize` from `message-list.tsx` removes the second parser.
- **The formatter extraction stops halfway.** `formatBytes` in `tool-call-item.tsx:187` (KB to 0 decimals) and the inline `toFixed(1)} KB` in `file-request-item.tsx:187` remain, so one expanded turn can show "12.5 KB" on the row and "13 KB" on the tool card.

### Minor

- `key={file.filePath}` collides when the same path is delivered twice in one turn. That predates the PR, but rows now carry per-call description/size that can differ between the two.
- `isFolder` / `getDisplayName` / `getRelativePath` (`file-delivery-row.tsx:20-31`) are the fourth copies of helpers in `file-download-pill.tsx`, `file-preview-context.tsx:97`, `workspace-file-url.ts:3`, and `tool-definitions/read.ts:11`.
- `tool-renderers/deliver-file.tsx:91` still renders `FileDownloadPill` for the same call, so an expanded turn shows one delivery twice with different affordances.
- `DeliveredFiles` sits outside `MessageErrorBoundary` and `description?.trim()` runs on an unvalidated cast, so a non-string description in a persisted transcript blanks the whole list rather than one message.
- The folder branch is unreachable from the producer (`deliver_file` errors on anything that isn't a regular file, and the collector filters `isError`); the folder unit test asserts behavior nothing can trigger.

### Nits

- Rows span the full column (like the turn-summary divider) while assistant paragraphs wrap at `max-w-[80%]` inside `MessageItem`, so a long answer's right edge and the row's Download button won't line up. Looks intended; flagging in case it isn't.
- `data-file-name`, `data-file-path`, `className`, and `data-testid="delivered-files"` have no readers; the E2E uses `file-delivery-row` + `hasText`.
- `message-list.tsx:627` could use `deliverFileDef.parseInput` instead of a widened inline cast.
- The test's `vi.mock` exports a `FilePreviewProvider` that is never rendered.

iddogino (approved):

PR #946, opened 2026-09-03
feat(renderer): restyle the file drawer header and tabs

westaskew:
## Summary

Stacked on https://github.com/SkillfulAgents/SuperAgent/pull/945 — merge that first.

- **Chrome-style tab strip.** Uniform-width tabs that shrink in lockstep; the active tab is a raised card whose open bottom hides the body's top border; labels fade instead of ellipsizing; per-tab close, always visible on the active tab; overflow arrows slide the row one tab at a time. The track is translated rather than scrolled so the active tab's 1px overhang onto the body is never clipped.
- **Header row removed.** The drawer's close/hide control rides at the right end of the tab strip, aligned with the body card. Icon is `panel-right`.
- **File body is a bordered, rounded card** inset 16px on the tab rail's gray. It's headed by the filename with its byte size (from a HEAD request's Content-Length via a new `useFileSize` hook) and the per-file copy/download actions, with tooltips.
- **No-preview state** drops the question-mark badge; download icons use `arrow-down-to-line`.

## Test plan

- [x] file-preview unit suites (120 passing): tab bar overflow/arrow tests with faked geometry, tray title/size/actions, copy button
- [x] E2E `file-preview.spec.ts` header/title/close assertions updated (not run locally)
- [ ] Visual check in the Electron app: tab strip at 1, 5, and 15 tabs; compact vs wide close control; dark mode

### DIFF
--- a/e2e/specs/file-preview.spec.ts
+++ b/e2e/specs/file-preview.spec.ts
@@ -447,7 +447,8 @@ test.describe('File Preview', () => {
       // right-side panel control with a left-side close button.
       const viewportWidth = page.viewportSize()!.width
       await expect(async () => {
-        for (const control of [header.getByTitle('Close file preview'), header.getByTitle('Download file')]) {
+        const title = page.getByTestId('file-preview-title')
+        for (const control of [header.getByTitle('Close file preview'), title.getByLabel('Download file')]) {
           await expect(control).toBeVisible()
           const box = await control.boundingBox()
           expect(box).not.toBeNull()
--- a/e2e/specs/file-preview.spec.ts
+++ b/e2e/specs/file-preview.spec.ts
@@ -514,7 +514,7 @@ test.describe('File Preview', () => {
     await expect(page.getByTestId('file-preview-copy')).toHaveCount(0)
   })
 
-  test('copies the open text file from the preview header', async ({ page }) => {
+  test('copies the open text file from the title row', async ({ page }) => {
     await page.context().grantPermissions(['clipboard-read', 'clipboard-write'])
     await agentPage.createAgent(`CopyFile ${Date.now()}`)
     const agentSlug = await getLatestAgentSlug(page)
@@ -527,8 +527,9 @@ test.describe('File Preview', () => {
     await filePill.click()
     await expect(markdown(page).getByRole('heading', { name: 'Test Report' })).toBeVisible({ timeout: 10000 })
 
-    const header = page.getByTestId('file-preview-header')
-    await header.getByTestId('file-preview-copy').click()
+    // The per-file actions live with the filename, not in the panel header.
+    const titleRow = page.getByTestId('file-preview-title')
+    await titleRow.getByTestId('file-preview-copy').click()
 
     // The real gate: a browser only honours the write while the click's user
     // gesture is live, so this fails if the handler awaits a fetch first.
@@ -536,7 +537,7 @@ test.describe('File Preview', () => {
       .toBe('# Test Report\n\nThis is a test with **bold** text.')
 
     await expect(page.getByText('Copied contents of “report.md”')).toBeVisible()
-    await expect(header.getByTestId('file-preview-copied-icon')).toBeVisible()
-    await expect(header.getByTestId('file-preview-copy-icon')).toBeVisible({ timeout: 5000 })
+    await expect(titleRow.getByTestId('file-preview-copied-icon')).toBeVisible()
+    await expect(titleRow.getByTestId('file-preview-copy-icon')).toBeVisible({ timeout: 5000 })
   })
 })
--- a/src/api/routes/agents.test.ts
+++ b/src/api/routes/agents.test.ts
@@ -668,6 +668,10 @@ async function getReq(app: Hono, url: string): Promise<Response> {
   return app.request(`http://localhost${url}`, { method: 'GET' })
 }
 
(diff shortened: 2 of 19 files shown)


iddogino:
Reviewed as part of the stack (main ← #949 ← #950 ← #945 ← #946 ← #948). Screenshots are from the E2E mock build at this PR's head, with the base (#945) alongside for comparison.

**Drawer with one tab**, header row → tab strip with a titled body card (size from the HEAD request):

Drawer before and after

**Tab strip at 5 and 14 tabs**, at the default drawer width (~450 px), light and dark:

Tab strip at five and fourteen tabs, before and after

### Blockers

- **HEAD requests leak a file descriptor in the API process.** `useFileSize` sends `HEAD` to `/:id/files/*`, which has no HEAD branch (`src/api/routes/agents.ts:6559`). Hono synthesizes HEAD by running the GET handler and discarding the body, so the `fs.createReadStream` opened there is never consumed or cancelled. Reproduced with the real Hono + `@hono/node-server` on Node 22: 50 HEADs against a 5 MB file left 47 extra fds open. Files under the 64 KB high-water mark don't leak (the stream drains into its buffer and ends), which is why a small-file probe looks clean. Every tab open, redelivery, and >60 s refetch pays this, and #948 multiplies it with one HEAD per sent chip. Fix: return `c.body(null)` for `c.req.method === 'HEAD'` after setting `Content-Length` and before opening the stream, or register the route with `agents.on(['GET', 'HEAD'], …)`.
- **The tab strip mis-measures while slid.** `measure()` derives `maxOffset` from `group.scrollWidth` (`src/renderer/components/file-preview/file-tab-bar.tsx:38`), but the track is a translated child and Chromium shrinks scrollable overflow by the translate. Verified in headless Chromium: at offset 390 on a 689 px track in a 300 px group, `scrollWidth` reads 300 and the computed max drops to 0. So any re-measure while `offset > 0` (closing a tab, resizing the drawer) clamps the offset and unmounts the arrows, leaving tabs unreachable. Repro: open 10 files in a narrow drawer, click "Later tabs" to the end, close the active tab. The unit test mocks `scrollWidth` statically, so it can't catch this. Measure the untransformed track (`trackRef.current.scrollWidth` or a sum of child widths) instead.
- **Stale E2E locators.** `e2e/specs/file-preview.spec.ts:531` still scopes `file-preview-copy` / `file-preview-copied-icon` / `file-preview-copy-icon` under `file-preview-header`, but the copy button moved to the sibling `file-preview-title` row (only the Download locator at line 451 was migrated). That test will time out. The comment at lines 446-447 about a "left-side close button" is also stale.

### Should fix

- **Labels vanish at five tabs.** At the default drawer width the uniform-basis tabs already hit the `min-w-[4.25rem]` floor with five open, and the fade mask leaves two letters per label ("cl", "vc", "re", "da" in the screenshot); at fourteen they are icon-only. Chrome gets away with lockstep shrinking because it keeps the active tab readable; here the active tab shrinks too. Either exempt the active tab from the shrink, or raise the floor and overflow sooner.
- **No size in Cloud Workspace mode.** The desktop cloud proxy strips `content-length` unconditionally (`src/api/routes/cloud-proxy.ts:93`) and never re-frames it for a body-less HEAD, so the hook resolves `null` there and the label silently never renders. Exempt it when the upstream body is empty.
- **Clipped tabs stay in the tab order with no focus reveal.** The old `overflow-x-auto` let the browser scroll a focused tab into view; `overflow-x-clip` is not a scroll container, and only `activeIndex` slides the track (`file-tab-bar.tsx:93`). With 8+ tabs, Tab moves through invisible buttons and `Close <name>` spans with no visible focus ring. A `focusin` handler on the track that reveals the focused tab the same way the active-index effect does would restore it.
- **Trackpad and touch scrolling of the strip is gone.** Two-finger swipe, shift+wheel, and drag did work with the native scroller; now hidden tabs are reachable only via the two chevrons, one tab per click. An `onWheel` on the group folding `deltaX` into `setOffset` is cheap.

### Minor

- The body card is `rounded-lg rounded-tl-none` unconditionally (`file-preview-tray-content.tsx:78`), but the square corner only makes sense when tab 0 is active at offset 0. Since opening a file activates the newly appended tab, the default multi-tab state shows a square top-left corner under a transparent inactive tab.
- `CopyFileButton` wraps itself in a `TooltipProvider` (`copy-file-button.tsx:56`) inside the tray's own provider, so hovering Copy then sliding to Download doesn't get the shared skip-delay.
- `file-preview-pagination.integration.test.tsx:65` now renders the real `useFileSize` under a real `QueryClientProvider` and issues an unmocked `fetch(..., { method: 'HEAD' })` from jsdom; it passes only because the relative URL is rejected and swallowed. Mock the hook as the tray-content test does.

### Nits

- `getAgentFileApiPath(activeTab…)` is computed twice per render and `activeTab?.kind === 'file'` re-narrowed four times (`file-preview-tray-content.tsx:27`); one `activeFile` const above the hook would serve the HEAD path, `fileUrl`, and `downloadUrl`.
- `getApiBaseUrl() + getAgentFileApiPath()` is now hand-concatenated in three places; a `getAgentFileUrl(agentSlug, filePath, { inline, version })` helper next to `getAgentFileApiPath` would cover them.
- `transition-transform` re-targets on every resize frame; the drawer already gates its width transition on `!isResizing`.
- `fakeLayout` in `file-tab-bar.test.tsx:48` returns a restore closure nothing calls.

iddogino (approved):

PR #948, opened 2026-09-03
feat(renderer): upload chips match the delivered-file treatment, in the composer and once sent

westaskew:
## Summary

Stacked on #946 (which is stacked on #945) — merge those first.

- **Composer chips** use the `FileIconTile` (gray square, category-colored stroke) on a white chip. The remove X is always visible on a gray circle in the top-right corner. The done state shows just the size (no check mark).
- **Image uploads in the composer** are a bare 44px square: full-bleed picture, no name/size text (hover for both), with spinner/error overlays instead of a text line.
- **Sent user messages** render attachments with a new `SentAttachmentChip` that mirrors the composer: tile + clean name + size (via `useFileSize`), click opens the drawer. Replaces the `FileDownloadPill` there.
- **Sent images are a special case.** Files and images are laid out in separate blocks so a tall picture never stretches the chips. Up to three images stack at native aspect under a 256px cap; four or more go into a right-aligned 3-column grid of cropped squares.
- **Upload names are cleaned.** A shared `displayNameForPath` strips the `<epoch ms>-` prefix the upload route adds, used by the sent chip and by the drawer's tab/title names.

## Test plan

- [x] messages + context + utils suites (854 / 54 / util tests passing): chip file/image/folder/grid cases, count → layout helper, prefix stripping, composer chip states
- [x] Existing upload E2E specs keep working — the sent chip keeps the `file-pill` test id and exposes the file name as text (`sr-only` for images)
- [ ] Visual check in the Electron app: send 1, 3, and 5 images with and without other files; composer with a mix; dark mode


iddogino:
Reviewed as part of the stack (main ← #949 ← #950 ← #945 ← #946 ← #948). Screenshots are from the E2E mock build at this PR's head, with the base (#946) alongside for comparison.

**Composer chips**: icon tile on a white chip, corner remove button, image as a bare square (light and dark):

Composer chips before and after

**Sent attachments**: pills → chips with clean names and sizes; the image shows as a picture:

Sent attachments before and after

**Sent images**: five images as a 3-column grid, two images plus a file stacked:

Sent image layouts

### Blockers

- **Composer image chips lose their upload feedback.** The image branch drops `UploadMeta` (`src/renderer/components/messages/attachment-preview.tsx:178`), so a failed image upload has no Retry control and its error is a hover-only `title` (unreachable on touch), and a queued image is indistinguishable from a done one. The PDF beside it still shows "waiting", the error text, and Retry. Dropping five screenshots and having one 500 leaves Send as the only recovery path, with no cue.
- **This PR multiplies the HEAD fd leak from #946.** Every `SentAttachmentChip` calls `useFileSize` (`sent-attachment-chip.tsx:48`), so a thread with 30 messages × 4 attachments issues ~120 HEADs on every session open, each leaking a descriptor in the API process for files over 64 KB (details on #946). Fix the route there; here, skip the size fetch for images, where it only feeds a hover title.

### Should fix

- **The upload-prefix strip is applied to every path, not just uploads.** `displayNameForPath` runs on every drawer tab and title via `getDisplayName` (`src/renderer/context/file-preview-context.tsx:99`), so agent-written `1725000000001-run.log` and `…002-run.log` both display as `run.log`. Folder uploads are stored without a prefix (`agents.ts` `handleFolderUpload` uses `uploads/<basename>/`), so the PR's own folder test encodes a path the server never produces, and a folder literally named `1700000000000-backup` is renamed. The original filename is already returned by the upload route; carrying it (or scoping the strip to `/workspace/uploads/` singles) avoids guessing from the path.
- **Same file, two names on one screen.** Only the sent chip and the drawer title/tabs use the stripped name; `unsupported-renderer.tsx:12-26` directly beneath the title, `image-renderer.tsx` alt, `file-delivery-row.tsx`, `file-download-pill.tsx`, and `home-bookmarks.tsx` all still show the raw prefixed name.
- **Sent images have no error fallback.** `` has no `onError` and the name is `sr-only` (`sent-attachment-chip.tsx:81`), so after the agent cleans `uploads/` or the container is recreated, every past message with images shows a bare broken-image square with no visible filename. The old pill always showed the name.
- **Name announced twice.** Inside one `role="button"`, `` plus the `sr-only` span compute an accessible name of "photo.png photo.png" (`sent-attachment-chip.tsx:90`). Keep the span for text lookups and give the image `alt=""`.
- **Full-resolution originals in 130 px squares.** The image `` loads the original with no `loading="lazy"` / `decoding="async"` from a route that sets `Cache-Control: private, no-store`, so every remount re-downloads multi-MB PNGs and decodes them on the main thread. `MARKDOWN_COMPONENTS.img` already sets both attributes.
- **Duplicate React keys.** `key={idx}` became `key={filePath}` (`message-item.tsx:494` and `:513`), but folder uploads at `/workspace/uploads/<basename>/` are not unique within a message: attaching `~/a/src` and `~/b/src` lists the same path twice.

### Minor

- "Is this an image" is defined two ways across the send boundary: MIME + preview in the composer (`attachment-preview.tsx:116`) vs `IMAGE_EXTS` by extension in the chip (`sent-attachment-chip.tsx:31`). A `.tiff` with an `image/*` MIME renders as a picture before send and as a file chip after. One `isPreviewableImage(name)` in `file-types.ts` would serve both.
- The inline URL is hand-assembled as `${getApiBaseUrl()}${fileApiPath}?inline=true` (`sent-attachment-chip.tsx:82`), the third copy of that recipe after `file-preview-tray-content.tsx:40` and `markdown-url-transform.ts:74`, and needs another `DIRECT_BASE_URL_CONSUMERS` entry.
- `SentAttachmentChip`'s clickable shell (folder detection, `open()`, `role=button` + Enter/Space, hover/focus classes, tile + truncated name + size line) is a near line-for-line copy of `FileDeliveryRow` and the composer's non-image branch, with `FileDownloadPill` as a fourth sibling. They already diverge on prefix stripping and hover color. A shared openable-file shell would keep the next chip tweak in one place.

### Nits

- `staleTime: 60_000` on an immutable upload path could be `Infinity`.
- The composer's remove button moved to an absolute corner; on the image chip it overlaps the picture at `right-0.5 top-0.5`, which reads fine on the gradients here but may need the `bg-background` ring on busy screenshots.

iddogino (approved):

iddogino:
## Demo: the image chip's upload state

Three 3 MB photos attached at once, with the uplink throttled to 180 KB/s so the states are visible. Watch the second chip: a progress bar fills along its bottom edge while the third sits behind it showing the waiting clock — the two things a picture-only chip previously had no way to say. At the end a forced 503 turns a fourth chip red with a retry glyph, and clicking it recovers.

https://github.com/user-attachments/assets/2fc093b7-6355-41b1-bfd2-241d8f4d2806

Recorded by `e2e/specs/upload-image-feedback.demo.spec.ts` under `playwright.demo.config.ts`.

PR #950, opened 2026-09-03
feat(renderer): file-type icons by category, and the file icon tile

westaskew:
## Summary

Stacked on https://github.com/SkillfulAgents/SuperAgent/pull/949 — merge that first.

- **`FileTypeIcon` draws lucide `file-*` icons chosen by a shared `fileCategory()`** in file-types: document, text, code, data, config, shell, spreadsheet, presentation, image, video/audio, archive, font, 3D. Plain `file` is the only fallback. The categories reuse the preview's existing image/video/audio/markdown/CSV sets.
- **One size token per text context** (`sm` 14 / `md` 16 / `lg` 20 / `xl` 24) and a `folder` variant, so every call site aligns the same way. Lucide's page fills about five-sixths of the box, which is why 20px exists (a ~16px page inside a 32px tile).
- **`FileIconTile`**: a 32px neutral tile with a 5% border and a category-colored icon stroke, for row and card layouts.
- **All ten call sites** move to the tokens; the hand-placed lucide Folder icons next to them go through the component's folder variant.
- **`react-file-icon` is removed.** Nothing renders it anymore.

## Test plan

- [x] Renderer suites (1316 passing across ui / file-preview / messages / agents / quick-dispatch); new tests for the icon (category → lucide class, fallback, sizes, folder) and the tile (stroke color, neutral cases)
- [ ] Visual check: pills, tabs, bookmarks, folder browser, attachment preview, quick-dispatch, light and dark


iddogino:
Reviewed as part of the stack (main ← #949 ← #950 ← #945 ← #946 ← #948). Screenshots are from the E2E mock build at this PR's head, with the base (#949) alongside for comparison.

**Delivered-file pills** (`react-file-icon` → lucide category icons):

Delivered-file pills before and after

**Composer chips**, same ten files:

Composer attachment chips before and after

**Sent-message pills** and **drawer tabs**:

Sent attachment pills before and after

Drawer tabs before and after

No blockers. The category classifier and the size tokens are a clear improvement over `react-file-icon`; the points below are about edge cases in the classifier and where the icon's own color gets in the way.

### Should fix

- **Extensionless names collide with category extensions.** `fileCategory` keys on the shared `getFileExtension`, which returns the whole lowercased name for dotless files (`src/renderer/components/file-preview/file-types.ts:75`). So a file named `key` gets the presentation icon, `lock`/`env`/`conf` the config icon, `obj`/`max` the 3D icon. Meanwhile `Dockerfile`/`Makefile`/`.gitignore` are in `TEXT_EXTS` (and preview as text) but sit in no category set, so they get the blank page. An explicit dotless-name map consulted before the extension lookup would make `README`/`LICENSE` a named case instead of a side effect of the helper.
- **The icon pins its own color.** `text-muted-foreground` is in the base class (`src/renderer/components/ui/file-type-icon.tsx:91`), so it no longer follows the parent: the folder pill's `hover:text-foreground` brightens the label but not the glyph (`file-download-pill.tsx:38`), and the failed-delivery state in `tool-renderers/deliver-file.tsx:82` renders a gray icon inside red text. Defaulting to `currentColor` and letting callers opt into muted would fix every site at once (the tab bar in #946 already has to pass `text-inherit` to escape it).
- **Two extension helpers in one component.** The local `getExtension` (`file-type-icon.tsx:62`) exists only to feed `data-file-ext`, which nothing outside the unit test reads, and it disagrees with `getFileExtension` on dotless names, so `README` renders `data-file-category="text"` with no `data-file-ext`. Dropping `getExtension`, `data-file-ext`, and its `describe` block leaves one source of truth.

### Minor

- Two raw lucide `Folder` icons remain at stroke width 2 next to the new stroke-1 file icons: `agent-home/home-bookmarks.tsx:105` (folder bookmarks look heavier than file bookmarks in the same list) and `file-preview-tray-content.tsx:49` (goes away in #946 anyway).
- `tool-renderers/deliver-file.tsx:29` mapped `size={20}` to `md` (16px); `lg` is the 20px token and is documented for exactly this row.
- `CATEGORY_SETS` is a linear scan over 14 sets on every render, and `folder-browser.tsx` renders up to 1,000 icons unvirtualized and re-renders on every filter keystroke. Inverting the sets into one `Map<ext, FileCategory>` at module load makes it a single lookup and also removes the silent ordering contract between sets. `fileCategory` also re-splits the filename right after `getExtension` did.
- `TEXT_EXTS` and the five new text-ish sets are two hand-maintained lists of mostly the same extensions (`file-types.ts:39`) and already disagree; deriving one from the other would stop them drifting.
- `data-tinted` on `FileIconTile` reports the prop, not whether a tint was applied, so `<FileIconTile folder />` advertises `data-tinted="true"` while rendering neutral (`file-icon-tile.tsx:46`).

### Nits

- At `sm` (14 px, stroke 1) the category glyphs are hard to tell apart in pills: `.csv`, `.md`, and `.mp4` read as near-identical outlines in the first screenshot. Worth checking whether `sm` should use stroke 1.5, since the base rendered colored badges there.
- `category ?? 'other'` at `file-type-icon.tsx:85` is unreachable, and `filename` is unused when `folder` is set; a discriminated prop would make that explicit.
- `FileIconTile` has no caller until #945, so within this PR it's dead code; the docstring already claims the delivered-file row uses it.
- Sizes are applied via inline `style`, so a caller's `className` can't override the box; `size-3.5`/`size-4`/… classes through `cn()` would compose like the rest of `components/ui`.
- Both new test files loop over cases with manual `unmount()`; `it.each` would name and isolate each case.

iddogino (approved):

PR #951, opened 2026-09-03
feat(composer): store secrets as chips instead of a sidecar (SUP-726)

JeremyBischoff:
The composer remembered secured pills in a sidecar list and paired them with painted text. A second chip kind cannot share that pairing, so this PR makes a chip an atom in the draft whose kind owns serialize, parse, paint, and send, with secrets as the first kind and mentions later.

Closes https://linear.app/datawizz/issue/SUP-726/generic-composer-chip-facility-extract-from-the-secured-secrets-pill

20 files, net down. Four new modules. The sidecar list and send-time string replace are gone.

A chip is `{ kind, payload }`. Position in the document is the instance. No id.

| Kind | Payload |
| --- | --- |
| secret | `key`, `envVar` |
| mention (later) | `userId`, `name` |

A kind owns two surfaces. Composer is required. Transcript fields are each optional.

| Field | Role | If omitted | Secret |
| --- | --- | --- | --- |
| `composer.raw` | draft, undo, clipboard | (required) | `[[secret:ENV\|Key]]` |
| `composer.parse` | lift a marker to a chip | (required) | `{ key, envVar }` |
| `composer.render` | paint in the composer | (required) | `[Key \| *********]` |
| `transcript.raw` | string sent to the agent | send `composer.raw` | `[Key saved to .env - ENV]` |
| `transcript.parse` | read a stored thread string | use `composer.parse` | omitted |
| `transcript.render` | React chip in the thread | plain text | omitted |

One marker grammar: `[[kind:referent|label]]`. The editor is the ProseMirror port. The hook writes the marker on save and rewrites on send. A new kind is a file plus one registry line. Grammar sits in its own file so the registry and a kind do not import each other.

Gone: the sidecar type, send-time string replace, the draft slot that stored the list, new-session carry of that list.

Intended deltas: draft string is the marker, paste of a marker is a live chip, clipboard text is the marker. No compat parse of old `[Key | *********]` drafts. They were in-memory only.

Mentions later add a kind file and one registry line. Do not edit the secret kind.

### DIFF
--- a/src/renderer/components/messages/chip-marker.ts
+++ b/src/renderer/components/messages/chip-marker.ts
@@ -1,13 +1,17 @@
 /** `[[kind:referent|label]]` — one grammar, kind name in the first capture. */
-export const CHIP_MARKER = /\[\[([a-z][a-z0-9_]*):([^|\]]+)\|([^\]]+)\]\]/g
+const CHIP_MARKER_BODY = String.raw`\[\[([a-z][a-z0-9_]*):([^|\]]+)\|([^\]]+)\]\]`
+
+export const CHIP_MARKER = new RegExp(CHIP_MARKER_BODY, 'g')
+export const CHIP_MARKER_ANCHORED = new RegExp(`^${CHIP_MARKER_BODY}$`)
+export const CHIP_MARKER_STICKY = new RegExp(CHIP_MARKER_BODY, 'y')
 
 export function formatChipMarker(kind: string, referent: string, label: string): string {
-  return `[[${kind}:${referent}|${encodeURIComponent(label)}]]`
+  return `[[${kind}:${referent}|${encodeURIComponent(label.toWellFormed())}]]`
 }
 
 export function parseChipMarker(raw: string): { kind: string; referent: string; label: string } | null {
-  const match = new RegExp(CHIP_MARKER.source).exec(raw)
-  if (!match || match[0] !== raw) return null
+  const match = CHIP_MARKER_ANCHORED.exec(raw)
+  if (!match) return null
   try {
     return { kind: match[1], referent: match[2], label: decodeURIComponent(match[3]) }
   } catch {
--- a/src/renderer/components/messages/chip-marker.ts
+++ b/src/renderer/components/messages/chip-marker.ts
@@ -1,0 +1,16 @@
+/** `[[kind:referent|label]]` — one grammar, kind name in the first capture. */
+export const CHIP_MARKER = /\[\[([a-z][a-z0-9_]*):([^|\]]+)\|([^\]]+)\]\]/g
+
+export function formatChipMarker(kind: string, referent: string, label: string): string {
+  return `[[${kind}:${referent}|${encodeURIComponent(label)}]]`
+}
+
+export function parseChipMarker(raw: string): { kind: string; referent: string; label: string } | null {
+  const match = new RegExp(CHIP_MARKER.source).exec(raw)
+  if (!match || match[0] !== raw) return null
+  try {
+    return { kind: match[1], referent: match[2], label: decodeURIComponent(match[3]) }
+  } catch {
+    return null
+  }
+}
--- a/src/renderer/components/messages/composer-chips.ts
+++ b/src/renderer/components/messages/composer-chips.ts
@@ -1,0 +1,39 @@
+import type { DOMOutputSpec } from 'prosemirror-model'
+import { CHIP_MARKER } from './chip-marker'
+import { secretChip } from './secret-chip'
+
+export { CHIP_MARKER, formatChipMarker, parseChipMarker } from './chip-marker'
+export { secretChip } from './secret-chip'
+
+export interface Chip<P extends Record<string, string> = Record<string, string>> {
+  kind: string
+  payload: P
+}
+
+export interface ChipSurface<P extends Record<string, string>, Paint> {
+  raw: (chip: Chip<P>) => string
+  parse: (raw: string) => Chip<P> | null
+  render: (chip: Chip<P>) => Paint
+}
+
+export interface ComposerChipKind<P extends Record<string, string> = Record<string, string>> {
+  kind: string
+  composer: ChipSurface<P, DOMOutputSpec>
+  transcript?: Partial<ChipSurface<P, unknown>>
+}
+
+export const COMPOSER_CHIP_KINDS: readonly ComposerChipKind<Record<string, string>>[] = [secretChip]
+
+export function rewriteChipsForSend(text: string): string {
+  const byKind = new Map(COMPOSER_CHIP_KINDS.map((kind) => [kind.kind, kind]))
+  return text.replace(
+    CHIP_MARKER,
+    (raw, kindName: string) => {
+      const kind = byKind.get(kindName)
+      if (!kind) return raw
+      const chip = kind.composer.parse(raw)
+      if (!chip) return raw
+      return kind.transcript?.raw?.(chip) ?? raw
+    }
+  )
+}
--- a/src/renderer/components/messages/composer-chips.ts
+++ b/src/renderer/components/messages/composer-chips.ts
@@ -2,7 +2,13 @@ import type { DOMOutputSpec } from 'prosemirror-model'
 import { CHIP_MARKER } from './chip-marker'
 import { secretChip } from './secret-chip'
 
-export { CHIP_MARKER, formatChipMarker, parseChipMarker } from './chip-marker'
+export {
+  CHIP_MARKER,
+  CHIP_MARKER_ANCHORED,
+  CHIP_MARKER_STICKY,
+  formatChipMarker,
+  parseChipMarker,
+} from './chip-marker'
 export { secretChip } from './secret-chip'
 
 export interface Chip<P extends Record<string, string> = Record<string, string>> {
@@ -24,15 +30,25 @@ export interface ComposerChipKind<P extends Record<string, string> = Record<stri
 
 export const COMPOSER_CHIP_KINDS: readonly ComposerChipKind<Record<string, string>>[] = [secretChip]
 
-export function rewriteChipsForSend(text: string): string {
-  const byKind = new Map(COMPOSER_CHIP_KINDS.map((kind) => [kind.kind, kind]))
+const chipKindsByName = new Map(COMPOSER_CHIP_KINDS.map((kind) => [kind.kind, kind]))
+
+export function getChipKind(name: string) {
+  return chipKindsByName.get(name)
+}
+
+export function isBackedSecretChip(chip: Chip, knownSecretEnvVars: ReadonlySet<string>): boolean {
+  return chip.kind !== 'secret' || knownSecretEnvVars.has(chip.payload.envVar)
+}
+
+export function rewriteChipsForSend(text: string, knownSecretEnvVars: Iterable<string>): string {
+  const known = knownSecretEnvVars instanceof Set ? knownSecretEnvVars : new Set(knownSecretEnvVars)
   return text.replace(
     CHIP_MARKER,
     (raw, kindName: string) => {
-      const kind = byKind.get(kindName)
+      const kind = getChipKind(kindName)
       if (!kind) return raw
       const chip = kind.composer.parse(raw)
-      if (!chip) return raw
+      if (!chip || !isBackedSecretChip(chip, known)) return raw
       return kind.transcript?.raw?.(chip) ?? raw
     }
   )
--- a/src/renderer/components/messages/composer-chips.ts
+++ b/src/renderer/components/messages/composer-chips.ts
@@ -1,7 +1,6 @@
 import type { DOMOutputSpec } from 'prosemirror-model'
-import { isKnownSecret } from '@renderer/lib/secret-detection'
-import { CHIP_MARKER } from './chip-marker'
-import { secretChip } from './secret-chip'
+import { CHIP_MARKER } from '@renderer/lib/chip-marker'
+import { secretChip } from '@renderer/lib/secret-chip'
 
 export {
   CHIP_MARKER,
@@ -9,8 +8,8 @@ export {
   CHIP_MARKER_STICKY,
   formatChipMarker,
   parseChipMarker,
-} from './chip-marker'
-export { secretChip } from './secret-chip'
+} from '@renderer/lib/chip-marker'
+export { secretChip } from '@renderer/lib/secret-chip'
 
 export interface Chip<P extends Record<string, string> = Record<string, string>> {
   kind: string
@@ -27,6 +26,7 @@ export interface ComposerChipKind<P extends Record<string, string> = Record<stri
   kind: string
   composer: ChipSurface<P, DOMOutputSpec>
   transcript?: Partial<ChipSurface<P, unknown>>
+  isBacked?: (chip: Chip<P>, knownSecrets: ReadonlyMap<string, string>) => boolean
 }
 
 export const COMPOSER_CHIP_KINDS: readonly ComposerChipKind<Record<string, string>>[] = [secretChip]
@@ -37,8 +37,8 @@ export function getChipKind(name: string) {
   return chipKindsByName.get(name)
 }
 
-export function isBackedSecretChip(chip: Chip, knownSecrets: ReadonlyMap<string, string>): boolean {
-  return chip.kind !== 'secret' || isKnownSecret(chip.payload, knownSecrets)
+export function isBackedChip(chip: Chip, knownSecrets: ReadonlyMap<string, string>): boolean {
+  return getChipKind(chip.kind)?.isBacked?.(chip, knownSecrets) ?? true
 }
 
 export function rewriteChipsForSend(text: string, knownSecrets: ReadonlyMap<string, string>): string {
@@ -48,7 +48,7 @@ export function rewriteChipsForSend(text: string, knownSecrets: ReadonlyMap<stri
       const kind = getChipKind(kindName)
       if (!kind) return raw
       const chip = kind.composer.parse(raw)
-      if (!chip || !isBackedSecretChip(chip, knownSecrets)) return raw
+      if (!chip || !isBackedChip(chip, knownSecrets)) return raw
       return kind.transcript?.raw?.(chip) ?? raw
     }
   )
--- a/src/renderer/components/messages/composer-chips.ts
+++ b/src/renderer/components/messages/composer-chips.ts
@@ -1,4 +1,5 @@
 import type { DOMOutputSpec } from 'prosemirror-model'
+import { isKnownSecret } from '@renderer/lib/secret-detection'
 import { CHIP_MARKER } from './chip-marker'
 import { secretChip } from './secret-chip'
 
@@ -36,19 +37,18 @@ export function getChipKind(name: string) {
   return chipKindsByName.get(name)
 }
 
-export function isBackedSecretChip(chip: Chip, knownSecretEnvVars: ReadonlySet<string>): boolean {
-  return chip.kind !== 'secret' || knownSecretEnvVars.has(chip.payload.envVar)
+export function isBackedSecretChip(chip: Chip, knownSecrets: ReadonlyMap<string, string>): boolean {
+  return chip.kind !== 'secret' || isKnownSecret(chip.payload, knownSecrets)
 }
 
-export function rewriteChipsForSend(text: string, knownSecretEnvVars: Iterable<string>): string {
-  const known = knownSecretEnvVars instanceof Set ? knownSecretEnvVars : new Set(knownSecretEnvVars)
+export function rewriteChipsForSend(text: string, knownSecrets: ReadonlyMap<string, string>): string {
   return text.replace(
     CHIP_MARKER,
     (raw, kindName: string) => {
       const kind = getChipKind(kindName)
       if (!kind) return raw
       const chip = kind.composer.parse(raw)
-      if (!chip || !isBackedSecretChip(chip, known)) return raw
+      if (!chip || !isBackedSecretChip(chip, knownSecrets)) return raw
       return kind.transcript?.raw?.(chip) ?? raw
     }
   )
--- a/src/renderer/components/messages/markdown-composer-editor.test.tsx
+++ b/src/renderer/components/messages/markdown-composer-editor.test.tsx
@@ -10,7 +10,13 @@ import {
 import { formatChipMarker } from './chip-marker'
 import { findPotentialSecrets } from '@renderer/lib/secret-detection'
 
-function ControlledEditor({ initialValue = '' }: { initialValue?: string }) {
+function ControlledEditor({
+  initialValue = '',
+  knownSecretEnvVars = [],
+}: {
+  initialValue?: string
+  knownSecretEnvVars?: readonly string[]
+}) {
   const [value, setValue] = useState(initialValue)
   return (
     <>
@@ -19,6 +25,7 @@ function ControlledEditor({ initialValue = '' }: { initialValue?: string }) {
         onChange={setValue}
         placeholder="Write a message"
         dataTestId="markdown-editor"
+        knownSecretEnvVars={knownSecretEnvVars}
       />
       <output data-testid="markdown-value">{value}</output>
     </>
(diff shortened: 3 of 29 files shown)


src/renderer/components/messages/markdown-composer-editor.tsx
```diff
@@ -67,15 +107,48 @@ const markdownTokenizer = new MarkdownIt('commonmark', {
   linkify: true,
 }).enable('strikethrough')
 
+const chipKindsByName = new Map(COMPOSER_CHIP_KINDS.map((kind) => [kind.kind, kind]))
+markdownTokenizer.inline.ruler.before('escape', 'composer_chip', (state, silent) => {
```
iddogino, line 132: **Chips inside inline code / fenced blocks never render as a pill.**

markdown-it's `backticks` and `fence` rules never run inline rules, so `composer_chip` doesn't fire when the marker sits inside a code context. `securePotentialSecret` splices the marker by string offset regardless of context, so a user who pastes an `.env` snippet in a ``` fence (or wraps a key in backticks) and clicks "Send securely" sees the literal `[[secret:API_KEY|My%20Key]]` in the editor (no `data-testid=secured-secret`, char-by-char deletion). Meanwhile `rewriteChipsForSend` is a bare regex over the whole string and still emits `[Key saved to .env - API_KEY]` on submit, so what the user sees and what the agent receives diverge.

Reproduced in jsdom: 0 pills for both code contexts, send text still rewritten. The previous decoration-based `findSecretMatches` walked every text node including `code_block`, so this is a regression.

src/renderer/components/messages/composer-chips.ts
```diff
@@ -0,0 +1,39 @@
...
+export function rewriteChipsForSend(text: string): string {
+  const byKind = new Map(COMPOSER_CHIP_KINDS.map((kind) => [kind.kind, kind]))
+  return text.replace(
+    CHIP_MARKER,
+    (raw, kindName: string) => {
+      const kind = byKind.get(kindName)
+      if (!kind) return raw
+      const chip = kind.composer.parse(raw)
+      if (!chip) return raw
+      return kind.transcript?.raw?.(chip) ?? raw
```
iddogino, line 52: **A chip's referent is never validated against the agent's saved secrets.**

Any `[[secret:ENV|label]]` that lands in the draft (pasted from a transcript, a web page, or copied out of another agent's composer, since `leafText` puts the raw marker on the clipboard) paints the amber "secured" pill and on send tells the agent `[Key saved to .env - ENV]` when no such key exists. The agent then acts on a credential that isn't in `.env`.

The old `replaceSecuredSecrets` only rewrote entries in the composer's own sidecar, so arbitrary bracketed text was untouched. The new e2e paste test codifies the unverified path (it never saves a key). `useAgentSecrets(agentSlug)` already exists in `use-secrets.ts` and could gate the lift and/or the rewrite.

src/renderer/components/messages/markdown-composer-editor.tsx
```diff
@@ -41,21 +47,55 @@ export interface MarkdownComposerEditorProps {
...
-  onRemoveSecuredSecrets?: (secrets: SecuredSecret[]) => void
   onEditorElement?: (element: HTMLDivElement | null) => void
 }
 
 const CARET_SENTINEL = '\u2063'
 
-const markdownSchema = new Schema({
-  nodes: commonmarkSchema.spec.nodes.addBefore('hard_break', 'soft_break', {
+function chipFromNode(kind: ComposerChipKind<Record<string, string>>, node: ProseMirrorNode): Chip {
+  return { kind: kind.kind, payload: JSON.parse(String(node.attrs.payload || '{}')) }
```
iddogino: **Unguarded `JSON.parse` on a payload that `parseDOM` copies verbatim from pasted/dropped HTML.**

`handlePaste` returns `false` when `text/plain` is empty and there's no `handleDrop`/`transformPastedHTML`, so an HTML-only clipboard or a drop of `<span data-chip-kind="secret" data-payload="nope">x</span>` goes through ProseMirror's default DOM parser into a chip node. The next `toDOM` hits `JSON.parse('nope')` and throws mid-dispatch, leaving the view desynced (`leafText` and the serializer throw on the same node). `data-payload="[]"` parses but renders `[undefined | *********]`, serializes `[[secret:undefined|undefined]]`, and sends `[Key saved to .env - undefined]`.

Suggested simplification: store the raw marker as the node's single attr. Then `leafText`/serializer become `node.attrs.raw`, `toDOM` is `render(parse(raw))`, `parseDOM.getAttrs` returns `false` when `parse` yields null, and `chipFromNode` plus the five `JSON.parse`/`stringify` hops disappear.

src/renderer/components/messages/markdown-composer-editor.tsx
```diff
@@ -41,21 +47,55 @@ export interface MarkdownComposerEditorProps {
...
+    atom: true,
+    selectable: false,
+    attrs: { payload: { default: '{}' } },
+    parseDOM: [{
+      tag: `span[data-chip-kind="${kind.kind}"]`,
+      getAttrs: (dom) => ({ payload: (dom as HTMLElement).getAttribute('data-payload') ?? '{}' }),
+    }],
+    toDOM: (node) => {
+      const spec = kind.composer.render(chipFromNode(kind, node))
+      if (!Array.isArray(spec) || spec.length < 2 || typeof spec[1] !== 'object' || spec[1] === null) return spec
```
iddogino: **`typeof [] === 'object'` lets a child spec in slot 1 through the guard.**

`DOMOutputSpec` permits `['span', ['b', 'text']]` with no attrs object. prosemirror-model's `renderSpec` excludes arrays from attrs explicitly, but after this spread the merged `{0:'b', 1:'text', 'data-chip-kind': …}` is a plain object and *is* treated as attrs, so `setAttribute('0','b')` throws `InvalidCharacterError` (reproduced in jsdom). Latent today since `secretChip` returns an attrs object, but the registry is the stated extension point for new kinds.

Add `|| Array.isArray(spec[1])` to the early return, or have the wrapper own the outer span and narrow `render` to inner content.

src/renderer/components/messages/chip-marker.ts
```diff
@@ -0,0 +1,16 @@
+/** `[[kind:referent|label]]` — one grammar, kind name in the first capture. */
+export const CHIP_MARKER = /\[\[([a-z][a-z0-9_]*):([^|\]]+)\|([^\]]+)\]\]/g
```
iddogino: **Referent and label classes accept whitespace and newlines.**

`[^|\]]+` / `[^\]]+` mean a marker can span soft breaks, and a raw credential placed in the label is exempted from secret detection:

- Paste `[[secret:FOO\nBAR|x]]` → lifted to a chip with referent `FOO\nBAR`, send emits `[Key saved to .env - FOO\nBAR]`.
- Paste `[[secret:A|sk-proj-<40-char credential>]]` → `findPotentialSecrets` protects the whole match, the "Is this a key?" prompt never fires, the pill shows the decoded credential as its label, and the credential is dropped from the send without being saved.

The old `[^\]\n]` pattern excluded newlines. Legit referents come from `keyToEnvVar` (`[A-Z0-9_]`) and labels are URI-encoded, so tightening to `[A-Za-z0-9_]+` / `[^\]\s]+` (or rejecting in `secretChip.composer.parse`) costs nothing.

src/renderer/components/messages/markdown-composer-editor.test.tsx
```diff
@@ -344,53 +319,64 @@ describe('MarkdownComposerEditor', () => {
...
-    )
+  it('lifts a secret marker to a pill and serializes the same marker', () => {
+    const marker = formatChipMarker('secret', 'GITHUB_TOKEN', 'GitHub Token')
+    render(<ControlledEditor initialValue={`Use ${marker}`} />)
+
+    expect(screen.getByTestId('secured-secret')).toHaveTextContent('[GitHub Token | *********]')
+    expect(screen.getByTestId('markdown-value').textContent).toBe(`Use ${marker}`)
+  })
+
+  it('leaves unknown and incomplete markers as text', () => {
```
iddogino: **These two tests never trigger a transaction, so the `markdown-value` assertions are tautological.**

`EditorState.create` + `new EditorView` never call `dispatchTransaction` (probe: `onChange` calls after mount = 0), so `markdown-value` is the untouched initial `useState`. Typing one character after `See [[foo:bar|baz]] [[secret:GITHUB_TOKEN` serializes to `See \[\[foo:bar|baz\]\] \[\[secret:GITHUB_TOKEN x` because prosemirror-markdown escapes `[`/`]`, so this test's claim is false under real editing. The chip serializer is only exercised incidentally by the paste/Backspace tests.

Add `await user.type(editor, ' ')` before asserting, and either assert the escaped form for unknown markers or decide that unknown markers shouldn't be escaped.

src/renderer/components/messages/chip-marker.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export function formatChipMarker(kind: string, referent: string, label: string): string {
+  return `[[${kind}:${referent}|${encodeURIComponent(label)}]]`
```
iddogino: **`encodeURIComponent` throws `URIError` on a lone surrogate, inside a React state updater, after the secret is already persisted.**

A key name containing an unpaired surrogate (malformed clipboard/IME payload) passes the dialog (`key.trim()` non-empty), the API schema is a bare `z.string()`, and `JSON.stringify`/`parse` round-trips `\ud83d` intact. So `saved.key` reaches `securePotentialSecret` → `secretChip.composer.raw` → `encodeURIComponent` → throws inside `setMessage(current => …)`. React rethrows during render, escaping `handleSubmit`'s try/catch and landing in the nearest error boundary with the secret saved but the draft unchanged. The decode side is guarded; wrap the encode side too, or use `label.toWellFormed()`.

src/renderer/lib/secret-detection.ts
```diff
@@ -67,9 +62,11 @@ function looksLikeSecret(value: string, precedingText: string): boolean {
  */
 export function findPotentialSecrets(text: string): PotentialSecret[] {
   const protectedRanges: Array<{ start: number; end: number }> = []
-  for (const match of text.matchAll(SECURED_DISPLAY_PATTERN)) {
-    const start = match.index
-    protectedRanges.push({ start, end: start + match[0].length })
+  for (const pattern of [new RegExp(CHIP_MARKER.source, 'g'), SECURED_DISPLAY_PATTERN]) {
```
iddogino: **`SECURED_DISPLAY_PATTERN` is dead after this PR but still exempts typed text from detection.**

Nothing writes `[Key | *********]` into the draft anymore: `secretDisplayText` is used solely for the chip's DOM label, every string path emits the `[[secret:…]]` marker, and the drafts store is a `useRef(new Map())` so no persisted legacy drafts exist. Yet a real key typed as `[sk-… | ****]` is currently exempted from detection for no reason, which also contradicts the PR's stated "no compat" stance.

Delete the pattern and collapse this loop to `text.matchAll(CHIP_MARKER)`. `matchAll` clones the regex, so the `new RegExp(…, 'g')` copy is redundant as well.

src/renderer/components/messages/chip-marker.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export function parseChipMarker(raw: string): { kind: string; referent: string; label: string } | null {
+  const match = new RegExp(CHIP_MARKER.source).exec(raw)
```
iddogino: **The grammar is rebuilt from `.source` at three sites, and the kind registry is built twice.**

- `chip-marker.ts:9` (no flags + hand-rolled `match[0] !== raw` anchor)
- `markdown-composer-editor.tsx:113` (`'y'` on a `state.src.slice(state.pos)` copy at every `[`)
- `secret-detection.ts:65` (`'g'` per keystroke)

`chipKindsByName` (editor) and `byKind` (`rewriteChipsForSend`) are parallel registries that must both be updated per new kind.

Suggest: export an anchored `^…$` variant for parse and a sticky one used with `lastIndex = state.pos` on `state.src` (no slice), pass the `/g` constant directly to `matchAll`/`replace`, and export one `getChipKind(name)` from `composer-chips.ts`.

src/renderer/components/messages/markdown-composer-editor.tsx
```diff
@@ -432,10 +486,29 @@ function buildInputRules() {
   })
 }
 
+const deleteAtomBefore: Command = (state, dispatch) => {
```
iddogino: **`deleteAtomBefore`/`deleteAtomAfter` duplicate prosemirror-view's built-in `captureKeyDown` → `stopNativeHorizontalDelete`.**

In a real browser the library already deletes a non-text `nodeBefore`/`nodeAfter` on Backspace/Delete; these ~20 lines exist only because `@testing-library/user-event` never sets `keyCode`, so the library path (which dispatches on `keyCode` 8/46) never fires in jsdom. The semantics also differ slightly: the library deletes soft/hard breaks too and skips Shift-Delete, the custom one targets only `spec.atom`.

Either keep them with a comment stating exactly that, or drop them and have the two new tests dispatch `fireEvent.keyDown(editor, { key: 'Backspace', keyCode: 8 })` so they exercise the production path.

iddogino (changes requested):
Ran a multi-angle review of this PR. Typecheck is clean and all 163 tests in the changed files pass, but there are ten findings below, six confirmed with reproductions. The two that matter most are the first two inline comments: both let the agent receive a `[Key saved to .env - X]` message that isn't true (a chip that didn't render inside code, and an unverified pasted marker).

Two cleanup-tier nits not commented inline:
- `secret-detection.ts` now has the codebase's first runtime `lib` → `components` import. Moving `chip-marker`/`secret-chip` under `src/renderer/lib` is cycle-free.
- `composer-chips.ts` declares `transcript.parse`/`render` with no consumer; the PR body says the raw-marker fallback is intended, so design nit only.

JeremyBischoff:
On the two body nits: left the files in components, since moving both into lib would put the pill UI in the utility layer. Left the unused transcript slots. The raw-marker fallback is still the intended path.

src/renderer/components/messages/markdown-composer-editor.tsx
JeremyBischoff: Done in `04e5c3e56` - leftover markers now lift to pills inside backticks and fences.

src/renderer/components/messages/composer-chips.ts
JeremyBischoff: Done in `04e5c3e56` - paint and send only treat a marker as a chip when this agent has that key.

src/renderer/components/messages/markdown-composer-editor.tsx
JeremyBischoff: Done in `04e5c3e56` - the chip attr is the raw marker, and HTML parse returns false when that marker is unreadable or unbacked.

src/renderer/components/messages/markdown-composer-editor.tsx
JeremyBischoff: Done in `04e5c3e56` - a child spec in slot 1 no longer gets spread as attrs.

src/renderer/components/messages/chip-marker.ts
JeremyBischoff: Done in `04e5c3e56` - secret parse refuses a newline name and a label that is a credential shape.

src/renderer/components/messages/markdown-composer-editor.test.tsx
JeremyBischoff: Done in `04e5c3e56` - those tests type a character first, then assert escaped unknown markers and unescaped known chips.

src/renderer/components/messages/chip-marker.ts
JeremyBischoff: Done in `04e5c3e56` - the label is made well-formed before encode.

src/renderer/components/messages/chip-marker.ts
JeremyBischoff: Done in `04e5c3e56` - one shared marker grammar and one kind lookup.

src/renderer/lib/secret-detection.ts
JeremyBischoff: Done in `04e5c3e56` - the old display skip is gone.

src/renderer/components/messages/markdown-composer-editor.tsx
JeremyBischoff: Done in `04e5c3e56` - the extra handlers are gone, and the tests send keyCode 8/46.

src/renderer/lib/secret-detection.ts
```diff
@@ -61,37 +81,34 @@ function looksLikeSecret(value: string, precedingText: string): boolean {
   return value.length >= 32 && entropy >= 3.5
 }
 
+export function parseSecretMarker(raw: string): { envVar: string; key: string } | null {
+  const parts = parseChipMarker(raw)
+  if (!parts || parts.kind !== 'secret') return null
+  if (!/^[A-Z0-9_]+$/.test(parts.referent)) return null
+  if (/[\r\n]/.test(parts.label) || secretTokensIn(parts.label, isCredentialShape).length > 0) return null
```
iddogino: **Regression: legitimate key names are now rejected.**

`secretTokensIn(label, isCredentialShape)` refuses any label containing a 20+ char token that starts with a `KNOWN_SECRET_PREFIX` (`sk-`, `pk_`, `npm_`, `github_pat_`, `AKIA`…). Ordinary user-chosen names hit this: `sk-openai-production-key`, `github_pat_for_deploys_2024`, `pk_live_stripe_key_prod`.

Probe: secure a token and name it `sk-openai-production-key` → secret is saved, `securePotentialSecret` inserts `[[secret:SK_OPENAI_PRODUCTION_KEY|sk-openai-production-key]]`, `parseSecretMarker` → `null`, no pill, `findPotentialSecrets` now flags the key *name* (re-opens "Is this a key?" on it), and `rewriteChipsForSend` leaves the marker untouched so the agent receives the literal marker (escaped `\[\[…\]\]` after any keystroke) instead of `[Key saved to .env - …]`.

Fix: only reject labels that `looksLikeSecret` would flag *and* that don't equal the saved key name, or validate the label at save time rather than parse time.

src/renderer/components/messages/markdown-composer-editor.tsx
```diff
@@ -67,15 +122,48 @@ const markdownTokenizer = new MarkdownIt('commonmark', {
   linkify: true,
 }).enable('strikethrough')
 
+markdownTokenizer.inline.ruler.before('escape', 'composer_chip', (state, silent) => {
```
iddogino, line 132: **Chip inside a heading drops the whole heading.**

The rule emits a chip token regardless of the parent block's content expression, but `withChipNodes` only widened `code_block`. A chip inside a heading makes prosemirror-markdown's `createAndFill` return `null` and the entire heading is silently dropped.

Probe: draft `## Deploy [[secret:API_KEY|My%20Key]]` with `API_KEY` known → editor renders `<p><br></p>`; after one keystroke `markdown-value` is `x`. Both the heading text and the secured reference are lost. Control with `known=[]` keeps `<h2>Deploy [[…]]</h2>`.

Either widen every inline-content textblock (heading etc.) in `withChipNodes`, or have the rule check the container and fall back to text.

**Also: chip inside a link label destroys the link.** Because the rule consumes `[[…]]` as one token and runs in silent mode inside `parseLinkLabel`, markdown-it treats it as a nested link and rejects the enclosing one. Draft `[see [[secret:API_KEY|My%20Key]]](https://x.com)` with `API_KEY` known renders with no `<a>`; after one keystroke the value is `x\[see [[secret:API_KEY|My%20Key]]\](https://x.com)`. With `known=[]` (and on main) the link is preserved. Guard the rule when `state.linkLevel > 0` / silent inside a label, or rely on `leftoverChipReplacements` there.

src/renderer/components/messages/markdown-composer-editor.tsx
```diff
@@ -645,18 +745,18 @@ export function MarkdownComposerEditor({
...
     if (!hostRef.current) return
 
-    const secretPlugin = buildSecretDecorationsPlugin(() => ({
-      potential: latestRef.current.potentialSecrets,
-      secured: latestRef.current.securedSecrets,
-    }))
-    const initialDoc = parseComposerMarkdown(lastMarkdownRef.current)
+    const secretPlugin = buildSecretDecorationsPlugin(() => latestRef.current.potentialSecrets)
+    lastKnownSecretEnvVarsRef.current = [...latestRef.current.knownSecretEnvVars].sort().join('\0')
+    const initialDoc = parseComposerMarkdown(
```
iddogino: **Escaped drafts are re-lifted on mount without a transaction, so the UI shows a pill but the escaped literal is sent.**

Whenever the user types while `useAgentSecrets` is pending or errored (first mount, react-query gc, 500 from `GET /secrets`, quick-dispatch with `agentSlug ''` until the agent list resolves), `onChange` stores the draft as `\[\[secret:…\]\]`. On the next mount `liftLeftoverChipMarkers` paints a pill from it, but no transaction is dispatched so `onChange` never fires; `message` stays escaped and `rewriteChipsForSend` does not match `\[\[`.

Probe: mount before secrets load with draft `Use [[secret:API_KEY|My%20Key]]` → type one char → stored as `xUse \[\[secret:API_KEY|My%20Key\]\]` → switch session → remount with `API_KEY` known: pill=true, `onChange` calls=0, value still escaped, `rewriteChipsForSend(value, ['API_KEY'])` returns it unchanged → clicking Send without typing sends the literal.

Fix: dispatch the lift (or call `onChange`) when the initial parse changed the doc, or paint unconditionally and gate only the send.

src/renderer/components/messages/markdown-composer-editor.tsx
```diff
@@ -811,16 +880,34 @@ export function MarkdownComposerEditor({
...
   useLayoutEffect(() => {
     const view = viewRef.current
-    if (!view || value === lastMarkdownRef.current) return
-    const nextDoc = parseComposerMarkdown(value)
+    if (!view) return
+    const knownKey = [...knownSecretEnvVars].sort().join('\0')
+    const knownChanged = knownKey !== lastKnownSecretEnvVarsRef.current
+    const valueUnchanged = value === lastMarkdownRef.current
+    if (valueUnchanged && !knownChanged) return
+    if (knownChanged && valueUnchanged) {
```
iddogino: **Shrinking `knownSecretEnvVars` never demotes an existing pill.**

This branch only calls `applyChipLifts` (adds lifts); `toDOM` never consults the known set, so the amber pill stays painted while `rewriteChipsForSend` now refuses to rewrite it.

Probe: quick-dispatch shares `draftKey 'quick-dispatch'` across agents. Paste/secure a marker for agent A (pill), switch the picker to agent B lacking `A_KEY` → pill still shows `[A Key | *********]`, `rewriteChipsForSend(value, B's set)` returns `Use [[secret:A_KEY|A%20Key]]` → agent B receives the raw marker. Same after deleting the key while the composer stays mounted. Reproduced in jsdom (`known ['A_KEY'] → []`).

src/renderer/lib/secret-detection.ts
```diff
@@ -38,6 +32,32 @@ function overlaps(start: number, end: number, ranges: Array<{ start: number; end
...
+      value = assignment[2]
+    }
+    if (test(value, text.slice(Math.max(0, start - 3), start))) {
+      tokens.push({ value, start })
+    }
+  }
+  return tokens
+}
+
+function isCredentialShape(value: string): boolean {
```
iddogino: **Entropy check is skipped for labels, so a generic high-entropy credential still passes (prior finding 5 only partially fixed).**

`isCredentialShape` only checks `KNOWN_SECRET_PREFIX` / `HEX_SECRET`, not the entropy rule `looksLikeSecret` applies.

Probe: paste `[[secret:GITHUB_TOKEN|Ab3dEf6hIj9kLm2nOp5qRs8tUv1wXy4zQwErTyUi]]` with `GITHUB_TOKEN` known → `findPotentialSecrets(bare)` = 1 but inside the marker = 0, `parseSecretMarker` accepts it, a pill shows the credential as its label, and send emits `[Key saved to .env - GITHUB_TOKEN]`. The credential is dropped without ever being saved and no "Is this a key?" prompt fires.

src/renderer/components/messages/markdown-composer-editor.tsx
```diff
@@ -41,21 +48,69 @@ export interface MarkdownComposerEditorProps {
...
+function chipNodeSpecs(): Record<string, NodeSpec> {
+  return Object.fromEntries(COMPOSER_CHIP_KINDS.map((kind) => [kind.kind, {
+    inline: true,
+    group: 'inline',
+    atom: true,
+    selectable: false,
+    attrs: { raw: { default: '' } },
+    parseDOM: [{
+      tag: `span[data-chip-kind="${kind.kind}"]`,
+      getAttrs: (dom) => {
```
iddogino, line 70: **Returning `false` from `getAttrs` does not drop the span.**

ProseMirror's `DOMParser` descends into it, so an HTML-only paste/drop of an unbacked chip inserts the masked display text `[GitHub Token | *********]` (neither chip nor marker) and the secret reference is silently lost.

Probe: drag a pill from one composer view into another editor whose agent lacks the key (default `handleDrop` → `parseFromClipboard` prefers `text/html` when present), or drop HTML from another app preserving `data-*` attrs: `editor.textContent` = `[GitHub Token | *********]`, `markdown-value` = `\[GitHub Token | \*\*\*…\]`, which never re-lifts even after the key is added. The new test only asserts `not.toContain('[undefined')`.

Demote to `schema.text(raw)` instead (e.g. in `transformPasted`) rather than returning `false` here.

src/renderer/hooks/use-message-composer.ts
```diff
@@ -45,13 +43,15 @@ export function useMessageComposer(options: UseMessageComposerOptions) {
...
 
   const [draft, setDraft] = useDraft<string>(draftKey)
-  const securedDraftKey = draftKey ? `${draftKey}:secured-secrets` : undefined
-  const [draftSecuredSecrets, setDraftSecuredSecrets] = useDraft<SecuredSecret[]>(securedDraftKey)
   const [message, setMessage] = useState(draft ?? '')
   const [dismissedSecretValues, setDismissedSecretValues] = useState<Set<string>>(() => new Set())
-  const [securedSecrets, setSecuredSecrets] = useState<SecuredSecret[]>(
-    () => options.initialSecuredSecrets ?? draftSecuredSecrets ?? []
-  )
+  const [justSecuredEnvVars, setJustSecuredEnvVars] = useState<string[]>([])
```
iddogino: **`justSecuredEnvVars` is append-only and never reconciled with `savedSecrets`.**

A key secured in this composer stays "known" for the composer's mounted lifetime even after it's deleted, re-opening the unverified-referent send from the first review.

Probe: secure X in a long-lived composer (agent home / session), delete X via another window, the API, or agent-side, refetch lands without X → pill still renders and `handleSubmit` emits `[Key saved to .env - X]` for a key not in `.env`.

The bridge only exists because `useCreateSecret.onSuccess` calls `invalidateQueries` without awaiting or `setQueryData`. Doing `setQueryData(['agent-secrets', slug], prev => [...prev, saved])` (or returning the invalidate promise) removes the need for the local list entirely.

src/renderer/components/messages/markdown-composer-editor.tsx
```diff
@@ -785,11 +850,15 @@ export function MarkdownComposerEditor({
         event.preventDefault()
         editorView.dispatch(
           editorView.state.tr
-            .replaceSelection(markdownClipboardSlice(text))
+            .replaceSelection(markdownClipboardSlice(text, new Set(latestRef.current.knownSecretEnvVars)))
             .scrollIntoView()
         )
         return true
       },
+      transformPastedHTML: (html) => {
```
iddogino: **`knownSecretEnvVarsForParse` is written here and never restored.**

Every other site uses `try/finally`; this one leaves the module global set, so any `DOMParser` path not preceded by `parseComposerMarkdown` or this editor's own `transformPastedHTML` validates chips against whichever editor last pasted HTML (or the initial empty set).

Scenario: session composer + create-agent-form (default empty set) mounted together; after an HTML paste in one, a chip span in the other that loses its `pmViewDesc` (native contenteditable undo via the Electron Edit menu, which bypasses the bound Mod-z and clones DOM) is re-read through `getAttrs` with a foreign/stale key set → a backed chip collapses to masked text or an unbacked one is accepted.

Main-line paths are covered by the view-desc short-circuit, so this is a design hazard rather than a demonstrated crash. Reset the global in `transformPasted`, or thread the set through markdown-it `env` / a per-view parser instead of a module global.

src/renderer/components/messages/markdown-composer-editor.tsx
```diff
@@ -89,53 +177,103 @@ const markdownSerializer = new MarkdownSerializer(
   }
 )
 
-function serializeLiteralMarkdown(value: string): string {
-  const paragraph = markdownSchema.nodes.paragraph.create(
-    null,
-    value ? markdownSchema.text(value) : undefined
-  )
-  return markdownSerializer.serialize(markdownSchema.nodes.doc.create(null, paragraph))
+function flattenCodeChipsForSerialize(node: ProseMirrorNode): ProseMirrorNode {
```
iddogino, line 242: **Runs on every `dispatchTransaction` and rebuilds the whole doc even when no code-marked chip exists.**

`node.copy(Fragment.from(children))` recursively allocates a full node/Fragment copy per keystroke on top of the serializer's own walk, for zero benefit in the common no-chip case. A large pasted draft (thousands of fence lines) pays this on every keystroke.

Also the hand-written adjacent-text merge duplicates what `Fragment.fromArray` already does (prosemirror-model joins `sameMarkup` text nodes).

Cheaper: a `doc.descendants` pre-scan that returns the doc untouched when no `isAtom`-with-`code`-mark node exists, and drop the `prev.sameMarkup(next)` merge branch since `Fragment.from(children)` performs it.

iddogino (changes requested):
Re-review of 04e5c3e5 against the first review. Typecheck is clean and all 180 tests in the changed files pass.

**Status of the 10 prior findings**

| # | Finding | Status |
|---|---|---|
| 1 | Chips in code not painted but still rewritten | Fixed, but the inline rule now regresses headings and link labels (see new comment on the rule) |
| 2 | Referent never validated | Partial: send and paint are gated, but `justSecuredEnvVars` is never pruned, a shrinking key set never un-lifts, and escaped drafts send the literal |
| 3 | Unguarded `JSON.parse` | Fixed |
| 4 | `toDOM` array guard | Fixed |
| 5 | Marker accepts whitespace / credentials | Partial: entropy check skipped for labels, and the prefix check now rejects legitimate key names |
| 6 | Tautological tests | Fixed for the original two (the new fenced-block test still asserts without a transaction; coverage gap only) |
| 7 | Lone surrogate encode | Fixed |
| 8 | Dead `SECURED_DISPLAY_PATTERN` | Fixed |
| 9 | Regex rebuilt, duplicate registries | Fixed |
| 10 | Duplicate delete commands | Fixed |

**New findings** are inline, ranked. The first five are the blockers: the label prefix check rejects real key names (a regression from the fix commit), chips inside headings/links break the block, escaped drafts re-lift without dispatching, and a shrinking key set never demotes a pill.

Cleanup notes not commented inline:
- `isBackedSecretChip` hard-codes `kind !== 'secret'` inside the generic registry; a per-kind `isBacked?` predicate on `ComposerChipKind` would keep the four parse+gate call sites in sync.
- `parseSecretMarker`'s `/^[A-Z0-9_]+$/` is a fourth copy of the env-var alphabet (`keyToEnvVar(x) === x` is the canonical check).
- `lib/secret-detection.ts` still has the runtime `lib` → `components` import from the first review.
- `useAgentSecrets` uses the default 5s `staleTime`, so every composer mount refetches `/secrets`.
- The e2e spec's slug regex duplicates `getCurrentAgentSlug` from two other specs with a looser pattern.

PR #961, opened 2026-09-04
feat(errors): placement/component keys on errorPresentation + composer placement (SUP-725)

yiw190:
## Why

**Bug:** a platform 402 ("insufficient balance") shows the plain red error card instead of the provider card with the billing link. The CLI tags that denial `apiErrorCode: 'unknown'`, which is not in `PROVIDER_ERROR_CODES`, so the server never asked the platform provider for a presentation.

Repro: exhaust a staging workspace's balance, send a message, observe the generic `ActivityErrorCard` rather than the "Insufficient Balance" card.

**Groundwork:** a provider error can only render one way today: the default `ProviderErrorCard`, inline in the chat stream. #914 needed a paywall above the composer and had to hard-wire that into `message-item`, `session-chat-column` and the composer itself. This PR adds the generic hooks so an error class can pick *where* it renders and *which* component renders it, with no paywall knowledge in the core files.

### The fix (user-visible)

- `BaseLlmProvider.presentationForTurnError(status, body, apiErrorCode)` replaces `parseErrorResponse` as the single entry point. It asks the provider override first, so a class the provider recognizes gets its presentation under any upstream SDK code (including `unknown`); the generic banner still only attaches under a `PROVIDER_ERROR_CODES` code.
- It is only consulted for upstream API failures. When the SDK reported no error code (container died, tool failure) or reported `max_output_tokens`, it returns `null` before the override runs. Those are the two paths that returned `null` on `main`; the widening is limited to real upstream codes outside `PROVIDER_ERROR_CODES`. This matters because the platform override matches on free text (`insufficient balance`, `spend cap`), and without the guard it would run on every failed turn.
- `isUpstreamApiErrorCode()` / `NON_UPSTREAM_ERROR_CODES` in `api.ts` hold that rule. `isProviderFacingError()` lets the renderer trust an attached presentation over the SDK code.
- `message-persister` and `agents.ts` call `presentationForTurnError`. `parseErrorResponse` had no callers left and is removed.

**Reviewer focus:** `parsePlatformErrorResponse`'s text patterns are now reached for `unknown`/`overloaded`/`model_not_found`/… turns, not only the five provider codes. Tests cover the two guarded paths (`max_output_tokens` + spend-cap text → null; no code + insufficient-balance text → null).

### Placement / component keys (no behavior change until #962)

- `errorPresentation` gains two optional keys: `placement` (`inline` | `composer`, default `inline`; `composer` renders **in place of** the composer, not above it) and `component` (registry key, default `default`). Older clients/servers are unaffected.
- `renderer/components/provider-error/`
  - `provider-error-registry.ts`: `component` key → React component. Adding a component = one row.
  - `provider-error-placement.tsx`: renders the session's current provider error iff its placement matches. Wraps the footer in `session-chat-column` at `composer`: no error → children (the composer) render as-is; a composer-routed error renders its component instead, and the component gets the displaced children so it can hand the composer back once resolved.
- `message-item` / `agent-activity-indicator` skip errors whose placement is not `inline`, so nothing renders twice.

No copy or styling changes; the platform provider recognizes the same classes as before. No dismiss.

## Stack

1. this PR
2. #962 — paywall as a registered component

Supersedes #914 / #931.

### DIFF
--- a/src/renderer/components/messages/message-item.tsx
+++ b/src/renderer/components/messages/message-item.tsx
@@ -1,7 +1,7 @@
 import { cn } from '@shared/lib/utils/cn'
 import { useState, useCallback, useRef, useLayoutEffect, useMemo, memo, type ReactNode } from 'react'
 import { Check, Copy, Link2 } from 'lucide-react'
-import { ProviderErrorCard } from '@renderer/components/ui/provider-error-card'
+import { resolveProviderError } from '@renderer/components/provider-error/provider-error-registry'
 import { ToolCallItem } from './tool-call-item'
 import { ThinkingBlockItem } from './thinking-block-item'
 import { SubAgentBlock } from './subagent-block'
@@ -366,20 +366,25 @@ function MessageItemComponent({ message, isStreaming, agentSlug, sessionId, isSe
 
   // Detect assistant messages that failed due to an LLM provider error (from SDK metadata)
   const isProviderErrorMessage = isAssistant && !!message.apiError && PROVIDER_ERROR_CODES.has(message.apiError)
+  // Errors routed to another placement (e.g. composer) are rendered there, not in the stream.
+  const providerError = resolveProviderError(message.errorPresentation)
+  const showInlineError = isProviderErrorMessage && providerError.placement === 'inline'
+  const hasInlineText = hasText && !(isProviderErrorMessage && !showInlineError)
 
   // Don't render assistant messages that have no text, no tool calls, and no
   // thinking (and aren't streaming). These are transient empty entries from
   // partially-persisted JSONL that will be filled in on the next refetch.
-  if (isAssistant && !hasText && toolCalls.length === 0 && thinking.length === 0 && !isStreaming) {
+  if (isAssistant && !hasInlineText && toolCalls.length === 0 && thinking.length === 0 && !isStreaming) {
     return null
   }
 
   // Skip rendering the text bubble for:
   // - assistant messages with only tool calls (no text) unless streaming
+  // - assistant provider errors routed to another placement
   // - user messages that only had attached files (text was fully stripped)
   const showMessageBubble = isUser
     ? (hasText || attachedFiles.length === 0)
-    : (hasText || isStreaming)
+    : (hasInlineText || isStreaming)
 
   return (
     <div
@@ -440,8 +445,8 @@ function MessageItemComponent({ message, isStreaming, agentSlug, sessionId, isSe
               )}
 
               {/* LLM provider error display */}
-              {hasText && !CustomUserRender && isProviderErrorMessage && (
-                <ProviderErrorCard message={text} presentation={message.errorPresentation} />
+              {hasText && !CustomUserRender && showInlineError && (
+                <providerError.Component message={text} presentation={message.errorPresentation} />
               )}
 
               {/* Text content */}
--- a/src/renderer/components/messages/message-item.tsx
+++ b/src/renderer/components/messages/message-item.tsx
@@ -2,6 +2,7 @@ import { cn } from '@shared/lib/utils/cn'
 import { useState, useCallback, useRef, useLayoutEffect, useMemo, memo, type ReactNode } from 'react'
 import { Check, Copy, Link2 } from 'lucide-react'
 import { resolveProviderError } from '@renderer/components/provider-error/provider-error-registry'
+import { ProviderErrorCard } from '@renderer/components/ui/provider-error-card'
 import { ToolCallItem } from './tool-call-item'
 import { ThinkingBlockItem } from './thinking-block-item'
 import { SubAgentBlock } from './subagent-block'
@@ -291,6 +292,9 @@ interface MessageItemProps {
   revealedToolCallIds?: ReadonlySet<string>
   /** Container-local image paths verified against image-bearing tool results. */
   embeddedImageAliases?: EmbeddedImageAliases
+  /** This row's provider error is the session's current one and a `ProviderErrorPlacement`
+   *  renders it elsewhere (e.g. composer). Skip the inline card so it does not show twice. */
+  suppressInlineError?: boolean
 }
 
 function resolveSubagentRun(
@@ -315,7 +319,7 @@ function resolveSubagentRun(
   }
 }
 
-function MessageItemComponent({ message, isStreaming, agentSlug, sessionId, isSessionActive, activeSubagents, completedSubagents, onRemoveMessage, onRemoveToolCall, readOnly, workDetailClassName, revealedToolCallIds, embeddedImageAliases }: MessageItemProps) {
+function MessageItemComponent({ message, isStreaming, agentSlug, sessionId, isSessionActive, activeSubagents, completedSubagents, onRemoveMessage, onRemoveToolCall, readOnly, workDetailClassName, revealedToolCallIds, embeddedImageAliases, suppressInlineError }: MessageItemProps) {
   useRenderTracker('MessageItem')
   const isUser = message.type === 'user'
   const isAssistant = message.type === 'assistant'
@@ -366,9 +370,11 @@ function MessageItemComponent({ message, isStreaming, agentSlug, sessionId, isSe
 
   // Detect assistant messages that failed due to an LLM provider error (from SDK metadata)
   const isProviderErrorMessage = isAssistant && !!message.apiError && isProviderFacingError(message.apiError, message.errorPresentation)
-  // Errors routed to another placement (e.g. composer) are rendered there, not in the stream.
+  // Only the session's current error is skipped here (its placement renders it). Older rows
+  // routed elsewhere stay in the transcript as the default inline card.
+  const showInlineError = isProviderErrorMessage && !suppressInlineError
   const providerError = resolveProviderError(message.errorPresentation)
-  const showInlineError = isProviderErrorMessage && providerError.placement === 'inline'
+  const InlineErrorComponent = providerError.placement === 'inline' ? providerError.Component : ProviderErrorCard
   const hasInlineText = hasText && !(isProviderErrorMessage && !showInlineError)
 
   // Don't render assistant messages that have no text, no tool calls, and no
@@ -446,7 +452,7 @@ function MessageItemComponent({ message, isStreaming, agentSlug, sessionId, isSe
 
               {/* LLM provider error display */}
               {hasText && !CustomUserRender && showInlineError && (
-                <providerError.Component message={text} presentation={message.errorPresentation} />
+                <InlineErrorComponent message={text} presentation={message.errorPresentation} />
               )}
 
               {/* Text content */}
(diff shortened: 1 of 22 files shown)


iddogino (commented):
The mechanism is the right shape: the provider owns recognition (`presentationForTurnError`), the presentation carries `placement` + `component`, the renderer resolves both through one registry, and the core files (`message-item`, `session-chat-column`, activity indicator) know nothing about any specific error class. Old client / old server compatibility is handled by defaults and by `providerErrorPresentationSchema` stripping unknown keys. The `isUpstreamApiErrorCode` guard is the right place to stop text-matching overrides from claiming `max_output_tokens` / container deaths.

Checks on the stacked head: touched test files 366/366, typecheck and lint clean.

One thing I want fixed here rather than in #962, because the rule lives here: a composer-routed error row renders nothing in the transcript, permanently (inline). Two nits inline.

Note, not a request: `ProviderErrorPlacement` wraps `PendingWakeBanner` and `StaleSessionNotice` as well as the composer, so a component that withholds `children` also hides those. Fine for the paywall, but worth a line in the `children` docstring so the next component author knows what they are displacing.

src/renderer/components/messages/message-item.tsx
```diff
@@ -365,21 +365,26 @@ function MessageItemComponent({ message, isStreaming, agentSlug, sessionId, isSe
   const streamingSplit = isStreaming && text ? splitStreamingMarkdown(text) : null
 
   // Detect assistant messages that failed due to an LLM provider error (from SDK metadata)
-  const isProviderErrorMessage = isAssistant && !!message.apiError && PROVIDER_ERROR_CODES.has(message.apiError)
+  const isProviderErrorMessage = isAssistant && !!message.apiError && isProviderFacingError(message.apiError, message.errorPresentation)
+  // Errors routed to another placement (e.g. composer) are rendered there, not in the stream.
+  const providerError = resolveProviderError(message.errorPresentation)
+  const showInlineError = isProviderErrorMessage && providerError.placement === 'inline'
+  const hasInlineText = hasText && !(isProviderErrorMessage && !showInlineError)
```
iddogino, line 378: **Should fix here.** `showInlineError` is false for any composer-routed row, regardless of whether that row is still the session's *current* error. Once the paywall clears (top-up, then a normal reply), `ProviderErrorPlacement` stops rendering it and this row still returns `null`, so the transcript shows the user's message followed by nothing. On `main` the inline card stayed in history. Your own test pins this ("renders nothing in the stream for a provider error routed to the composer placement") without the "while it is current" qualifier.

Suggest: suppress inline only for the row `currentProviderError` is showing. `MessageList` already has the messages and stream state, so it can compute the current error's message id (export a small `currentProviderErrorId` next to `currentProviderError`) and pass `suppressInlineError` to that one item; every other composer-routed row falls back to the default inline card. That also keeps the transcript honest when an error is routed to a placement the current layout does not mount.

src/shared/lib/types/api.ts
```diff
@@ -236,6 +236,27 @@ export const PROVIDER_ERROR_CODES = new Set([
   'server_error',
 ])
 
+// SDK codes for turn failures that are not an upstream API error. A provider
+// override must never claim these, whatever the error text happens to contain.
+export const NON_UPSTREAM_ERROR_CODES = new Set(['max_output_tokens'])
```
iddogino, line 237: Nit: with the SDK's assistant error enum, `NON_UPSTREAM_ERROR_CODES` and `PROVIDER_ERROR_CODES` are complements except for `unknown`, so there are now two hand-maintained sets whose relationship is implicit. When the SDK adds a code, it lands in neither set and silently becomes "upstream" here. Either derive one from the other (`isUpstreamApiErrorCode = code != null && !NON_UPSTREAM.has(code)` is fine, but then say in the comment that `PROVIDER_ERROR_CODES` is `upstream minus unknown`), or add a test in `api.test.ts` that asserts the two sets are disjoint and that their union plus `unknown` equals the SDK enum, so drift fails CI instead of changing routing.

src/renderer/components/provider-error/provider-error-placement.tsx
```diff
@@ -0,0 +1,75 @@
...
+    if (item.type !== 'assistant') continue
+    const isProviderError =
+      !!item.apiError && isProviderFacingError(item.apiError, item.errorPresentation) && !!item.content.text
+    return isProviderError
+      ? { message: item.content.text, presentation: item.errorPresentation, live: false }
+      : null
+  }
+  return null
+}
+
```
iddogino, line 47: Nit: the persisted branch decides on the last assistant row only, which is what you want, but it silently depends on `useMessages` returning the trailing page with the newest item last. That is true today (`message-list` and the activity indicator rely on the same thing). A one-line comment here saying "trailing page, newest last" would save the next reader a trip into `use-messages.ts`.

src/renderer/components/messages/message-item.tsx
yiw190: Fixed in 644a157. `currentProviderError` now returns the row's `messageId`, and `currentProviderErrorId` (next to it) gives `MessageList` the id of the one persisted row a placement is showing; only that row gets `suppressInlineError`. Every other composer-routed row renders the default inline card, so the transcript keeps the error after the paywall clears. `currentProviderErrorId` returns null for inline-routed errors, since the transcript is the inline placement and nothing else mounts it. Tests: message-item (suppressed vs. fallback card), placement (id/null cases), message-list (current row skipped; kept after a normal reply; older routed row kept while a newer one is current).

src/shared/lib/types/api.ts
yiw190: Went with the derive-and-document route plus a test, with one correction to the premise: the SDK enum has 11 codes (`oauth_org_not_allowed`, `account_on_hold`, `overloaded`, `model_not_found` are in neither set), so 'union plus unknown equals the enum' would fail today. `api.test.ts` now pins the enum from the pinned `sdk.d.ts` and asserts NON_UPSTREAM ∪ PROVIDER ∪ {the five upstream-without-banner codes} == enum and that the two sets are disjoint. A new SDK code fails the test until someone places it. The host can't import the SDK types (only agent-container depends on it), so the enum is a literal in the test with a note to update on bump.

src/renderer/components/provider-error/provider-error-placement.tsx
yiw190: Added the 'trailing page, newest last' line. Also added a note on `children` that the composer placement displaces the pending-wake / stale-session banners too, per your closing note.

yiw190:
Thanks — all three addressed, plus one adjacent gap I found while fixing the first.

**Composer-routed row vanished from the transcript** (fixed, 644a157 + 0248443). `MessageList` now knows which single row a placement is currently showing (`currentRoutedProviderError`) and suppresses only that one. Every other composer-routed row renders the default inline card, so history keeps the error after the paywall clears. Inline errors are never suppressed since the transcript is the inline placement. Same fix covers the streaming row, which was showing the raw `API Error: 402 {...}` text next to the placement until the persisted row arrived.

**Two hand-maintained Sets** (fixed). One correction to the premise: the SDK enum has 11 codes, not 7 — `oauth_org_not_allowed`, `account_on_hold`, `overloaded`, `model_not_found` are in neither set, so "union + unknown == enum" would fail today. `api.test.ts` pins the full enum (host can't import SDK types; literal with a note to update on bump) and asserts disjoint + NON_UPSTREAM ∪ PROVIDER ∪ {those five} == enum. A new SDK code is red until someone places it.

**Comments** (done): "trailing page, newest last" on the scan, and the `children` docstring now says the composer placement displaces the wake/stale banners too.

Touched tests green, lint + typecheck clean, CI green.

src/renderer/components/provider-error/provider-error-placement.tsx
```diff
@@ -0,0 +1,91 @@
...
+// Live error wins. Otherwise the last assistant message, if it is a provider
+// error and the agent is not working again. A normal reply after it expires it.
+export function currentProviderError(
+  live: LiveErrorState,
+  messages: readonly ApiMessageOrBoundary[] | undefined,
+): CurrentProviderError | null {
+  if (live.error && isProviderFacingError(live.apiErrorCode, live.errorPresentation)) {
+    return { message: live.error, presentation: live.errorPresentation ?? undefined, live: true, messageId: null }
```
iddogino: **Blocking.** When the live error wins, `messageId` is `null`, so `MessageList` suppresses only the streaming row. But `session_error` keeps `error` set until the next `stream_start` / `session_active` / reconnect, and its handler in `use-message-stream.ts` explicitly expects the persisted JSONL row to arrive in the meantime ("until the persisted JSONL data arrives"). All three `/messages` responses attach the presentation, so the steady state after a 402 is: `error` set **and** an assistant row with `apiError` + composer presentation at the end of `messages`. For that row `item.id === null` is false, `showInlineError` is true, and `InlineErrorComponent` falls back to `ProviderErrorCard`, so the transcript shows a full "You need more usage credit" card directly above the paywall until the user sends something. The new message-list tests never have both at once (`[user]` with a live error, or `[user, routedError()]` idle).

Suggest: keep live-wins for the placement, but always report the persisted row the live error was written into. Scan for the last assistant row first; if it is a provider-facing error, use its id even when returning the live variant:

```ts
const lastAssistantError = lastAssistantProviderError(messages) // { id, text, presentation } | null
if (live.error && isProviderFacingError(live.apiErrorCode, live.errorPresentation)) {
  return { message: live.error, presentation: live.errorPresentation ?? undefined, live: true, messageId: lastAssistantError?.id ?? null }
}
```

Then `MessageList` suppresses both the streaming row (`live === true`) and that persisted row (`item.id === messageId`), and a test with `mockStreamState.error` set **and** `mockMessagesData.data = [user, routedError()]` should assert zero `provider-error-card`s.

iddogino (changes requested):
All three landed as asked, and the SDK enum correction is right: thanks for pinning the full 11 and making placement of a new code a deliberate choice. The `suppressInlineError` shape (MessageList knows the one current row, MessageItem stays dumb) is the right split, and covering the streaming row too was a good catch.

Checks on the stacked head: 1342/1342 across the renderer/messages, provider-error, llm-provider, types and billing test dirs; typecheck and lint clean.

One remaining gap, which is why I'm holding: the live error and its persisted row coexist, and only one of them is suppressed. Inline below.

src/renderer/components/provider-error/provider-error-placement.tsx
yiw190: Fixed in 91c53e3. `lastAssistantProviderError(messages)` is scanned first; the live variant now reports that row's id (`messageId: lastAssistantError?.id ?? null`), so `MessageList` suppresses both the streaming row (`live === true`) and the persisted row (`item.id === messageId`). New message-list test has `mockStreamState.error` set **and** `[user, routedError()]` and asserts zero `provider-error-card`s (confirmed red before the fix). Placement tests updated: live + persisted → `messageId` is the row id; live + only a user row → null; live + a normal reply row → null. #962 rebased on top.

iddogino (approved):

PR #962, opened 2026-09-04
feat(paywall): platform 402 as a registered provider-error component (SUP-725)

yiw190:
## Why

A platform 402 leaves users with a generic provider error and no next action. Repro: exhaust an unsubscribed or subscribed staging workspace, send a message.

Reworks #914 onto the placement/component keys from the wiring PR. The platform provider tags a 402 with `component: 'paywall'`, `placement: 'composer'`; the renderer registry maps that key to `PaywallCard`. `message-item` / `session-chat-column` know nothing about paywalls.

## Proxy dependency

The clear path reads `access.allowed` and the CTA reads `hasPaymentMethod` from `GET /v1/billing`. Both come from platform [#363](https://github.com/SkillfulAgents/platform/pull/363) (`feat(billing): SuperAgent paywall hand-off + resume gate`), merged to `main` 2026-09-03. `Deploy Proxy Worker` succeeded for `main` and `staging` after it, and a live `GET /v1/billing` against the production proxy returns both fields. No merge-order constraint remains: this PR fails open against a proxy without them (below).

## Behavior

The card **fails open**. The composer is withheld only while a **fresh** billing snapshot positively says `access.allowed === false`. In every other state (snapshot without `access`, stale cached fallback, fetch error, platform disconnected, still loading) the card renders **above** the composer and the composer stays usable. **Dismiss** always hands the composer back for that mount.

- `cleared` = fresh snapshot with `access.allowed === true` (for a live 402, fetched after the paywall appeared) → card gone, composer back.
- Known window: after Dismiss or after clearing, the transcript row for that 402 is still the "current" routed error (see #961), so nothing shows for it until the next assistant reply lands. Old inline card in history after that. Accepted; not a permanent loss.
- Live 402: the composer is visible while the first billing fetch runs, then withheld once the fresh deny lands. Consequence of failing open; accepted.
- **Add usage** opens the platform (`intent=topup`). That button becomes **Recheck**. The card also refetches billing every 5s while it is visible. No `billing-updated` deep link / callback.

## What changed

- `platform-error-presentation.ts`: 402 → paywall component at the composer placement; copy tailored to `subscription_required` when the body kept it. `isInsufficientBalance` is `status === 402` or the two `insufficient` substrings; no bare `402` text match. `ORG_BILLING_LINK` exported.
- `error-presentation.ts`: `resolveOrgLink(path, org)` is the single `{orgId}` resolver, used by `resolvePresentationMarkdown` and the paywall CTA.
- `provider-error/paywall-cta.ts` (+ `paywall-cta.test.ts`, node): pure CTA resolution — subscribe / add card / top up / fix payment / ask admin / go to billing. One `PAYMENT_NEEDS_ATTENTION` set. Top-up URL keeps `intent=topup` and does not send `return_app`.
- `provider-error/use-paywall-billing.ts`: `{ cta, loading, blocked, cleared, recheck }`. Refetches on mount (live), on Recheck, and every 5s while the card is up. `seenAt` / `dataUpdatedAt` guard so a snapshot cached before a live 402 cannot clear it.
- `provider-error/paywall-card.tsx` (+ `paywall-card.test.tsx`, jsdom): the card, Dismiss, fail-open rendering, and Add usage → Recheck.
- `PlatformBillingInfoSchema`: optional `hasPaymentMethod` and `access`.
- Registry: one row, `paywall → PaywallCard`.
- `provider-error-card.test.tsx`: the 402 inline-fallback test asserts the paywall copy (it pinned the retired billing-link card).

No callback path: no `billing-updated` scheme, IPC, or `RootLayout` handler.

Not carried over from #914/#931: auto-resume of the session after billing recovery. The composer stays available, so the user retries manually.

## Stack

1. #961 — placement/component keys + composer placement
2. this PR

Supersedes #914 / #931.


src/renderer/components/provider-error/paywall-card.tsx
```diff
@@ -0,0 +1,322 @@
...
+    </div>
+  )
+}
+
+// Platform 402. An invitation, not a failure: neutral card, title + muted
+// subtitle, one role/billing-aware CTA. Displaces the composer; hands it
+// back once the proxy gate allows access again.
+export function PaywallCard({ message, presentation, children, live = true }: ProviderErrorComponentProps) {
+  const billing = usePaywallBilling(extractSubscriptionRequired(message), live)
+  if (billing.cleared) return <>{children}</>
```
iddogino: **Blocking.** `cleared` is the only way `children` (the composer) come back, and it requires `fresh && access.allowed === true`. Each of these leaves the session with no composer and no dismiss:

- Proxy that does not return `access` (see the schema comment: "newer proxies"): `cleared` is never true, top-up or not.
- Platform disconnected or provider switched while this row is the last assistant message: `useBillingInfo(false)` is a disabled query, so in react-query v5 `isLoading` is false, `data` is undefined, `cleared` is false, and `billingHref` is null so the button is disabled too. (The persisted path self-heals on the next `/messages` refetch because `attachProviderErrorPresentations` uses the active provider, but the live one does not until the SSE reconnects.)
- Billing fetch failing → server serves the cached snapshot with `stale: true` → never fresh.
- Any drift between the proxy's 402 gate and its `access.allowed` verdict.

Suggest failing open: withhold the composer only when a fresh snapshot positively says `access.allowed === false`; in every other state render the card **above** `children` (which is what #914 did). Then the worst case is the old behaviour, not a dead session. A "Dismiss" that hands the composer back for this mount would also be cheap insurance.

src/shared/lib/types/skillset-schema.ts
```diff
@@ -269,6 +269,15 @@ export const PlatformBillingInfoSchema = z.object({
   orgPool: z.object({
     poolBalanceCents: z.number(),
   }),
+  // Newer proxies include this so the 402 paywall can choose add-card vs top-up.
+  hasPaymentMethod: z.boolean().optional(),
+  // Proxy gate verdict. `reason` stays free-form so a new proxy reason can't fail the parse.
+  access: z
```
iddogino, line 275: **Blocking question.** Which proxy change adds `access` and `hasPaymentMethod`, and is it live on staging/production? The clear path in `usePaywallBilling` is `access.allowed === true` and nothing else, so against a proxy without it the deep link refetches a snapshot that cannot change the answer. Please link the proxy PR in the description and note the merge order; if it is not deployed yet this PR should wait for it or fail open (previous comment).

src/shared/lib/llm-provider/platform-error-presentation.ts
```diff
@@ -6,27 +6,105 @@ import {
...
 
+// Streaming 402s often drop the JSON body and surface as "socket closed".
+// Any 402 from this provider is a billing deny; the card picks the CTA.
 function isInsufficientBalance(status: number | undefined, raw: string): boolean {
   const lower = raw.toLowerCase()
   return (
-    lower.includes('insufficient balance')
+    status === 402
+    || /\b402\b/.test(lower)
+    || lower.includes('insufficient balance')
```
iddogino: `inferred` already comes from `inferErrorStatus`, which matches `\b402\b` when there is no parenthesized status, so this clause is redundant in that case. When there *is* a parenthesized status it is wrong: `"Request rejected (500) · upstream id 402-abc"` infers 500 and this line still routes it to the paywall, replacing the composer. Now that #961 runs this override for every upstream code, the text path is reached far more often than before. Drop the clause and keep `status === 402` plus the two `insufficient` substrings.

src/renderer/components/provider-error/paywall-card.tsx
```diff
@@ -0,0 +1,322 @@
...
+    if (!billingError) return
+    console.warn('[Paywall] billing snapshot fetch failed:', billingError)
+    captureRendererException(billingError, { tags: { area: 'paywall', op: 'billing-info' } })
+  }, [billingError])
+
+  return useMemo(() => {
+    const billingHref = resolveOrgBillingUrl(platformAuth)
+    if (billingQuery.isLoading) {
+      // Resolve from the 402 flag alone when the snapshot cannot change the answer.
+      const canResolveFrom402 = platformAuth !== undefined && (
```
iddogino: Dead computation: while `isLoading` this returns `cta` alongside `loading: true`, but `PaywallCard` uses `'Checking billing'` for the heading and `PaywallActions` returns the spinner whenever `loading`, so the early CTA is never rendered. The test "resolves subscribe from the 402 body before the snapshot arrives" asserts only the spinner and the checking copy, so it would pass with `cta: null`. Either render the early CTA (then the test should assert the Subscribe button) or delete `canResolveFrom402` and return `cta: null` here.

src/main/index.ts
```diff
@@ -666,6 +668,15 @@ ipcMain.handle('flush-pending-menu-commands', () => {
   return flushPendingMenuCommands()
 })
 
+ipcMain.handle('flush-pending-billing-updated', () => {
```
iddogino: Simplify: `pendingBillingUpdated` + `flush-pending-billing-updated` + `billing-updated-consumed` exist so a paywall that mounts *after* the deep link still refreshes. The same outcome with no main-process state: register one `onBillingUpdated` listener at app level (wherever `onPlatformAuthCallback` is wired) that does `queryClient.invalidateQueries({ queryKey: ['platform-billing'] })`. A paywall mounting later runs its own mount-time invalidate anyway, and the Account screen's billing tab gets the refresh for free. That removes the flag, one `handle`, one `on`, and the `finally` in preload.

src/renderer/components/provider-error/paywall-card.tsx
```diff
@@ -0,0 +1,322 @@
...
+import { captureRendererException } from '@renderer/lib/error-reporting'
+import { openExternalUrl } from '@renderer/lib/open-external'
+
+import type { ProviderErrorComponentProps } from './provider-error-registry'
+
+// ---------------------------------------------------------------------------
+// CTA resolution (pure)
+// ---------------------------------------------------------------------------
+
+const ORG_BILLING_PATH = '/dashboard/organizations/{orgId}?tab=billing'
```
iddogino: Reuse nits: (1) `ORG_BILLING_PATH` + `resolveOrgBillingUrl` re-implement the `{orgId}` substitution and trailing-slash handling that `resolvePresentationMarkdown` already does for `ORG_BILLING_LINK` in `platform-error-presentation.ts`; export one `resolveOrgLink(path, org)` from `error-presentation.ts` and use it in both. (2) `UNRESOLVED_SUBSCRIPTION` and `PAYMENT_NEEDS_ATTENTION` (L61-62) are the same three strings. (3) The description lists `paywall-cta.ts` and `use-paywall-billing.ts` as separate files with their own tests; they are all in this 320-line file. The pure CTA section is testable without jsdom, so the split is worth doing rather than just fixing the description.

iddogino (changes requested):
Good use of the #961 hooks: one registry row, one provider branch, and the paywall logic lives entirely in its own component. The `seenAt` / `dataUpdatedAt` guard against a stale `allowed` snapshot is careful and well tested.

Holding on one thing: the card fails closed. It withholds the composer until a **fresh** billing snapshot says `access.allowed === true`, and there is no dismiss. Every way the clear path can fail leaves the user unable to type in that session, which is a worse outcome than the generic card this replaces. Details inline, plus a question about the proxy dependency that the whole clear path rests on.

Checks on this head: touched test files 366/366, typecheck and lint clean.

**Blocking**
1. Composer lockout when billing cannot be confirmed (inline at `paywall-card.tsx`).
2. `access` / `hasPaymentMethod` are described as coming from "newer proxies". Which proxy change, and is it deployed? Without `access` the paywall never clears, even after a successful top-up.

**Should fix**
- `/\b402\b/` in `isInsufficientBalance` is redundant with `inferErrorStatus` and wrong when a different parenthesized status precedes it.

**Simplify / reuse**
- CTA computed during `isLoading` is never rendered, and the test named for it only asserts the spinner.
- The pending-flag + two IPC channels for `billing-updated` can be one app-level listener that invalidates the billing query.
- `resolveOrgBillingUrl` / `ORG_BILLING_PATH` duplicate `resolvePresentationMarkdown` / `ORG_BILLING_LINK`; two identical status sets.
- PR body says `paywall-cta.ts` and `use-paywall-billing.ts` are separate files; they are all in `paywall-card.tsx`.

yiw190:
Both blockers addressed, plus the rest. Description rewritten to match.

**Composer lockout** (fixed, 7f1fd8e). The card now fails open: the composer is withheld only while a fresh snapshot says `access.allowed === false`. No `access` field, stale fallback, fetch error, platform disconnected, loading → card sits above a usable composer. Dismiss always hands it back. Each of your four cases has a test asserting the composer is present. Two accepted consequences, noted in the description: after Dismiss/clear the transcript shows nothing for that 402 until the next reply lands (then the inline card returns per #961); and for a live 402 the composer is visible during the first billing fetch, then withheld.

**Proxy dependency**: platform #363, merged to main 2026-09-03, deployed to staging and production (`Deploy Proxy Worker` green; live `GET /v1/billing` on prod returns `access` and `hasPaymentMethod`). Linked in the description. With fail-open there's no merge-order constraint left.

**`/\b402\b/`**: dropped; `status === 402` + the two `insufficient` substrings. Test for `"(500) · upstream id 402-abc"` → null.

**Dead CTA during loading**: deleted; test now asserts spinner only, no button.

**IPC**: one `BillingUpdatedHandler` in `RootLayout` invalidates `['platform-billing']`. Pending flag, both extra channels, and the preload `finally` are gone. (Put it in `RootLayout` rather than next to `onPlatformAuthCallback` — that one is only wired inside `usePlatformConnect`, not app-wide.)

**Reuse**: `resolveOrgLink` is the single `{orgId}` resolver, `ORG_BILLING_LINK` exported not re-declared, one `PAYMENT_NEEDS_ATTENTION` set, and the file is actually split now: `paywall-cta.ts` (pure, node test), `use-paywall-billing.ts`, `paywall-card.tsx`.

Also fixed the one red CI test (`provider-error-card.test.tsx` still pinned the old "Insufficient Balance" card).

src/shared/lib/llm-provider/error-presentation.ts
```diff
@@ -68,19 +68,27 @@ export function defaultParseErrorResponse(
...
 const ORG_PLACEHOLDER_LINK = /\[([^\]]+)\]\(([^)]*\{orgId\}[^)]*)\)/g
 
+export interface OrgLinkContext {
+  connected?: boolean
+  platformBaseUrl?: string | null
+  orgId?: string | null
+}
+
+// `{orgId}` path or absolute URL → concrete URL on the connected org; null without org context.
+export function resolveOrgLink(pathOrUrl: string, org: OrgLinkContext | null | undefined): string | null {
```
iddogino: this still feels like org related logic that lives where it shouldn't (breaking the abstraction). Why? "Org" is a platform LLM provider concept, which is one type of provider. 

If we added open router, will we need an open router link context too??

iddogino (changes requested):

iddogino (approved):

PR #971, opened 2026-09-04
One description per workspace file, one way to open a row (SUP-771)

iddogino:
Closes SUP-771. Pure refactor against a quiet main, as the ticket sequenced it — the file-UI stack (#949, #950, #945, #946, #948) has landed.

## The problem

Ten surfaces draw "a workspace file as a clickable thing" — the delivery row, the `deliver_file` renderer, the download pill, home bookmarks, the folder browser, the preview tab strip, the drawer's header, its unsupported-file screen, the composer's chips and a sent message's. Each re-derived the same handful of facts from a raw path, and by now the answers disagreed rather than merely repeated:

| Question | Answers before | Answers after |
| --- | --- | --- |
| What URL serves this file | 4 | 1 |
| Is this an image | 3 | 1 |
| How big is it, formatted | 2 | 1 |
| What is this file called | 2 (`read` vs `write`) | 1 |
| How does a file row activate | 4 | 1 |

## What this does

**One descriptor.** `describeWorkspaceFile(path, agentSlug)` answers display name, relative path, folder-or-not, category, the drawer's renderer for it, whether to draw it as a picture, and its download/inline/API-route URLs. The delivery row, the pill, the sent chip, the drawer's header and the unsupported screen ask it instead of assembling those from three or four helpers each. The point isn't the lines it saves — it's that the next surface asks rather than re-derives, which is now also written down in `CLAUDE.md`.

**One activation.** `openableProps()` carries the click, keyboard and focus behaviour for a row that opens the drawer but can't be a `<button>`. The three hand-rolled copies disagreed about `preventDefault`, `stopPropagation` and the target guard — and the guard is the part that matters: without it a key event bubbling from a nested link gets `preventDefault()`ed by the row, which is exactly how Enter on a delivery row's Download button opened the drawer instead of downloading (#945). The tab strip's copy is deleted rather than converted: its close control was a `role="button"` span nested inside the tab's own `<button>`, invalid content whichever way you read it, so the tab is now a shell holding two real buttons and the browser handles Enter.

**Tab identity includes the agent.** The key was the path alone. Two agents can each have a `/workspace/report.md`, and a drawer that outlives one agent — the chat-integration views — would have shown the first agent's bytes under the second's name; opening the second file wouldn't even have made a second tab.

**`Bookmark` is a union.** It was a hand-written interface with `link`, `file` and `folder` all optional while the server enforced "exactly one of" in Zod, so every use rebuilt the invariant with a ternary chain and a `!` per branch. A type-level assertion in the agents route now fails to compile if what the renderer can build stops being something the schema accepts.

## Drifts this closes (the only behaviour changes)

- An `.svg` was a previewable image to the drawer and not an image at all to the Markdown transform, so the same picture rendered inline in one place and not the other; an `.avif` was the reverse. Both read the one table now, and `avif` moves into the previewable group — where Chromium has always had it.
- The Markdown transform carried its own workspace-prefix strip and traversal check beside the shared encoder's. The check lives once now, in `isSafeWorkspaceFilePath` next to the builder it guards: the encoder escapes, it can't sanitise, and `..` comes through `encodeURIComponent` unchanged.
- `toWorkspaceRelativePath` now anchors its prefix to a segment boundary. A `/workspaceX/report.pdf` was coming back as `X/report.pdf`, which reads as a path inside the workspace and is not one.
- `write`'s `getDisplayPath` replaced `/workspace/` *anywhere* in the string rather than only at the start and left a trailing slash on, so `read` and `write` named one path two ways.
- `formatBytes` in the tool-call item was `formatFileSize` with one fewer decimal on the KB branch, so a PDF read `195 KB` there and `195.3 KB` in the drawer. It reads `195.3 KB` in both now.

`file-types.ts` also moves from `components/file-preview/` to `lib/` — it holds no React, and a lib module reaching up into a component directory for a shared table is how the table stops getting reused.

## Verification

Typecheck clean, lint clean, 4293 unit tests pass, and `file-preview.spec.ts` + `file-upload.spec.ts` (17 e2e) pass against the restructured tab strip. New tests cover the descriptor, the activation primitive's target guard, `isSafeWorkspaceFilePath`, the closed `.svg`/`.avif` drift, the cross-agent tab collision, and the tab close as a real button.

## Screenshots

Delivered-file rows (the descriptor picks the chevron for what the drawer can render and the Download button for what it cannot) and the preview drawer's restructured tab strip.

| Light | Dark |
| --- | --- |
| Delivered-file rows (light) | Delivered-file rows (dark) |
| Preview drawer tab strip (light) | Preview drawer tab strip (dark) |

https://claude.ai/code/session_01HZzjPxnddXJX46bLy5LHEu


iddogino:
Both valid, both mine, and they're the same mistake twice: I gave *tabs* a `(agentSlug, path)` identity and left everything hanging off a tab keyed by path alone. That's worse than not having touched it — before the change two agents could never have two tabs for one path, so the aliasing was unreachable. Separating the tabs is what made it reachable.

Fixed in 6f29ad8f.

**P1 — comments.** This was the serious one, and the draft-submission consequence you named is real: `CommentBar.handleSubmit` formats whatever `comments` it was handed into one session's composer draft, so a bucket shared across agents could be pasted into the wrong composer.

The map is now keyed by `getWorkspaceFileKey(agentSlug, filePath)` — the same function `getPreviewTabKey` builds the tab key from, so the two identities can't drift apart again. `FileComment` carries its `agentSlug`, `commentsFor(filePath, agentSlug)` is the single place the key is composed, and `closeTab` / `renameFilePath` / `renameDirectoryPath` / `removeDirectoryPath` each act on one agent's bucket. That meant threading the slug through the seven renderers that draw comments plus `CommentOverlay` — wide, but mechanical.

One thing worth flagging: the directory rename/delete paths read the bucket's identity off `fileComments[0]` rather than parsing it back out of the key, since a path can contain the separator. Every comment in a bucket shares its file, so that's invariant, but it's the one non-obvious line.

**P2 — folder state.** `toggleFolder`, `setFolderQuery` and `selectFolderEntry` now take the slug and share an `isFolderTabFor` predicate with `openFolder`, next to the file one. Single call site each (`folder-browser.tsx`), which already had `folder.agentSlug`.

**Tests.** Six cases across both, and I verified each fails against the reintroduced bugs (reverted `getWorkspaceFileKey` to ignore the slug, and `isFolderTabFor` likewise — 6 failures, 0 with the fix). The context test now reads comments back through the exported key builder instead of restating its format.

Full run: typecheck clean, lint clean, 4298 unit tests, and the 17 e2e in `file-preview.spec.ts` + `file-upload.spec.ts` still pass.

On the local-run problem — `node_modules` isn't inherited by a worktree checkout, so `vitest`/`tsc` won't resolve there. `npm ci` in the worktree fixes it; the repo also has a note about linking the main checkout's `node_modules` in if you'd rather not pay a full install.

PR #980, opened 2026-09-05
feat(voice): read agent replies aloud with Deepgram text-to-speech

iddogino:
First of two stacked PRs toward a voice-to-voice "call mode". This one adds the text-to-speech side and a visible win on its own: agent replies get a speaker button that reads them out, with the spoken words lit as playback reaches them. The second PR wires the call loop (auto-send after speech, auto-play replies, barge-in interrupt) on top of the streaming primitive introduced here. A rename PR (`stt` → `voice` across paths and routes) lands between them; this PR deliberately keeps the existing `stt` names so the diff stays about behavior.

Deepgram only (directly or via the platform proxy). OpenAI reports `supportsTts: false` and shows nothing.

## What's here

**Provider capability** — `BaseSttProvider` gains `getTtsVoices()` / `getTtsToken()` next to the existing voice-agent pair, with `supportsTts()` derived from having a non-empty catalogue. The voice ids are the provider's own (they go straight into its synthesis request), so the catalogue lives with the implementation: Deepgram's Aura-2 list is in `deepgram-voices.ts`, shared by the platform provider since it proxies to Deepgram; OpenAI returns none. Deepgram and platform mint the same short-lived grant token they already use for listening. `/api/stt/configured` reports `supportsTts`, the provider's voices, and the deployment's default among them (that endpoint is member-readable; the settings endpoint is not), so the renderer never imports a catalogue; new `/api/stt/tts-token` returns the token plus the voice and speed to use.

**Per-user voice and speed** — each person keeps their own voice and speaking rate in user settings (`voice.ttsVoice`, `voice.ttsSpeed`). The token endpoint asks the provider to resolve the caller's pick, then the deployment's default, then the provider's own default, so the client still needs no settings round-trip. Speed rides Deepgram's `speed` parameter (0.7–1.5) and is provider-neutral. Stored voice ids are plain strings (which ids exist is the provider's business, not the schema's); both PUTs check a voice against the configured provider's catalogue and refuse an unknown one rather than silently dropping it, and a stored id the current provider no longer offers falls through at read time. Writing `ttsVoice: null` unsets a personal pick, which is what the picker's **Workspace Default (Zeus)** entry does in shared deployments, so a member can go back to following the admin's choice.

**Settings › Voice is open to everyone.** A personal section (voice, speed, preview) sits at the top; provider, API key, and the STT test stay admin-gated below it. A member's tab never requests the admin-only settings endpoint; the default voice it labels comes from `/api/stt/configured`. Saving a provider or key invalidates that status, so the personal section and the speaker buttons appear as soon as the key lands. Shared deployments get an admin-only "Default voice", and a member on a deployment without speech configured sees a note to ask an admin.

**Streaming synthesis** — `TtsAdapter` (`src/renderer/lib/tts.ts`) mirrors `SttAdapter`: text goes in via `speak()`/`flush()`, PCM comes back, `flushed` events mark batch boundaries. Text is buffered until the socket opens, so a caller can queue and go. Deepgram Aura over `wss://api.deepgram.com/v1/speak`, linear16 @ 24 kHz.

**Player** — `SpeechPlayer` is push-based (`append(words)` … `end()`), so a finished message and a message still streaming in take the same path. Words are cut into sentence-sized batches by `SpeechSegmenter` (abbreviation-aware, min/soft/hard limits); each batch is one Speak+Flush. Batches are handed to the synthesizer only about ten seconds ahead of playback (two in flight), so stopping or changing speed early in a long reply pays for a sentence or two rather than the whole message. Audio is scheduled back-to-back on an `AudioContext` that is created inside the click, before the token round-trip, so autoplay-blocked browsers unlock it; a refused resume, and a socket the server closes before the reply is in, both surface as errors instead of a silent or stuck "speaking". No word timestamps come from the provider, so the position is derived: the server's per-batch `Flushed` event ties each batch's audio to its words, and position inside a batch is interpolated. `getWordCursor()` is what the highlight reads.

**Word-exact highlight without alignment heuristics** — `spoken-words.ts` does one walk over the message's HAST tree. `markdownToSpokenWords` collects the prose words (skipping fences, image alt text, raw HTML) and `rehypeSpokenWords` wraps those same words in `<span data-spoken-word={i}>`. Same walker, so span *i* is always word *i*. A word split by inline markup (`**bold**,`) stays one spoken word and its spans share an index. `useSpokenWordHighlight` runs a rAF loop only while reading and toggles `data-spoken` on just the spans that crossed the cursor.

**UI** — `ReadAloudControls` under settled assistant replies (hover-revealed, pinned while reading; hidden entirely when the provider can't speak). Idle: a speaker button. While reading: **pause/resume**, **stop**, and a **speed picker** that persists to the user's own settings. A Deepgram connection is fixed to one speed, so a speed change mid-reply carries on from the word being spoken at the new rate, with the words already said staying lit (the player takes a first-word offset); made while paused it stays paused, and once playback has ended it changes nothing. A failure (revoked key, dropped socket, blocked audio) shows as red text next to the speaker button on the message it happened to, cleared by the next play. Pause suspends the `AudioContext`, which freezes its clock, so the cursor holds and scheduled audio waits in place. The prose is dimmed and lights up word by word; the cursor is monotonic so it never retreats while a segment's audio is still arriving. One reader app-wide: playing another message stops the first. Every control stays mounted and only toggles display — swapping the row's children on play tripped the same WebKit re-clamp described below, by the row's height.

## The WebKit scroll jump, and why the spans are rendered up front

Safari threw the viewport up by a message's worth whenever play or stop was pressed on a reply at the live edge. Reproduced in Playwright WebKit: re-rendering a reply's Markdown with the word spans makes WebKit re-clamp the scroll container mid-commit (React edits the kept text nodes in place before inserting the spans; hand-built spans don't trigger it, React's diff does). The scroll engine then reads the upward, size-stable move as the reader escaping — the click is fresh input evidence — and drops follow. Chromium commits synchronously and never shows it.

So readable replies (TTS configured, settled assistant text) render their word spans up front, and play/stop change only classes and attributes, which never move WebKit. Verified with real play and stop clicks in WebKit on the last and a middle reply at the live edge, and in Chromium. The cost is a span per word in readable replies; the windowed list keeps it bounded.

## Verified live

Against the real Deepgram endpoint with a configured key: the `/v1/auth/grant` token is accepted by `/v1/speak` with the `['bearer', token]` subprotocol, every `Flush` yields one `Flushed` (the highlight advances in step), and the per-user voice/speed round-trip works end to end (invalid speed → 400, valid pick stored, token endpoint returns it).

## Tests

- `tts.test.ts` — adapter: URL/auth/binaryType, speed parameter, pre-open buffering, audio vs. event demux, error/close mapping.
- `spoken-words.test.ts` — Markdown → words (fences, images, HTML, headings, lists, tables), span indices match collection, markup-split words.
- `speech-segmenter.test.ts` — sentence/clause/block cuts, min-length merge, abbreviations, hard cap.
- `speech-player.test.ts` — batching, the send window, scheduling, cursor interpolation across segments, pause/resume (frozen clock, deferred finish), first-word offset, monotonic cursor, done/stop/error, server-initiated close, refused resume, odd-byte stitching.
- `use-read-aloud.test.tsx` — controller lifecycle (stale token round-trip, replacement, errors), pause/resume, restart-from-cursor with fresh credentials (no-op after done, paused stays paused), per-message error, hook, highlight loop.
- `message-item.test.tsx` — control gating (assistant only, not streaming/error), spans up front, dimming only for the active message, pause/resume/stop/speed visibility per status, error text.
- `voice-tab.test.tsx` — member vs. admin sections, members never request admin settings, Workspace Default until picked and to unset, personal writes go to user settings.
- `tts-support.test.ts` / `user-settings-schema.test.ts` — provider capability and voice catalogue, token minting, voice and speed resolution, lenient reads vs. strict writes.

`npm run typecheck`, `npm run lint`, and the touched suites pass. Also checked in Playwright WebKit against the dev server: the spans and controls appearing when the configured-status query resolves after the transcript mounted (direct load, and clicking into a session) keep the reader at the live edge; pause → speed change stays paused on the same word; a failed token fetch shows the red text; a 74-word reply plays through at 1.5× with a monotonic highlight; a speed write after playback ended replays nothing; the scroller never moved.

## Screenshots

Mock server; the reading state is driven through the store (no audio) so the frame is deterministic.

| Light | Dark |
| --- | --- |
| Speaker button on hover (light) | Speaker button on hover (dark) |
| Reading, words lit as spoken, with pause, stop, and speed (light) | Reading, words lit as spoken, with pause, stop, and speed (dark) |
| Settings › Voice, personal section above the admin sections (light) | Settings › Voice, personal section above the admin sections (dark) |

https://claude.ai/code/session_0196tnfhLZANb3NCZJioKQhf


src/shared/lib/stt/tts-voices.ts
iddogino, line 1: Voices are presented as a general file, but are an implementation specific detail (deep gram in this case) and should be encapsulated in the implementation.

stt provider should return supported voices

Speeds etc can remain generic (not provider dependent)

iddogino: Addressed in 260108765: the Aura-2 catalogue moved into `deepgram-voices.ts` (shared with the platform provider), `BaseSttProvider` gained `getTtsVoices()` / `getDefaultTtsVoice()` / `hasTtsVoice()` / `resolveTtsVoice()`, stored voice ids are plain strings validated against the configured provider at the write boundary, and `/api/stt/configured` reports the provider's voices so the renderer never imports a catalogue. Speeds stay generic in `tts-preferences.ts`.

yiw190 (approved):

PR #994, opened 2026-09-08
feat(billing): reuse platform cookies in a cross-site billing iframe

yiw190:
## Summary

Keep the paywall card compact: Web cloud owners/admins click a platform-hosted iframe CTA, then manage billing inside the existing SuperAgent Dialog. Companion: SkillfulAgents/platform#408. Electron and non-cloud handoff behavior stay unchanged.

This update aligns the host with the hosted Checkout refinement. It does not add Payment Element, a publishable Stripe key, or a SuperAgent subscription Dialog.

- Only the business CTA is embedded in the card (`surface=cta`); no inline payment form or extended card layout. The iframe CTA is compact and native-sized. Hosting the button in the platform iframe lets its click directly request Storage Access instead of trying to transfer a parent-page click across origins.
- While cookie permission is needed, the platform iframe shows only `Allow access to billing to continue.` After permission reload, the host still requires a real Upgrade confirmation against the actual quote. Access permission alone never authorizes a charge.
- Existing-card upgrade stays on the compact card (no subscription Dialog). No-card upgrade uses a single hosted Stripe setup popup, then the existing Metronome subscribe. Credits open an amount Dialog only after missing-card setup completes and that popup is closed and verified. No auto-purchase.
- Accept `open-billing` and status/theme messages only from the exact iframe window, configured platform origin, and matching organization. The existing BillingEmbedFrame is reused inside the existing Dialog; no new payment form or dialog framework.
- On `billing-updated`, recheck the existing billing snapshot. Keep the Dialog open while access remains blocked. Once a fresh snapshot allows access, show success feedback for 1.2 seconds before closing.
- Dialog content follows iframe height reports and the viewport limit. Long content scrolls inside the iframe, not both the iframe and Dialog. Preserve ready/timeout/retry/new-tab fallback and owner/admin eligibility.
- Continue using platform cookies directly. No OIDC billing bootstrap, org-bearer embed identity, or host API/page proxy. All production changes in this update are renderer-side.

## Validation

- 50 focused SuperAgent tests pass, including forged-message rejection, close/reopen, recheck gating, success delay, and unchanged Electron/non-cloud behavior.
- Changed implementation lint passes. Full SuperAgent typecheck was not rerun for this update.
- Real component UI fixture render tested with actions/SAA/Stripe mocked: compact Upgrade iframe, permission hint, quote-after-reload confirmation, existing-card upgrade, and credits popup-closed-before-amount-dialog sequencing.
- No real Stripe, 3DS, or Safari verification. Screenshots were delivered in chat and are not attached here.

## Browser evidence and limitations

Full Chromium 145 with distinct HTTPS sites and real third-party-cookie blocking confirms native Storage Access activation and reuse: after a browser-protocol permission grant, the CTA click restores cookie access; a new same-origin Dialog iframe can activate the existing grant without another click. A protocol-denied decision leaves cookies blocked. The Storage Access methods were not mocked.

CDP supplied the permission decision; manual browser-prompt acceptance was not tested. Safari/Firefox, deployed login sessions, and real Stripe checkout still need staging validation. The first-party login cookie must exist, browsers may show a permission confirmation, and permission lifetime is browser-managed. Fallbacks remain for missing sessions or denied/unsupported access.

Deploy with platform #408. The companion changes measure the entire padded content root and render the compact authorization CTA.


yiw190:
**Update (cd134abc7): no more Dialog; the CTA iframe expands in place.**

Why: the Dialog mounted a second iframe (new document) that had to re-request storage access. On Safari that was another click plus another browser prompt.

- On `open-billing` the launcher frame takes the full card row and honours `resize` (clamped). Dismiss moves below the panel.
- Frame is keyed by view (not CTA kind) and its `src` is frozen per attempt, so `add_card → topup` after saving a card keeps the same iframe document and its grant.
- Card removal + toast on `billing.cleared` unchanged.

Pairs with SkillfulAgents/platform#408 (ac3d7dcc). 49 card tests, typecheck (src) and lint pass.

src/renderer/components/provider-error/platform-paywall-card.tsx
```diff
@@ -102,25 +158,91 @@ function PaywallActions({
...
   )
+  // Electron keeps the system-browser hand-off. Web on a cloud workspace (the
+  // only place the platform will frame its billing page) embeds it in-app.
+  const inApp = !isElectron() && platformAuth?.platformControlled === true
+  const view = billing.cta ? EMBED_VIEW[billing.cta.kind] : undefined
+  const embedded = inApp && view !== undefined
+  const { recheck } = billing
+  const handleBillingUpdated = useCallback(() => {
+    billingChanged.current = true
+    recheck()
```
iddogino, line 242: [P2] Preserve recovery when saving auto-refill fails

The companion platform #408 top-up flow emits `billing-updated` after a successful charge even when saving auto-refill fails, leaving an error and a `Save auto-refill` retry button inside the iframe. This recheck obtains an allowed billing snapshot, so the `billing.cleared` early return removes the entire iframe and the success toast replaces that recovery UI. The user can therefore believe the operation succeeded while auto-refill remains unsaved. Please coordinate completion/error state with the platform and keep the iframe available until pending settings errors are handled; an allowed billing snapshot alone is insufficient to close it.

src/renderer/components/provider-error/billing-embed-frame.tsx
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  // Same iframe element in both layouts; only the box around it changes.
+  if (launcher) return (
+    <div className={cn('flex min-w-0 flex-col gap-1', expanded || stretch ? 'w-full items-stretch' : 'items-end')}>
+      {!onHintChange && (hint || failure) && <p className="max-w-xs text-right text-xs text-muted-foreground" data-testid="billing-cta-hint">{failure ? FAILURE_MESSAGE[failure] : hint}</p>}
+      {failure ? (
+        <Button size="sm" title={FAILURE_MESSAGE[failure]} disabled={!fallbackHref} onClick={openFallback}>Open billing in a new tab</Button>
+      ) : (
+        <div className={cn('relative', expanded || stretch ? 'w-full' : 'shrink-0')} style={expanded ? { height } : undefined} data-testid="billing-cta-body" data-expanded={expanded}>
```
iddogino: [P2] Limit expanded launcher height to the available viewport

The expanded launcher uses the reported height (up to 640px) without a viewport cap; the viewport limit below only applies to the non-launcher branch. This card lives in the bottom-pinned SessionThread footer. In Chromium at 1024×600 with a 640px height report, the iframe starts 77px above the viewport even without the app header; at 844×390 it starts 287px above it. The amount and close controls at the top become unreachable, and scrolling the transcript cannot move the pinned footer. Please cap the expanded launcher to the available viewport height and let its content scroll within the iframe.

src/renderer/components/provider-error/platform-paywall-card.tsx
```diff
@@ -102,25 +158,91 @@ function PaywallActions({
...
+          {PRO_BENEFITS.map((benefit) => (
+            <li key={benefit} className="flex items-start gap-2">
+              <Check className="mt-0.5 h-3.5 w-3.5 shrink-0 text-brand" aria-hidden="true" />
+              {benefit}
+            </li>
+          ))}
+        </ul>
+        {hint && <p className="mt-3 text-[11px] leading-4 text-muted-foreground" data-testid="billing-cta-hint">{hint}</p>}
+      </div>
+      <div className="flex min-w-[200px] shrink-0 flex-col items-start justify-center gap-3 self-stretch border-l border-border/70 py-1 pl-10 pr-3">
```
iddogino: [P2] Stack subscription columns on narrow screens

The unwrapped flex row, 200px non-shrinking price column, gap, and card padding leave only 69px for the benefits at a 375px viewport, and 14px at 320px. Chromium checks of the rendered component show overflowing text; at 320×568 the card is 682px tall and its heading is above the viewport in the pinned composer footer. Please switch to a stacked layout at narrow widths and relax the desktop-only minimum width/padding so the subscription explanation and action remain readable and reachable on phones.

iddogino (changes requested):
Requesting changes for three issues: the host removes auto-refill failure recovery after a successful charge, expanded billing controls can become unreachable in short windows, and the subscription layout breaks on narrow phones. Details are in the inline comments.

Validation: 107 focused tests passed. Changed implementation lint reported no errors and one warning. Chromium layout checks used the rendered PR components with platform iframe content mocked; no real Stripe transaction was performed.

src/renderer/components/provider-error/platform-paywall-card.tsx
yiw190: Fixed in 9759484ea: `SubscribeBody` is `flex-col` below `sm` and the price column becomes a full-width block with a top border; `min-w-[200px]`, `shrink-0`, `border-l` and `pl-10` are now `sm:` only, and the card padding is `px-4` on narrow widths. Test asserts the responsive classes.

src/renderer/components/provider-error/billing-embed-frame.tsx
yiw190: Fixed in 9759484ea: the expanded launcher body now has `maxHeight: calc(100dvh - 160px)` alongside the reported height, so the iframe shrinks to the viewport and the embed document scrolls inside it (the embed page sets no `overflow: hidden`). Test covers the cap being set only while expanded.

src/renderer/components/provider-error/platform-paywall-card.tsx
yiw190: Fixed in coordination with SkillfulAgents/platform#430. The platform now sends `billing-updated` with `pending: true` when the charge succeeded but the auto-refill save failed, and settles with a plain `billing-updated` (retry succeeded) or `close` (new `Skip for now`). Host side here: `readEmbedMessage` reads `pending`; while pending the card skips its `cleared` early return so the composer unblocks but the iframe with the retry UI stays mounted, and the success toast / `paywall_cleared` are held until the settle message. Two tests cover the hold-then-settle and hold-then-close paths (both fail without the card change). Hosts on the old protocol are unaffected: the field is ignored.

src/renderer/components/provider-error/platform-paywall-card.tsx
```diff
@@ -102,25 +158,102 @@ function PaywallActions({
...
   )
+  // Electron keeps the system-browser hand-off. Web on a cloud workspace (the
+  // only place the platform will frame its billing page) embeds it in-app.
+  const inApp = !isElectron() && platformAuth?.platformControlled === true
+  const view = billing.cta ? EMBED_VIEW[billing.cta.kind] : undefined
+  const embedded = inApp && view !== undefined
+  const { recheck } = billing
+  const handleBillingUpdated = useCallback(({ pending }: { pending: boolean }) => {
+    billingChanged.current = true
+    setPendingSettings(pending)
```
iddogino: [P2] Hold the iframe before the charge becomes visible to billing polls

`pendingSettings` is only set when the platform sends `billing-updated`, but the merged platform #430 calls `payment.pay()`, then awaits `saveAutoReload()`, and only afterwards calls `onSuccess()` to send `pending: true` on failure. During that interval the existing five-second billing poll (or focus refresh) can already return `access.allowed: true`. Because the hold is still false, line 270 unmounts the iframe before it can report the failed settings save, losing the recovery UI again. The new tests always deliver the pending message before the allowed snapshot and therefore miss this ordering.

Reproduced with the existing test harness: expand the iframe, change the mocked billing result to ALLOWED without sending a completion message (representing a completed charge with a settings request still outstanding), advance `PAYWALL_RECHECK_INTERVAL_MS`, and assert the same iframe remains mounted. That assertion fails on this revision.

Please coordinate a hold that starts before the purchase/settings operation can affect billing, or otherwise prevent an open transaction panel from being removed by background refreshes until its completion state is known. Cover the allowed-poll-before-settings-result ordering, including a later settings failure.

iddogino (changes requested):
Re-reviewed 3224ba9ea. The two layout findings are addressed: Chromium checks confirm the expanded top-up iframe fits at 1024×600 and 844×390, and the subscription card remains readable at 320px, 375px, and 390px widths.

The auto-refill recovery hold works after `pending: true` arrives, but there is still a race before that notification: an automatic billing refresh can clear the card while the platform is awaiting the settings save. See the inline finding.

Validation: all 109 focused tests pass, and lint of the two revised implementation files passes. An additional targeted regression test fails because the iframe is removed when the five-second poll observes allowed access before the platform reports its settings result. Browser checks used the rendered components with iframe content mocked; no real Stripe transaction was performed.

src/renderer/components/provider-error/platform-paywall-card.tsx
yiw190: Confirmed, thanks. Fixed in 637eb3b5f: the hold now starts on open-billing instead of on a pending message, so a poll/focus refresh can't unmount an open panel. Only a settled billing-updated or close releases it; pending: true keeps holding.

Added tests for allowed-poll-first → late pending: true → settled update, plus the focus-refresh and close variants. All fail on the previous revision and pass now.

src/renderer/components/provider-error/platform-paywall-card.tsx
```diff
@@ -102,25 +158,107 @@ function PaywallActions({
...
   )
+  // Electron keeps the system-browser hand-off. Web on a cloud workspace (the
+  // only place the platform will frame its billing page) embeds it in-app.
+  const inApp = !isElectron() && platformAuth?.platformControlled === true
+  const view = billing.cta ? EMBED_VIEW[billing.cta.kind] : undefined
+  const embedded = inApp && view !== undefined
+  const { recheck } = billing
+  const handleBillingUpdated = useCallback(({ pending }: { pending: boolean }) => {
+    billingChanged.current = true
+    setPanelHold(pending)
```
iddogino, line 241: [P2] Preserve the purchase hold through the card-setup notification

The current platform `EmbedBillingCta` completion effect calls `openPanel()` and then `notifyParent("billing-updated")` when missing-card setup finishes. That delivers `open-billing` followed by a plain `billing-updated`, so this unconditional `setPanelHold(pending)` immediately clears the hold that `expand()` just established. The top-up panel remains open because access is still denied. When the user subsequently clicks Purchase and save, no new `open-billing` is emitted; a billing poll can again see the charge and unmount the iframe while `saveAutoReload()` is outstanding, losing a later settings failure/retry UI.

Reproduced on this revision with a focused test: start with `hasPaymentMethod: false`; switch the snapshot to a saved card with access still denied; deliver `open-billing` then plain `billing-updated`; wait for the refetch and verify the same expanded iframe remains; switch the snapshot to ALLOWED and advance `PAYWALL_RECHECK_INTERVAL_MS` without a purchase completion message. The assertion that the iframe remains mounted fails.

Please distinguish the card-saved update from final purchase completion, or re-establish the hold before every purchase. Add this card-setup → open panel → card-saved update → purchase → allowed poll → late settings-result sequence to the regression coverage.

iddogino (changes requested):
Final check of 637eb3b5f: the existing-card polling/focus race is fixed, and all 112 focused tests plus lint of the revised implementation pass. The previously verified layout fixes remain unchanged. One missing-card variant of the same race remains; a targeted test using the current platform message sequence reproduces it. Details below. No real Stripe transaction was performed.

src/renderer/components/provider-error/platform-paywall-card.tsx
yiw190: Fixed via platform #434: the card-saved completion now sends billing-updated { pending: true }, so the hold from open-billing survives it; only the purchase (or a no-panel upgrade) sends the plain completion. Host needed no logic change; 8f92133aa adds your exact sequence as a regression test (fails with a plain card-saved update, passes with pending: true). Old hosts treat pending: true as a normal update, so deploy order doesn't matter.

https://github.com/SkillfulAgents/platform/pull/434/changes 
I can deploy platform side after you think both PRs are valid.

iddogino (approved):
Re-reviewed 8f92133aa together with the protocol fix in SkillfulAgents/platform#434 at 1d0734d6. No remaining findings from this review.

The card-saved update now carries pending: true, so it preserves the hold established by open-billing through the subsequent purchase and settings save. The previously failing independent reproducer now passes with that payload. The host regression also covers a later allowed billing poll, pending settings failure, and settled completion. The earlier layout fixes remain unchanged and were verified in Chromium during the previous review.

Validation: all 113 focused SuperAgent tests pass; the independent missing-card reproducer passes; lint of the reviewed implementation passes. Platform #434 CI checks also pass. I inspected the companion diff but did not rerun its tests locally or perform a real Stripe transaction.

Deployment dependency: platform #434 is still open at review time and must also be merged and deployed for the missing-card race fix to take effect. PR #994 alone does not change the platform notification payload.

PR #1008, opened 2026-09-09
fix(automation): release settled streams and safely roll back failed sends

yiw190:
## Problem

Completed chat conversations can retain a host-to-container WebSocket indefinitely. A failed follow-up also has a race: when the previous turn has emitted its result but has not reached final idle, provisional activity can clear that result and leave the session marked active after the send fails. X-agent follow-ups have a related missing rollback. Reconnecting also needs a replay boundary: WebSocket open can precede the previous turn’s replayed result and idle, which must not settle a newly sent turn.

This generalizes connection cleanup across unpromoted chat, scheduled, webhook, and x-agent sessions. Related to [SUP-788](https://linear.app/datawizz/issue/SUP-788/microvm-ingress-throttling-slack-session-failures-and-false-unhealthy); the exact AWS quota behind the reported ingress 429 remains unverified.

## Change

- Use one release policy for unpromoted automated sessions. Release the transport after authoritative settlement of the current result, or a matching process eviction, once active work, pending sends, input waits, background work, and recovery no longer hold it open. Interactive and promoted sessions keep their existing behavior.
- Preserve chat state and request registrations when detaching its transport, so replies reconnect to the same conversation. Other automation types retain their existing full cleanup on release.
- Route chat replies, scheduled wake delivery, and x-agent follow-ups through a shared send helper. Reserve and reconnect the transport before marking the next turn active, serialize delivery attempts, and roll back failed sends without overwriting accepted output or newer runtime state.
- Resolve stream readiness only on the guest’s existing acknowledgement, sent after listener attachment and terminal replay. Sends also wait for an attachment already in progress; cleanup waits until attachment finishes. Closing, cancelling, replacing, or stopping an attachment rejects its pending readiness promise.
- Mark an already-started initial turn active before attachment, and attach chat forwarding before consuming replay, so fast first turns can finish normally.
- Handle final idle arriving during a failed follow-up, delayed session metadata, queued send failures, reconnect failures, and background wake grace.
- Keep the guest process-eviction fallback: emit the process instance only after a successful idle stop, and ignore stale process identities on the host. Existing idle thresholds remain unchanged.

## Expected behavior

A completed unpromoted automation releases its connection when settlement is confirmed. A subsequent chat reply reconnects before sending to the same session. Failed follow-ups preserve the previous turn's result and allow final idle to settle it. Pending work and promotion prevent automatic release.

## Validation

- 746 focused host tests passed across the persister, stream client, request lifecycle, chat integrations, scheduler, triggers, and x-agent routes.
- 133 guest lifecycle tests passed during review; guest code is unchanged by the follow-up fix.
- The original failed-follow-up reproduction passes. A real host WebSocket connected to a local server reproduced delayed-replay detachment for chat, scheduled, webhook, and x-agent sessions before the fix; all four now stay subscribed through the new reply.
- Host TypeScript checking, ESLint on changed TypeScript files, and `git diff --check` passed.
- Coverage includes all automation categories, promotion, delayed metadata, repeated reconnects, overlapping failures, accepted output, recovery, and background work.
- Live Slack/Telegram/iMessage delivery and the production AWS quota were not validated.

## Deployment / scope

Settlement cleanup uses existing authoritative runtime idle events. Replay-aware readiness uses an existing acknowledgement and requires no additional guest protocol change. The process-eviction fallback additionally requires the updated agent-container image. No production deployment, quota adjustment, retry-policy change, health-probe change, or UI change is included.

### DIFF
--- a/agent-container/src/session-manager-idle-eviction.test.ts
+++ b/agent-container/src/session-manager-idle-eviction.test.ts
@@ -176,6 +176,43 @@ describe('SessionManager idle eviction', () => {
     expect(mockReleaseBrowserLock).toHaveBeenCalledWith(id)
   })
 
+  it('announces eviction only after the process stops and before a racing send resumes', async () => {
+    const { id, proc } = await createIdleSession()
+    const processInstance = manager.getProcessInstanceId(id)
+    const events: unknown[] = []
+    manager.subscribe(id, (message) => events.push(message))
+    let finishStop!: () => void
+    const stopStarted = vi.spyOn(proc, 'stop').mockImplementation(async () => {
+      await new Promise<void>((resolve) => { finishStop = resolve })
+      proc.running = false
+    })
+    await pastThreshold()
+    const eviction = manager.evictIdleSessions()
+    expect(stopStarted).toHaveBeenCalledOnce()
+    expect(events).toEqual([])
+    const sentBefore = proc.sendMessageCalls
+    const send = manager.sendMessage(id, 'continue the same conversation')
+    expect(proc.sendMessageCalls).toBe(sentBefore)
+    finishStop()
+    await Promise.all([eviction, send])
+    expect(events[0]).toMatchObject({ type: 'system', subtype: 'process_evicted', process_instance: processInstance })
+    expect(proc.sendMessageCalls).toBe(sentBefore + 1)
+    expect(proc.sentContents.at(-1)).toBe('continue the same conversation')
+    expect(manager.hasActiveSession(id)).toBe(true)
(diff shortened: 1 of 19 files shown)


yiw190:
There’s a reconnect race here: ready resolves on WebSocket open, not after replay finishes. If the previous turn’s replayed result + idle arrive after withSessionSend() marks the new turn active, they incorrectly settle it. Once the send completes, cleanup closes the stream, potentially losing the new reply in chat delivery.

The existing test assumes replay arrives before ready. I added a delayed-replay test that reproduces the incorrect detachment.

PR #1025, opened 2026-09-10
End the platform login launch when the stored token changes (SUP-807)

JeremyBischoff:
In a browser window, clicking Reconnect on the Account tab left the button on "Opening browser…" forever, and adding an access key afterwards did not clear it. The platform login callback is a deep link that lands in the desktop app, so the launch state in the web app never heard back. Now any change to the stored token's timestamp ends the launch in a browser window, while on desktop the callback keeps ownership as before. A login window closed by mistake had no way back either, on desktop or web, so after 10 seconds the same Cancel link the OAuth connection flows use appears next to the button and ends the launch so Reconnect can be clicked again.

Closes https://linear.app/datawizz/issue/SUP-807/reconnect-from-a-browser-window-leaves-the-account-tab-on-opening

How it works:

- The connect hook watches the platform auth status query's `updatedAt`. When it changes after mount, the launching flag clears.
- The effect returns early when the desktop callback bridge exists, so a metadata refresh mid-login cannot end a launch the callback owns.
- The connect hook owns the cancel: it holds the login popup, runs the existing `useDelayedOAuthAbort` timer off the launching flag, and exposes `cancelConnect` and `canCancel`. Cancel closes the popup and clears the launch. A launch request that resolves or fails after Cancel is ignored, so it cannot navigate or clobber a retry.
- The Account tab (both the empty state and the Reconnect row) and the onboarding wizard render the shared `OAuthFlowCancel` link.
- Hook tests: a token change ends the launch; Cancel appears at 10 s (off at 9,999 ms) and ends the launch; a cancelled request that fails late neither navigates nor errors.

## Screenshots

| Light | Dark |
| --- | --- |
| Account tab (connected), Cancel after 10 s (light) | Account tab (connected), Cancel after 10 s (dark) |
| Account tab, Cancel after 10 s (light) | Account tab, Cancel after 10 s (dark) |
| Onboarding wizard, Cancel after 10 s (light) | Onboarding wizard, Cancel after 10 s (dark) |

### DIFF
--- a/src/renderer/components/settings/platform-tab.tsx
+++ b/src/renderer/components/settings/platform-tab.tsx
@@ -8,6 +8,7 @@ import { Popover, PopoverContent, PopoverTrigger } from '@renderer/components/ui
 import { Progress } from '@renderer/components/ui/progress'
 import { ErrorBoundary } from '@renderer/components/ui/error-boundary'
 import { RequestError } from '@renderer/components/messages/request-error'
+import { OAuthFlowCancel } from '@renderer/components/connections/oauth-flow-cancel'
 import { ProfileSection } from './profile-section'
 import { useUser } from '@renderer/context/user-context'
 import { usePlatformConnect, useSavePlatformAccessKey } from '@renderer/hooks/use-platform-auth'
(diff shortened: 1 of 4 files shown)


iddogino:
@JeremyBischoff there's still no retry path. If I mistakingly close the browser window it opened - it'll still be stuck on opening forever. 

We have a pattern in OAuth connection flows where after a few seconds we'll drop in a "cancel" button - what do you think about generalizing that into a reusable pattern and using it here?

JeremyBischoff:
Done in `d0e0912` - the pattern already exists as two shared pieces (`useDelayedOAuthAbort` + `OAuthFlowCancel`), so this wires the platform login onto them: after 10 s a Cancel appears on the Account tab and the wizard, closes the window and ends the launch so Reconnect works again.

Mapping this turned up an opportunity for a bigger refactor. The timer and the link are shared, but the popup + cancel closure around them (hold the popup, close it, clear the pending flag, ignore late responses) is hand-rolled five times: the Composio reconnect hook, the add-integration dialog twice, the integration row's MCP re-auth, and now the platform hook. The three in-chat connect cards (connect account, MCP re-auth, connect MCP server) open a window with the same inline code and have no Cancel at all. One hook for the closure and the timer would collapse the five and give the cards Cancel, and since it would no longer be OAuth-specific the two pieces should be renamed with it. I can do this as a follow-up PR if that sounds right.

PR #1029, opened 2026-09-10
feat(accounts): X as a connected account on Gamut's Composio (SUP-809)

JeremyBischoff:
Adds X as a connected account on hosts that run on Gamut's Composio, so an agent can act on a member's own X data (bookmarks, home timeline, DMs, posting) with the member's token while every call rides the platform hop and gets metered there. Composio dropped managed Twitter auth, so X runs on Gamut's own OAuth app behind a custom Composio auth config. That config returns full tokens, which would flip the host into calling X directly and unmetered, so the platform redacts the X token on its shared key and the host's existing probe keeps X on the hop, the same path Slack and Gmail take. The host hides X on hosts with a personal Composio key, where Gamut's auth config does not exist. The platform half (route table, pricing, metering, redaction) is SkillfulAgents/platform#416 and deploys first, or X calls cross the hop unmetered. A host that read an X token before that deploy keeps calling X directly for up to five minutes, the token cache lifetime.

Closes https://linear.app/datawizz/issue/SUP-809/x-as-a-connected-account-on-gamuts-composio

- Catalog entry `twitter` with a `platformOnly` mark, which lists X in the connect dialog only on Gamut's Composio. The provider has no X-specific path: the platform's redaction lands X in proxy mode through the existing token probe.
- Scope map rows mirror the platform's user-lane table, one distinguishing scope per row because the policy resolver takes the most permissive decision across matched scopes. A snapshot test pins the row set to a copy of the platform table, since the repos cannot import each other. A platform-side change needs a manual refresh of that copy.
- The container offers X in the service search and the prompt only when the host set `COMPOSIO_PLATFORM_MODE`, the same gate the Composio trigger tools use. Platform auth alone is not enough: a host with a personal Composio key sends every Composio call to its own org.
- `agent-container/docs/x.md` carries both lanes: the full call table with a price per row, the JSON-only chunked media upload at 512 KB per segment, and the error meanings. The prompt states the $0.200 URL-post price and asks first. A routing rule in the guide picks the built-in reads for public data with no account, the account for the member's own data or writes, and the account for everything once one is connected, since its rate limit is per user rather than the shared app pool.

Known from X's published spec, left for the platform table. Recent post counts and trends by location are app-only, so they fail on the user lane. The by-username posts path is absent from the spec. The hide-replies scope in the connect list has no route.


agent-container/src/system-prompt.md
```diff
@@ -511,6 +519,9 @@ Before video, music, 3D, talking-head, or voice cloning, tell the user the cost
 ## Built-in X reads
 
 Search recent public X (Twitter) posts and read public profiles, timelines, mentions, and follower lists through the platform without asking the user for an X account or API key. Before using this capability, read `/opt/gamut/docs/x.md`. Every post and user object returned costs money, so request only what the task needs. Never invent an X endpoint; the guide's table is the only allowlist. Before followers or following, tell the user it is $0.01 per person, up to $1 per page, and get an OK.
+<%#composioTriggers%>
```
iddogino: @JeremyBischoff is `composioTriggers` the right filter? I though that was for Composio managed web hooks?

agent-container/src/tools/search-connected-account-services.ts
```diff
@@ -58,6 +58,10 @@ export const SERVICES: ServiceInfo[] = [
   // Social Media
   { slug: 'linkedin', displayName: 'LinkedIn', category: 'Social Media', description: 'Professional networking' },
   { slug: 'instagram', displayName: 'Instagram', category: 'Social Media', description: 'Photo and video sharing' },
+  // X exists only through Gamut's Composio; the host sets this env at container start.
+  ...(process.env.COMPOSIO_PLATFORM_MODE === 'true'
```
iddogino: Nit - create a separate list of `GAMUT_PLATFORM_ACCOUNTS`, then append into the main list if connected

iddogino (changes requested):

agent-container/src/system-prompt.md
JeremyBischoff: Same env under an old name: `composioTriggers` is just `COMPOSIO_PLATFORM_MODE`, which the host sets when connected accounts run through Gamut's Composio. The trigger tools were its first consumer, so it got their name. X needs the same gate because its auth config lives on Gamut's Composio, not on a personal key. Split it in dcf821b9c: `composioPlatform` reads the env, `composioTriggers` and `composioX` both derive from it, and the X blocks read their own flag.

agent-container/src/tools/search-connected-account-services.ts
JeremyBischoff: Done in dcf821b9c - X lives in `GAMUT_PLATFORM_ACCOUNTS`, appended to the catalog when the host runs on Gamut's Composio.

agent-container/src/system-prompt.md
JeremyBischoff: Renamed in 0919d3582 after Slack: `composioX` is now `platformAccounts`, one switch for the blocks that only apply when the host has platform-only accounts, so no account name reaches the prompt vars. The shared runtime catalog is the follow-up.

iddogino (approved):

PR #1031, opened 2026-09-10
Route every agent operation through agentRegistry.get(slug) and complete the actor contract

iddogino:
## Summary

First PR of the agent-actor refactor: **routing only**. Adds `src/shared/lib/agent-actor/` and points every consumer outside the container layer at `agentRegistry.get(slug)`. One actor method is one underlying call; routes still compose the same calls they composed before. No state moves and the API contract is unchanged, with one deliberate exception noted under "Behaviour change".

This is the first step of the remote-contract unit: the interface has to be complete and honest enough that an actor whose container and files live elsewhere can implement it. Spec: `agent-actor-remote-contract-spec.md` (repo root, untracked); design: `agent-actor-architecture.md`.

No UI change.

## What is in here

- **`agent-actor/`**: `AgentRegistry` + `AgentActor` interfaces (`container`, `sessions`, `messages`, `inputs`, `usage`, `files` groups), `LocalAgentActor` with 1:1 passthroughs to the container manager, message persister, session service, usage loaders, and the user-input / review / computer-use / MCP re-auth registries, an in-process registry keyed by `agent.slug`, and `files.workspacePath()` as the only file op for now. Every op reads its dependency at call time, so tests that mock one leaf module keep intercepting exactly the call they intercepted before.
- **Every stored-session read and write goes through the actor.** `sessionExists`, `getSessionMessagesDelta`, `readSessionMetadata` and the rest are no longer called directly from routes, services, or sweeps (59 call sites across `agents.ts`, `x-agent.ts`, the auto-delete monitor, widget repair, notifications, and the cross-agent readers). `session-service` is fenced.
- **Cross-agent readers fan out through the actor, server-side.** Home graph, home-card health, usage, auto-delete, and the notification summaries call each agent's handle; no projection tables. Usage gets `actor.usage.daily()` and `actor.sessions.usage(id)`.
- **Input ops are scoped to the handle's agent.** `register` stamps the actor's slug onto the request scope; `get`, `claim`, `releaseClaim`, `resolve`, `recentResolution`, and `enrich` treat another agent's request as not found. Before this a handle for agent A could read, claim, and settle agent B's request (routes checked, the actor did not).
- **Connection env sync** is `actor.container.syncConnectionEnvironment(kind)`; the fan-out over assigned agents moved out of the container layer into `services/connection-sync-service.ts`.
- **One WebSocket op.** `container.openWebSocket(path, { search, protocols, headers })` replaces `webSocketBaseUrl` + `hostAuthHeaders` in both stream proxies; the actor targets the container's port and adds its auth headers last. `hostBridgeIp` and `probeHostPort` stay as host capabilities (the container reaching Chrome on this machine), not transport.
- **Fence**: `eslint-rules/no-agent-internals-outside-actor.js` (error) blocks imports of the runtime singletons, the transcript modules (`session-service`, `session-transcript-append`, `session-summary-cache`, `session-media`, `workflow-tree`, `connection-runtime-sync`), the agent path helpers, and the path-taking usage loaders, including namespace imports and re-exports of fenced names. Type-only imports are not fenced. The allowlist `eslint-rules/agent-internals-allowlist.json` only shrinks: an unlisted import errors, and a listed import that no longer exists errors too.
- **Meter**: `eslint-rules/agent-actor-escape-hatch-tally.js` (warn) counts `container.fetch` and `files.workspacePath` per file.

## Burn-down baseline

Allowlist: 54 entries in 36 files, by the PR that removes them:

| Goes in | Files |
|---|---|
| App-level persister hub (`broadcastGlobal`, `addGlobalNotificationClient`, `broadcastSessionUpdate`) — decide in B | agent-bootstrap, agents, browser, notifications, scheduled-tasks, x-agent, chat-integration-manager, client-broadcast-channel, platform-notifications-manager, wake-delivery, widget-refresh, widget-repair, account-reauth-manager |
| B — `ContainerHost` split; `completeMcp` / `rejectAll` | startup, settings, runtime-status, debug, mcp-proxy, remote-mcps |
| C — operation-shaped `FileOps` | chrome-provider (downloads dir), mount-service, artifact-service |
| D — transcripts behind the actor | agents (session-media, session-summary-cache, `getSessionJsonlPath`, `getAgentSessionsDir`, remaining session-service helpers), workflows, x-agent-chat, session-fork-service, and the cross-agent session helpers in scheduled-tasks, webhook-triggers, task-scheduler, startup |
| E — config docs | agent-service, agent-preferences-service, agent-template-service, secrets-service, agent-hooks-service, skillset-service, skillset-reconcile, skills/index |
| F — graph edges as policy projection | home-graph-service (`getAgentSessionMetadataPath` stat cache) |

Escape-hatch tally: 42 in `agents.ts` (`container.fetch` ×30, `files.workspacePath` ×12), 10 across browser, chat-integration-manager, agent-service, widget-refresh, widget-service.

## Behaviour change

One, and it is the point of the input-ownership guard: a decision route (`answer-question`, `provide-*`, `capability-review`, `run-script`, `computer-use`, `complete-browser-input`) probed with a request id that belongs to **another agent**, or one with no agent in its scope, now answers the outcome-less already-settled shape (`200 { success: true, alreadySettled: true }`) instead of `404`. The actor cannot see a foreign request, so it is indistinguishable from an unknown id, and unknown ids already got this shape so stale cards dismiss themselves. Nothing runs, the foreign request stays open, and no outcome is disclosed; the twelve gate tests assert exactly that. A 404 here told a probing caller that the id exists for someone else, so the new answer discloses less.

## Deviations from the spec

- **Mocks changed in existing test files.** Routes used the client that `ensureRunning()` returned, and those tests mock `ensureRunning` without `getClient`. The actor drops that return value and reaches the client through `getClient`, so the mocks now expose one. Three tests moved a `vi.mock` path from `connection-runtime-sync` to `connection-sync-service` with the helpers. No assertions changed except the twelve gate tests above.
- Passthroughs forward only the arguments they were given (`(...args) =>`); otherwise `toHaveBeenCalledWith(slug)` sees a trailing `undefined`.
- Extra actor methods beyond the spec's list: the two env-sync wrappers, `syncConnectionEnvironment`, `openWebSocket`, `usage.daily`, `sessions.usage`, and `withSend`, `stopTask`, `turnGeneration`, `isWaitingBackground`, `hasOnlyUntrackedBackgroundWork` for calls that landed on main during the rebase.
- **`/files/*` body not moved into `local-file-ops.ts`.** It is a single download route with range, HEAD, 416 and 206 handling interleaved with the fs work; no list/write/delete/mkdir routes exist. That reshape is PR C (operation-shaped FileOps).

## Verification

- `npm run typecheck`: clean.
- `npm run lint`: 0 errors, 21 warnings (pre-existing plus the tally meter).
- `npx vitest run`: 712 files, 11821 tests pass.
- Mocked E2E (`E2E_MOCK=true E2E_PORT=3177`): 340 pass, 24 skipped, 10 fail in the 15.1-minute run under 4 workers. All 10 were timeouts under load (browser-stream ×2, proxy-review, search, thinking-display, home-graph ×5, the last as pointer-interception and outside-viewport retries on nodes that had rendered) and all 10 pass when re-run together in isolation (12 of 12 in 27 s, the home-graph cases in about a second each). Same pattern as the two earlier full runs on this branch.
- Earlier CI on the routing-only version: `e2e-tests` failed once on `template-onboarding.spec.ts:14`, a pre-existing flake that also fails on main runs 34410741946 and 34403297402 and passes 3 of 3 locally; the re-run passed.

## Review fixes

- Rebased onto main. The two connection-replacement routes main added since this branch was cut (`account-reauth.ts`, `mcp-reauth.ts`) go through the actor; the MCP replacement gets the one passthrough the contract was missing, `inputs.mcpReauth.replace`.
- `perf/harness.ts` is seeded into the fence allowlist. `npm run lint` covers `perf/`, so CI's unit-tests job was failing at the lint step and vitest never ran in CI for this stack; the Verification section above used a narrower glob. `npm run lint` is now the gate: 0 errors.
- The auth E2E for a cross-agent reauth dismiss asserted the old 404. Another agent's route cannot see the request at all now, so it answers the settled shape unknown ids always got (200, `alreadySettled`, no outcome) and leaves the request open; the spec pins that, and the owner's own dismiss proves the request was untouched. The shared-connections auth suite passes 13/13 locally.


src/api/routes/activity.ts
```diff
@@ -22,7 +22,7 @@ activityRouter.use('*', Authenticated())
 // actually processing — a persisted 'running' with an inactive session is a
 // dead or stopped run and must downgrade to failed instead of pulsing forever.
 const isSessionLive = (agentSlug: string, sessionId: string) =>
-  messagePersister.isSessionActive(agentSlug, sessionId)
+  agentRegistry.get(agentSlug).sessions.isActive(sessionId)
```
iddogino: shouldn't this be something like:

`agentRegistry.get(agentSlug).session(sessionId).isActive()`

src/api/routes/agents.ts
```diff
@@ -758,13 +749,13 @@ async function enrichAgentsWithSummary(
...
-      const hasAgentLevelReviews = reviewManager.getPendingReviewsForAgent(agent.slug).length > 0
+      const hasAgentLevelReviews = agentRegistry.get(agent.slug).inputs.reviews.pending().length > 0
       for (const sessionId of sessionSummary.sessionIds) {
-        const isActive = messagePersister.isSessionActive(agent.slug, sessionId)
+        const isActive = agentRegistry.get(agent.slug).sessions.isActive(sessionId)
         if (isActive) {
           hasActiveSessions = true
         }
-        if (messagePersister.isSessionAwaitingInput(agent.slug, sessionId)) {
+        if (agentRegistry.get(agent.slug).sessions.isAwaitingInput(sessionId)) {
```
iddogino: same here, in theory this could be `agentRegistry.get(agent.slug).session(sessionId).isAwaitingInput()`

iddogino (commented):
👍

yiw190 (approved):

PR #1038, opened 2026-09-10
Dissolve ContainerManager into a per-agent ContainerRuntime and a ContainerHost

iddogino:
## Summary

PR B of the agent-actor refactor, stacked on #1031: **state moves, no behaviour change**. The container manager's seven per-agent maps become one `ContainerRuntime` object per agent, its host-level half becomes `ContainerHost`, and `ContainerManager` is deleted. The actor's container ops delegate to the agent's runtime; the four host-level callers (startup, settings, runtime status, debug) import `containerHost` from the actor package. The API contract is unchanged.

Spec: `agent-actor-remote-contract-spec.md` §4 B (repo root, untracked). The mailbox is not part of this PR or this unit.

No UI change.

## What is in here

- **`container/container-runtime.ts`** — `ContainerRuntime`, one per agent: client, cached status, start time, keep-alive, health warnings, stopping flag, in-flight start. `getClient`, `ensureRunning`, `stopContainer`, `restartContainer`, `syncAgentStatus`, `getCachedInfo`, `updateCachedStatus`, `markAsStopped`, `runHealthCheck`, `handleUnexpectedDeath`, `dispose`, `stopSync` — the manager's methods with the slug bound. Code moved verbatim; `this.clients.get(agentId)` became `this.client`.
- **`container/container-host.ts`** — `ContainerHost` + singleton `containerHost`: owns the runtimes (`runtime(slug)` creates on first use), readiness and image pulls, `initializeAgents`, `syncAllStatuses`, the status-sync and health-monitor loops, `stopAll` / `stopAllSync`, and `afterForceStop` (the fallout of force-killing the shared VM, which used to sit inside `stopContainer`). Host-level method names are unchanged; `removeClient` → `dropRuntime`, `clearClients` → `clearRuntimes`.
- **Actor**: `LocalActorDeps.containerHost` replaces `containerManager`; container ops call `deps.containerHost.runtime(slug).…`. `syncAgentConnectionEnvironment` takes the runtime as a parameter instead of importing the manager. `registry.all()` is renamed `running()` — it was always running agents only, and the old name invited cross-agent reads over it.
- **Host-level callers**: `startup.ts`, `settings.ts`, `runtime-status.ts`, `debug.ts` import `containerHost` from `@shared/lib/agent-actor`; `settings.ts` calls `agentRegistry.evictAll()` where it cleared the manager's clients. The persister's lazy import and the E2E mock browser scenario point at the host. Their allowlist entries are gone; `container-host` and `container-runtime` are fenced.
- **Test doubles** in `agent-actor/testing/`: `createFakeAgentActor` / `createFakeRegistry` (every method a `vi.fn()` on first access, overrides per group) for tests written against the actor contract, and `hostFromManagerMock`, which presents a manager-shaped mock as a host so the 42 existing test files that mocked the manager keep their assertions.

## Deviations from the spec

- **Tests re-targeted through an adapter, not rewritten.** The spec had every `vi.mock('…/container-manager')` become a mock of the registry with a fake actor. Those tests also mock `session-service`, the persister, and other leaf modules and assert on them, so a fake actor would have meant re-expressing every assertion. Instead each file swaps its mock to `container-host` wrapped in `hostFromManagerMock` (two lines per file) and keeps its assertions. The fake actor exists for new tests and for the conformance suite.
- **Conformance suite deferred to PR C.** Running `LocalAgentActor` and an in-memory fake through one suite needs operations with real semantics on both sides; the first such group is `FileOps` in C. B ships the fake and the registry double.
- **The manager's own tests** (75 + 23 + 4) were renamed with `git mv` to `container-host.test.ts`, `container-runtime-concurrency.test.ts`, `container-runtime.reserved-env.test.ts` and re-spelled (`containerManager.ensureRunning('x')` → `containerHost.runtime('x').ensureRunning()`); same expectations.

## Verification

- `npm run typecheck`: clean.
- `npm run lint`: 0 errors, 21 warnings (pre-existing plus the escape-hatch tally meter). Allowlist: 50 entries in 33 files (the four host-level `container-manager` entries are gone).
- `npx vitest run`: 713 files, 11825 tests pass. The 42 re-targeted test files and the manager's renamed 104 tests all pass with their assertions unchanged.
- Mocked E2E (`E2E_MOCK=true E2E_PORT=3177`): 341 pass, 24 skipped, 9 fail in the 14.0-minute run under 4 workers. All 9 were timeouts under load (activity-spark-charts, browser-stream ×2, capability-review, read-aloud-follow on WebKit, home-graph ×4); 8 pass when re-run together in isolation and the WebKit read-aloud case passes 3 of 3 when repeated alone. Same set and pattern as the runs on #1031.

## Review fixes

- A dropped runtime is inert. The stopping/starting guards moved from slug-keyed maps into per-runtime fields here, so a client the persister still held kept calling a runtime that `dispose()` had reset, and the delete route made it concrete by evicting before `deleteAgent` stopped the container: the dying container's socket error ran unexpected-death recovery mid-stop. `dispose()` now marks the runtime and its old client's callbacks do nothing; the delete route leaves stop-then-forget to `deleteAgent`. Pinned by `container-runtime-dispose.test.ts`.
- `connection-replacement.ts` (from main) asks the host for the agent's runtime; its tests and the two replacement route tests mock the host instead of the dissolved manager.

## Second review round

- Runner change during a start: the settings route clears every runtime when the container runner changes. A start already in flight finished on the runtime it began on, which the host had dropped, so the container came up running with nothing counting it as running or stopping it on quit (the old manager kept per-agent status maps, so the start's outcome landed somewhere). A runtime whose container is starting now stays known to the host and only loses its client, so the next one is built for the new runner. Pinned in `container-host.test.ts`, proven to fail without the change.


iddogino (commented):
👍

yiw190 (approved):

PR #1043, opened 2026-09-11
fix(browser): stop the default snapshot from silently hiding page text

iddogino:
## Why

Theme 1 of the transcript-mining report (`docs/browser-harness-friction-themes-2026-09.md`, 142/200 reviewed production sessions): the default `browser_snapshot` drops every static text node and says nothing about it. Agents concluded prices, results and error messages were not on the page and fell back to screenshots or `browser_eval` innerText scrapers. `fullText` was discovered late (call 284 of 377 in one session) and returned the raw unfiltered tree.

Root cause is in agent-browser 0.27.2 (`cli/src/native/snapshot.rs`): `-i` skips every ref-less node at render time and `-c` is a line filter keeping only lines with `ref=` or `: `. There is no flag combination for "interactive elements plus the page's text".

## What changed

- **One page probe per snapshot** (`PAGE_PROBE_SCRIPT` replaces `IFRAME_ENUM_SCRIPT`, same single eval round trip): iframes as before, plus visible text length, live-region text (`role=alert/status`, `aria-live`, `<output>`, visibility via `getClientRects` so fixed toasts count) and the first 240 chars of `<main>`.
- **Footer on the default view**: live regions, how much text was dropped, how the page starts, and that `fullText:true` reads it with identical refs. Silent in fullText mode and on pages with under 200 chars of text.
- **`fullText` is now compact**: `compactWithText()` keeps refs, StaticText, values and named images with their ancestors, collapses inline wrappers (`strong`/`emphasis`/…) so a sentence is one line, and drops text that repeats a sibling control's name or the parent's value. The test fixture is real `snapshot` output captured from agent-browser 0.27.2 inside the container image.
- **`browser_get_state`** takes `scope`/`fullText`/`includeUrls` and `screenshot:false`.
- Tool descriptions, the web-browser subagent prompt and `docs/browser-use.md` name `fullText` (with `scope`) as the way to read a page.

## Verified

- `vitest` on `snapshot-format.test.ts`: 25 tests, including the probe script run against a fake DOM and the captured-tree fixture.
- `tsc --noEmit` and `eslint` clean on the changed files.
- Probe script and all three snapshot modes exercised with the real CLI in `superagent-container:browser-stack-1001` against a checkout-like test page: probe returned the iframe, 225 text chars, both live regions (the hidden one excluded) and the preview.

Not done: no end-to-end run through the `/browser/snapshot` endpoint in a rebuilt image. The server glue is a few lines; happy to build and run it if you want that before merge.

No UI change.

### DIFF
--- a/agent-container/docs/browser-use.md
+++ b/agent-container/docs/browser-use.md
@@ -34,19 +34,23 @@ available.
 
 ## Observe Efficiently
 
-Use `browser_snapshot(interactive: true, compact: true)` for normal observation.
-It returns actionable refs such as `@e1`.
+Use `browser_snapshot()` for normal observation. It returns actionable refs
+such as `@e1` and, in the default interactive view, no static text at all. A
+footer reports how much page text was dropped and what the live regions
+(alerts, status, toasts) currently say.
 
 Useful snapshot options:
 
+- `fullText: true`: add the page's static text (prices, prose, validation
+  errors, table values) to the same compact tree. This is the way to read a
+  page; refs are identical in both views;
 - `scope`: restrict a large page to a form, dialog, or other CSS-selected
-  region without invalidating refs elsewhere;
-- `fullText: true`: include static text such as validation errors, prices,
-  instructions, and toasts;
+  region without invalidating refs elsewhere. Combine with `fullText`;
 - `includeUrls: true`: inline link URLs when labels are ambiguous.
 
 Use `browser_get_state` when the URL, screenshot, and accessibility snapshot
-are all useful together. Use `browser_screenshot` only for pixel-level facts
+are all useful together; it takes the same `scope`/`fullText` knobs and
+`screenshot: false` skips the image. Use `browser_screenshot` only for pixel-level facts
 that the accessibility tree cannot express, such as visual layout, charts,
 images, or color.
 
--- a/agent-container/src/server.ts
+++ b/agent-container/src/server.ts
@@ -815,7 +815,7 @@ import { prepareEvalScript, finalizeEvalOutput, evalErrorHint } from './eval-scr
 import { judgeSelectCommit, SELECT_COMMIT_SETTLE_MS } from './select-verify';
 import { resolveCommittedValue } from './field-value-readback';
 import { capBrowserOutput, redactCdpUrls, MAX_BROWSER_OUTPUT_CHARS, MAX_BROWSER_ERROR_CHARS } from './browser-output';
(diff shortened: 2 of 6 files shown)


iddogino:
**Replay of six mined production sessions with a real model** (Sonnet 5 as main and browser model; container built from the full stack #1043 → #1045 → #1047 → #1049 → #1051; same task prompts as the originals, delegated verbatim to the web-browser subagent). Originals for the two Beatport cases ran on Sonnet 4.6, the other four on Sonnet 5.

| Case | Theme | Original → replay: tool calls | snapshots | screenshots | evals (innerText scrapes) | duration |
|---|---|---|---|---|---|---|
| Beatport afro-house top 100 | 1 | 96 → 10 | 3 → 1 (fullText) | 2 → 0 | 48 → 6 | 583s → 67s |
| Beatport house top 100 | 2 | 92 → 20 | 10 → 2 | 5 → 1 | 53 → 12 | 865s → 120s |
| Meta Ad Library "Wispr Flow" | 1 | 115 → 58 | 20 → 18 (7 fullText) | 17 → 13 | 23 → 0 | 815s → 215s |
| Meta Ad Library 5 brands | 3 | 108 → 92 | 31 → 10 | 16 → 21 | 0 → 1 | 699s → 319s |
| absorbmore.com checkout | 3 | 127 → 46 | 11 → 5 | 28 → 10 | 21 → 4 | 1039s → 231s |
| Craigslist listings scan | 1 | 83 → 179 | 0 → 76 (56 fullText) | 0 → 0 | 46 → 0 | 463s → 569s |

What the lines did in practice: first snapshot with `fullText:true` on 4 of 6 cases (the originals discovered page text was reachable at call 90, or never); `browser_open` on Facebook reported "still loading" and the agent waited instead of looping; `Effect: dialog "Filters" opened`, `+13 interactive elements`, `alertdialog "Cookie consent" closed` were acted on without confirmation snapshots; one Beatport open hit a Cloudflare 403 and came back as `⚠ bot-block: Cloudflare` with `isError`.

Craigslist is the one case that got more expensive: the replay replaced 46 innerText evals with 76 snapshots (56 fullText) and searched each listing by title, where the original opened listing URLs directly and scraped text. Different strategy choice by the model rather than a harness regression, but worth watching in the next dump.

Things the runs exposed that this stack does not address: `scope:'[role="dialog"]'` failing right after the effect line announced a dialog (theme 8); an agent retrying a "covered by <div>" click three times (theme 5); the eval wrapper's "add `return`" note firing on every eval that did return a value (theme 4).

iddogino:

PR #1045, opened 2026-09-11
fix(browser): report page identity and readiness on open and every snapshot

iddogino:
**Stacked on #1043** (it extends the page probe that PR adds). Base is `fix/snapshot-text-visibility`; retarget to `main` once #1043 merges.

## Why

Theme 2 of the transcript-mining report (130/200 reviewed sessions): `browser_open` echoed the requested URL for a 200, a 429, a login redirect and `about:blank` alike, and no snapshot carried URL, title, HTTP status or readiness. `(no interactive elements)` was the same string for a hydrating SPA, a 401 JSON body, a Chrome net-error page and an Imperva ban. One session spent 120 calls on a rate-limited site before `network requests` revealed the 429.

## What changed

- **Probe** (the one eval per snapshot from #1043) now also reads `location.href`, title, `readyState`, `document.contentType`, the navigation response status via `PerformanceNavigationTiming.responseStatus` (Chrome 109+, no CDP), visible loading indicators, Chrome net-error pages, and bot-wall signatures (Cloudflare, Imperva, Akamai, Google sorry, PerimeterX, DataDome, CAPTCHA vendors).
- **Status line on every snapshot / get_state**: `[page] URL · "title" · HTTP 200 · complete · 84 refs`, then `⚠` lines ordered by how much each invalidates the tree, each with the action to take (request_browser_input for a bot wall, browser_wait when still loading). Trees with fewer than 5 refs get a 1,500-char text preview.
- **`browser_open` reports the landing, not the intent**: the discarded `get url` read becomes the probe. Text is `Loaded "Sign in" at <final url> (redirected from <requested>) · HTTP 200`; a net error, bot wall or HTTP ≥ 400 is returned as a tool error with the page text attached. If the probe cannot run, the old text is used with an explicit "landing page could not be read".
- **Prompt and docs**: read the status line first; `browser_wait` guidance no longer says "browser_open already waits" as the reason never to wait.

Cut from the report's proposal on purpose: an `observe:` parameter on open, in-flight request counts, cookie-based classifiers, viewport size, and a header on screenshot results.

## Verified

- 57 tests across `snapshot-format.test.ts` and `tools/browser.test.ts`, including the probe script executed against a fake DOM for every signature and the three new open-tool outcomes.
- `tsc --noEmit` and `eslint` clean.
- Real CLI in the container image (`superagent-container:browser-stack-1001`):
  - `example.org/definitely-missing-page` renders a normal-looking tree (heading + link) while `responseStatus` reports **404**. Exactly the case the header is for.
  - A dead host lands on `chrome-error://chromewebdata/` with an empty body and no `#main-frame-error`, so detection also keys on the URL (found by this check, fixed, test added).
  - A normal page: `HTTP 200 · complete`, both live regions, iframe, preview.

No UI change.

### DIFF
--- a/agent-container/docs/browser-use.md
+++ b/agent-container/docs/browser-use.md
@@ -39,6 +39,15 @@ such as `@e1` and, in the default interactive view, no static text at all. A
 footer reports how much page text was dropped and what the live regions
 (alerts, status, toasts) currently say.
 
+Every snapshot starts with a status line: `[page] URL · "title" · HTTP status ·
+readyState · N refs`, followed by `⚠` warnings when the site is unreachable, a
+bot challenge is up, the server answered 4xx/5xx, the document is a raw file, or
+content is still loading. `browser_open` reports the same facts for the page it
+actually landed on (final URL, redirect, HTTP status) and is marked as an error
+when that page is a net error, a bot wall or an HTTP error. Act on a warning
+before reading the tree: a bot wall or login page needs the user, a loading page
+needs `browser_wait` for the element you need.
+
 Useful snapshot options:
 
 - `fullText: true`: add the page's static text (prices, prose, validation
--- a/agent-container/src/server.ts
+++ b/agent-container/src/server.ts
@@ -815,7 +815,10 @@ import { prepareEvalScript, finalizeEvalOutput, evalErrorHint } from './eval-scr
 import { judgeSelectCommit, SELECT_COMMIT_SETTLE_MS } from './select-verify';
 import { resolveCommittedValue } from './field-value-readback';
 import { capBrowserOutput, redactCdpUrls, MAX_BROWSER_OUTPUT_CHARS, MAX_BROWSER_ERROR_CHARS } from './browser-output';
-import { capSnapshot, compactWithText, formatIframePlaceholders, formatTextFooter, parsePageProbe, EMPTY_PROBE, PAGE_PROBE_SCRIPT } from './snapshot-format';
+import {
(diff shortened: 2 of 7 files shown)


iddogino:
**End-to-end validation** (image built from the stack, run standalone, real MCP tool handlers driven against real sites):

- `example.com` → `Loaded "Example Domain" at https://example.com/ · HTTP 200`; snapshot header `[page] … · HTTP 200 · complete · 2 refs`.
- GitHub 404 (`/SkillfulAgents/definitely-not-a-repo-xyz-123`) → open is `isError` with `⚠ HTTP 404 (not found — check the URL)` and the page text; the snapshot header repeats the warning above an 83-ref tree that otherwise looks normal.
- Dead host → the CLI's own `net::ERR_NAME_NOT_RESOLVED` error (unchanged path).
- `http://github.com/login` → `https://github.com/login` with no "redirected from" (scheme upgrade ignored, as intended).
- Snapshot taken 7ms after `Enter` on Wikipedia search → `untitled · HTTP 200 · loading · 0 refs` + `⚠ page still loading`; DuckDuckGo results scoped snapshot → `interactive · 10 refs` + warning.

**Bug found and fixed by this run** (commit 870474d6): `browser_open` re-parsed the server's normalized `page` object with the raw-shape parser, dropping `httpStatus` and `liveRegions` — the GitHub 404 initially came back as a plain "Loaded …" with no warning. The unit tests mocked the raw shape and passed; they now mock what the server sends. Same commit switches the probe's visibility test to `checkVisibility` (see #1047 for why).

iddogino:

PR #1046, opened 2026-09-11
Type the agent's configuration documents and the agent catalog

iddogino:
## Summary

PR E1 of the agent-actor refactor, stacked on #1041: **configuration documents and the agent catalog**. Every per-agent configuration file lives inside the workspace, so `actor.config` is a typed layer over `actor.files`: `get`, `put`, and a serialized `update` for the documents that are read-modify-written, each with a Zod schema at the boundary. Which agents exist is a host-level question, so it gets a host-level answer: `agentCatalog`. Behaviour is unchanged; the API contract is unchanged.

Spec: `agent-actor-remote-contract-spec.md` §4 E1 (repo root, untracked).

No UI change.

## What is in here

- **`agent-actor/config-schema.ts`** — the document table: `instructions` (`CLAUDE.md`, text), `secrets` (`.env`, text, shared with the container, mode `0o666`), `preferences` (`agent-preferences.json`, `agentPreferencesSchema`), `claudeSettings` (`.claude/settings.json`, `claudeSettingsWithHooksSchema`, unknown keys pass through). `ConfigDocError` with `corrupt` (stored JSON does not parse or fit) and `invalid` (a caller tried to store something that does not fit).
- **`agent-actor/config-ops.ts`** — one implementation over any `FileOps`: decode and validate on read, validate and pretty-print on write, serialize `update`. A corrupt document makes `get` throw and `update` abort before writing, so a broken file is never silently replaced — the property the preferences and mounts code had to build by hand.
- **`agent-actor/local-config-ops.ts`** — what is local: a document the container also writes (`.env`) is serialized with the on-disk lock the container honours, and `.env` is healed back to `0o666` on read when this process owns it (the heal used to live in `getSecretEnvVars` and only ran at session start). Documents only this process writes are serialized in-process and leave no lock files in the workspace.
- **`FileOps.putDoc`/`write` take `WriteOptions { mode }`** — advisory; the filesystem implementation forces it after the rename (the container must read `.env`), anything else ignores it.
- **`agentCatalog`** (`AgentCatalog`: `list`, `exists`, `mint`, `resolve`, `remove`) over today's directory scan and id helpers. Creating an agent is `mint` + `files.mkdir('')` + `config.put('instructions')`; deleting is stop + `remove` + evict.
- **`containerHost.agentHostPath(slug)`** — the agent directory above the workspace, host-only, for the one thing kept there: `mounts.json`, the host folders bind-mounted into the container.
- **Contract suite** `describeConfigOpsContract`, run against the local and in-memory implementations; the local file adds the mode-heal and lock cases.
- **Ported consumers**: `agent-service` (parse, list, get, create, update, delete, exists, raw instructions), `agent-preferences-service`, `secrets-service`, `agent-hooks-service`, `skills/index`, `mount-service`.

## Deviations from the spec

- **`skillsetMetadata` is not a config document yet.** `.skillset-agent-metadata.json` belongs to the template service, which is E2's; it joins the table there.
- **The `getAgentWorkspaceDir` definitions stay** (fenced). Their only importers are the container layer and the actor package.

## Verification

- `npm run typecheck`: clean.
- `npm run lint`: 0 errors, 19 warnings. Allowlist: 24 entries in 20 files. What remains for storage paths: `agent-template-service` (2), `skillset-service` (1), `skillset-reconcile` (2) — all E2 — and `home-graph-service`'s metadata stat cache (F). Everything else left is the app-level persister hub and the host-level MCP re-auth / review registries.
- `npx vitest run`: 716 files, 11890 tests pass. New: the `ConfigOps` contract suite over both implementations plus the local mode-heal and lock cases. Two route tests widened partial mocks of the preferences and hooks schema modules with `importOriginal`, since the config table imports the real schemas. One assertion in `agent-service.test.ts` changed error class: a read of a directory where `CLAUDE.md` should be now throws the actor's `WorkspaceFileError` (`not-a-file`) instead of a raw `EISDIR`; the claim (throws, not null) is unchanged.
- Mocked E2E (`E2E_MOCK=true E2E_PORT=3177`): 339 pass, 24 skipped, 11 fail in the 15.4-minute run under 4 workers. All 11 were timeouts under load (browser-stream ×2, chat-flow ×3, capability-review, session-long-sleep, home-graph ×4); 15 of 17 pass when re-run together in isolation and the last two (browser-stream, chat-flow "see tool call in response") pass 2 of 2 and 3 of 3 alone. A first attempt at this run died half-way because the machine lacked Playwright's full Chromium build (`chromium-1208`) and the host-browser spec's spawn failure took the dev server down; `npx playwright install chromium` fixed the environment. That unguarded spawn is a pre-existing robustness bug, not touched here.

## Behaviour notes

- `ConfigOps.update` writes nothing when the mutator returns the current document, so the secrets not-found / conflict / invalid branches leave `.env` untouched as before.
- The `.env` mode heal now runs on every secrets read, not only at session start.
- `updateAgentPreferences` and `removeAgentHook` serialize in-process per document instead of through a lock file beside the document; `.env` keeps the on-disk lock because the container writes it too.

## Review fixes

- The `.env` mode heal before a secrets read used `stat`/`chmod`, which follow links; a link an agent planted as `.env` would have had its target's mode changed on every read. The heal uses `lstat` and leaves links alone.
- The test fake gained the `config` group this PR added to the contract.
- Home perf budgets: reading each agent's CLAUDE.md through the actor adds one realpath per agent per home read, pinned alongside the artifact probes.

## Second review round

- The in-memory workspace keeps the mode a write asked for, so the contract now says what every store reports: the mode asked for is what `stat` reports afterwards, and a rewrite without one keeps it. #1048's export relies on it.

## Real-path containment pulled out

Follows #1040: the in-memory store no longer reports `resolvedPath` (it answers `resolve` with the path itself), and the `.env` heal test reads through a planted link the way `main` does while still pinning that the heal never chmods what the link points at. The mode a write asks for is kept as before.

Every document type round-tripped byte-identically against `main` (secrets with comments, quotes, `export`, CRLF and duplicates; preferences; hooks with unknown keys; `CLAUDE.md` frontmatter), through a tolerance matrix (empty, whitespace, invalid JSON, BOM, arrays, directories, links, mode 000) and 20–30-way concurrent writes. Two findings, fixed here with tests that failed first:

- **A directory in place of `.env` answered 500 where `main` answered 404.** The secrets service's pre-check asked only whether something was at `.env`; it now asks for the file, so an unknown-secret update or delete is `not_found` again.
- **The mode heal chmod'ed a directory planted as `.env` on every secrets read.** The guard covered only a link; it now leaves anything that is not the file alone, so a directory keeps its execute bits and its contents stay reachable.


iddogino (commented):
👍

yiw190 (approved):

PR #1047, opened 2026-09-11
fix(browser): report what a click did, not just whether the URL moved

iddogino:
**Stacked on #1045** (which is stacked on #1043). Base is `fix/page-status-line`; retarget as the chain merges.

## Why

Theme 3 of the transcript-mining report (118/200 reviewed sessions): click, press, select and hover returned only a URL comparison. On any single-page app the result read "URL unchanged … Re-snapshot only if you need to see resulting DOM changes" whether a dialog opened, a toast fired, a request failed or the click was swallowed. Agents either trusted it and shipped wrong work (a "See more" click that did nothing became "5 groups is the complete set") or paid a snapshot per click (150 `ArrowUp` presses on one URL, with 80 evals interleaved to find out what happened). The tree diff had been deferred because a snapshot per action rotates the CLI's ref registry.

## What changed

- **`action-effect.ts`**: one eval before the action and one after the settle fingerprint the page with no snapshot and no ref rotation. Fingerprint: open dialogs / alertdialogs / popovers with accessible names, live-region text, a DOM census of visible interactive elements, body-text length + hash, the focused element, and requests that failed since the action via `PerformanceResourceTiming.responseStatus`. The "before" read clears the resource-timing buffer so a long session cannot fill the default 250-entry buffer and hide a later failure. The "after" read supplies the URL, so this replaces the post-action `get url` at no extra round trip.
- **`Effect:` line on click / press / select / hover**: `dialog "Delete project?" opened · announced: "Cart Error" · failed request: fetch /cart/add.json → 404 · +12 interactive elements`, or `no DOM change within 300ms — the element may not be handling clicks (disabled, covered, or needs a different target)…` with a verb-specific hint. Press always reports focus. Nothing is claimed across a navigation.
- **Removed** "Re-snapshot only if you need to see resulting DOM changes". Prompt rule 4 and the docs now say what the effect line carries and that "no DOM change" is a signal to check the element rather than click again.

Cut from the report's proposal: the page-init MutationObserver (needs a script injected on every new document via CDP, which the harness only holds for the host browser) and the subtree hash around the target (needs the daemon's ref-to-element map). The before/after read at the settle point catches any toast that persists past the window, which covers the evidence.

## Verified

- 20 new tests in `action-effect.test.ts` (fingerprint script against a fake DOM including resource entries and a browser without `:popover-open`, parser, diff, formatter for every verb) plus the existing digest and tool tests. `tsc --noEmit` and `eslint` clean.
- Real CLI in the container image (`superagent-container:browser-stack-1001`):
  - `showModal()` click → `top: [{dialog, "Delete project?"}]` (name via `aria-labelledby`), interactive 4 → 6, focus `button "Cancel"`.
  - Click that appends a `position:fixed` `role=alert` toast and sets a `role=status` region → both texts in `live`.
  - Same-origin `fetch('/cart/add.json')` 404 (page served over HTTP in the container) → `failed: [{"/cart/add.json", 404, fetch}]`. From `file://` the same fetch reports nothing, as expected for cross-origin without Timing-Allow-Origin.
  - A button covered by the open dialog is refused by the CLI itself ("covered by <dialog#dlg>"); fingerprints before/after are identical, which is the "no DOM change" case.

Cost: two evals per action (~2–4ms each in the container), no snapshot.

No UI change.

### DIFF
--- a/agent-container/docs/browser-use.md
+++ b/agent-container/docs/browser-use.md
@@ -79,9 +79,15 @@ Use the most specific tool:
 - `browser_hover` for hover menus and tooltips;
 - `browser_scroll` for page or container scrolling.
 
-Trust the action result. Click and key results report navigation; fill results
-report the value the page actually committed. A fill warning means the page
-kept a different value—fix it before moving on.
+Trust the action result. Click, key, select and hover results report
+navigation plus an `Effect:` line — dialogs opened or closed, live-region
+announcements (toasts, validation errors), failed requests since the action,
+and the change in the number of interactive elements. `Effect: no DOM change
+within 300ms` means the page did not react: the element is probably disabled,
+covered, or the wrong target, so check its state or `browser_wait` for what you
+expect instead of clicking again. Fill results report the value the page
+actually committed. A fill warning means the page kept a different value—fix it
+before moving on.
 
 Navigation makes existing refs stale. Re-snapshot when a result reports
 navigation, when a dialog or dynamic view changes the relevant controls, or
--- a/agent-container/src/action-effect.test.ts
+++ b/agent-container/src/action-effect.test.ts
@@ -1,0 +1,185 @@
+import { describe, it, expect } from 'vitest'
+import {
+  fingerprintScript,
+  parseFingerprint,
+  diffFingerprints,
+  formatActionEffect,
+  type Fingerprint,
+} from './action-effect'
+
+// Minimal fake DOM for the fingerprint script: querySelectorAll by selector,
+// activeElement, body.innerText, resource timing entries.
+type FakeEl = {
+  tag?: string; role?: string | null; text?: string; visible?: boolean; label?: string | null
+  labelledby?: string | null; heading?: string | null; matchesPopover?: boolean
+  placeholder?: string | null; name?: string | null; labels?: Array<{ innerText: string }>
+}
+function el(o: FakeEl) {
+  const attrs: Record<string, string | null> = {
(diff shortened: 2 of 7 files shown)


iddogino:
**End-to-end validation** (image built from the stack, run standalone, real MCP tool handlers driven against real sites):

- Wikipedia "Main menu" click → `Effect: +13 interactive elements`; clicking a heading afterwards (which closes the menu) → `-13 interactive elements`.
- `Tab` → `Effect: focus: link "Wikipedia"`; `Enter` in the search box → `Page NAVIGATED`.
- MDN nav button: hover → `no DOM change within 300ms — nothing opened on hover. Try browser_click…`; click → `+12 interactive elements`; `Escape` → `-12 interactive elements · focus: button "HTML"`.
- DuckDuckGo: fill + Enter → navigation.

**Bug found and fixed by this run** (commits 870474d6 on #1045 and b7df4eed here): the visibility test used `getClientRects`, which is non-empty for `visibility:hidden` / `opacity:0` elements. Wikipedia's Vector dropdown hides its closed menu exactly that way (`display:block; visibility:hidden; opacity:0`), so opening it read as `page text changed` with +0 interactive elements. Both probes now use `checkVisibility({visibilityProperty, opacityProperty})` with `getClientRects` as the fallback. Also from the run: `page text changed` now carries a size (`+1,234 chars`), and focus reports implied roles (`link` rather than `a`).

iddogino:

PR #1048, opened 2026-09-11
Move skillsets and templates onto the actor's file operations

iddogino:
## Summary

PR E2 of the agent-actor refactor, stacked on #1046: **skillsets and templates on the file operations**. The skills tree under `.claude/skills` and the template import/export of a workspace are plain files, so the skillset service, the reconcile sweep, and the template service read and write them through `actor.files`, and the template's `.skillset-agent-metadata.json` becomes the `skillsetMetadata` config document. The skillset cache (git clones under the data dir) is host-level and stays where it is. Behaviour is unchanged; the API contract is unchanged.

Spec: `agent-actor-remote-contract-spec.md` §4 E2 (repo root, untracked).

No UI change.

## What is in here

- **`agent-actor/copy-into-workspace.ts`** — `copyHostDirIntoWorkspace(files, hostDir, workspaceDir, { exclude, followSymlinks })`: the one way a host directory (a skillset cache clone) gets copied into a workspace, through `files.mkdir` and `files.write`, with the same exclusion and symlink rules `copyDirectoryFiltered` had. Both services use it.
- **`skillsetMetadata` config document** — `.skillset-agent-metadata.json`, validated with the existing `InstalledAgentMetadataSchema` (loose on unknown keys so a read-modify-write keeps what this build does not know).
- **`skillset-service`**: the agent's skills tree (`.claude/skills/<dir>/…`) is read and written through `files.list/stat/getDoc/putDoc/write/delete/mkdir`; installs and refreshes copy from the cache with the helper; zip import writes entries through the actor (entry-name sanitation kept, workspace containment now the actor's); export gathers through the actor. The skillset cache (git clones, PR and publish writes into the clone) stays on `fs`.
- **`skillset-reconcile`**: the sweep iterates `agentCatalog.list()` and reads each agent's skill and template metadata through its actor; the prune helpers take the agent's `files` and delete through it. New unit tests (10) cover both prune helpers and the sweep.
- **`agent-template-service`**: export walks the workspace with `files.list` and streams each file into the archive from `files.read` (bounded in-flight entries); import writes zip entries with `files.write`; install and update copy from the cache with the helper; onboarding-skill and template-prompt reads are ranged `files.read`s; `CLAUDE.md` goes through `config.get/put('instructions')` and the template metadata through `config.get/put('skillsetMetadata')`; hashes are computed over the same bytes as before (byte-identical digests). Cache-clone writes stay on `fs`.
- **Allowlist**: every `file-storage` path helper entry outside the actor package is gone except one, `home-graph-service`'s metadata stat cache, which PR F removes with the invocation counts. What remains is the app-level persister hub and the host-level MCP re-auth / review registries.

## Deviations from the spec

- **The `getAgentWorkspaceDir` definitions stay** (fenced). Their only importers are the container layer and the actor package.

## Verification

- `npm run typecheck`: clean.
- `npm run lint`: 0 errors, 19 warnings. Allowlist: 19 entries in 17 files — the persister hub, the MCP re-auth / review registries, and `home-graph-service`'s metadata stat cache (F).
- `npx vitest run`: 718 files, 11904 tests pass. New: `skillset-reconcile.test.ts` (10). One assertion in the template export test changed from matching a raw `ENOENT` to the actor's not-found error when a file disappears mid-export; the claim (the export rejects rather than finalizing an incomplete archive) is unchanged.
- Mocked E2E (`E2E_MOCK=true E2E_PORT=3177`): 338 pass, 24 skipped, 12 fail in the 16.1-minute run under 4 workers. All 12 were timeouts under load (agent-lifecycle, chat-flow, capability-review, composer-failure-recovery, file-preview, stop-session, home-graph ×6) and all pass when re-run together in isolation (13 of 13 in 36 s). Same pattern as the runs on #1031, #1038, #1040, #1041 and #1046.

## Behaviour notes

- Zip import buffers each entry before writing it through the actor (the zip reader has no streaming entry API); entry sizes are capped as before.
- Template export keeps at most four archive entries in flight, which bounds open file descriptors; an archiver error now also destroys the archive rather than finalizing an incomplete one.

## Review fixes

- Copied skill files lost their mode (a skill's `scripts/run.sh` landed non-executable, and every refresh stripped it again), were copied one after another through `write`, and a link back into the tree recursed until the path ran out. The copier now copies a directory's files concurrently, hands each file's mode on, remembers directories by real location when following links, and on a workspace on this machine takes each file in one filesystem copy (`LocalFileOps.copyHostFile`, mode preserved, same destination containment as any write). Test: `copy-into-workspace.test.ts`.
- The workspace-side read of a skill package for hashing went file by file; the files of a directory are read concurrently now.
- Install perf, under the harness's per-operation latency: 347 ms before this stack, 3580 ms as first pushed, 771 ms now. What remains is destination containment the plain copy never had (a link check and a parent resolution per written file, a real-path check per file the hash reads back, about 140 operations). The install budget is set to 900 ms with that rationale in the file; a bulk import that resolves the destination directory once, and a batched read for the hash, would bring it back under the old 500 ms target as a follow-up.
- Folder upload keeps each file's mode (the executable-bit regression from #1040's route rewrite).
- Export: archiver neither listens on the sources it is handed nor ends them when it stops. A workspace read failing mid-export used to be an unhandled stream error with the host export lock left held; destroying the archive left opened sources reading. The export now owns its sources: a source error fails the export like the archive's own errors, and any stop cancels every opened source. Both pinned by tests that drive a real export with a failing and a never-ending source.

## Second review round

- Import memory: each zip entry was read whole into memory before being written, where the extraction it replaced streamed to disk. One 128 MiB entry grew the process by 127 MiB; 30 MiB now that the entry streams through the actor's `write` (`ZipReader.openEntryStream`, with the size limit failing the stream). A failure the write reports is no longer skipped: only a destination outside the workspace is, as before; anything else (a name the filesystem rejects, a full disk) fails the import instead of dropping the file in silence.
- Export modes: archiver was handed a stream and a date but no mode, so a 0755 script came out 0644 and failed with EACCES after extraction. The mode the actor reports (`FileStat.mode`, from #1040) goes along; zip entries report theirs, and both exports are pinned to keep a script executable (proven to fail without the change).
- The folder upload's stream fallback (a workspace that is not on this machine) carries the source's mode through `copy-into-workspace.ts`; on this machine the file is one copy, which keeps it.

## Real-path containment pulled out

Follows #1040: `copyHostFile` copies into a directory the walker has already created, one filesystem copy per file as the plain copy did, with no link check or parent resolution per file. The install perf budget is back to `main`'s 500 ms and its operation count is pinned at 106: `main`'s 99 plus the atomic writer's temp file and rename for the three documents the install writes (two of which `main` wrote in place) and one `realpath` for the cycle guard on the source tree.

Template export and import round-tripped byte-identically against `main` in all four cross-import combinations (nested and empty directories, dotfiles, unicode names, a 240-byte name, modes, a hard link, links inside, outside and dangling, a 64 MiB file, 2000 small files); hostile path entries stayed contained on both; `openEntryStream` limits hold and leak no descriptors. One finding, critical: **a zip whose entry inflates past its declared size, or whose deflate stream is corrupt, took the whole host process down** on import, where `main` answered 500. The cause is in the actor's streamed write and is fixed on #1040; this PR adds the import-level test, `fails the import, without taking the process down, when an entry inflates past its declared size`, which builds such an archive and pins a rejected import with no unhandled error. It fails on the previous head of this branch.


iddogino (commented):
👍

yiw190 (approved):

PR #1049, opened 2026-09-11
fix(browser): wait up to 2s for a loading document before taking a snapshot

iddogino:
**Stacked on #1047** (→ #1045 → #1043). Retarget as the chain merges.

## Why

Validating the stack against real sites: a snapshot taken right after `Enter` on Wikipedia's search box returned 7ms later as `[page] … · untitled · HTTP 200 · loading · 0 refs` with the still-loading warning. Correct, but the agent then pays a `browser_wait` plus a second snapshot for the most common sequence there is.

## What changed

- `waitForDocumentReady()` in `snapshot-format.ts`: poll a reader until `complete`, up to a timeout; stops at once if the read fails.
- `/browser/snapshot` polls `document.readyState` (one ~3ms eval) every 150ms for up to 2s before reading the tree. `browser_get_state` inherits it.
- The still-loading warning names the wait: `page still loading after waiting 2.0s — …`, so a genuinely slow page is distinguishable from an instant read.

Cost on an already-loaded page: one tiny eval. Nothing changes for `browser_open` (the CLI already waits for load) or for actions.

## Verified

- 5 new unit tests with an injected clock: immediate return, poll-until-complete with the waited time, timeout, dead-read stop, and the warning text.
- `tsc --noEmit` and `eslint` clean.

No UI change.

### DIFF
--- a/agent-container/docs/browser-use.md
+++ b/agent-container/docs/browser-use.md
@@ -42,7 +42,8 @@ footer reports how much page text was dropped and what the live regions
 Every snapshot starts with a status line: `[page] URL · "title" · HTTP status ·
 readyState · N refs`, followed by `⚠` warnings when the site is unreachable, a
 bot challenge is up, the server answered 4xx/5xx, the document is a raw file, or
-content is still loading. `browser_open` reports the same facts for the page it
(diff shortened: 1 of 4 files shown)


iddogino:

PR #1050, opened 2026-09-11
Draw agent-to-agent edges on the home graph from invoke permissions only

iddogino:
## Summary

PR F of the agent-actor refactor, stacked on #1048: **agent-to-agent edges on the home graph are a projection of the invoke permissions**. The graph no longer counts sessions to decide which agent pairs to draw or how to style them: a pair with a permission row gets one solid edge, with no count chip. The server stops walking every agent's session metadata for the counts, which removes the last filesystem read outside the actor package and the last storage entry in the fence allowlist.

Spec: `agent-actor-remote-contract-spec.md` §0 (decision: edges are a pure policy projection) and §4 F.

## What changes for the user

- Agent-to-agent edges are always solid. Before, a pair was dashed until a session had been invoked and solid afterwards, with a hover chip like "12 invocations". That distinction is gone; the edge means "a relationship you configured".
- A pair that has talked only under the implicit `review` default, with no policy row, no longer draws an edge. Accepted in the spec: the edge now means a configured relationship, and a per-agent "who invoked this agent" read is the hover follow-up if that is wanted.
- Resource edges (accounts, MCPs, chats, webhooks, crons) are unchanged: still dashed until used, still counted from the audit tables.

## What is in here

- `home-graph-service`: `countInvocations`, the per-agent metadata stat cache, and the session-metadata read are removed. The `invocations` field stays on the wire as an empty array for one release so a renderer cached from the previous build keeps working; the schema comment says so, and it can be removed next release.
- `use-graph-data`: agent edges come from `permissions` only, one per unordered pair, `variant: 'permission'`, no weight.
- `agent-graph`: a permission edge is solid because it exists (resource edges keep the exercised / unused rule), and carries no count chip. `EDGE_UNIT` loses its `agent` row.
- Tests: the renderer tests that pinned activity edges now pin the permission projection and that a populated `invocations` field changes nothing; the service test pins the empty field and the permission it derives from.
- Allowlist: `home-graph-service` is gone. No entry for any storage path remains outside the actor package; what is left is the app-level persister hub and the host-level MCP re-auth / review registries.

## Screenshots

Mock instance, three agents, one invoke permission (Release Notes Writer → Inbox Triager). The permitted pair gets one solid edge with no count chip; the pair with no policy row gets none.

| Light | Dark |
| --- | --- |
| Home graph (light) | Home graph (dark) |

## Verification

- `npx tsc --noEmit`: exit 0.
- `npx eslint 'src/**/*.{ts,tsx}'`: exit 0, 0 errors (19 pre-existing warnings). Fence allowlist down to 18 entries in 16 files, none of them a storage path.
- `npx vitest run`: 718 files passed, 6 skipped; 11 904 tests passed, 45 skipped, 1 todo.
- Mocked E2E (`E2E_MOCK=true E2E_PORT=3177 npx playwright test`): 340 passed, 10 failed, 24 skipped in 15.1 min. All 10 pass in isolation:
  - `web-graph` project alone: 7/7 in 21 s. The full-run errors were "element is outside of the viewport" during the layout-settle retries, the file's documented load flake.
  - `chat-flow:84`, `session-long-sleep:87`, `settings:302` (Voice tab) and WebKit `read-aloud-follow:119`: pass.
  - `browser-stream`: 4/4 with `--workers=1`. Run in parallel with each other, three of its four tests fail because they share one mocked host browser (the canvas never appears). The same two tests failed in the full run of every earlier PR in this stack, and this PR touches nothing on that path.
- Screenshots above are from a fresh mock instance seeded through `/api` (three agents, one invoke policy).
- Not run: the manual `dev:electron` smoke.

## Review fixes

- One more caveat for the policy-only edges: a global invoke policy (no target, "may invoke any agent") draws no edge either. The server has always filtered null-target rows out of the graph; before this PR the activity edges stood in for those pairs. If that matters, fanning a global row out to every visible agent is a small server-side follow-up.

## Second review round

- Nothing in this PR's own scope changed; the stack below it carries the round's fixes (#1038 to #1048).
- E2E: `expandAgent` in `e2e/pages/agent.page.ts` probed once for the sidebar chevron and gave up silently. The chevron renders once the agent's session count is non-zero, which the agents endpoint reports only once the session's transcript lands, a beat after the sessions endpoint lists it from its metadata. Without the home graph's metadata walk the page is ready sooner, so on CI the probe ran before the count refreshed and the session-row drop test failed three of three here while passing on every branch below (the speed and model selection specs flaked the same way). The helper now waits for the chevron.

The edge table for fifteen permission and invocation fixtures matches the description above exactly on both trees, and forty instrumented page loads never showed a sidebar row without its chevron on either tree, so the E2E helper change masks nothing. One finding, cosmetic, fixed here: **a connector route or anchor the user had dragged for a pair that had invoked each other was lost once after upgrade**, because that pair's edge id changed from the activity form to the permission form and the saved geometry was keyed by id. `savedEdgeGeometryFor` looks the geometry up under the current id and then the earlier one; the next drag saves it under the current id and the persist prunes the old key.


iddogino (commented):
👍

yiw190 (approved):

PR #1051, opened 2026-09-11
fix(browser): re-read before claiming "no DOM change"; report typed field values

iddogino:
**Stacked on #1049** (→ #1047 → #1045 → #1043).

## Why

Replaying mined production sessions with a real model (Sonnet 5) against the stack showed two ways the effect line could be wrong:

- absorbmore.com add-to-cart: the cart drawer opened after the 300ms read. The result said `no DOM change within 300ms — the element may not be handling clicks`. The agent recovered via a screenshot, which is the call the effect line exists to save.
- Meta Ad Library search box: `Control+a` and typed text reported `no DOM change within 50ms — the key may have been ignored`. The value changed; neither the page text hash nor the interactive census can see input values.

## What changed

- **Second read on silence.** If the first post-settle fingerprint shows nothing, wait until 1.2s after the action and read again before reporting. The reported window is the real one (`no DOM change within 1200ms`). Cost lands only on actions that would otherwise be reported as inert.
- **Focused field value in the fingerprint.** Reported as `field value now "monday"` when it changes with focus unchanged. Counts as a change for the second-read decision.
- Press hint reworded: a silent key is normal for a modifier combo.

## Verified

- 79 tests across the effect, formatter and tool suites; `tsc --noEmit` and `eslint` clean.
- Replay of the absorbmore case on an image with this change is in progress; result will be posted here.

No UI change.

### DIFF
--- a/agent-container/src/action-effect.test.ts
+++ b/agent-container/src/action-effect.test.ts
@@ -173,6 +173,12 @@ describe('focused field value', () => {
   it('effectHasChange is false for an identical fingerprint', () => {
     expect(effectHasChange(diffFingerprints(base, { ...base }))).toBe(false)
   })
+
+  it('a focus move counts as a change for press but not for click (the click itself moves focus)', () => {
+    const moved = diffFingerprints(base, { ...base, focus: 'button "Add to cart"' })
+    expect(effectHasChange(moved, { countFocus: true })).toBe(true)
+    expect(effectHasChange(moved, { countFocus: false })).toBe(false)
+  })
 })
 
 describe('diffFingerprints + formatActionEffect', () => {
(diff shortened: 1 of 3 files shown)


iddogino:
**Validated with a real model** (absorbmore.com mobile-checkout replay, Sonnet 5, image built from this PR head):

- Add-to-cart click now reads `Effect: +12 interactive elements` (the cart drawer). On the previous build the same click read `no DOM change within 300ms` and the agent recovered via a screenshot.
- The two silent actions in the run read `no DOM change within 1206ms` / `1207ms`, i.e. the second read happened and the reported window is the real one. With the first cut of this PR a click's own focus move suppressed the re-read (fixed in faa89caf).
- Whole run: 46 tool calls, 10 screenshots, 4 evals, 137–231s across two runs, versus 127 calls, 28 screenshots, 21 evals, 1039s in the original session.

iddogino:

PR #1052, opened 2026-09-11
refactor(browser): page observer, status line and action settle as reusable primitives; wait for busy pages

iddogino:
**Stacked on #1051** (→ #1049 → #1047 → #1045 → #1043).

## What this is

The two things every browser result now carries — the status line at the top and the effect line at the bottom — were built ad hoc across four PRs: two page-side scripts, a hand-wired settle loop in the server, and four endpoints calling it by hand. This PR extracts them into three modules so they can be attached to any action in one line, and gives both the affordance a person gets from the browser's spinner: **wait while the page is working, and say so when it still is**.

- **`page-observer.ts`** — one page-side read, one Zod-validated `PageObservation`: identity and readiness, text, live regions, iframes, interactive census, dialogs, focus and field value, requests in flight and failed, busy signals, bot wall, net error. Install-once instrumentation wraps `fetch`/`XMLHttpRequest` transparently (counting only), which is the only way to see requests *in flight* and cross-origin statuses.
- **`page-status.ts`** — `formatStatusLine`, `pageWarnings`, `waitForQuiet`, `landedElsewhere`. The snapshot endpoint holds on `waitForQuiet` and uses that single read for header and footer; there is no separate probe any more.
- **`action-settle.ts`** — `ACTION_POLICIES` (one row per verb: settle, recheck, poll, cap, whether focus counts, the no-change hint) and `observeAction` (baseline → act → settle → poll while busy until quiet and two reads agree, or the cap; silence gets one recheck at 1.2s). `browser_scroll` is attached as the first new verb.

## Busy, as seen from the page

Semantic first (indeterminate `role=progressbar`, `aria-busy`, loading status text, `cursor: wait`), then infinite CSS animations via `document.getAnimations()` classified by geometry: thin strip pinned to the top = top bar, small square = spinner, several wide blocks = skeleton. Animations count only if they **started after the action** (`currentTime ≤ elapsed`, or their target was hidden at the baseline), or, for a status read, in the last 3s **and not with the page's own load** — so a spinning logo is never a spinner. Permanent widgets (progressbar, aria-busy) are reported but never waited on.

What the agent sees:

```
[page] https://… · "Results" · HTTP 200 · complete · 12 refs
⚠ still busy: top bar, spinner, 4 skeletons — content still arriving. browser_wait for the element you need, then re-snapshot.

Effect: still busy after 2.0s (spinner, 1 request in flight) — the page is still working. So far: +3 interactive elements. browser_wait for what you expect, then re-snapshot.
```

[...]

PR #1053, opened 2026-09-11
fix(browser): tool text states observations only — no inferred negatives, no advice unless certain

iddogino:
Stacked on #1052. Seventh and last PR of the browser-harness stack.

## Why

Two adversarial reviews of #1043–#1052 (stability, and "can this text set the agent astray") found no crash, hang or 500 path, but found that every misleading line an agent could receive was an **inference**, never a fact. Reproduced on a container built from the stack:

| Agent saw | Reality | Likely next move |
|---|---|---|
| `Effect: no DOM change — the element may not be handling clicks` | a checkbox / aria-pressed toggle / click into a text field | click again → un-checks the box |
| `Effect: no DOM change — you are at the end of the content` (×3) | viewport 300–1620 of 23,857px on a Wikipedia article | stop scrolling |
| `⚠ bot-block: CAPTCHA … more scraping will not get past it` + `isError` | "This site is protected by reCAPTCHA…" on a login form | hand off to the user |
| `⚠ HTTP 404 (not found — check the URL)` on every snapshot | an SPA served from a 404 fallback, 61 refs | abandon the task |
| `Effect: field value now "hunter2-secre"` | Backspace in a host-autofilled password field | secret in the transcript |
| `Effect: page text changed (same length …)` | a clock ticked; the click's real effect (a navigation at 700ms) was masked | trust the wrong effect |

## Principle (added to CLAUDE.md)

Browser tool text states **observations only**: what was seen, or that nothing was and what was looked at. Never infer a negative or give advice unless the observation leaves no doubt. Agents can establish those states themselves with a snapshot or screenshot; a hint only saves steps, while a wrong hint turns a recoverable run into a failed one. When in doubt, say less.

## What changed

**Observed more, so silence is rarer**
- `stateHash` over visible controls (checked, aria-checked/pressed/expanded/selected, disabled, chosen option) → `control state changed (checked/pressed/expanded/selected)`.
- Focus is reported for a click that moved it (`focus: textbox "Password"` shows which element took the click), not only for a press.

**Fewer false positives**
- Text diffs use the content region (`main`, else body) and ignore same-length changes (clocks, counters). The footer's whole-page char count is unchanged.
- Failed requests count on the site's own hosts only.
- Bot-wall signatures are the challenge page's own wording (product names dropped) and count only on a page with ≤3 visible controls or a 403/429/503.
- The 50ms slack that let an always-on animation count as new is gone.
- `focusValue` is `''` for password fields and secret `autocomplete` kinds.

**No verdicts**
- `Effect: none observed within 1200ms (dialogs, live regions, page text, control state, focus)` replaces every per-verb hint. A scroll that changed nothing gets no line and no 1.2s recheck (0.3s per scroll, was 1.2s).
- `Effect: page still active after 2.0s (spinner, 1 request in flight). So far: …` — no "browser_wait then re-snapshot".
- HTTP 4xx/5xx is a fact in the status line on any page and a ⚠ / `browser_open` error only when the tree is thin. `401 (authentication required)`, `404 (not found)`; no "check the URL".
- `still loading — content may be incomplete.` / `still active: …` carry no advice. PDF → `browser_download it and Read the file`.
- A click/press that opened a new tab reports the tab, not an effect line about the old one.
- Prompt and docs reworded; the contradictory "re-snapshot after a moment" rule removed.

## Validation

Unit: 94 tests in the five touched files, `tsc --noEmit` and eslint clean.

Real path (image built from this branch, real tool handlers, fixtures + real sites):
- checkbox, aria-pressed, aria-expanded, native `<select>` → `control state changed`; same value selected twice → `none observed …`
- Wikipedia article, 3 scrolls → viewport line only, 0.3s each
- click that navigates via JS 700ms later on a page with a ticking clock → `Page NAVIGATED` (the stack-6 build said `page text changed (same length …)`)
- Backspace in a password field → `none observed … — focus: textbox "Password"`
- login form with the reCAPTCHA disclosure → opens normally; Pinterest, Instagram, Stack Overflow, Cloudflare logins → no ⚠
- GitHub 404 page (83 refs) → `HTTP 404` in the status line, no ⚠, `browser_open` not an error
- slow (3s) form submit via click and via Enter → `NAVIGATED` (the CLI waits for the navigation it triggers)
- Wikipedia menu / Tab / search, DuckDuckGo search, MDN hover+click+Escape → same facts as before, new wording

Refuted during review (no change needed): slow full-page navigations reading as "no change" (the CLI waits), pages stuck at `readyState=interactive` (a hanging subresource fails `agent-browser open` itself, pre-stack), CodeMirror's cursor blink as a spinner.

Not addressed here: `browser_open` on a URL that downloads (PDF in the headless shell) lands on `about:blank` and says so — misleading but pre-existing and outside this stack.

No UI change.


iddogino:
## Second commit: recheck only where a late effect is plausible (9051e428)

Each verb's policy now decides from the settle read whether silence warrants the 1.2s recheck. No opt-out parameter — the agent cannot know in advance when it is safe to skip, and a knob that reintroduces "URL unchanged" as the whole result is the regression this stack fixed.

| Action | Recheck | Measured (previous commit → now) |
|---|---|---|
| click that lands focus in a text field | no | 1.22s → 0.32s |
| click on a button / link | yes | 1.23s when silent; a JS navigation 700ms later is still caught as NAVIGATED |
| press Enter / Space | yes, 300ms settle | Enter on a form: 3.0s (the CLI waits for the navigation it triggers) |
| press Tab / Escape / arrows / combos / characters | no, 50ms settle | 1.21s → 0.06s |
| select | yes | 0.32s on change, 1.22s when silent |
| hover | no | 1.23s → 0.33s (MDN nav button: 2.0s → 0.32s) |
| scroll | no | 0.33s |
| any verb on a page still working | busy poll to the 2s cap, unchanged | `page still active after 2.0s (1 request in flight)` |

Measured on a container built from this commit with the real tool handlers, on the fixtures and on Wikipedia, MDN and CodeMirror. 98 unit tests, typecheck and lint clean.

iddogino:
## Replay of the six mined sessions on the full stack (#1043 → #1053 @ 9051e428), real model

Same harness as the stack-5 replay on #1043: Sonnet 5 as main and browser model, the original task prompts delegated verbatim to the web-browser subagent, container built from this branch. Stack-5 numbers are from that earlier run for reference.

| Case | Original | Stack-5 replay | Stack-7 replay |
|---|---|---|---|
| Beatport afro-house top 100 | 96 tools · 583s | 10 · 67s | 12 · 93s |
| Beatport house top 100 | 92 · 865s | 20 · 120s | Cloudflare 403 challenge on open → `⚠ bot-block: Cloudflare` + `⚠ HTTP 403` → `request_browser_input` in 11s, then idle (correct; no user in a replay) |
| Meta Ad Library "Wispr Flow" | 115 · 815s | 58 · 215s | 51 · 248s |
| Meta Ad Library 5 brands | 108 · 699s | 92 · 319s | 93 · 330s |
| absorbmore.com checkout | 127 · 1039s | 46 · 231s | 61 · 259s |
| Craigslist listings scan | 83 · 463s | 179 · 569s | 88 · 531s |

**What the new lines did in practice** (every harness line of interest paired with the agent's next call):
- 9 `Effect: none observed …` lines across the runs. The next call was a type, a press, a screenshot or a snapshot every time; no re-click of the same element.
- 32 `control state changed` lines, mostly on Ad Library filter and search interactions that read as "no DOM change" on stack 5.
- Craigslist is back near the original: 179 → 88 calls. The stack-5 run had turned every listing into a fullText snapshot; this run scrapes with eval like the original did and takes 4 fullText snapshots.
- Beatport house's Cloudflare challenge is a real wall (403, "Performing security verification", nothing to interact with), so the tightened detector still fires there and the hand-off is the right move. Cloudflare on Beatport is intermittent; this case is not a stable regression test.

**Two things this run showed, fixed in 91dfd80e:**
- A feed of loading cards printed `progressbar, progressbar, progressbar, progressbar, progressbar, progressbar.` — now `6 progressbars`.
- After `Effect: page still active after 2.0s (1 request in flight). So far: +13 interactive elements · control state changed`, the absorbmore agent followed the prompt's "wait for the element you expect" with a guessed selector and lost 25s to the wait timeout. The prompt now says re-snapshot, and use `browser_wait` only for a selector you already know. That wait is most of absorbmore's 46 → 61 difference; the rest is screenshots, which the CRO task asks for.

**Still outside this stack** (same as the stack-5 notes): `scope` selectors that match nothing right after a dialog or listbox was announced (theme 8), clicks on covered elements retried (theme 5), and model-side mistakes like `browser_run("mouse click …")` or Playwright calls inside `browser_eval`.

Transcripts are in the session scratchpad under `replays/` (stack 7) and `replays-prev-stack4to6/`.

iddogino:

PR #1055, opened 2026-09-11
fix(browser): show agents what a page is — status line, landing report, text footer, fullText that keeps text

iddogino:
Read-side half of the browser-harness work from the transcript mining ([themes 1 and 2](docs/browser-harness-friction-themes-2026-09.md)). Squashed and reduced from the seven-PR stack #1043–#1053, keeping only what is a plain observation with evidence of benefit. The act-side half (effect line after click/press/select/hover) follows as a separate PR on this branch.

## What the agent sees

Every snapshot and `browser_open` result now opens with a status line:

```
[page] https://en.wikipedia.org/w/index.php?search=… · "Accessibility tree - Search results - Wikipedia" · HTTP 200 · complete · 113 refs
```

followed by a `⚠` line only when something is certain from the observation:
- `site unreachable (ERR_NAME_NOT_RESOLVED) — this is Chrome's error page, not the site.`
- `Cloudflare challenge page — the site is challenging automated access; what follows is the challenge, not the site.` (challenge wording, on a page with ≤3 controls or a 403/429/503 — never a product name, so the reCAPTCHA disclosure on a login form does not fire it)
- `raw application/pdf document, not a web page — the tree shows Chrome's viewer.`
- `page still loading after waiting 2.0s — content may be incomplete.`
- `no interactive elements — read the page text below before concluding the page is empty; …`

An HTTP 4xx/5xx is a **fact in the line, never a warning**: GitHub's 404 page has 83 refs and reads as `HTTP 404 · complete · 83 refs`; an SPA served from a 404 fallback is not called an error page. `browser_open` reports the landing page (`Loaded "Sign in to GitHub" at https://github.com/login (redirected from …) · HTTP 200`) and is an error only for a net error.

The default (interactive) snapshot ends with a footer:

```
Page text (~9,072 chars on this page) is NOT in this interactive view — prices, prose, errors, table values are all dropped. Starts: "Main Page Talk Read …"
To read it: browser_snapshot with fullText:true (add scope:"<css selector>" to keep it small). Refs are the same in both views.
```

plus the live regions (alerts, status, toasts) when the page has any. `fullText: true` now fetches the unfiltered tree and compacts it in the harness, so it keeps static text and its refs match the default view. `browser_get_state` takes `scope`/`fullText`/`includeUrls` and `screenshot: false`.

A snapshot waits up to 2s for `readyState` to reach complete, so "press Enter, snapshot" lands on the results page instead of `loading · 0 refs`.

## How

- `page-observer.ts`: one eval, one Zod-validated observation. It reads the page and changes nothing in it — no globals, no wrapped APIs (the earlier stack's fetch/XHR instrumentation and animation heuristics are deliberately not here).
- `page-status.ts`: `waitForLoaded`, `pageWarnings`, `formatStatusLine`, `landedElsewhere`.
- `snapshot-format.ts`: `compactWithText`, `countRefs`, `formatTextFooter`; the iframe placeholders now come from the same observation instead of a second eval.
- Prompt and guide: step 0 (read the status line; a ⚠ is an observation to weigh, an HTTP status is a fact) and the fullText rule.

## Principle

Browser tool text states observations only. It never infers a negative or gives advice unless the observation leaves no doubt (Chrome's own error page). Agents can establish the rest with a snapshot or screenshot; a wrong hint turns a recoverable run into a failed one. Adversarial reviews of the earlier stack found every misleading case was an inference, which is why HTTP-status warnings, bot-block hand-off instructions and the "check the URL" advice are not in this PR.

## Validation

- 62 unit tests across `page-observer`, `page-status`, `snapshot-format` and `tools/browser` (fake DOM, virtual clock, a hostile document whose DOM throws everywhere); `tsc --noEmit` and eslint clean (three pre-existing warnings in server.ts).
- Real path, image built from this branch, real tool handlers: GitHub 404, dead host, http→https redirect, Wikipedia menu/Tab/search via Enter, DuckDuckGo search with a scoped snapshot.
- Model replays of six mined sessions were run on the full earlier stack (Beatport 96→10 calls, Ad Library 115→51, absorbmore 127→61, Craigslist 83→88); the gains came from the footer/fullText and status line kept here, plus the effect line in the follow-up PR.

No UI change.


iddogino:
Review fixes landed in b9c3dde1 (this PR) and 88b99c37 (#1056), each verified on a fixture page through the real CLI in a container built from the branch:

1. **Secret autocomplete tokens** — the attribute is now split into tokens and each checked (`section-billing shipping cc-number` masks; `cc-exp-month`/`cc-exp-year` added). Backspace in that field now reads `Effect: none observed … — focus: textbox "Card"`, no value.
2. **fullText word boundaries** — two runs that meet without whitespace get a space unless punctuation or a currency sign attaches to its neighbour. Real CLI on the fixture: `Never delete backups before` / `Price: $42.00 per month` (was `Neverdeletebackups` / `$42 .00` after the first attempt).
3. **Multi-line textarea** — the CLI renders the value with raw newlines and repeats the lines beneath with `LineBreak` nodes; the tree is now read as logical lines, `LineBreak` survives the merge as `\n`, and the repeated run is dropped as the parent's value. Output keeps `textbox "Notes" [ref=e3]: line one⏎line two⏎line three…` once, with nothing beneath it.
4. **Challenge-wall gate** — also requires ≤2,000 chars of text on the status-200 path (a challenge page is a few hundred); a 3,000-char article mentioning Turnstile with one link is no longer a wall.
5. **Focus identity and full-value hash** — the observation now carries `focusId` (`tag#id@census-index`) and `focusValueChars`/`focusValueHash`; #1056 uses them so Tab between same-named fields is a focus move and an edit beyond the 120-char preview is detected and labelled `(174 chars, first 120 shown)`.

One thing outside this PR that the fixture made visible: the CLI's own snapshot prints the value of a non-password text input, so `textbox "Card" [ref=e4]: 4242424242424242` appears in the tree regardless of the harness. Worth a separate look at masking `cc-*` and `one-time-code` fields in the snapshot too.

iddogino:
Second round, in b7650b1a — all three verified through the real CLI on a fixture built from your cases:

1. **Whitespace is no longer reconstructed.** The tree keeps the whitespace between inline elements as its own `StaticText " "` node (confirmed in the raw CLI output); compaction now keeps that node and merges runs exactly, with no separator heuristics. Real CLI on `Visit example.<strong>com</strong> and never run <code>rm</code> <code>-rf</code> there. <strong>Never</strong> <em>delete</em> <code>backups</code>.`:
   ```
   - StaticText "Visit example.com and never run rm -rf there. Never delete backups."
   ```
   A `<pre><code>` block keeps its newlines and indentation (`"def main():\n    print(\"hi\")\n"`), and the multi-line textarea case from round one still holds. A blank text node under an otherwise empty container is dropped as before.
2. **Walls need evidence a challenge is being served, not wording.** On a 200 page the only trigger is the vendor's own interstitial DOM — `#challenge-form`/`#challenge-running`/`#challenge-stage`/`#challenge-error-text` (Cloudflare), the `_Incapsula_Resource` frame, `#px-captcha`, `#datadome-captcha`/`captcha-delivery.com` — on a page with ≤3 controls. Deliberately not markers: `.cf-turnstile`/`cf-chl-widget` (the embedded widget on login forms) and the `/cdn-cgi/challenge-platform/` script (loaded by every protected page). Wording counts only with a 403/429/503, plus Google's `/sorry/` URL. Your short 200 article titled "How Cloudflare Turnstile can verify you are human" with one link now opens with no ⚠. Beatport's real challenge (`"Just a moment..." · HTTP 403`) still fires via the status path.
3. **Field values are raw.** Hash, length and preview are taken from the value as the page holds it; the preview is escaped for display by the effect line, so a newline reads as `\n` and indentation survives. Inserting a newline into `alpha beta` is now a change (verified on #1056 after rebase).

65 unit tests, typecheck and lint clean.

PR #1061, opened 2026-09-12
fix(browser): report an exec timeout as a timeout, never as the agent's own argv

iddogino:
## What

`execBrowser` builds its agent-visible error from `error.stdout || error.stderr || error.message`. On an exec timeout (30s kill) or a missing binary Node populates neither stdout nor stderr, so the agent got `Command failed: agent-browser <every argv element>` — its own eval script or typed text echoed back as the error, no exit code, and the word "timeout" nowhere.

New `describeExecFailure(err, verb, timeoutMs)` in `browser-output.ts`:

- process killed by the exec layer → `agent-browser eval produced no result within 30s and was stopped.` (plus anything the CLI did write)
- maxBuffer kill → `… was stopped: its output exceeded the buffer limit.`
- CLI wrote diagnostics → those, unchanged (`✗ Wait timed out after 25000ms`, `✗ Unknown ref: e31`)
- nothing written, no kill → `agent-browser open failed (exit ENOENT) without output.`

`error.message` (argv, CDP socket) never reaches agent text; it still goes to the container log.

## Why

Transcript-mining theme 21 (≈50/200 sessions): a 29-second eval kill read as "the fetch with return in async context is failing" (9 calls on a wrong hypothesis); a `keyboard type` echo read as "avoid all special characters" and the agent shipped hyphens instead of em-dashes to a live site. Facts only: the verb and the failure class the exec layer can vouch for.

## Tests

- `browser-output.test.ts`: timeout kill names verb + duration and contains neither `Command failed` nor the script nor `ws://`; CLI output kept alongside a kill; diagnostics pass through; exit code without output; buffer-overflow kill distinguished.
- typecheck + lint clean.

No UI change.

### DIFF
--- a/agent-container/src/browser-output.test.ts
+++ b/agent-container/src/browser-output.test.ts
@@ -40,45 +40,48 @@ describe('redactCdpUrls', () => {
 
 describe('describeExecFailure', () => {
   const script = '(async () => { const rows = await fetch("/api/rows").then(r => r.json()); return rows.length })()'
-  const timeoutKill = {
-    killed: true,
-    signal: 'SIGTERM',
-    code: null,
-    stdout: '',
-    stderr: '',
-    message: `Command failed: agent-browser --cdp ws://192.168.5.2:58686/devtools/browser/abc eval ${script}`,
-  }
+  const argvMessage = `Command failed: agent-browser --cdp ws://192.168.5.2:58686/devtools/browser/abc eval ${script}`
+  const NOT_KNOWN = 'whether the page-side action completed is not known'
 
-  it('names the verb and the timeout when the exec layer killed the process, and never echoes the argv', () => {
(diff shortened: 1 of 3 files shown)


iddogino:
Fixed.

- **Timeout vs external kill.** Only `err.killed === true` (Node's own timeout/maxBuffer kill) is treated as the exec ceiling. A signal without `killed` is reported as `agent-browser eval was terminated by SIGTERM after 13 ms with no output.`
- **Measured, not assumed.** `execBrowser` times the call; the text says `after 30012 ms`, not "within 30s".
- **No cancellation claim.** Both lines end with `The CLI call was stopped; whether the page-side action completed is not known.` — which is exactly what the exec layer can vouch for.

Tests: the 13 ms external SIGTERM case, the measured timeout, the buffer kill, and the unchanged pass-through cases.

PR #1062, opened 2026-09-12
fix(browser): fill and select read-backs state values, not causes

iddogino:
## What

**Fill read-back.** On divergence the line was `⚠ Field value is now "X" — differs from the requested "Y". The site reformatted, truncated (maxlength), or rejected the input. If this matters, fix it before moving on; keystroke-listening widgets may need browser_type instead.` It is now `⚠ Field value is now "X" — differs from the "Y" you sent.` The two values are the observation; the cause is not.

**Select read-back.** `judgeSelectCommit` compares the requested string against the option *value* read back. When the agent asked by visible label for the option that was already selected (`requested "California", element value is still "CA"`) it reported "did not commit" and asserted a cause ("the site reverted the change (React-controlled select) or the target is not a native <select>"). The route now runs one in-page check in that exact case — is there a `<select>` whose value is `CA` and whose option with that value is labelled `California`? — and reports it committed. The remaining failure text states what was observed (the value did not change) plus the existing custom-dropdown recipe, without naming a cause. Strings are JSON-embedded in the script.

`Selected "X" in @e3 — committed value verified: "…"` → `— the element's value is now "…"`, which is what was read. Prompt step 6 and `docs/browser-use.md` say the same.

## Why

Transcript-mining theme 24 (≈45/200 sessions): `'⚠ Field value is now "Reels" — differs from the requested "speakeasy". The site reformatted...' — the site did nothing of the sort; ref e10 had been rebound`; `'select reported success but the value did not commit (requested "California", element value is still "CA")' — CA *is* California`; one agent wrote "always re-snapshot after typing rather than trusting the browser_type readback" into its compaction summary. Facts only, per the browser-hints principle in CLAUDE.md.

Verified on the real CLI (agent-browser 0.27.2) that `get value` on a `<select>` returns the option value and `get text` returns every option's label, so the label has to come from the page; and that the select is not focused after `select`, so the check goes through `document.querySelectorAll("select")`.

## Tests

- `select-verify.test.ts`: label-match case passes; unchanged value without label match still fails with "did not change"; the reason no longer contains "reverted"; script embeds quotes safely; parser handles the CLI's double-encoded boolean.
- `browser-digest.test.ts`: divergence line contains both values and none of `reformatted`/`truncated`/`rejected`/`browser_type`.
- prompt tests (`agent-browser-upgrade`, `system-prompt-vars`) pass; typecheck + lint clean.

No UI change.

### DIFF
--- a/agent-container/docs/browser-use.md
+++ b/agent-container/docs/browser-use.md
@@ -90,8 +90,9 @@ number of interactive elements, and focus. `Effect: none observed within Nms
 failed: a highlight, a change inside an iframe or canvas, or a slow server
 all look like this. Snapshot or screenshot if you need to know; do not click
 again on that basis, which would undo a toggle. Fill results
-report the value the page actually committed. A fill warning means the page
-kept a different value—fix it before moving on.
+report the value the page actually committed. A fill warning means the field
+now holds a different value than you sent; both are shown and the cause is
+not known—check the field before moving on.
 
 Navigation makes existing refs stale. Re-snapshot when a result reports
 navigation, when a dialog or dynamic view changes the relevant controls, or
--- a/agent-container/src/browser-digest.test.ts
+++ b/agent-container/src/browser-digest.test.ts
@@ -83,7 +83,7 @@ describe('formatFillReadback', () => {
   it('warns on divergence (maxlength truncation, reformatting, rejection)', () => {
     const msg = formatFillReadback('x'.repeat(100), 'x'.repeat(75))
     expect(msg).toContain('⚠')
-    expect(msg).toContain('differs from the requested')
+    expect(msg).toContain('differs from the')
   })
 
   it('says so when the value cannot be read back', () => {
@@ -117,5 +117,16 @@ describe('parseScrollInfo / formatScrollDigest', () => {
     expect(formatScrollDigest({ y: 0, viewportHeight: 800, pageHeight: 5400 })).toContain('(top of page)')
     expect(formatScrollDigest({ y: 4600, viewportHeight: 800, pageHeight: 5400 })).toContain('(bottom of page)')
     expect(formatScrollDigest(null)).toBe('')
+  })
+})
+
+describe('formatFillReadback states values only', () => {
+  it('names both values on divergence and asserts no cause', () => {
+    const msg = formatFillReadback('speakeasy', 'Reels')
+    expect(msg).toContain('⚠ Field value is now "Reels"')
+    expect(msg).toContain('"speakeasy" you sent')
+    for (const claim of ['reformatted', 'truncated', 'rejected', 'browser_type']) {
+      expect(msg).not.toContain(claim)
+    }
   })
 })
--- a/agent-container/src/browser-digest.ts
+++ b/agent-container/src/browser-digest.ts
@@ -90,6 +90,13 @@ function displayValue(value: string): string {
  * programmatically and reports success regardless of what the page kept —
  * maxlength truncation, JS reformatting, and keystroke-only widgets all
  * silently diverged in the audit (F6).
+ *
+ * On divergence the line states the two values and nothing else. The old
+ * text asserted a cause ("The site reformatted, truncated (maxlength), or
+ * rejected the input") that the read-back never established — a rebound
+ * ref reading a neighbouring element produced the same line, and agents
+ * carried the invented site behaviour for hundreds of calls (mining theme
+ * 24). The agent can see the difference; what it means is its call.
  */
 export function formatFillReadback(requested: string, committed: string | null): string {
   if (committed === null) {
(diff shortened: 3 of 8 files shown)


iddogino:
Both fixed.

**Sibling dropdown (P1).** The page-wide `querySelectorAll("select")` probe is gone. In the one ambiguous case (value unchanged and not the requested string) the route now runs `focus @ref` and reads `document.activeElement` — its own value and `selectedOptions[0].text` — and matches only when *that* element holds the read-back value under the requested label. A focus that lands anywhere else (another element, a frame's `<iframe>`, a detached ref) yields null, so the label stays unknown and the result is the plain "did not change" failure. Your scenario — target reverted to New York, another dropdown on California — reads `{value:"NY", label:"New York"}` and fails.

**Unreadable read-back (P2).** `select reported success but the element's value could not be read back afterwards, so the selection is unverified.` No element-type diagnosis; the custom-dropdown recipe stays as the conditional it always was.

Tests cover the reverted-target case, the focus-elsewhere null, and that the null-read text names no cause. Re-verified on the merged image.

iddogino:
Both addressed.

**Target identity (P1).** Focus is gone from the check. The route asks the CLI for the ref's rectangle (`get box @ref --json`, which resolves the ref the same way `select` did) and the in-page script reads `document.elementFromPoint` at that rectangle's centre, accepting it only if it is a `<select>` whose own `getBoundingClientRect` matches the box within 2px (both viewport- and document-relative readings of the CLI box are tried). Nothing else on the page is consulted and an onfocus redirect has nothing to redirect. If what sits at that point is not a `<select>` with that rectangle, the label stays unknown and the result is the plain "did not change" failure.

Verified on the merged image with your scenario — target `<select>` whose `onfocus` moves focus to the sibling dropdown holding California, and a change handler that reverts to New York:

```
browser_select(@e2, "California") → Error: select reported success but the element's value did not change (still "NY"; requested "California").
browser_select(@e2, "New York")   → Selected "New York" in @e2 — the element's value is now "NY".
```

**Option label (P2).** The script reads `option.label` (label attribute, falling back to text). One finding while verifying it: the pinned CLI's own `select` does not match a `label` attribute — `select @e4 East` on `<option value="EA" label="East">EA</option>` returns `✗ No option matched ["East"]. Available options: WE ("WE"), EA ("EA")` before the read-back ever runs. So with this CLI the already-satisfied case can only be reached by value or by text, both of which pass. `.label` is the right read regardless and is what ships.

iddogino:
Fixed, and this time without any page-side lookup — the third one (page-wide search, focus, geometry) to identify the wrong element made the pattern clear: nothing that runs as page script can resolve a CLI ref.

The route now reads the target's own option list with `get html @ref` — resolved by the CLI against the same ref as the `select` and the `get value` read — and checks whether the option holding the read-back value carries the requested label (`option.label` semantics: label attribute, else text; entities decoded). A sibling, a focused, or a covering dropdown cannot answer for the target because the page is never asked.

Your scenario on the real CLI — target `<select>` and a covering `<select>` with identical bounds (`position:absolute`, same box), the cover holding California, the target reverting to New York:

```
- combobox [expanded=false, ref=e2]: New York
- combobox [expanded=false, ref=e3]: California
browser_select(@e2, "California") → Error: select reported success but the element's value did not change (still "NY"; requested "California").
browser_select(@e2, "New York")   → Selected "New York" in @e2 — the element's value is now "NY".
```

The onfocus-redirect and sibling-dropdown pages from the earlier rounds still fail/pass the same way on the same image. Tests parse the captured `get html` shape, attribute order and quoting variants, `label` attributes, entities and optgroups, and the NY-vs-California mismatch.

iddogino:
Both fixed in 10289db41.

- **Duplicate values.** `targetOptionMatches` now requires every option carrying the read-back value to carry the requested label; any conflict leaves the verification unknown (no match). Duplicates that agree on the label still verify.
- **Comments.** HTML comments and the inert contents of `<template>`, `<script>` and `<style>` are stripped before the option scan.

Both reproduced on the real CLI with an image built from the branch (`Dup`: California and New York both value `"0"`, reverting to New York; `Com`: California only inside a comment, reverting to New York):

```
browser_select(@e2 "Dup", "California") → Error: … value did not change (still "0"; requested "California").
browser_select(@e3 "Com", "California") → Error: … value did not change (still "NY"; requested "California").
browser_select(@e3 "Com", "New York")   → Selected "New York" in @e3 — the element's value is now "NY".
```

The overlapping-select, onfocus-redirect and sibling-dropdown pages from the earlier rounds behave the same on the same image. Tests cover the conflicting-label case, agreeing duplicates, and option-shaped text inside comment/template/script.

iddogino:
Fixed in 11072a5e5. The attribute regex is gone; the tag is tokenized character by character (quotes honoured, `>` and `=` allowed inside values, first occurrence of a name wins as in the DOM), and an option's text now ends at the next option/optgroup/select tag, so the `</option>` HTML lets authors omit is handled as well.

Your case on the real CLI, image built from the branch (`<option title="Search label='California'" value="0" selected>New York</option>`, reverting to New York):

```
browser_select(@e2, "California") → Error: … value did not change (still "0"; requested "California").
browser_select(@e2, "New York")   → Selected "New York" in @e2 — the element's value is now "0".
```

Tests: `label=` inside another attribute's value, `value=` inside a `data-*` value, `>`/`=` inside quoted values, unquoted values, repeated attributes, attribute-looking text content, and omitted closing tags. The duplicate-value, comment, overlap, onfocus and sibling pages from the earlier rounds still behave the same on the same image.

PR #1065, opened 2026-09-12
fix(browser): browser_wait reports how long it waited; refuses syntax the CLI cannot take

iddogino:
## What

New `wait-target.ts` classifies `browser_wait`'s `for` argument and words the result:

| input | result |
|---|---|
| `.loaded` | `Selector ".loaded" matched after 1840 ms.` |
| `body` (already present) | `Selector "body" matched after 3 ms.` |
| `1500` | `Waited 1503 ms.` |
| `networkidle` reached | `Load state "networkidle" reached after 120 ms.` |
| `networkidle` timed out | `Load state "networkidle" was not reached within 25010 ms.` (not an error, as before — but no longer a silent success) |
| `text=Trigger`, `role=button`, `button:has-text("Save")`, `div >> x` | refused before the CLI runs: not CSS; names `browser_run(["wait","--text","<text>"])` / `["wait","--url","<pattern>"]` |
| selector timeout | the CLI's `✗ Wait timed out after 25000ms` plus `Page: <url> · readyState <state>` |

The route measures wall-clock around the exec; the tool renders from the same classifier. Tool description and prompt line updated (the CLI's `wait --help` confirms the `<ms>`, `--text`, `--url`, `--load` modes).

## Why

Transcript-mining theme 10 (≈74/200 sessions, ~380 wasted calls): *"used a tool that does nothing 102 times, saying 'let me wait longer' after each one … The harness never once said 'this didn't wait'"*; *"'Wait condition "body" satisfied.' — technically true, functionally a lie; it resolved in 0ms and the agent believed its rate-limit mitigation was in effect"*; *"repeated `text=Trigger` nine times across three hours, treating the 25s failure as a sleep primitive"*.

Everything printed is a measurement or a syntax fact. No claim about the page beyond its URL and readyState at the moment of a timeout.

## Tests

- `wait-target.test.ts`: classification (CSS, ms, load states, Playwright prefixes and pseudo-selectors, CSS with `=` not refused) and wording (0 ms visible, sleep, load timeout never "reached").
- `tools/browser.test.ts`: measured elapsed rendered; Playwright syntax refused without a server call; load timeout reported as not reached.
- prompt tests pass; typecheck + lint clean.

No UI change.

### DIFF
--- a/agent-container/src/server.ts
+++ b/agent-container/src/server.ts
@@ -813,7 +813,10 @@ import { resolveRunCommandArgs } from './browser-command-args';
 import { validatePressKey } from './press-key';
 import { prepareEvalScript, finalizeEvalOutput, evalErrorHint } from './eval-script';
 import { judgeSelectCommit, parseSelectOptions, targetOptionMatches, SELECT_COMMIT_SETTLE_MS } from './select-verify';
-import { classifyWaitTarget } from './wait-target';
+import { classifyWaitTarget, WAIT_PAGE_PROBE_SCRIPT, parseWaitPageProbe } from './wait-target';
+
+/** Budget for the page probe around a wait — independent of the exec ceiling. */
+const WAIT_PAGE_PROBE_TIMEOUT_MS = 3000;
 import { resolveCommittedValue } from './field-value-readback';
 import { capBrowserOutput, redactCdpUrls, describeExecFailure, BROWSER_EXEC_TIMEOUT_MS, MAX_BROWSER_OUTPUT_CHARS, MAX_BROWSER_ERROR_CHARS } from './browser-output';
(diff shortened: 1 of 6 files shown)


iddogino:
Both fixed.

- **Quoted strings.** The `:has-text(` / `>>` test now runs on the selector with quoted strings (double, single, with escapes) blanked out. `button[data-note="a>>b"]` and `[title=':has-text(x)']` pass through as CSS; a bare `div >> text=Hi` is still refused. The `text=`/`role=` prefix test stays on the raw start, where no CSS selector can begin.
- **Probe budget.** The page probe after a failed wait runs only when the CLI itself printed `Wait timed out` (so the browser was answering), and through `execBrowser`'s new `timeoutMs` option at 3 s. A hung browser now costs one exec ceiling, and the result simply has no `Page:` line.

Also, since the #1070 review pointed out that a wait's success result carried no URL: a successful wait now ends with `Page: <url>`, read through the same 3 s probe.

PR #1067, opened 2026-09-12
fix(browser): browser_get_state is one observation — URL from its own snapshot, one line per failure cause

iddogino:
## What

`browser_get_state` ran `get url`, `screenshot` and `snapshot` in `Promise.all` and concatenated whatever came back, `isError` never set. Now:

- **Snapshot first, then screenshot**, sequentially. The URL comes from the snapshot's own page probe (the same one that prints the `[page]` status line), so the `**Current URL:**` line can no longer disagree with the tree under it. One fewer CLI round trip.
- **One line per distinct failure cause.** A dead browser fails every leg with the same message; that is one fact: `**Error (snapshot and screenshot):** ✗ Auto-launch failed: … 410 Gone`. A partial failure stays a section error (`**Error (screenshot):** …`) with the rest of the result intact.
- **`isError: true` when every requested leg failed** (with `screenshot: false`, that is the snapshot alone).

## Why

Transcript-mining theme 23 (≈48/200 sessions): *"'Current URL: …CorpSearchResults.aspx' next to a screenshot and snapshot that are unambiguously the Rhode Island … entity search"*; *"browser_get_state on a dead browser printed the same 'CDP WebSocket connect failed: HTTP error: 410 Gone' three times as three separate field errors, obscuring the one actionable fact"*; *"'Current URL: Error - ✗ Auto-launch failed…' alongside a complete, correct snapshot"*.

## Tests

- `tools/browser.test.ts`: URL taken from the snapshot's page and `get url` never called, snapshot before screenshot; one cause failing both legs → one error line + `isError`; partial failure → section error, no `isError`; `screenshot:false` single failure is total.
- typecheck + lint clean.

No UI change.

### DIFF
--- a/agent-container/src/tools/browser.test.ts
+++ b/agent-container/src/tools/browser.test.ts
@@ -315,3 +315,101 @@ describe('browser_wait result carries the page URL', () => {
     expect(result.content[0].text).toBe('Selector "#dash" matched after 900 ms. Page: https://a.com/dashboard')
   })
 })
+
+describe('browser_get_state coherence', () => {
+  afterEach(() => vi.unstubAllGlobals())
+
+  const json = (body: unknown, status = 200) =>
+    new Response(JSON.stringify(body), { status, headers: { 'Content-Type': 'application/json' } })
+
+  it('takes the URL from the snapshot it shows, and never calls get url', async () => {
+    const calls: string[] = []
+    vi.stubGlobal('fetch', vi.fn(async (url: string) => {
+      calls.push(String(url))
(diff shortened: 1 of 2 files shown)


iddogino:
Fixed. A screenshot leg now fails in three ways, all counted: the route errored, no path came back, or the file could not be read/resized (`**Error (screenshot):** screenshot file could not be read: <path>`). Your case — snapshot failed and the file unreadable — renders both error lines and sets `isError`. Tests cover that case and the missing-path case (partial, no `isError`).

PR #1070, opened 2026-09-12
docs(browser): correct four web-browser prompt claims that are false today

iddogino:
## What

Four statements in `web-browser-agent-prompt.md` describe behaviour the harness does not have. Each is corrected to what is true; no rule is loosened.

| was | now |
|---|---|
| "NEVER close the browser. **You do not have the browser_close tool.**" | "NEVER close the browser. Do not call `browser_close` — the parent agent manages browser lifecycle." The subagent's tool list (`mcpToolNames('browser', browserMcpTools)`) does include `browser_close`. |
| "ALWAYS report the current URL … (**use `browser_run("get url")`**)" | Same rule; take the URL from the `[page]` status line of the last snapshot or the last action result — no extra call. |
| "**Check tabs every 5 actions.** … The snapshot footer also shows your tab count." | Replaced by what the runtime does: new-tab notices on click/press results, `[Tabs: N open]` only when more than one tab is open, warnings when the count is high. List tabs when you need an id or a result says the count is high. The MANDATORY label goes; the limit/close/dedupe rules stay. |
| "`browser_screenshot(full?)` — Take a screenshot (returns file path; **use Read to see the image**)" / "`Read` — Read screenshot files" | The image is returned inline; `annotate: true` maps labels to refs. `Read` is for workspace files such as downloads. |

## Why

Transcript-mining "prompt guidance that cost time": *"24+ sessions of no-op `tab` listings, one 30 s timeout, and one session that halted a task citing 'risk exceeding the browser tab limit'"* — `formatTabStatus` returns `''` at ≤1 tab, so the mandated probe never had a footer to confirm it. The URL rule was *"one guaranteed extra call per session restating what the last digest printed, and a fabricated URL when answered from memory"*. Theme 25: the prompt's tool block is the agent's only view of the tools at decision time; a claim it cannot verify becomes a belief.

## Tests

- `agent-browser-upgrade.test.ts` and `system-prompt-vars.test.ts` (prompt/tool consistency) pass.
- Prompt-only change; typecheck + lint clean.

No UI change.

### DIFF
--- a/agent-container/src/web-browser-agent-prompt.md
+++ b/agent-container/src/web-browser-agent-prompt.md
@@ -17,7 +17,7 @@ You are a web browser automation agent. You receive high-level objectives and ac
 - `browser_upload(filePath, selector?)` — Upload a local file into an `<input type="file">`. Use this for Dropbox, Box, Dropzone, and any file picker flow.
 - `browser_download(url, filename?)` — Download a file/image/asset through the browser (cookies + login state apply) into `/workspace/downloads/`. To save an image: get its URL first (`browser_run("get attr @e5 src")` or `browser_eval`), then download it. Report the returned path to the parent agent.
 - `browser_wait(for)` — Wait for a CSS selector to appear, or a number of milliseconds. The result reports how long it actually took (a selector already on the page matches in ~0 ms — that is not a delay). Not for load states — `browser_open` already waits for the page to load. To wait for text: `browser_run(["wait","--text","<text>"])`.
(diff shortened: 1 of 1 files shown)


iddogino:
Reworded in both places (Critical Rules and Response Format): reuse the URL from your most recent snapshot or action result; if anything since then could have navigated — a wait, a click whose result did not show the URL, a redirect you were expecting — take a snapshot first and report the URL from its status line. Separately, #1065 now puts `Page: <url>` on every successful `browser_wait` result, which closes the exact sequence you reproduced.

PR #1071, opened 2026-09-12
fix(browser): iframe footer lists only frames the tree could not read; drop the coordinate-click recipe

iddogino:
## What

**The footer under the tree.** It listed every cross-origin frame as `contents NOT in this snapshot (cross-origin)` and appended "click into it by coordinates, then use browser_type". Now `formatIframePlaceholders(iframes, tree)` checks the tree itself: a visible DOM iframe is listed only when the tree has no `Iframe` node with that accessible name carrying children. The line is `  - iframe "Secure payment input frame" (js.stripe.com) · cross-origin` under "Frames on this page whose contents are not in this tree:", with no recipe. A frame that is in the tree with refs gets no footer at all.

**The origin check.** `sameOrigin` was `!!f.contentDocument`, which is null for a sandboxed frame on the page's own host. It is now that OR `new URL(f.src, location.href).origin === location.origin`.

**The advice.** Three tool descriptions and the eval error hint prescribed a coordinate click. The CLI has no such command (`mouse` is move/down/up; the mined sessions got `Unknown subcommand: click`). They now say what is true: frame fields normally have refs in the snapshot — use them; if a frame is listed as unreadable, `browser_run mouse move/down/up` then `browser_type`. `docs/browser-use.md` and the prompt line agree.

## Why

Verified on the real CLI (agent-browser 0.27.2) with a page embedding a cross-origin frame (`localhost` page, `127.0.0.1` frame): `snapshot -i -c` prints `- Iframe "Secure card frame" [ref=e2]` with `textbox "Card number" [ref=e7]` under it; `fill @e7 4242…` then `get value @e7` returns the number; `click @e6` on the inner button works. The same probe shows `contentDocument` false for a sandboxed same-host frame while the URL origin matches. (`frame <selector>` returned "Frame not found" for both `iframe` and `iframe[title=…]`, so it is deliberately not recommended.)

Transcript-mining theme 11 (≈70/200 sessions): *"The snapshot listed the iframe's buttons with usable refs (e50, e51, e52) and then said 'iframe … contents NOT in this snapshot (cross-origin)'" → tried JS, then raw mouse coordinates, then finally @e50, which worked*; *"the agent's single wrongest belief ('the iframe is cross-origin') was harness-induced"*; ~60 lines of credit-card advice on a Gmail read task.

## Tests

- `snapshot-format.test.ts`: merged frame → no footer; unreadable frame → fact line with host and `cross-origin`, none of the old claims; frame in tree but empty → listed; same-origin unreadable frame without the label; untitled frames; blank frames; empty input.
- `page-observer.test.ts`: sandboxed same-host and relative-src frames report `sameOrigin: true`; Stripe stays `false`.
- `eval-script.test.ts`: hint names refs, not coordinates. Prompt tests pass; typecheck + lint clean.

No UI change.

### DIFF
--- a/agent-container/docs/browser-use.md
+++ b/agent-container/docs/browser-use.md
@@ -127,9 +127,10 @@ Do not ask the user to paste credentials into chat. The browser profile retains
 successful sessions. Follow the system prompt's confirmation rules for
 submissions with financial, legal, destructive, or externally visible impact.
 
-Cross-origin iframe content is unavailable to `browser_eval` and may not appear
-as actionable snapshot fields. For embedded payment fields, click the iframe
-field by the available mechanism and use `browser_type`.
+Cross-origin iframe content is unavailable to `browser_eval`, but the snapshot
+normally merges frames — payment frames included — with working refs. Use those
+refs; for keystroke-driven fields, click the ref and use `browser_type`. A frame
+the tree could not read is listed under the snapshot.
 
 ## Uploads and Downloads
 
--- a/agent-container/src/eval-script.test.ts
+++ b/agent-container/src/eval-script.test.ts
@@ -125,7 +125,8 @@ describe('evalErrorHint', () => {
   it('adds the top-frame hint for null-property errors', () => {
     const hinted = evalErrorHint("✗ Evaluation error: TypeError: Cannot read properties of null (reading 'click')")
     expect(hinted).toContain('TOP frame only')
-    expect(hinted).toContain('browser_type')
+    expect(hinted).toContain('refs from browser_snapshot')
+    expect(hinted).not.toContain('coordinate')
   })
 
   it('adds the quoting hint for invalid-selector errors', () => {
--- a/agent-container/src/eval-script.ts
+++ b/agent-container/src/eval-script.ts
@@ -124,7 +124,7 @@ export function finalizeEvalOutput(output: string): string {
 /** Append a hint to known-confusing eval error shapes. */
 export function evalErrorHint(error: string): string {
   if (/Cannot read propert(y|ies) of (null|undefined)/i.test(error)) {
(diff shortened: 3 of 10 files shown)


iddogino:
Fixed. Three changes:

1. The observer reports each frame's accessible name as `aria-label || title` (the tree prints the accessible name, so the comparison has to use it).
2. The tree parser unescapes the quoted name (`\"` and `\\`), so `Iframe "Card \"main\" frame"` matches the DOM frame `Card "main" frame`.
3. The claim is suppressed whenever the match is ambiguous. A frame is listed only when it has a name, no `Iframe` node with that name carries children, **and** the tree holds fewer child-bearing `Iframe` nodes than the page has visible frames. If every visible frame is accounted for by a merged node, a name mismatch is my matching problem, not the tree's, and nothing is printed. Unnamed frames never produce a line.

Tests cover the aria-label override, the escaped-quote title, and the "all frames accounted for" suppression, alongside the earlier cases. Re-verified on the merged image.

iddogino:
Fixed. Two changes:

1. The observer computes the accessible name the way the tree does: `aria-label`, then `aria-labelledby` (the referenced elements' text, joined), then `title`. Your aria-labelledby frame now matches its tree node by name.
2. The count is gone. A frame is listed only when the identification is complete: it has a name, no tree `Iframe` node carries that name with children, **and every tree `Iframe` node's name matches some visible frame** — so no unmatched node could be this frame under a name I failed to compute. One unaccounted-for tree node suppresses every claim. Unnamed frames never produce a line.

Your scenario on the merged image (readable frame named via `aria-labelledby="lbl"` → "Billing card", beside an empty frame titled "Empty one"):

```
- Iframe "Billing card" [ref=e2]
  - textbox "Card number " [ref=e6]
  - button "Inner OK" [ref=e5]
- Iframe "Empty one" [ref=e3]
- button "Pay" [ref=e4]
Frames on this page whose contents are not in this tree:
  - iframe "Empty one" (127.0.0.1:8765) · cross-origin
```

The readable frame gets no line; the empty one is listed because its own node, matched by name, has no children. Tests cover this scenario, the unmatched-node suppression, and the earlier cases.

PR #1072, opened 2026-09-12
fix: carry compaction state to late-joining session streams (SUP-822)

JeremyBischoff:
A session stream that opens after compaction has started shows "Working" until the summary lands, and one that misses the finish keeps a spinner until the next turn, because `compact_start` and `compact_complete` are one-shot broadcasts and nothing else told a late-joining client where the session stood. The stream's connected snapshot and keep-alive ping now carry `isCompacting`, the client takes its value from them, and the server keeps that flag honest across the turn's lifecycle instead of only at the next turn start. Found while building the fork-and-compact flow in #1044, which opens a fresh copy and compacts it: the copy's stream connects after the compaction begins.

Closes https://linear.app/datawizz/issue/SUP-822/late-joining-session-streams-do-not-learn-compaction-state
Related to https://linear.app/datawizz/issue/SUP-793/stale-notice-continue-in-a-compacted-fork-fork-compact

Footprint: production +75 / -35 across 4 files, tests +328 / -69 across 5.

```
stream opens
  → connected { isActive, isCompacting, ... }   ← was missing isCompacting
  → every 30s ping { isActive, isCompacting }    ← same
client
  → snapshot value wins. An older host that omits it falls back to the local
    value while the turn runs, and clears it once the turn's output ended
```

- Server: `isSessionCompacting` reads the persister's flag only while the session is active. A stop, a turn result, a replaced process and a runtime-started turn reset it. A transport reattach to the same live turn keeps it. The two paths that end a compaction with no idle frame following (a soft stop that spares background work, a replaced process) broadcast `compact_complete` so open tabs drop their indicator too.
- Client: the ping corrects a missed `compact_start` or `compact_complete` within one interval, and a finished compaction invalidates the transcript so its boundary shows.
- A tool result that follows a summary missed across a reattach still reaches the tool-result handler: the compact-summary branch clears the flag and falls through instead of exiting early.
- The message list's boundary-count safety net is removed. Its baseline started at zero, so a page opened mid-compaction cleared the snapshot's value the moment a transcript with earlier boundaries loaded. The ping covers the missed-frame case it existed for, with fewer lines.

A mock-runtime e2e reloads the page mid-compaction and expects the indicator to read "Compacting" from the snapshot. Live smoke against the staging Platform: reload 0.3s into a real compaction shows "Compacting..." on the fresh stream, then the boundary lands.

No UI change beyond the indicator showing "Compacting..." where it showed "Working".

### DIFF
--- a/e2e/specs/message-queueing.spec.ts
+++ b/e2e/specs/message-queueing.spec.ts
@@ -78,11 +78,9 @@ test.describe('Message queueing while agent is working', () => {
     // "Compacting..." status must survive it, with the ghost below the line —
     // the queued message is picked up on the far side of the boundary.
 
-    // Settle a first turn before starting the one that compacts. compact_start
-    // is a one-shot broadcast and the connected frame carries no compaction
-    // snapshot, so a client that subscribes late never learns compaction began.
-    // The first message is what creates the session and navigates to it — under
-    // CI load that hand-off outran the mock's compacting status and the whole
+    // Settle a first turn before starting the one that compacts. The first
+    // message is what creates the session and navigates to it — under CI load
+    // that hand-off outran the mock's compacting status and the whole
     // compaction came and went before the stream was listening.
     await sessionPage.sendMessage('hello there')
     await expect(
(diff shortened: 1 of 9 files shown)


JeremyBischoff:
Closing. The remaining symptom on main is a late-joining tab showing "Working" instead of "Compacting..." for the length of one compaction, and it corrects itself. Main already clears a stale "Compacting..." on session idle, session error, the ping, and the message list's boundary check. That does not justify a turn-lifecycle change in the message persister.

PR #1081, opened 2026-09-14
fix: stop reporting a sleeping/waking cloud workspace as an error (SUP-849)

yiw190:
Fixes [SUP-849](https://linear.app/datawizz/issue/SUP-849/desktop-app-reports-cloud-workspace-sleepingwaking-503-deployment).

## What was broken

A desktop app connected to a cloud workspace keeps polling after the workspace goes to sleep. Every poll fails with `Error: deployment_unavailable` (or a hook's fallback text like `Failed to fetch settings`), each failure is reported to Sentry, and mutation toasts show the raw wire code.

Sentry: one root cause split across 18 issue groups in `electron` — [ELECTRON-9B](https://datawizz.sentry.io/issues/ELECTRON-9B) (248 events / 25 users), [ELECTRON-6G](https://datawizz.sentry.io/issues/ELECTRON-6G) (3109 events / 59 users), [ELECTRON-KX](https://datawizz.sentry.io/issues/ELECTRON-KX) (1603), [ELECTRON-A3](https://datawizz.sentry.io/issues/ELECTRON-A3) (1496), [ELECTRON-10X](https://datawizz.sentry.io/issues/ELECTRON-10X), and more: [search](https://datawizz.sentry.io/issues/?project=4511191116218368&query=deployment_unavailable).

## Why

When an org's workspace is not `ready` (cold-idle put it to `sleeping`, or it is `waking` / `provisioning` / `stopping`), the Cloudflare router worker answers every non-HTML request itself:

```
503 {"error":"deployment_unavailable","state":"sleeping"}
x-workspace-unavailable: sleeping
```

`cloud-proxy.ts` only special-cases 401 and passes this through. Every renderer hook does `if (!res.ok) throw new Error(err.error || 'Failed to load X')`, and the global `QueryCache` / `MutationCache` `onError` in `query-client.ts` reports every failure to Sentry. Nothing in `SuperAgent/src` referenced the header or the error code before this PR.

## Repro

1. Connect the desktop app to a cloud workspace.
2. Let the org idle until the route is `sleeping` (or set it via the route admin API).
3. Leave the app open, or open Settings → Account.
4. Every query fails, Sentry gets one event per poll.

## What changed

- `renderer/lib/workspace-unavailable.ts` (new): `WorkspaceUnavailableError(state)` with a human message per state, and `isWorkspaceUnavailableError`.
- `renderer/lib/api.ts`: `apiFetch` classifies the reply once, keyed on the `x-workspace-unavailable` header (not on body text or status — the router also marks a 502 `unreachable`), leaves a breadcrumb, and rejects with `WorkspaceUnavailableError` before any hook can turn the body into an `Error` of its own. Exported as `handleWorkspaceUnavailableResponse` for the other boundary.
- `renderer/lib/upload.ts`: the XHR path's `checkResponse` runs the same classification.
- `api/index.ts`: the local API's CORS policy exposes `x-workspace-unavailable` (`Access-Control-Expose-Headers`). The packaged renderer is `file://` calling loopback cross-origin, so without this the browser hides the header from `headers.get()` / `getResponseHeader()` and classification never runs. Header name lives in `shared/lib/workspace-unavailable-header.ts` so API and renderer read one constant.
- `renderer/lib/error-reporting.ts`: Sentry `beforeSend` drops `WorkspaceUnavailableError`, so query, mutation, and unhandled-rejection paths are all covered in one place.
- Tests: `api/cors-expose-headers.test.ts` asserts the real app's cross-origin response lists the header (fails without the CORS change); new cases for the header path (state carried, human message, header-not-status, plain 503 left alone, not treated as a rejected session), the `beforeSend` filter, the XHR boundary, and the error class. Existing bare `fetch` fakes gained a `headers` field since the real `Response` always has one.

Behavior for a user: mutations against a sleeping workspace now toast "Your cloud workspace is asleep. Please try again in a moment." instead of `deployment_unavailable`. Queries still fail silently and render their inline states, as before.

## Not in this PR

Surfacing a single "workspace is asleep / starting" state in the desktop UI, pausing polling, and calling the router's `/__gamut/wake` so the app self-heals — tracked as the follow-up in SUP-849.

## Verification

- `npm run typecheck` clean
- `eslint` clean on touched files
- `vitest run src/renderer`: 323 files / 3666 tests passing


src/renderer/lib/api.ts
```diff
@@ -15,10 +19,25 @@ export async function apiFetch(
 ): Promise<Response> {
   const baseUrl = getApiBaseUrl()
   const response = await fetch(`${baseUrl}${path}`, init)
```
iddogino, line 21: ### [P1] Expose the workspace header through CORS

The packaged desktop renderer calls loopback across origins, but the API's CORS policy (`src/api/index.ts:69`) does not expose `x-workspace-unavailable`. Consequently, both `response.headers.get()` here and XHR's `getResponseHeader()` return `null`. Classification never runs, so the original Sentry noise and raw upload errors remain.

Please expose the header on the local API/proxy response. Chromium reproduced both failures using this PR's actual fetch and upload helpers; exposing the header corrected both.

iddogino (changes requested):
Requesting changes because the desktop renderer cannot read the header used to classify unavailable workspaces. Please expose `x-workspace-unavailable` through the local API/proxy's CORS policy and cover the browser boundary.

Validation: 112 focused tests passed. A Chromium reproduction using the PR's actual fetch and upload helpers confirmed that the current CORS policy hides the header; exposing it corrected both paths.

src/renderer/lib/api.ts
yiw190: Fixed in 9cb9ef7ca. `cors()` in `src/api/index.ts` now sets `exposeHeaders: [WORKSPACE_UNAVAILABLE_HEADER]`; the header name moved to `src/shared/lib/workspace-unavailable-header.ts` so the API and renderer share one constant. Added `src/api/cors-expose-headers.test.ts`, which requests the real app with `Origin: null` and asserts `Access-Control-Expose-Headers` lists `x-workspace-unavailable` — verified it fails without the CORS change and passes with it.

PR #1087, opened 2026-09-14
SUP-831: Refactor chat integrations into AgentIntegration hierarchy

iddogino:
Application lifecycle, event routing, and actor sessions now use `AgentIntegration`, so another integration family can reuse the coordinator without exposing chat transport methods to application callers. `ChatAgentIntegration` supplies the existing chat behavior, and Slack, Telegram, and iMessage inherit from it.

- Introduce a provider registry with metadata available before connection, normalized input/response events, routing and connection-independent access/session policy, family tools, and explicit turn completion/failure output.
- Move attachment preparation, streaming delivery, indicators, commands, and callback normalization into the chat family. Move the setup-only contact-card hook into iMessage.
- Migrate startup, desktop resume, and API callers to the shared manager and the latest `AgentActor` boundary. Preserve reconnect/retry handling, model overrides, access approvals, and session recovery.
- Use `onEvent` as the single event subscription API. Log and report event-processing failures without changing connection status; retain error status and notifications for connector failures. Outbound session recording works during reconnection and still enforces current access restrictions.
- Keep existing credentials, storage rows, session mappings, HTTP paths, and setup flows compatible. No database migration or customer reinstallation is required.

Based on latest `main` at `570c981c9`. Implements [SUP-831](https://linear.app/datawizz/issue/SUP-831/refactor-introduce-agentintegration-hierarchy-and-migrate-chat). TaskManagerAgentIntegration and Linear are the separate [SUP-832 follow-up](https://linear.app/datawizz/issue/SUP-832/linear-add-taskmanageragentintegration-and-per-agent-linear-identities).

Architecture and contract reference: `docs/agent-integrations.md`.

Validation:

- Full unit suite: **12,203 passed**, 48 skipped, 1 todo, across 742 files. Run with `SUPERAGENT_DATA_DIR=$(mktemp -d) npm run test:run -- --maxWorkers=4` so host-token fixtures have an isolated writable data directory.
- New provider-neutral tests exercise routing, reply destinations, complete-turn delivery, access enforcement, installation isolation, teardown, event-error isolation, and outbound sessions without a connection using an adapter with no chat methods. Chat regression tests also cover persisted timeouts, session metadata, and access denial during reconnection.
- `npm run typecheck` and `npm run lint` pass (0 lint errors).
- `npm run build:api` and `npm run build:electron` pass, including runtime dependency checks.


iddogino:
1. Response-event errors now flip the integration to error status. emitEvent awaits the full handleInteractiveResponse (the doc comment at agent-integration.ts:38 says it only awaits queue acceptance). A rejection flows to emitError, and the manager's error handler writes
     status = 'error' and fires a notification. The old code only logged and reported. Uncaught paths are DB reads and inputs.claim, so it is unlikely, but the escalation is a semantic change. Same for a throw inside resolveRoute on the input path. Wrap the handler in
     connectIntegration with the same catch-and-report the old subscriptions had.
  2. Dual emission is dead weight. emitMessage and friends emit through the new emitEvent and then loop the old handler arrays. Nothing outside tests subscribes to onMessage, onInteractiveResponse, or onTypingHint any more. Delete them, plus the four unused unsubscribe fields
     on ManagedConnector and the unused instance describeTarget at chat-agent-integration.ts:250.
  3. ensureSession assumes a live connection with two non-null assertions at agent-integration-manager.ts:346. The old version read the timeout from the row. A health-check rebuild between the send and the transcript write throws a TypeError instead of logging the message.

PR #1096, opened 2026-09-15
SUP-861: Preserve Slack thread participation across restarts

iddogino:
Slack's “only respond when mentioned” filter forgot joined threads whenever the connector was recreated. A thread that accepted unmentioned follow-ups before restart silently ignored them afterward, even though its agent session survived.

Persist the 1,000 most recently joined thread IDs per integration and authenticated Slack bot, and restore them before Socket Mode accepts events. This covers both per-thread sessions and multiple threads sharing one channel session. The Slack provider factory supplies the store through the AgentIntegration registry. State is saved on message receipt, survives disconnect, and is removed when the installation is deleted; switching bot identities resets participation.

A new SQLite table stores this provider state. Existing active per-thread session mappings seed it once on upgrade. Older shared-session conversations did not store thread anchors, so those threads need one mention after upgrading; subsequent restarts preserve participation. Most added lines are the generated Drizzle schema snapshot.

Fixes [SUP-861](https://linear.app/datawizz/issue/SUP-861/slack-ignores-unmentioned-thread-replies-after-app-restart).

### Validation

- Validated against `main` including the merged AgentIntegration refactor (#1087).
- Fifteen regression tests cover registry-based creation/recreation, startup event ordering, all thread-routing modes, write failures, disk reopen, legacy recovery, integration/bot isolation, bounded state, and deletion cleanup. The original four restart cases failed before the fix.
- **900 tests passed, 7 skipped** across 51 integration/service/API/migration test files.
- Typecheck, lint, API build, and Electron build passed, including runtime dependency checks.
- **Real Slack + real Sonnet agent:** an unmentioned follow-up after cold restart recalled `CEDAR-861` in the original thread and agent session. Pause/resume and recovery from legacy session mappings also passed. Two threads sharing a channel session both recalled `SHARED-861` in their respective threads after restart. An unjoined thread remained unanswered and created no thread session.
- The isolated test app stopped, its private Slack channel was archived, and temporary credential copies were removed. Original dev/prod profiles were read only.


JeremyBischoff:
Validated live against real Slack: the unmentioned thread reply after restart is dropped on the base and recovered on this PR (upgrade, restart, pause/resume, shared-session threads).

Blocking: `factory-reset.sup206.test.ts` fails because factory reset doesn't delete `slackThreadState`. It cascades at runtime, but the test requires the explicit delete next to `chatIntegrationSessions` in `src/api/routes/settings.ts`.

Optional: the seed `save()` inside `load()` (`slack-thread-state.ts:47`) isn't guarded like the receipt-path save, so a write failure there fails `connect()` before Socket Mode starts.

PR #1099, opened 2026-09-15
Recover chat replies when the container refuses the session stream (SUP-859)

JeremyBischoff:
A reply to a chat whose agent session the container does not have now reaches the existing chat self-heal and gets answered, instead of failing with "Failed to send your message to the agent" on every message. The common trigger is an agent-initiated `send_chat_message`, which maps the chat to a session that only exists on the host. The stream readiness change that waits for the container's acknowledgement dropped the container's "Session not found" refusal and rejected with a generic close, which the self-heal does not match, so the dead mapping stayed forever on Slack, Telegram and iMessage.

Closes https://linear.app/datawizz/issue/SUP-859/chat-integrations-replies-fail-forever-after-a-proactive-dm

- The container refuses a stream for an unknown session with an error frame, then closes. `subscribeToStream` now rejects readiness with that frame's message, so the attach fails with "Session not found".
- The chat self-heal already matches that text, archives the mapping, prepares the input and starts a fresh session with the reply. No manager change.
- Every other pre-acknowledgement failure (close, socket error, replaced or cancelled attach) keeps the generic error, so transient drops still get the retry prompt and never rotate. Frames after the acknowledgement cannot reject an already resolved readiness.
- The mock container resolved readiness for any session id, so tests reached the HTTP 404 path production no longer takes. It now refuses unknown sessions the way the real container does.
- The agent's outbound message does not carry into the replacement session. That is unchanged behavior from before the readiness change.

No UI change.


yiw190:
Thanks for tracking this down — the regression is mine, from the readiness await I added in #1008. The fix here is right: the container's `Session not found` frame is the correct signal, and I confirmed the new branch only fires on the pre-ack refusal.

It does leave the root cause though: `ensureSession` still maps the chat to a session the container never had, so every proactive DM's first reply goes fail → close → archive → new session. Fine to land this as the guard; a follow-up should make that path not fail in the first place:
- `outbound_only` flag on `chat_integration_sessions`, set when `ensureSession` creates a host-only session.
- `notifySessionOfOutboundMessage` branches on it (host-only → `appendAssistant` directly, no silent catch).
- `handleIncomingMessage` treats an `outboundOnly` mapping as the new-chat path and archives it only after the live session exists.

yiw190:
One user-visible gap to flag: after the rotate the agent has no memory of what it proactively sent — "Should I proceed with X?" → "Yes, go ahead" → fresh session only sees "Yes, go ahead". Pre-#1008 had the same problem, but it undercuts the point of a proactive DM.

Suggest handling it in the follow-up: read the assistant text from the host-only session's transcript, strip `SYSTEM_MESSAGE_PREFIX`, cap it, and pass it into `startNewChatSession` as a short "already sent on your behalf" system-prompt section. The e2e test should then assert the new session's prompt contains the outbound text.

PR #1100, opened 2026-09-15
Add OpenAI Live conversation voice and standard read-aloud with BYOK

iddogino:
OpenAI BYOK now supports both GPT-Live conversations inside agent sessions and standard read-aloud using `gpt-4o-mini-tts`. Live uses the configured summarizer to translate spoken requests and agent results; standard TTS reads message text directly with selectable voices and speed. API keys stay on the host.

Deepgram and OpenAI share the conversation interface, React integration, and agent coordinator. Host routes delegate through optional provider capabilities, with named errors for unsupported operations. Cleanup stays bound to the provider that created each session.

- Add user-bound Live sessions, bounded request bodies, serialized and timed cancellation, safe pause/resume, transient WebRTC recovery, cleanup retries, and expiry warnings.
- Show interlaced two-line subtitles and play holding music only during eligible silent work; the adapters provide speaking state. Live stops and blocks standalone read-aloud for the duration of the call, including connection setup and pauses.
- Generalize TTS initialization to describe either a token-based WebSocket or server-streamed HTTP connection. The new authenticated synthesis route validates provider, voice, text, and speed before delegating to the provider.
- Reuse the existing PCM player, text segmentation, word highlighting, pause/resume, stop, speed, and voice controls for OpenAI read-aloud. HTTP batches preserve order, split oversized text, cancel outstanding work, and reject stale requests after a provider change. Shared streaming deadlines bound waiting for headers and each chunk, with no fixed total lifetime.

- Return capability errors directly so they remain 400s under the production error handler. Expose provider-owned configuration/access messages safely and log upstream failures. Share provider validation, canonical provider IDs, and Unicode-safe speech chunking; select TTS adapters by transport. Keep `/tts-token` as a thin compatibility wrapper for older clients.

## Validation

- This follow-up passes 331 tests across 29 unit/integration files, including production-mounted route errors, healthy streams beyond 30 seconds, stalls and cancellation, Live audio ownership, provider configuration, both conversation engines, settings, and playback.
- Seven Chromium browser tests pass: OpenAI read-aloud controls, Live stopping/hiding read-aloud and restoring it on exit, existing read-aloud scroll behavior, and four Deepgram voice-mode tests.
- Typecheck passes. Full lint has no errors; touched-file lint is clean. `git diff --check` is clean.
- Screenshots use synthetic data and mocked audio. Live BYOK was tested interactively during development; the new standard TTS integration is validated with mocked provider responses and real browser PCM playback. Live OpenAI TTS quality and latency still need a smoke test.

Known limits: terminal Live failures require exiting/re-entering voice mode; cleanup handles are in process memory; agent response correlation uses existing stream state rather than durable command/turn IDs. Read-aloud voice/speed preferences do not alter Live's Marin voice. Architecture and setup: `docs/openai-live-voice-spike.md`.

## Screenshots

| Light | Dark |
| --- | --- |
| OpenAI voice and read-aloud settings (light) | OpenAI voice and read-aloud settings (dark) |
| OpenAI read-aloud playback (light) | OpenAI read-aloud playback (dark) |
| Live conversation and subtitles (light) | Live conversation and subtitles (dark) |
| Live reserves audio and hides read-aloud (light) | Live reserves audio and hides read-aloud (dark) |


iddogino:
## Review of 36cc9c0 (merged with current main)

**Verification**
- Lint on all 38 changed TypeScript files: clean.
- Typecheck: clean after a fresh install.
- Tests in the 13 changed test files: 194 passed, 0 failed.

**Findings, most severe first**

1. **Leaked Live session handles lock users out.** `src/api/routes/voice.ts:95` keeps the map entry and its 60-minute timer when the upstream hangup throws, and the renderer never retries. Page reloads never release at all (no `pagehide` handler). Four leaks hit the per-owner cap and `POST /live/session` returns 429 for up to an hour. Outside auth mode the owner is the constant `local`, so every device on the host shares that cap. Delete the entry regardless of hangup result, and release on `pagehide`.

2. **Pause during a submit leaves a phantom request.** `src/renderer/lib/voice-agent-coordinator.ts:101` returns early when a request card pauses the coordinator but leaves `awaiting` true. After the card closes the agent's floor is held for 15 s, then a false "request was sent, no activity" warning fires. Same at line 80 for a queued submit. The existing pause test never resumes, so it misses this.

3. **Empty assistant turns break session creation.** `src/renderer/components/messages/message-input.tsx:355` seeds Live history without filtering empty text, and tool-call-only assistant turns carry `text: ''` (message-transform.ts:612). The provider forwards them as `{ type: 'text', text: '' }` input items, which OpenAI rejects with a 400, so voice mode cannot start in most tool-using conversations. `voice-agent-feedback-dialog.tsx:71` filters these for the same reason.

4. **Stale cancel error swallows the next utterance.** `voice-agent-coordinator.ts:90` applies `cancellationError` to the next submit whether or not it was queued behind the failed cancel (the comment says "already-queued" but nothing checks that). One transient interrupt failure loses the user's next real request.

5. **Stale manual delegation flag.** `src/renderer/lib/openai-live-bridge.ts:94` sets `pendingDelegation = 'manual'` even when there is nothing to map, and only a successful map clears it. A stray mic press means the next casual utterance is force-mapped and sent to the agent.

6. **Transient WebRTC state treated as fatal.** `src/renderer/lib/voice-conversation-openai.ts:105` tears down on `disconnected`, which commonly self-recovers. A brief network blip ends the call and hangs up the host session. Only `failed` should be terminal; arm a 5–10 s grace timer for `disconnected`.

7. **Unbounded wait on interrupt.** The old 5 s `INTERRUPT_WAIT_MS` bound was removed. `voice-agent-coordinator.ts:100` awaits the interrupt POST with no timeout anywhere in the chain, so a stalled container leaves the UI in "thinking" with no error and no 15 s warning.

8. **Unconditional cancel on idle sessions.** `src/renderer/lib/voice-conversation-deepgram.ts:127` sends `cancel` even when nothing is running (the old hook guarded on `isActive || awaitingTurn`). Talking over a read-aloud tail POSTs `/interrupt`, which also runs the agent-scoped `reviews.denyAll()`. Guard on `snapshot().active || pendingAccepted`.

9. **Hold music no longer yields to foreign read-aloud.** `voice-conversation-deepgram.ts:194` only counts audio when `activeId === streamId`. Clicking read-aloud on an older message plays over the hold loop until the agent's own reply is audible. Main stopped it within 200 ms via `readAloud.isAudible()`.

10. **Missing PR screenshots.** The PR changes renderer UI (voice-tab copy, the new `voice-conversation-preview.tsx`, `voice-mode-controls.tsx`) but attaches no light/dark captures and does not say "No UI change". Per CLAUDE.md, run the `pr-screenshots` skill and `gh pr edit 1100 --attach`.

**Lower priority, confirmed**
- Stale `['voice-configured']` query cache builds the wrong adapter for up to 60 s after switching openai→deepgram.
- Voice tab hides the "TTS not set up" note for members under OpenAI Live with no replacement.
- Fixed 60-minute server-side hangup with no user warning.
- Two sites call `res.json()` before checking `res.ok`.
- Efficiency: `history` array rebuilt every `MessageInput` render; `setSnapshot`/`setAgent` receive fresh objects per stream token (second render per token); two RMS `reduce` loops every 20 ms over fftSize 2048; `ResizeObserver` rebuilt per transcript delta.
- `mapLiveConversation` strips code fences with a regex where `skillset-service`/`agent-template-service` use `output_config.format.type: 'json_schema'`.

**Refuted during review**
- Replacing the bespoke `limitJsonBody` with Hono's `bodyLimit`: it 500s on under-limit chunked bodies under `@hono/vite-dev-server` with `overrideGlobalObjects: false`, so the custom middleware is justified.
- `content.text` undefined crash: producer invariant guarantees a string.
- Resume-after-card floor handoff idle gap: the premise does not occur.

Server side otherwise looks right: endpoints are authenticated, owner-bound, body-limited to 128 KB, and Zod-validated at the boundary. The 502 on session creation echoes the provider's error message verbatim; fine today since the provider builds its own strings, worth keeping in mind if upstream response text is ever passed through.

iddogino:
## Follow-up review of 598be4895 (commits 51d8c96a3 + 598be4895)

**Verification**
- Typecheck: clean. Lint: clean.
- Tests in the 16 affected files: 290 passed, 0 failed.
- Screenshots attached (light + dark).

All nine code findings from the first review are fixed in 74ad1651c and confirmed against the code. Below covers only the two newest commits.

**Findings, most severe first**

1. **Provider-capability 400s become 500s in production.** `src/api/routes/voice.ts:49` `requireLiveConversation()` throws `HTTPException(400, { res })`. The global handler at `src/api/index.ts:285` returns `{ error: 'Internal server error' }` 500 for every thrown error, and a sub-app mounted with `app.route()` inherits it (Hono copies sub-app handlers raw when the sub-app has the default `errorHandler`). Reproduced by mounting the router under a parent with the production `onError`: a Deepgram/unset deployment calling `/live/session` or `/live/map` gets 500 instead of the intended 400 with the named message. `voice-live.test.ts` passes only because `voice.request()` runs the router standalone. Return `c.json(..., 400)` from the helper, as every other route in `src/api` does (this is the only `HTTPException` import in the API).

2. **Read-aloud now plays over an active Live call.** `supportsTts()` flipping to true for OpenAI makes the "Read aloud" menu item (`message-item.tsx:425`, gated only on `useIsTtsConfigured()`) reachable during a Live WebRTC session. `voice-conversation-deepgram.ts` subscribes to `readAloud` and ducks/stops it; `voice-conversation-openai.ts`, `voice-conversation-live.ts` and `use-conversation-mode.ts` never reference it. Result: `HttpTtsAdapter` playback on top of the Live `<audio>` output, nothing pauses either, and the reader's output can reach the open Live mic and be treated as user speech. Hide/stop read-aloud while Live is active, or subscribe the way the Deepgram adapter does.

3. **Whole-body 30 s timeout cuts long synthesis mid-stream.** `openai-provider.ts:38` and `tts-http.ts:81` use `AbortSignal.timeout(30_000)` on the fetch, which bounds the entire streamed PCM body, not time-to-first-byte. A batch is one segment up to the 4096-char cap (minutes of audio at 24 kHz), and 80 words at speed 0.7 breaches it on a slow upstream. `reader.read()` at `tts-http.ts:95` then rejects, the adapter closes with `error`, and the finished-message path (`use-read-aloud.ts:172-175`) has no retry. Use a per-chunk idle timeout (re-armed on each read) or a TTFB-only deadline, on both sides.

4. **Generic `/tts-session` error loses the actionable message.** `voice.ts:194` collapses every `getTtsConnection()` failure into one 502 string and logs nothing. The `/tts-token` path it replaces surfaced the provider's message (e.g. "No API key configured for Deepgram. Add one in Settings > Voice.") and `console.error`'d it. Users now see identical text for a missing key, a bad key and an outage. Surface the thrown message (400 for the two configuration errors, 502 + logging for upstream).

5. **Dead route and duplicated preambles.** `GET /tts-token` has no callers left (`use-read-aloud.ts` and all e2e specs moved to `/tts-session`) yet gained a new "use /tts-session" branch. "No voice provider configured…" appears six times in `voice.ts` and the voice/speed resolution at 238-242 is copied verbatim into 188-191. Delete `/tts-token` and factor a `requireConfiguredProvider(c)` helper shared by `/tts-session`, `/tts` and the Live helper.

**Simplification, lower priority**
- `createTtsAdapter()` (`tts.ts:197`) dispatches on provider id although the server now reports `connection.transport`; switching on transport removes both defensive "wrong transport" branches (`tts.ts:75`, `tts-http.ts:23`) and their tests.
- The 4096-char splitter at `tts-http.ts:70` is a third copy of a chunking loop (`openai-live-bridge.ts:14-27 liveTextChunks`, `chat-integrations/utils.ts:101-122 splitChatMessage`); extract one `splitText` helper.
- `ttsSynthesisSchema.provider` (`tts-types.ts:14`) is a hand-copied enum of `VoiceProvider`, used only so the client can echo the provider back for the 409 check. `hasTtsVoice()` and `getTtsSynthesis()` already reject cross-provider requests; drop the field or derive the enum from the settings type.
- `use-hold-sound.ts:58` now polls `readAloud` on a 200 ms interval, a second source of truth beside the `speaking` prop. Folding read-aloud into the OpenAI adapter's speaking state fixes this and finding 2 together.
- `voice-live.test.ts:20` mock still declares `supportsTts: () => false` for OpenAI and the test at line 66 asserts it, so nothing covers the new production `/configured` shape for OpenAI (`supportsTts: true`, voice catalogue, default `marin`).

PR #1102, opened 2026-09-15
feat(accounts): connect Shopify stores from the App Store install (SUP-847)

JeremyBischoff:
Shopify only lets merchants install Gamut from its App Store listing, so a Shopify connection cannot start inside Gamut like every other service does. This PR handles the other direction: Shopify installs the app, Platform hands the store to the merchant's workspace, and the workspace connects it through Composio without the merchant clicking anything else. Agents then call the store's GraphQL Admin API through the proxy.

Closes https://linear.app/datawizz/issue/SUP-847/connect-shopify-in-the-gamut-app-from-the-app-store-install

## The flow

From Shopify (the merchant's path, and the reviewer's):

1. Install from the App Store, or open an installed app from the store admin.
2. Shopify redirects to Platform's `/install/shopify`, which deeplinks into the workspace as `Connections?shop=<store>` (SkillfulAgents/platform#442 and SkillfulAgents/platform#444, merged).
3. Connections starts the Composio grant in the same tab with the store pre-filled, so Composio's "enter your store" page is skipped and Shopify approves an already-installed app without a prompt.
4. The callback page returns to the new account. A store that is already connected goes straight to its account instead.

From Gamut: hidden for now. The App Store listing 404s until Shopify approves the app, so Shopify is marked `unlisted` and the directory, the agent's service search, the system prompt, and the chat card do not offer it. Once we pass App Store review, a one-line flip switches it on: the Shopify tile appears and opens the listing.

Reconnect on a lapsed store runs the same pre-filled grant in a popup.

## What Shopify forces

- **One account per store.** Shopify issues one token per app and store. A second grant kills the first, so a duplicate account would look healthy and fail on every call. Connecting a store that already has an account updates that account instead. Background sync does not import Shopify grants, so it cannot create the duplicate either.
- **The account is named after its store, and cannot be renamed.** The name is how the connect path finds the existing account to update. Rename it and the next install creates a duplicate, which kills the original. A dedicated store column would avoid this, at the cost of a migration for one provider. X and Plaid shipped without one, so this PR keeps the name as the key and guards it in four places.
- **GraphQL only.** New public apps may only use the GraphQL Admin API, so the proxy refuses anything else.
- **Platform only.** Like X and Plaid, Shopify exists only through Gamut's Composio (a custom auth config on Gamut's own app), so it is hidden on a personal Composio key.

Desktop-local workspaces are out of scope: `/deeplink` opens cloud workspaces only.


src/api/routes/connected-accounts.ts
iddogino: This is a global route and it is littered with shopify specific references. Single-integration logic does not belong in a generic route. 

It also seems out of scope here -> I thought the flow on the app is - get a toolkit slug + token, save. Not sure why we are doing a bunch of vendor specific validaion

PR #1104, opened 2026-09-15
feat(voice): platform provider switches from Deepgram to OpenAI voice via the proxy

yiw190:
> **Depends on** [platform#446](https://github.com/SkillfulAgents/platform/pull/446) being deployed first — the proxy must serve `/v1/openai/*` or cloud voice 404s.

## Why

#1100 makes the OpenAI voice stack (GPT-Live conversation, `gpt-4o-mini-tts` read-aloud, realtime dictation) available as BYOK and it performs much better than the Deepgram stack. Platform-connected workspaces should get the same experience by default without supplying their own OpenAI key. Today the `platform` voice provider is hard-wired to Deepgram through `/v1/deepgram/*`.

## What changed

**`PlatformVoiceProvider` is now `OpenaiVoiceProvider` behind the proxy.** It subclasses the OpenAI provider and overrides only what differs: the credential (platform token instead of a stored key), the base URL (`${proxy}/v1/openai`), and the user-facing copy for missing-connection / auth / balance failures. Everything else — TTS synthesis, Live session create/close, summarizer-based Live mapping, client-secret minting, whisper transcription — is inherited unchanged.

To make that possible `OpenaiVoiceProvider` gained two protected hooks: `apiBaseUrl()` and `messages()` (an `OpenaiVoiceMessages` record), plus a shared `failureMessage(status, fallback)` so 401/403 → auth copy and 402/429 → quota/balance copy in one place. Its `id`/`name` are widened from literals to `VoiceProvider`/`string` so the subclass can override them. No behaviour change for BYOK OpenAI beyond one unified 401/403 message.

**Renderer:** `/token` and `/voice-agent-token` return `protocol: 'deepgram' | 'openai-realtime'`. The STT, voice-agent, and iframe polyfill factories switch on that protocol (not the provider id). TTS already selects by `connection.transport`.

**Settings copy:** the Platform option in Settings › Voice now reads `GPT-4o Mini Transcribe · GPT-Live` and describes OpenAI voice via the platform connection.

Deepgram BYOK is untouched (parked, not removed). Platform login already defaults `sttProvider: 'platform'`, so no onboarding change.

## Effect on a cloud user

Before: dictation and read-aloud used Deepgram Nova/Aura via the proxy; no conversation voice.
After: dictation uses OpenAI realtime transcription, read-aloud uses `gpt-4o-mini-tts` with the OpenAI voice picker, agent sessions get GPT-Live conversation voice, and Telegram/chat voice notes transcribe through whisper — all billed to the workspace through the proxy.

## Screenshots

| Light | Dark |
| --- | --- |
| Settings Voice with Platform (light) | Settings Voice with Platform (dark) |
| Settings Voice Platform picker (light) | Settings Voice Platform picker (dark) |

## Validation

- New `platform-provider.test.ts`: token-only configuration (stored OpenAI key is never used), every endpoint routed to `https://proxy.test/v1/openai/...` with the platform bearer, platform wording for missing connection / 403 / 402 / Live failure.
- Updated `tts-support.test.ts`, `tts.test.ts`, `openai-tts.test.ts`, `voice-live.test.ts` for the new mapping (platform → OpenAI catalogue, `http` transport, Live supported).
- Voice-related tests pass. Typecheck clean for `src/` (the `tools/` errors are pre-existing missing optional SDKs). ESLint clean on touched files.
- Not verified: live end-to-end against a deployed proxy — needs platform#446 deployed and an `OPENAI_API_KEY` project with GPT-Live access.


iddogino (changes requested):
## Review of ee1de4842

**Context.** #1100 is already squash-merged into main (f83abc14a), so this PR is a single commit on top of main and can be retargeted now.

**Verification.** Typecheck and lint clean; 212 tests across the 16 voice-related files pass. The `/v1/openai/*` proxy allowlist is not verifiable in this repo (platform#446).

The subclass shape is right: credential source, base URL, copy, and key validation are the only overrides, and every OpenAI endpoint goes through one `apiBaseUrl()` hook. The items below are what I'd like addressed before merge.

### Requested changes

1. **Iframe dictation is broken for platform.** `src/api/speech-recognition-polyfill.ts:201` is a third provider dispatcher: `if (provider === "openai") return createOpenaiAdapter(); return createDeepgramAdapter();`. A dashboard iframe on a platform deployment fetches `/api/voice/token`, gets `{ provider: 'platform', token: 'ek_…' }`, opens `wss://api.deepgram.com/v1/listen` with an OpenAI client secret, and fails. `stt.ts:556` and `voice-agent.ts:72` were updated; this one was not.

2. **Missing-token copy points at a key field platform has no access to.** `voice-provider.ts:47/81/104/146/169` still throw "No API key configured for Platform. Add one in Settings > Voice." for a disconnected platform account, reached from `/token`, `/voice-agent-token` and `/tts-session`. The Voice tab renders no key input for platform and its note says "No API key required". `messages().missingKey` has the right wording but only `synthesizeSpeech` and `createLiveSession` use it; route the base-class checks through it.

3. **Raw gateway bodies reach the composer.** `openai-provider.ts:234` (`mintClientSecret`) embeds `await res.text()` in the thrown message for any status other than 401/403/402/429, and `/token` / `/voice-agent-token` return `error.message` verbatim. With the proxy lane not yet deployed (404) or the gateway down (502), a platform user starting dictation sees an HTML error page in `VoiceInputError`. The removed platform `mintEphemeralToken` cancelled the body and used a fixed message; do the same here.

4. **Inconsistent failure copy on Live start.** `createLiveSession` (`openai-provider.ts:117`) bypasses `failureMessage()` and hard-codes `OpenAI Live session failed (…) + liveSessionHint`, so a revoked platform token yields "Sign in again…" on dictation but "OpenAI Live session failed (401). Check your platform voice access and workspace balance." on voice mode. Use `this.failureMessage(res.status, …)` there, and `${this.name}` in the six other strings (120, 130, 210, 234, 238) that still hard-code "OpenAI" now that `name` is widened.

5. **Platform disconnect leaks an open Live session.** `closeLiveSession` (`openai-provider.ts:126`) does `if (!apiKey) return`; for platform that means a disconnect during an open GPT-Live session skips `/hangup`, `LiveSessionRegistry.attempt` treats the resolved promise as closed, and the billable session stays open on OpenAI's side with nothing left to close it. Pre-existing for BYOK, but the platform token lifecycle is independent of voice settings so the window is real now. Capture the token in the registry's close callback at create time (voice.ts `liveSessions.add(owner, () => …)`), or at least return `closing` rather than silently succeeding.

6. **Renderer quota copy is BYOK-only.** `stt.ts:407` `friendlyRealtimeError` and its twin at `voice-agent-openai.ts:249` map `insufficient_quota` / `rate_limit_exceeded` to "check your OpenAI account balance and billing settings". Platform users now go through these adapters; the host already defines `PLATFORM_MESSAGES.quotaExceeded` for the same condition. Pass the credential owner into the adapter (or share the message constants) and collapse the two copies.

7. **`PlatformVoiceProvider.validateKey` is unreachable.** `POST /api/settings/validate-stt-key` rejects providers other than `deepgram`/`openai`, and `voice-tab.tsx` only renders a key input for `isApiKeyProvider()`. The override and its test cover dead code that would make a metered `/realtime/client_secrets` call per click if ever wired. Delete the override, or wire the route first.

8. **Protocol is still chosen client-side from the provider id.** STT, voice-agent and the polyfill each encode "platform ⇒ OpenAI realtime", so the fact must ship in lockstep with the server — finding 1 is the demonstration, and a stale web tab or older Electron client against a newer deployment builds `DeepgramAdapter` with an `ek_` secret. TTS already selects by `connection.transport` and conversation by the server-reported `conversationEngine`. Have `/token` and `/voice-agent-token` return `protocol: 'deepgram' | 'openai-realtime'` from a `getSttProtocol()` on `BaseVoiceProvider` (inherited free by the subclass) and switch all three factories on it.

### Also, not blocking

- The only route test exercising `platform` on `/live/session` and `/live/map` was deleted without a positive replacement (`voice-live.test.ts:82`), and the mock still models platform as lacking `getLiveConversation`. Add `providerName = 'platform'` to the positive create/map/close cases.
- Settings > Voice copy changed with no light/dark screenshots and no "No UI change" line; per CLAUDE.md run `pr-screenshots` and `gh pr edit 1104 --attach`. The model label and the "Conversation voice uses GPT-Live…" sentence are now duplicated verbatim between the platform and openai entries.

Refuted during review: token freshness during long TTS/Live (re-read per request), the renderer client-secret format (passed through verbatim), the legacy `/tts-token` 400 for platform (designed migration path), attribution for chat transcription (same context as the LLM provider).

yiw190:
Addressed the eight requested items.

Token routes now return protocol, and the polyfill / STT / voice-agent factories switch on that. Missing-credential copy goes through messages().missingKey. Mint failures no longer leak gateway bodies. Live start uses failureMessage(), hangup captures the token at create, and renderer quota copy shares the platform strings. Removed the dead validateKey override.

Also added platform to the Live route tests and attached Settings › Voice light/dark screenshots.

iddogino (approved):

PR #1107, opened 2026-09-16
Actor-owned idle alarm: each container sleeps itself (SUP-835)

iddogino:
Closes SUP-835.

## What

Auto-sleep is decided by the actor. The central once-a-minute monitor is gone; each container arms its own idle alarm from the events that define activity, and the alarm puts it to sleep.

Before, each minute and for every running agent, the monitor read `sessions.hasActive()`, `container.startedAt()`, `container.lastKeepAliveAt()` and `sessions.list()`. The last one read the agent's session metadata from disk. The actor already sees every event that defines activity, so those reads bought nothing, and a remote actor could not have answered them where the monitor runs anyway.

## How

- **Clock** (`src/shared/lib/container/activity-clock.ts`): the runtime owns an `ActivityClock` in place of its two loose timestamps. Start, keep-alive and session activity mark it; stop and dispose reset it. Session marks only move forward.
- **Alarm** (`src/shared/lib/container/idle-alarm.ts`): every mark re-arms one unref'd timer for the auto-sleep timeout counted from that mark. On fire, a busy container (a session active or awaiting input) gets another full window, a quiet-but-recent one waits the remainder, and an idle one is stopped with the same options the sweep used, `escalateToForceStop: false` included. A stop that does not complete leaves the clock marked and the alarm retries a minute later. This is the local form of a Durable Object alarm: activity and decision live with the container.
- **Session activity funnel** (`session-store.ts`, `session-summary-cache.ts`, `local-agent-actor.ts`): the actor's session store carries an `onActivity` hook that marks its runtime, and `recordSessionActivity` calls it. That is the one path every session write passes: stream frames, provisional sends, transcript appends, `sessions.recordActivity`. `sessions.markIdle` marks the clock too.
- **Contract** (`types.ts`): `ContainerOps.idleSince(): number | null` replaces `startedAt()` and `lastKeepAliveAt()`, which had no other reader. It is the clock the alarm reads, exposed for a status view or a fallback sweep. The runtime answers it directly, so the actor op is a passthrough.
- **Lifecycle**: the alarm reads the clock as empty unless the container is running and neither starting nor stopping. A session of a stopped agent is still written to (a message deleted, a transcript appended) and that marks the clock, but it arms nothing, and a fire during a start does nothing. The alarm is armed once a start settles, re-armed from the existing marks whenever a status sync sees the container running, and disarmed when a sync reports it stopped.
- **Timer range**: every armed delay is clamped to Node's maximum (2^31 - 1 ms); a longer timeout proceeds in legs and every fire rechecks the deadline.
- **Settings**: the timeout is read at every arm. The settings PUT route calls `containerHost.rearmIdleAlarms()` when it changes so running containers pick up a shorter or longer window at once. Zero disables.
- **Startup**: no monitor to start or stop. The boot-signals test drops its log marker.
- **Test adapter** (`host-from-manager-mock.ts`): every runtime the adapter hands out gets a real clock and `idleSince`, and the host accepts `rearmIdleAlarms`, because actor session writes now mark the clock.

## Behaviour

Matches the sweep in intent. An agent with an active session, a session awaiting input, or a keep-alive inside the window is never stopped. A warm-started agent with no sessions sleeps one timeout after its start. An agent with nothing on its clock is left alone. Two differences: the container sleeps at the timeout rather than up to a minute later, and a timeout change applies immediately instead of at the next tick.

One deliberate widening: a provisional send that `markSessionIdle` later reverts still counts for the clock, so at worst a container stays awake one extra window. The summary cache still reverts it for session ordering.

## Tests

- `idle-alarm.test.ts`: the alarm on a fake container with fake timers. Arm from a mark, push-out on every mark, remainder for a past mark, immediate fire for a stale one, busy re-arm, disabled, shortened timeout, retry after an incomplete or failed stop, cancel, and re-entrancy.
- `container-runtime-idle-alarm.test.ts`: the real runtime through start, keep-alive, session write, busy and awaiting, disabled, host re-arm on a shorter timeout, retry after an incomplete stop, an offline write to a stopped agent, a stale mark during an in-flight start, a sync that reports stopped then running, and drop.
- `activity-clock.test.ts`: the clock on its own.
- `registry.test.ts`: the actor's `idleSince` passthrough, the store hook marking the right agent's runtime, and `markIdle` marking the clock.
- `session-service.summary-cache.test.ts`: the hook fires for every finite record.
- `perf/idle-alarm.perf.ts`: fires every running container's alarm by hand over the small profile, once with nothing idle and once with everything idle and stopped. Both pinned at zero filesystem operations.

Typecheck and lint pass.

No UI change.


iddogino:
CI note: `e2e-webkit-macos` fails on one repeat of `e2e/specs/openai-read-aloud.spec.ts:64` ("Live stops read-aloud and reserves audio until the call exits"), a spec added to main today in #1100. The same spec fails the same job on `codex/renderer-voice-providers` (run 35040631801) and passed on earlier repeats within these runs, so it is flaky and unrelated to this PR, which changes nothing under `src/renderer/`. Every other check is green.

iddogino:
Both P2s fixed in 055da1dc4.

**Idle stops against stopped or starting containers.** The alarm now reads the clock through `sleepableSince()`, which is empty unless the container is running and neither starting nor stopping. An offline write to a stopped agent still marks the clock but arms nothing, and a fire during an in-flight start does nothing. Because the start's own mark happens while the start is in flight, `trackStart()` arms the alarm once the start settles (a failed start leaves the container not running, so that arm is a no-op). Tests: `a session write to a stopped agent marks the clock but arms nothing, and the next start is unaffected` and `a stale mark during an in-flight start neither arms nor fires against the start` in `container-runtime-idle-alarm.test.ts`.

**Timer range.** Every armed delay is clamped to `MAX_TIMER_MS` (2^31 - 1). A longer timeout proceeds in legs and every fire rechecks the deadline, so a 30-day setting arms once per leg instead of looping. Test: `waits out a timeout longer than a Node timer in legs, without firing early` in `idle-alarm.test.ts`, which also asserts the delay handed to `setTimeout`.

iddogino:
Fixed in 18b600c22.

**Re-arm on the return to running.** `syncAgentStatus` now arms the alarm on every running report, from whatever the clock holds, and disarms it on a stopped report. The marks are kept either way, so a transient inspect failure followed by a running report arms from the original start mark; a container already past the timeout sleeps at once instead of earning a fresh window from the rediscovery floor. Tests in `container-runtime-idle-alarm.test.ts`: `re-arms when a status sync sees the container running again after reporting it stopped` (fires during the stopped window, then stops on the first running sync) and `a periodic sync that keeps seeing the container running does not push the alarm out`.

iddogino:
Local end-to-end run of the alarm on 18b600c22: the real web server in mock-container mode (`E2E_MOCK=true`, so only Docker is replaced; settings, persister, runtime and alarm are real), fresh data dir, `autoSleepTimeoutMinutes=2`, three agents. A is started and left alone, B gets a message at 60 s, C gets a keep-alive at 100 s. Statuses from `GET /api/agents` every 10 s:

```
  0s settings: autoSleepTimeoutMinutes=2 -> 200
     stored value: 2
  0s created agents: A=kqlctslbwr B=iw5ywxllq9 C=pkegv23qop
  0s started all three:  C=running  B=running  A=running
 10s C=running  B=running  A=running
 ...
 50s C=running  B=running  A=running
 60s B: message sent (HTTP 201)
 60s C=running  B=running  A=running
 ...
100s C: keep-alive (HTTP 200)
100s C=running  B=running  A=running
110s C=running  B=running  A=running
120s C=running  B=running  A=stopped
 ...
180s C=running  B=running  A=stopped
190s C=running  B=stopped  A=stopped
 ...
210s C=running  B=stopped  A=stopped
220s C=stopped  B=stopped  A=stopped
250s C=stopped  B=stopped  A=stopped

server log lines from the alarm:
     [ContainerRuntime] Agent kqlctslbwr idle for >2m, stopping...
     [ContainerRuntime] Agent iw5ywxllq9 idle for >2m, stopping...
     [ContainerRuntime] Agent pkegv23qop idle for >2m, stopping...
```

A sleeps at 120 s (start + 2 min), B at 190 s (message + 2 min, the reply frames landing a few seconds after the send), C at 220 s (keep-alive + 2 min). One alarm log line per container, no central sweep involved.

PR #1112, opened 2026-09-16
refactor(db): database handle behind a seam, with bundled migrations (SUP-864)

iddogino:
Linear: SUP-864 (first sub-issue of SUP-863). Follows #1108, which landed first; this PR is rebased on it.

Behaviour-neutral for the desktop app and single-tenant web: the same file opens with the same pragmas and runs the same migrations. What changes is where the driver is chosen, where migrations come from, and how tests build a database.

## What changed

- **One seam.** `src/shared/lib/db/drivers/` wraps the driver packages: `better-sqlite3.ts` (desktop, web) and `libsql.ts` (devDependency, CI only, never in the production import graph; `build:api`'s runtime-dependency check confirms). Both return an `AppDatabase`: drizzle over the schema with the result kind left open. Outside tests, only `drivers/` imports a driver package now.
- **Explicit open.** `openDatabase()` in `open-database.ts` creates the handle, migrates from the bundle, runs the data migrations and publishes it; `getDb()` returns it and throws before then. Electron main, the web server, the Vite dev server (through its lifecycle plugin, which also gates `/api` requests on the open) and the perf harness call it first. `db` stays a proxy over `getDb()`, typed as the synchronous view until SUP-865, so the 62 importers are untouched. First-access initialisation is gone; a failed open or migration now fails the boot (Electron quits with a fatal report) instead of the first request.
- **Bundled migrations.** `npm run db:bundle` (chained into `db:generate`) writes `migrations/bundle.ts` via `readMigrationFiles`, the exact `MigrationMeta[]` the folder migrator reads; `bundle.test.ts` fails on drift. The Dockerfile copy and the Electron extra-resource are removed; nothing reads `process.cwd()` or `process.resourcesPath` for migrations. `open-database.test.ts` migrates a fresh database and upgrades one written by the previous release's journal, asserting the `__drizzle_migrations` rows.
- **Raw bypasses through drizzle.** The chat access bootstrap is an `insertWhere()` checked with `changesOf()`; the auth startup probes are `db.get(sql)`; the audit-log prune takes an `AppDatabase` and issues its 5,000-row slices with `db.run(sql)`, keeping the `setImmediate` yields. `ApiLogAutoDeleteDb` and the `sqlite` export are deleted. Run results are opaque on the handle, so every `.changes` read (27 sites) goes through `changesOf()`.
- **Data-migration runner** typed against `AppDatabase`, awaiting its statements so it runs on either driver. The one migration awaits its two statements as well, the only place this PR converts statements to `await`.
- **Test helper and CI matrix.** `createTestDatabase()` builds an in-memory database from the bundle on the driver `DB_DRIVER` names. `npm run test:libsql` runs `scripts/libsql-must-pass.txt` with `DB_DRIVER=libsql`; `test.yml` gains the `unit-tests-libsql` job. The list starts with the database package tests and the prune service (five files) and only grows. Real-driver tests adopt the helper as they are touched; the five integration tests that seeded through the `sqlite` export open their own second connection to the same file.
- **Lint.** `no-db-transaction` also matches `getDb().transaction(` (the note left on the ticket from #1108).

## Verification

- Typecheck: only the pre-existing `session-service.test.ts:1007` error. Lint: 0 errors (48 pre-existing warnings).
- Full unit suite on the branch: everything passes except the machine-specific container tests and the pre-existing `session-service` fork-stamp test. Re-run after rebasing on main: 228 files / 4699 tests across the database-touching directories, all green.
- `npm run test:libsql`: 5 files / 23 tests green against libsql.
- `npm run build:api` with the runtime-dependency closure check passes; no `@libsql` in the server bundle.
- Wizard E2E suite run locally after the dev-server fix: 11 passed. The other E2E suites run in CI; `test:perf` run locally, result in a comment.

No UI change.


iddogino:
Local `npm run test:perf` on this branch (rebased on main): 5 files passed, no query- or file-read-count pin moved.

iddogino:
CI: 14 checks pass, including the new `unit-tests-libsql` job. The one red check, `e2e-webkit-macos`, is `openai-read-aloud.spec.ts:64` failing 1 of its 3 repeats (it passed on repeats 2 and 3, and it is the read-aloud audio-reservation test, not a database path). The same test fails 2 of 3 repeats on main's own latest run (0.5.26-rc.1, run 35048845818), so it is a pre-existing WebKit flake rather than something this branch introduced.

iddogino (commented):
👍

PR #1113, opened 2026-09-16
fix(voice): keep the Live reply audible while a request card is up

yiw190:
## Bug

In voice mode on the OpenAI Live engine (what the platform RC uses), the moment any request card appears mid-reply — remote MCP, MCP re-auth, connected account, AskUserQuestion, … — the voice cuts out mid-sentence and never comes back for that reply.

## Repro

1. Enter voice mode on a deployment whose voice provider is OpenAI / platform (`conversationEngine: 'openai-live'`).
2. Ask for something that makes the agent call a user-input tool while it is talking (e.g. "connect my X MCP server").
3. The agent starts saying "I need access to …", the card pops, the speech stops instantly. Answering the card does not replay it.

## Cause

`OpenAILiveConversation.setPaused(true)` did `audio.muted = true` on the WebRTC output element and appended a "Wait silently" instruction to the Live model. The audio kept streaming into a muted element, so unmuting on resume could not bring it back. The chained engine (Deepgram) already does the right thing: it closes the mic and reads the reply to its end (`readAloud.endStream`).

## Change

`live-session.ts`
- Pause no longer mutes the output element; only the mic track + `session.input_audio.mute`.
- Instruction changed from "Wait silently" to "Finish what you are saying, then wait; do not delegate until the application resumes."
- `monitorSpeech()` no longer forces the speaking indicator off while paused, so the UI reflects the audio that is actually playing.

`live-bridge.ts`
- `receive()` lets `session.output_transcript.delta` through while paused so the assistant subtitles keep updating; input transcripts and delegations are still dropped.

Tests updated for the new contract (`live-session.test.ts`, `live-bridge.test.ts`).

## Verified
- `npm run typecheck` clean, `eslint` clean on changed files
- `vitest`: `providers/openai/*` (27), `use-voice-mode`, `use-openai-live-voice-mode`, `voice-mode-engine` (45) all green

Not verified: end-to-end against a real OpenAI Live session — needs a manual run on staging.


yiw190:
## Follow-up (next PR) — what this PR does not cover

Testing this branch locally (dev mode, both changes live) still showed two things after a `request_secret` card. Evidence: the voice screen's subtitle (= the Live model's own `output_transcript`) read *"…smiling for no reason. The walls didn'tANOTHER_SECRET"* while the agent's actual reply was a different story ("A kid taped a paper star…"). Session transcript confirms the agent never sent the first story (the message before the tool call has `stop_reason: tool_use`; the story arrives in a later message after the tool result).

### 1. The reply is still cut mid-sentence at pause (mechanical)

After removing the mute, pause still sends `session.instructions.append` ("finish what you are saying…"). Per the [GPT-Live delegation guide](https://developers.openai.com/api/docs/guides/live-delegation), `instructions.append` is the channel for *"system-level instructions… such as a direction to stop speaking"*, so sending one mid-reply plausibly interrupts generation. The resume instruction ("Continue listening and responding normally") also invites the model to speak, which reads as a "restart".

Proposed: send both state changes as `session.thinking.append` (documented as *"not spoken on append"*, the channel for UI-state context) and drop the `instructions.append` on pause/resume. Prototype was `002e5c569` (reverted from this branch to keep it to one change).

### 2. The Live model improvises content the agent never sent

`session.commentary.append` is documented as *"the model is trained to paraphrase the appended text"*. When the agent says "here's a tiny story while that prompt comes up", the model treats it as its own promise and tells a story. A system-prompt rule ("say only what the commentary says…") was tried in `10b335c27` and verified ineffective in a live run.

Candidates, untested:
- Prefix each relayed reply in `OpenAILiveBridge.reply()` with a relay frame (e.g. `Relay this agent reply in your own words and add nothing to it: …`) so the constraint rides with the content.
- Hold back open-ended teasers on the host side, or send them via `thinking.append` and only voice completed replies.

### 3. Cosmetic

Transcript deltas are concatenated exactly; after an interruption the next utterance is glued to the previous one (`didn'tANOTHER_SECRET`). `live-bridge.ts` `receive()` could insert a space when a new assistant utterance follows a cut.

PR #1119, opened 2026-09-16
fix(voice): stop the Live request mapper from turning its own clarification into the user's request

yiw190:
## Bug

In voice mode the backend agent sometimes receives a user request the user never made. The Live request mapper (`mapLiveConversation`) is the only writer of that bubble.

Two live examples, same mechanism:

1. User: *"Do we have Supabase access set up on this agent?"*  
   Voice assistant: *"Just to clarify — are you asking if there's an existing credential, or do you want to connect to Supabase?"*  
   Agent received both sentences as one user message:

2. User: *"You are the new casual greeting agent"* → *"Yes"*  
   Voice assistant had asked whether to save a Supabase API key and request a project URL.  
   Agent received: *"Save my Supabase API key and ask me for my project URL to complete the connection."*

In the same session the user said *"just do it"* six times and only got another clarifying question each time.

This is not an STT speaker-label error. Input/output labels on the data channel were correct. The mapper mixed `assistant:` speech into the user request.

A different, valid job of the same mapper is to **understand intent**: fix speech-to-text, apply corrections, and write one clear request. Example: *"super basic grade" / "super base key"* → *"Set up a Supabase connection and save a Supabase key."* That is not this bug. Forwarding the raw utterance would "fix" the leak by destroying that job.

## Root cause

Traced with a full voice-trace plus `/live/map` I/O, then replayed against the live summarizer (wire dump: prompt + body unchanged, `stop_reason=end_turn`, not truncated):

1. The mapper invents `clarify` questions. The voice model speaks them. They re-enter the next mapping as `assistant:` lines.
2. The user says "Yes" (or the next mapping runs while those lines are still in the transcript). The mapper treats the assistant's offer as the user's request.
3. The bridge cleared the user's accumulated words after every mapping, including `clarify`. Each "just do it" arrived alone, so the mapper clarified again.

`history` is also captured once at bridge construction and stays `[]`, and the transcript window is 24 lines, so earlier context is gone. Those make the input worse; they are not what writes the wrong bubble.

Replays (Haiku and Sonnet): the captured 2132 body fabricated 6/6. Stripping `assistant:` lines stopped the fabrication. Changing the prompt, adding `utterance` / `lastClarify`, or clearing `previousRequest` did not, as long as those assistant questions stayed in the transcript. Feeding `lastClarify` back made the screenshot-shaped body worse.

## Fix

- Keep the mapper as the writer of `message.text`. It should understand intent from the user's words: clean STT, apply corrections, combine turns. It must not paste the utterance verbatim, and it must not import tasks from `assistant:` lines or from its own previous clarification.
- Pass `utterance` (everything the user said since the last request actually *sent*; survives `clarify`) and `lastClarify` so the model can see the user's own words separately from the spoken transcript.
- After `MAX_CONSECUTIVE_CLARIFY` (2) clarifications in a row, send the user's accumulated words so the loop can break. The backend can ask a better question.
- Do **not** replace `message.text` with the raw utterance. That blocks the leak and also blocks summarization (observed live).

The UI subtitle (`onUtterance`) still resets per mapping; the accumulator is a separate field.

## Tests

Bridge and provider tests cover: utterance accumulation across clarify; mapper-written intent reaching `onRequest`; STT-style rewrite kept; consecutive-clarify cap; empty `message` / `clarify` rejected; `/live/map` schema for the new fields.


yiw190:
superseed by 1126

PR #1121, opened 2026-09-16
Actor-owned per-agent state: pending inputs, reviews, re-auth waits, computer-use grants, summary cache (SUP-836)

iddogino:
Moves the per-agent in-memory state that the actor stack (#1031 → #1050) left in process-wide singletons onto the agent's actor, so that `agentRegistry.evict(slug)` is the one operation that ends everything an agent holds in this process, and the singletons keep only routing.

Linear: SUP-836

## What moved

| State | Was | Now |
| -- | -- | -- |
| Pending user-input requests, claims, resolution trail | `userInputRequestManager` maps keyed by tool-use id | `AgentInputRequests`, one per actor (`user-input/agent-input-requests.ts`) |
| Proxy / x-agent review settlers and timers | `reviewManager.settlers` | `AgentReviews`, one per actor (`proxy/agent-reviews.ts`) |
| Account and MCP re-auth waits | `accountReauthManager` / `mcpReauthManager` group maps keyed by `slug\0subject` | `AgentReauthWaits`, one per actor per flavour (`proxy/reauth-waits.ts`, one generic class with an account spec and an MCP spec) |
| Computer-use grants and the grabbed app | `computerUsePermissionManager` maps keyed by slug | `AgentComputerUse`, one per actor (`computer-use/agent-permissions.ts`); persisted "always" grants are read and written per agent entry |
| Warm session summary | module `Map` keyed by workspace path | `SessionStore.summaryCache`, one per store, bound to the store's storage key (`services/session-summary-slot.ts`) |

They are composed in `agent-actor/agent-state.ts` and held as `LocalAgentActor.state`. `agentRegistry.evict` rejects every parked review and re-auth wait, settles every open request as `invalidated` (so the wire hears of it), drops the grants, the grab and the summary, then drops the runtime and the handle. `evictAll` is now test-only: the container-runner change in the settings route calls `containerHost.clearRuntimes()`, which is what it always did for runtime state, and no longer implies "the agents are gone".

## The singletons are routers

Each manager keeps its public API but holds no per-agent state: it gets a directory over the actors' stores from the registry (`agent-actor/store-directory.ts`) and dispatches. `userInputRequestManager` keeps the one index the ticket describes (request id → the agents holding it, maintained from the stores' transitions) plus the transition listeners; a settlement sequence makes its cross-agent trail lookup deterministic. Process-wide sweeps (`rejectAll` at shutdown, `completeAccount` / `completeMcp` on reconnect, `resolveRequestsByParent`) iterate the live handles. A registration that names no agent has no store to live in and is dropped (logged, never thrown); every persister registration carries the slug.

The actor's own `inputs` ops read and write its store directly. The review and re-auth ops go through their routers, which dispatch straight back to the actor's store: the router is the process-wide API those flows already expose (the proxy routes, the mock container client, startup), and the actor's ops are its agent-scoped face.

`account-reauth-manager` is now fenced too, with `AccountReauthOps` on the actor (`inputs.accountReauth`); the proxy, agents and account-reauth routes go through it. The fence gained the new per-agent modules and lost the `account-reauth-manager → request-manager` and `perf/harness → session-summary-cache` entries; `connected-accounts.ts` and `startup.ts` keep their process-wide `completeAccount` / `rejectAll` calls on the allowlist.

## Tests

- `registry.test.ts` plants state of every kind (request, review, account and MCP wait, once-grant, grab, summary revision), evicts, and asserts nothing survives into the next handle, every settlement went out as a transition, and the routers' directories read through the handles only.
- The existing manager suites run against real stores via `attachInMemoryAgentState()` (`agent-actor/testing/in-memory-agent-state.ts`), which is also what the persister-only suites attach, next to `attachSessionStores`.
- `permission-manager.test.ts`'s settings mock now feeds writes back into reads, since each agent persists only its own entry.
- Two `agents.test.ts` cases were re-pointed: an agent-less registration is refused rather than parked, and the attention-failure case spies on the actor's store instead of the router.

## Left out

Step 5 of the ticket, the persister's per-session maps (activity, awaiting cache, stream subscriptions, SSE clients, coalescing, send serialisation, last fatal), is not in this PR. The ticket puts it last, and SUP-837 splits the persister first so that move is smaller; the routers and `AgentState` are the seam it will slot into.

## Verification

- `npx tsc --noEmit` clean; `npm run lint` 0 errors (48 pre-existing warnings).
- Targeted suites green: user-input, proxy, computer-use, agent-actor, summary cache, the persister and replay suites, chat integrations, agent-service, and the route suites that touch these managers (`agents`, `x-agent`, `proxy`, `mcp-proxy`, `security-repro`, `session-scope`, …).
- Full `vitest run`: see the CI job; locally only the two known environment-specific container-client files differ from CI.

No UI change.

## Review round 1

- **Same request id in two agents.** A session cloned into another agent replays its source's tool-use ids, so an id can be open in two agents' stores. The router's index is now id → the agents holding it (in registration order), and every id-addressed router method takes the agent when the caller knows it; every persister path passes its slug, so `completeInputRequest(A, …)` settles A's copy and only A's, and clearing A leaves B reachable. A bare id answers with the first registrant still holding it, as the one global store used to. `submitDecision` with an expected agent consults only that agent's store.
- **Dev-server reload.** The host, persister and routers survive a backend module reload on `globalThis`; the actors' state now does too (`agentActorStates`), and the reloaded registry wraps each surviving state in a fresh handle, so pending inputs stay open, parked reviews stay decidable, and a reconnect completed after the reload releases the request parked before it. `registry-reload.test.ts` drives a real `vi.resetModules` reload for each case.

## Review round 2

- **Bare-id routing after a reload.** The router rebuilds its id → owners index from the stores' open requests when the registry attaches them, so after a dev-server reload a bare `getOpenRequest(id)` and the id-only `submitDecision(id, decision)` still find the owner; nothing is announced a second time. Covered in `registry-reload.test.ts` and `request-manager.test.ts`.
- Rebased onto `main` (the memories ops and the runtime's `stale()` landed in `local-agent-actor.ts` meanwhile).

## Review round 3

- **Owner order after a reload.** Each store records the process-wide sequence at which it first registered an open request (the sequence lives on `globalThis` in development, next to the stores), and the rebuild on attach replays the stores in that order. So when two agents hold one id, the first registrant stays first across a reload however the handles were made. Covered in `registry-reload.test.ts` and `request-manager.test.ts`.


iddogino:
Both findings addressed in 52cfc52, with the review's probes carried into the suite:

- **Duplicate ids across agents:** the router index is now id → the agents holding it (registration order), every id-addressed router method takes the agent when the caller knows it, and every persister path passes its slug, so `completeInputRequest(A, …)` settles only A's copy and clearing A leaves B reachable by bare id. Regression tests: `request id ownership across agents` in `message-persister.request-lifecycle.test.ts` (both review probes) plus two cases in `request-manager.test.ts`.
- **Dev-server reload:** the actors' state persists on `globalThis` next to the host, persister and routers, and the reloaded registry wraps each surviving state in a fresh handle. Regression tests: `registry-reload.test.ts` (the three review probes, plus eviction after a reload), and a `createAgentRegistry(deps, { states })` case in `registry.test.ts`.

Typecheck and lint clean; the actor, store, persister, chat-integration and route suites pass locally.

iddogino:
Fixed in a5323e2 (branch rebased onto main; the conflict in `local-agent-actor.ts` was the new memories ops and `stale()` from #1024): `attachAgents` now rebuilds the id → owners index from the stores' open requests, in handle order, without re-emitting `created`. Both round-two probes are in `registry-reload.test.ts` (bare-id lookup and the id-only `submitDecision` after a real `vi.resetModules` reload), and `request-manager.test.ts` checks that re-attaching stores with open requests rebuilds the index silently.

Typecheck and lint clean; 84 store/actor/persister/integration files and 16 route files pass locally after the rebase.

iddogino:
Fixed in 18c84d7: each store now records the process-wide sequence at which it first registered an open request (kept on `globalThis` in development, like the stores), and `attachAgents` replays the stores in that order, so the first registrant of a shared id stays first across a reload regardless of handle order. Your `reload-owner-order` probe passes on this head; the same case is in `registry-reload.test.ts` (via a real `vi.resetModules` reload) and `request-manager.test.ts` (re-attach with b registered before a).

iddogino:
CI on 18c84d7: every check passes except `e2e-webkit-macos`, which fails (twice, including a rerun) on `e2e/specs/openai-read-aloud.spec.ts:64` — "Live stops read-aloud and reserves audio until the call exits" (30 s timeout). The same job fails on the same spec on `main` at this PR's merge base 6051958a1 (Tests run 35161703524), and this PR's last commit only changed the server-side request-index ordering, so it is not introduced here.

PR #1129, opened 2026-09-17
feat(sessions): noninteractive flag for unattended sessions, notify_user tool (SUP-884, SUP-887)

yiw190:
Implements [SUP-884](https://linear.app/datawizz/issue/SUP-884/automated-session-add-a-variant-of-system-prompt) (unattended-session guidance + `notify_user`) the way review asked for in [SUP-887](https://linear.app/datawizz/issue/SUP-887): one `noninteractive` flag, set by cron / trigger / widget-repair, read by host and container, cleared by promotion.

## Why

Scheduled and webhook sessions run with nobody watching. When one hits a soft failure or finishes with something the user should see, the outcome sits unread in a hidden session: the model has no guidance on how far to retry, and when it has nothing to *ask*, no way to make the session visible.

The first cut of this PR bolted that on with a host-side `systemPrompt` appendix, a container `isAutomated` flag, a host `promotedToInteractive` marker and a five-flag hidden set. Review: make it one flag, pass it to the container, let the container own the tool and the prompt section, and let promotion clear it.

## What the flag means

`noninteractive: true` = nobody is looking at this session.

| Session | `noninteractive` | notify_user tool | unattended prompt section | hidden in sidebar |
|---|---|---|---|---|
| Normal | false | no | no | no |
| Cron / trigger | **true** | yes | yes | yes (until promoted) |
| Widget repair | **true** | yes | yes | yes (until promoted) |
| Chat Integration | false | no | no | yes, by `isChatIntegrationSession` (unchanged) |
| x-agent target | false | no | no | yes, by `invokedByAgentSlug` (unchanged) |

Promotion (a `request_*` / `AskUserQuestion`, `notify_user`, or a human message) writes `noninteractive: false` on both sides. It never flips back.

### Host
- `SessionMetadata.noninteractive` (schema + type). `task-scheduler`, `trigger-manager`, `widget-repair-service` write it on register and pass `metadata: { noninteractive: true }` to the container. `x-agent.ts` passes `noninteractive: false`.
- `session-visibility.ts`: `isNoninteractiveSession(meta)` — explicit flag when present, else derived for pre-existing rows from `(isScheduledExecution || isWebhookExecution || isWidgetRepair) && !promotedToInteractive`. `isHiddenAutomatedSession(meta)` = that, plus chat-integration / x-agent until promoted. Chat Integration is hidden but **not** noninteractive.
- `promoteAutomatedSession`: writes `noninteractive: false`; `promotedToInteractive: true` is only still written for the chat / x-agent classes (hidden by their own flag). It is the one place a session becomes visible, so after its own write it tells the container (`client.promoteSession` → `POST /sessions/:id/promote`) for sessions that were noninteractive. A failed container call is logged + captured, not fatal (see Decisions).
- Push channels: `session_notify` added to `web-push-channel` (`PUSHABLE_TYPES`, 1h TTL) and `apns-relay-channel` (`VISIBLE_TYPES` / `SILENT_TYPES`). APNs alerts only the session's origin device, and a cron / trigger session has none, so `session_notify` alerts **every eligible device** (`ALERT_ALL_DEVICES_TYPES`); owner prefs and, in auth mode, agent access still gate per device.
- `notification-manager`: `triggerNotification()` returns `'created' | 'suppressed'` (settings turned the type off) and a failed promotion now **throws** instead of `console.error` + continue. `triggerAgentNotify` gates on `isNoninteractiveSession` and returns the outcome; the `notify_user` handler tells the agent "notified" vs "visible, but no notification was sent". Per-turn flag renamed `notifyUserHandledThisTurn`.
- `wake-delivery.ts`: scheduled wakes send `noninteractive: true`. **Bug fix**: the container read every `schedule_resume` wake as a human message and promoted the session, dropping its unattended class.
- `SendMessageOptions.isAutomated` → `noninteractive` ("not from a human"); x-agent follow-ups keep sending it.
- Deleted `scheduler/automated-session-prompt.ts` (+ test).

### Container
- `metadata.isAutomated` → `noninteractive`; a record persisted with the old spelling is mapped on load. Eviction class, browser-lock release and fork clearing keyed on it as before.
- `ClaudeCodeProcess` takes `noninteractive`. `createUserInputMcpServer(getProcess, { noninteractive })` registers `notify_user` only when true. `system-prompt.md` gains a `<%#noninteractive%>` **Unattended Session** section (the former appendix text; now says that requesting input or calling `notify_user` ends unattended mode, and drops the scheduled/webhook split). Rendered just before Agent-Specific Instructions.
- Promotion follows the host: `POST /sessions/:id/promote` → `session-manager.promoteToInteractive` clears the flag, persists it, and calls `process.setNoninteractive(false)`; a cold (not loaded) session gets the persistence update only, so its resume is interactive-class. The process rebuilds its query (tool list + prompt) on the **next** send through the same interrupt + re-query branch a capability block flip uses, so the in-flight tool call finishes first. A human message (`sendMessage` without `noninteractive`) goes through the same method before the send, which also covers a container that was down for the host's call. The container does **not** infer promotion from a request being raised: the host may auto-approve one (computer-use / script-run with a stored grant) without showing it.
- `warmProfileSchema` gains `noninteractive`; `nextWarmProfile` warms interactive when the host sends `prewarmDefaults`.

### Decisions
- **x-agent is interactive** (per discussion): the calling agent reads the result and can send follow-ups. Consequence: x-agent target sessions leave the container's eager idle-eviction class (they were `isAutomated` before) and now idle out on the interactive threshold. Sidebar hiding / `request_*` promotion for them is unchanged.
- **Widget-repair is noninteractive**: system-initiated, nobody watching.
- Legacy rows are handled at read time by `isNoninteractiveSession`; no write-back migration. `promotedToInteractive` stays readable for those rows and for the chat / x-agent classes.
- Host still rejects a stray `notify_user` in an interactive session — covers the window between promotion and the container's re-query, and any host/container disagreement.
- Host write first, then the container call. If the container call fails, the session is visible and the host rejects a stray `notify_user`: the safe side. The reverse order could leave the container interactive while the host session stays hidden (the tool gone, nobody watching), which is the state review flagged.
- `session_notify` on APNs alerts every eligible device rather than picking a "most likely" phone: notify_user exists to reach the user, and the in-app notification row is visible to the same audience.

## Verification
- `agent-container`: `tsc` clean; vitest **1194 pass** (81 files). New: `mcp-server.test.ts` (tool gated on the flag), `claude-code-effort.test.ts` (noninteractive query bakes section + tool; re-query on the next send after `setNoninteractive(false)`; no-op when unchanged), `session-manager-idle-eviction.test.ts` (flag survives persistence and resume; legacy `isAutomated` maps; human message promotes once and tells the process; `noninteractive` wake / `shouldQuery:false` don't; raising a request — `secret`, `script_run`, `notify_user` — does **not** promote; host promote on a live session tells the process, persists, is idempotent; host promote on a cold session flips persistence without a resume; unknown session → 404), `warm-profile.test.ts`, `system-prompt-vars.test.ts` (section rendered only when flagged, before agent instructions), fork test for both spellings.
- Host: `tsc` shows only the pre-existing optional-module errors in `tools/` and `libsql.ts`; full vitest **12765 pass** (793 files) after updating the assertions this change moves (`promotedToInteractive: true` → `noninteractive: false`; wake options; x-agent / widget metadata). New: `session-visibility.test.ts` (derivation, x-agent hidden-but-interactive), `notification-manager.test.ts` (suppressed outcome, promotion failure throws, chat / x-agent / promoted refused), `message-persister.test.ts` (suppressed message to the agent, completion alert still suppressed; host promote calls the container after its own write, skips chat sessions, and a failed container call is logged without undoing the host promotion), `web-push-channel.test.ts` / `apns-relay-channel.test.ts` (`session_notify` pushes — these 6 tests fail on the previous commit with 0 sends; APNs alerts all devices, honors prefs, honors agent access in auth mode).
- `eslint` on changed files: 0 errors.

Not verified: end-to-end in a running container. Single check: schedule a task whose prompt is "call notify_user with message 'test'" → the session appears in the sidebar with one OS notification and no second completion alert; then reply in that session → the container's next query has no `notify_user` tool and the prompt has no Unattended Session section (container log: `Restarting query (noninteractive true -> false)`).


src/shared/lib/notifications/notification-manager.ts
```diff
@@ -323,6 +325,34 @@ class NotificationManager {
     })
   }
 
+  // Agent-initiated `notify_user`: no input is awaited, the user just needs to see it.
+  async triggerAgentNotify(
```
iddogino, line 334: success reported when the notification was dropped.
Returns `{ ok: true }` even when triggerNotification silently drops the notification

src/shared/lib/notifications/notification-manager.ts
```diff
@@ -323,6 +325,34 @@ class NotificationManager {
...
+    message: string,
+    title?: string,
+  ): Promise<AgentNotifyResult> {
+    // Trade-off (SUP-884): the tool's purpose is "promote a hidden session and
+    // alert". An interactive session is already visible and the user reads
+    // its replies, so a notification there would only duplicate the chat.
+    // Refuse instead of silently degrading. Delete this check to allow
+    // notify_user everywhere; nothing else depends on it.
+    const meta = await agentRegistry.get(agentSlug).sessions.metadata(sessionId)
+    if (!isHiddenAutomatedSession(meta)) {
```
iddogino: I don't think this should include chat integrations -- those aren't truly hidden. I guess there are two classes:

- Interactive Automated (user still sees response, but session is hidden in UI) - chat integrations
- Non Interactive Automated (user sees nothing) - triggers, crons

So maybe to standardize across the system, we should call them Automated and Interactive vs Non Interactive. Notifications apply for Non Interactive sessions

agent-container/src/mcp-server.ts
```diff
@@ -87,7 +88,7 @@ export function createUserInputMcpServer(getProcess: () => RemoteMcpInjectionTar
       scheduleTaskTool, scheduleResumeTool, listScheduledTasksTool, updateScheduledTaskTool,
       cancelScheduledTaskTool,
       pauseScheduledTaskTool, resumeScheduledTaskTool,
-      deliverFileTool, deliverSessionTool, requestFileTool, requestBrowserInputTool,
+      deliverFileTool, deliverSessionTool, requestFileTool, requestBrowserInputTool, notifyUserTool,
```
iddogino: To not confuse the agent, I think we should enable the tool only on noninteractive sessions. 

Combined with other comment -- I think we need to pass a new flag when starting a session (and in session metadata): `noninteractive`. This is how we know a session is an automated cron / trigger job. Then we can also clean up the promote logic

iddogino (changes requested):

agent-container/src/mcp-server.ts
yiw190: Yeah that would be cleanest for this

PR #1133, opened 2026-09-17
docs(agent): teach OpenAI audio endpoints

yiw190:
## Summary

- The proxy now serves OpenAI TTS and transcription at `/v1/openai/audio/speech` and `/v1/openai/audio/transcriptions`, but the container agent still only knew about audio through Deepgram.
- The agent guide and system prompt now advertise **OpenAI only** (`audio.md`, `/v1/openai`). Deepgram is omitted so new agent usage lands on OpenAI.
- The Deepgram backend endpoint is unchanged and stays up for existing callers. We can check ClickHouse later for leftover Deepgram traffic.

## Test plan

- [x] `npx vitest run src/system-prompt-vars.test.ts` in `agent-container` (66 passed)
- [x] Guide does not mention Deepgram (`rg -i deepgram agent-container/docs/audio.md` is empty)
- [ ] Not verified: an agent actually exercising the OpenAI examples end-to-end through the proxy from inside a container


yiw190:
Staging check on `platformproxy.gamutstaging.com`: the OpenAI lane covers the Deepgram surfaces we used to advertise.

- Deepgram `/listen` → `/v1/openai/audio/transcriptions` (`whisper-1`) — 200, transcribed the TTS clip
- Deepgram `/speak` → `/v1/openai/audio/speech` (`tts-1`) — 200, returned mp3
- Deepgram `/auth/grant` (app STT shim) → `/v1/openai/realtime/client_secrets` — 200, secret issued

`PlatformVoiceProvider` already points at `/v1/openai` only. Deepgram speak/listen/grant still return 200 if someone calls them; we are just not teaching new users those paths.

PR #1136, opened 2026-09-18
fix(voice): seed OpenAI Live with a token-budgeted window of chat history (SUP-888)

yiw190:
Linear: [SUP-888](https://linear.app/datawizz/issue/SUP-888/voice-mode-openai-live-session-starts-with-almost-no-chat-context)

## Bug

Activating OpenAI voice mode halfway through a chat gives the voice model almost no knowledge of what was already discussed, so the user has to re-explain everything.

Cause: the renderer sent the last 24 user/assistant turns (each cut to its **last** 4,000 chars), and `createLiveSession` then kept only the last **12** turns, each cut to its **last 800 chars**. That is ~2.4k tokens of context, and long assistant replies lost their opening (where the answer usually is).

## Repro

1. Have a chat with 15+ back-and-forth turns, including at least one long assistant answer.
2. Enter voice mode (OpenAI Live) and ask "what did we decide earlier?" or refer to something from turn 3.
3. Voice model has no idea; it can only delegate.

## What changed

- **Host** (`openai-provider.ts`): GPT-Live `input` is built by `windowVoiceHistory` — newest → oldest under a **7,000-token** budget estimated with `tokenx` (o200k-calibrated, 2 kB), at most **128 turns**, each clipped to its **first 1,500 chars** with a `[…]` marker. This stays under the API's hard cap of 128 messages / 8,192 tokens for startup `input` ([docs](https://developers.openai.com/api/docs/guides/live-conversations)).
- **Renderer** (`message-input.tsx`): only the transport bound (128 turns × head 4,000 chars); no budgeting. A later commit also fills newest-first under a shared 40 KiB serialized budget so `/live/*` bodies stay under the host's 128 KiB limit.
- The same window bounds the history handed to the request-mapping summarizer, so widening the schema does not blow up that prompt.

Tool calls / tool results / thinking were already excluded and remain excluded.

## Why `tokenx` instead of `gpt-tokenizer`

Head-to-head on 1,996 Cursor transcripts (122k clipped turns) vs `gpt-tokenizer` `o200k_base`:

- Chat-text MAPE **12.4%** (chars/4 was 26.7%). tokenx slightly overcounts on average (safer).
- At budget **7,000**, 0 sessions exceeded 8,192 actual tokens (max framed 8,174). At 7,500, 1 session overflowed.
- Coverage vs exact tokenizer at 7,500: p50 turns 23 vs 24; p90 54 vs 59; whole-session fit 1,468 vs 1,521.
- Package: **53 KB** vs **26 MB**. No lazy import, no tsup `external`, no `<|endoftext|>` throw workaround.

chars/4 is not safe: 272 sessions overflowed 8,192 at budget 7,500.

## Why these numbers

Measured on long agent transcripts (user + assistant text only) as a proxy — no long SuperAgent session was available locally:

- p50 session = 4.1k tokens / 25 turns → fits entirely; ~68% of sessions fit entirely.
- p90 = 24.8k tokens / 141 turns → window keeps ~54 turns.
- A single pasted 2.9M-char message defeats a naive window (0 turns kept); the per-turn head cap is what makes it work.

## Verification

- `vitest`: voice-history-window, voice-history-transport, and openai-live suites green (32 tests).
- `eslint` clean on changed files.
- Not verified: `npm run build:api`, or a live spoken session against the real API. Suggested smoke test: 20-turn chat → voice mode → "what did we decide earlier?".

## Follow-up (not here)

Summarize older history into a `developer` message for sessions that exceed the budget.


yiw190:
## tokenx vs gpt-tokenizer (why we dropped the full tokenizer)

Compared `tokenx` 2.1.0 against `gpt-tokenizer` `o200k_base` (and a flat chars/4 heuristic) on 13 crafted samples plus **1,996 Cursor transcripts / 122,085 clipped turns**. Then replayed the real window — newest-first, 1,500-char head cap, +4 tokens/message — and measured **actual** o200k tokens against the 8,192 cap.

We're not close on typical chats: p50 is 4.1k of the 8,192 cap; ~68% of sessions fit whole. The tokenizer was only there so a chars/token heuristic wouldn't undercount CJK/code and get the session rejected.

### Accuracy (chat corpus)

| Estimator | MAPE | Underestimates | Mean bias |
|---|---|---|---|
| tokenx | **12.4%** | 18.8% | +7.8% (safer) |
| chars/4 | 26.7% | 62.1% | −13.2% |

English chat is essentially exact. tokenx overcounts dense 中文 (safe). It still badly undercounts base64 (~−71%), same as its docs.

### 8,192-cap safety

| Estimator | Budget | Sessions over 8,192 | p50 / p90 turns kept |
|---|---|---|---|
| gpt-tokenizer | 7,500 | 0 | 24 / 59 |
| **tokenx** | **7,000** | **0** (max framed 8,174) | **23 / 54** |
| tokenx | 7,500 | 1 (max 8,617) | 24 / 56 |
| chars/4 | 7,500 | **272** | — |

tokenx at 7,000 loses about one median turn and five p90 turns vs the exact tokenizer, and drops a 26 MB rank table to 53 KB. chars/4 is not usable.

Swapped in `1fcc1ffb0`: sync `tokenx`, budget **7,000**. No lazy import, no tsup `external`, no `<|endoftext|>` workaround.

iddogino (approved):

PR #1137, opened 2026-09-18
fix: keep the composer mounted across provider error toggles (SUP-890)

yiw190:
Fixes [SUP-890](https://linear.app/datawizz/issue/SUP-890/voice-mode-exits-immediately-when-the-sessions-last-reply-is-a).

## Bug

In a session whose **last assistant message is a provider error routed to the composer** (e.g. a persisted platform 402 with `placement: "composer"`), clicking the voice button enters voice mode and exits again within ~100 ms, every time. Other sessions work.

## Repro

1. Open a session where the last assistant row is a `platform-paywall` error (any old 402 with no assistant reply after it).
2. Click the voice button.
3. Transcript shows an “Entered Voice Mode” / “Exited Voice Mode” pair; voice never starts. A “Checking billing” card also flashes on first open.

## Cause

`ProviderErrorPlacement` returned either `<>{children}</>` (no current error) or `<div><Card>{children}</Card></div>` (error). `currentProviderError` is `null` while the session is active, so:

1. Voice on → “switched to voice mode” notice → session active → error hidden → composer moves from inside the card to the fragment → **React remounts `MessageInput`**.
2. Unmount cleanup sends “left voice mode”; the fresh instance starts with `voiceModeOn = false`.
3. Turn ends → error back → remount again.

The cards themselves also moved `children` between fragment positions on clear/dismiss.

## Change

- `ProviderErrorPlacement` always renders the same tree; children live in one fixed slot. The card is a sibling above it, never its parent.
- Cards no longer receive `children`. `ProviderErrorComponentProps.children` → `onDisplaceChildren(displaced)`. The paywall reports `displaced` while billing is positively blocked and the placement sets `hidden` on the slot — same hide-not-unmount approach `session-chat-column.tsx` already uses for request cards, so drafts and voice state survive.
- `PlatformPaywallCard` renders nothing for a **persisted** (`live=false`) row while the billing snapshot is loading. “Checking billing” still shows for a live 402. This removes the flash the user saw.

## Tests

- New: `does not remount children when the current error comes and goes (SUP-890)` — mount counter stays at 1 across active→idle toggles.
- New: `hides the children slot while the component reports them displaced`.
- New: `renders no card for a persisted 402 until the snapshot arrives, then blocks`.
- Paywall suite: composer assertions changed from presence to visibility (`toBeVisible` / `not.toBeVisible`), via a small harness that mirrors the placement slot.

`typecheck`, `eslint`, and the renderer suites under `components/{provider-error,ui,layout,messages}` pass (1269 tests).

Not verified: manual run against the original “Hey man” session in the app.


yiw190:
The change fixes the remounting pattern, but there's an integration gap: when the paywall blocks the composer, it doesn't suspend the voice session, so the microphone can stay live while the controls are hidden. The targeted repro shows this issue even though the 116 focused tests pass. I'd recommend blocking this PR on that high-priority issue. The likely fix is wiring the displacement to existing voice suspension behavior without reintroducing remount flicker. I haven't verified real browser mic behavior yet.

yiw190:
Confirmed and fixed in 0c1902c. Repro: new column test `pauses MessageInput while a composer-placed error withholds it, keeping it mounted` failed on the previous head (`data-suspended="false"` while the slot was hidden).

Fix: `ProviderErrorPlacement` accepts function children and passes `{ displaced }`; `session-chat-column.tsx` folds it into `MessageInput`'s existing `suspended` prop (`pendingRequestCount > 0 || displaced`), so a blocked paywall takes the same pause path as a pending request card: mic closed, hold sound off, voice resumes when the card clears. No remount, the slot stays hidden-not-unmounted.

Not verified: real mic behavior in a browser.

iddogino (approved):

PR #1141, opened 2026-09-18
refactor(renderer): existing sign-in buttons use the shared login button (SUP-894)

JeremyBischoff:
The app's seven other sign-in buttons each draw their busy state and Cancel by hand: the spinner comes before the label on some and after it on others, and Cancel sits after the neighbouring button on four of them. This draws all seven with the shared login button from the previous PR, so each gets the same busy shape and a Cancel beside it on the side that keeps the button still. Nothing else about them changes: same size, colour, icon, wording and behaviour.

Closes https://linear.app/datawizz/issue/SUP-894/existing-sign-in-buttons-use-the-shared-login-button
Related to https://linear.app/datawizz/issue/SUP-814/one-shared-login-button-and-cancel-on-the-in-chat-connect-cards

Stacked on #1140.

| button | visible change |
| --- | --- |
| Wizard, Get Started | "..." becomes "…". The wizard keeps its Cancel on the help line below. |
| Account tab, Connect Account | Spinner moves before the label. Cancel moves beside the button. |
| Account tab, Reconnect | Spinner moves before the label. Cancel moves to its left, so "Add key" stays put. |
| Add-integration dialog, "+" per API | None. Screen readers now hear "Connecting …" while it is busy. |
| Add-integration dialog, Connect Server | "..." becomes "…". |
| Detail page, Reconnect | Gains "Reconnecting…". Cancel moves to its left, so "Rename" stays put. |
| Detail page, MCP Test connection | "..." becomes "…". Cancel moves to its left, so "Tools" stays put. |

The shared button gains an optional test id for its Cancel link, so the existing test ids survive.


JeremyBischoff:
Closing for now to keep the review stack short. This reopens on top of #1140 once it lands.

